PR #1, opened 2020-04-24
Add first draft for visualize class

orelgueta:
This is a first piece of code for visualisation. Very simple for now, but it is already useful for my reports code and the validation codes. Requires a lot more work and a bit more documentation to be more useful for gammasim.


simtools/tests/test_visualize.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/python3
+
+import logging
+import os
+import sys
+import numpy as np
+from astropy import units as u
```
RaulRPrado, line 7: I'm not using 'u' as namespace for units because I don't think single letter namespace is a smart thing to do. We can live well with 'units'.
I'm trying to keep single letters or double repeated letters variables for very short scopes, like short loops.

simtools/tests/test_visualize.py
```diff
@@ -0,0 +1,89 @@
...
+from astropy.io import ascii
+from .. import visualize
```
RaulRPrado, line 9: Better `from simtools import vizualize`

simtools/tests/test_visualize.py
```diff
@@ -0,0 +1,89 @@
...
+                   'formats': ('f8', 'f8')}
+    title = 'Test 1D plot'
+
+    dataIn = np.loadtxt('tests/testData/ref_200_1100_190211a.dat',
+                        usecols=(0, 1), dtype=headersType)
+
+    # Change y-axis to percent
+    if '%' in yTitle:
+        if np.max(dataIn[yTitle]) <= 1:
+            dataIn[yTitle] = 100*dataIn[yTitle]
```
RaulRPrado, line 31: Pep8: space in between operators

simtools/tests/test_visualize.py
```diff
@@ -0,0 +1,89 @@
...
+    data['Reflectivity'] = dataIn
+    for i in range(5):
+        newData = np.copy(dataIn)
+        newData[yTitle] = newData[yTitle]*(1 - 0.1*(i + 1))
+        data['{}%% reflectivity'.format(100*(1 - 0.1*(i + 1)))] = newData
+
+    plt = visualize.plot1D(data, title=title, palette='autumn')
+
+    plotFile = 'tests/testPlots/plot_1D.pdf'
+    if os.path.isfile(plotFile):
```
RaulRPrado, line 42: Can we try to use only pathlib for paths for consistency?
You can do:
`from pathlib import Path`
`if Path(plotFile).exists():`

simtools/tests/test_visualize.py
```diff
@@ -0,0 +1,89 @@
...
+        os.remove(plotFile)
+    plt.savefig(plotFile)
+    if not os.path.isfile(plotFile):
+        logger.critical('Did not create {}!'.format(plotFile))
+        sys.exit(1)
```
RaulRPrado, line 47: Shouln't we raise and error here, instead of exit(1)?

simtools/tests/test_visualize.py
```diff
@@ -0,0 +1,89 @@
...
+    return
+
+
+def test_plot_table():
+
+    logger.debug('Testing plotTable')
+
+    title = 'Test plot table'
+
+    table = ascii.read('tests/testData/Transmission_Spectrum_PlexiGlass.dat')
```
RaulRPrado, line 60: See my general comment on the testData

RaulRPrado:
This is my general comment.

1) About the testData. Do we want to keep any kind of data inside this repository? If not, my suggestion is the following:
I implemented a feature in 'config' that allows to search for files in the directories given in the config file by the key modelFilesLocations.
By now I have:
`modelFilesLocations:[
/home/prado/Work/Projects/CTA_MC/svn/Simulations/SimulationModel/ModelDescription/trunk/configReports,
/home/prado/Work/Projects/CTA_MC/svn/Simulations/SimulationModel/ModelDescription/trunk/datFiles,
/home/prado/Work/Projects/CTA_MC/svn/Simulations/SimulationModel/Verification/dataFromInstrumentTeams
]`, which are the directories with the yml files, dataFiles and data from instrument teams.

I can then do:
`from simtools.util import config as cfg  # we should use like this consistently

# if I want a file called Transmission_Spectrum_PlexiGlass.dat
# that is inside any of this directories, I do
file = cfg.findFile('Transmission_Spectrum_PlexiGlass.dat')
# this will first search locally and then inside those directories listed
# in the config file` 

With this we can use any file from those directories for testing. You just have to use your own config.yml file.

There is an assumption here that the files in these directories are not duplicated, which should be always true.

I'm using this feature to get any file, including the yml files with the model parameters, It's very useful and simple.

One can always start his application with `cfg.setConfigFileName('my_config.yml')`. If this is not done, the default config.yml will be used.

2) The vizualize module is generic, right? It's not related specifically to model parameters, right?
This is just a matter of structure, where we put the modules and how we use namespaces etc.

RaulRPrado:
My previous comment went misformatted, I'm still learning how to do this.
I can accept your pull request, since there is no conflict, or you can approach the comments first. How should we do that?

simtools/tests/test_visualize.py
orelgueta: Agreed. I took the use of a single letter u from ctapipe, but I don't mind using "units". I also don't use any other single/double letter variables, even in short loops.

simtools/tests/test_visualize.py
orelgueta: Yes!

simtools/tests/test_visualize.py
orelgueta: Do you mean the *? 
PEP8 recommends adding space around operators with lowest priorities, but there's no such case here. I prefer to have multiplications without spaces around them.

simtools/tests/test_visualize.py
orelgueta: Sure, but is there a specific reason to use pathlib?

simtools/tests/test_visualize.py
orelgueta: I didn't know which error to raise in this case, so I decided not to. However, I changed it now to raise a RuntimeError and removed the exit. Still not very informative, but better.

simtools/tests/test_visualize.py
RaulRPrado: Yes, the *
Ok, I though we were always suppose to add spaces. It's fine then.

simtools/tests/test_visualize.py
RaulRPrado: Not a strong one. I've been using it since the beginning and it never gave me a reason to change.

orelgueta:
I agree we don't want to keep data in the repository in the future. In addition, I would also like to avoid paths. So in the future we will simply read from the DB which doesn't require paths. 

However, for now I don't mind us using the paths and having a bit of necessary data in the repository. In this case, the data is simply test data in order to be able to run pytest on the visualize module. I can't read this data from the usual database because it doesn't exist there. It was created specifically for this test (we don't have ECSVs in the database yet).
So, my suggestion is for now to keep the data in the repository, and later move it to the DB and remove it. I don't expect it to change anyway so it's not an issue for change control.
What do you think?

orelgueta:
> My previous comment went misformatted, I'm still learning how to do this.
> I can accept your pull request, since there is no conflict, or you can approach the comments first. How should we do that?

Lets first agree on everything and then you accept the pull request.

simtools/tests/test_visualize.py
RaulRPrado: We can create our own exceptions if we happen to need something specific.
In this case I'm not actually convinced that we need to do this verification at all.

orelgueta: The verification is simply a way for pytest to know it failed. Usually you do an assert for it to fail, but I cannot assert anything for this test, so I do something like this. The RuntimeError is "general" enough for this purpose I guess.

simtools/tests/test_visualize.py
GernotMaier: But this is standard and I disagree. Almost every code you find using units is using 'as u' (same is numpy is np). 

Using units already produces a lot of 'clutter' in the code. I would advocate to allow the single letter import here.

RaulRPrado:
I would still prefer to use an existing data file instead of creating these test data files.
But it's fine if you want to keep it.
I would suggest then to create a directory in the main repository directory called 'data' and maybe a subdirectory called test-data [1] (maybe not). Then I would write a function in io_handler called getTestDataFile(fileName) where the path to the test-data directory is given inside. This way we hardcode this path only in this function and not in many places throughout the code.

[1] I know you like 'testData' naming style and I agree on using it for coding. However we don't need to use the same convention for file, directories and modules names. The convention that I've been using is no capital letter and words split by dash in case of folders and underscore for modules.

GernotMaier:
For the data - why don't we put that in a separate repository as e.g. ctapipe is doing (ctapipi-extra)?

simtools/tests/test_visualize.py
RaulRPrado: I agree with using the standard ones for np, plt etc. But u really does not seem like a good idea to me.

simtools/tests/test_visualize.py
RaulRPrado: Ok, you are right.

orelgueta:
> I would still prefer to use an existing data file instead of creating these test data files.

I would also be happy to use existing data, but at the moment I can't. Once we start diverging a bit from the sim_telarray format, I definitely will.

> I would suggest then to create a directory in the main repository directory called 'data' and maybe a subdirectory called test-data [1] (maybe not). Then I would write a function in io_handler called getTestDataFile(fileName) where the path to the test-data directory is given inside. This way we hardcode this path only in this function and not in many places throughout the code.

OK, will do that.

> [1] I know you like 'testData' naming style and I agree on using it for coding. However we don't need to use the same convention for file, directories and modules names. The convention that I've been using is no capital letter and words split by dash in case of folders and underscore for modules.

Sure, this is just my usual way of naming, but I can change for this.

orelgueta:
> For the data - why don't we put that in a separate repository as e.g. ctapipe is doing (ctapipi-extra)?

We might do that at some point, but I won't start with it already since currently we have very little data (only used for testing) and the rest will simply go into the database.
The reason I say we might do that in the future is that I foresee other data we might use, like a Cherenkov spectrum for re-weighting, where we would like to keep this data in the repo. Not sure though.

RaulRPrado:
> > For the data - why don't we put that in a separate repository as e.g. ctapipe is doing (ctapipi-extra)?
> 
> We might do that at some point, but I won't start with it already since currently we have very little data (only used for testing) and the rest will simply go into the database.
> The reason I say we might do that in the future is that I foresee other data we might use, like a Cherenkov spectrum for re-weighting, where we would like to keep this data in the repo. Not sure though.

Agree that we will probably have to do it in the future. By now we still have to read files from svn and if we need extra files, I would just

simtools/tests/test_visualize.py
orelgueta: Now Gernot managed to convince me that u is actually a better idea in this case =)

orelgueta:
I changed the data structure as you suggest @RaulRPrado, but I kept the hardcoded paths in the test code for now. Once you implement the function you mention, we can update the test code.

PR #2, opened 2020-04-30
Reports

orelgueta:
Improve 1D plotting and include 2D plotting as well.


orelgueta:
> Because there was no plotHist2D in visualize. Are you pushing your commits frequently? Commiting only updates your local repository, not the global one.

I realised afterwards that I did not commit visualize while I committed everything else. I am not yet used to the "add" thing. Sorry about that.

PR #4, opened 2020-05-01
Reorganizing the plotting functions

RaulRPrado:
I did some reorganization (at the master) on the ray_tracing and psf_analysis modules in order to keep the plotting function and still avoid duplication. Functions to get data for the visualize module are now called get*ForVisualize().


RaulRPrado:
> This function seems to be here just to explicitly avoid using the visualize module, not clear why.
> Either way, even if you want to use the simple plot of python, you can still use it with structured arrays like
> 
> `plt.plot(data['Radius'], data['Intensity'], **kwargs)`
> 
> So the duplication with the get data functions doesn't make sense to me.

I see which duplication you are talking about. I didn't realize before that the structured data could be used like that. It's not so complicated, indeed.

RaulRPrado:
I'll merge the getData functions then.

RaulRPrado:
Duplications of getData functions solved by https://github.com/gammasim/gammasim-tools/commit/1a5e577eac313859757146de6e2787d2caafd37d

RaulRPrado:
If it's all fine now, please merge pull request.

RaulRPrado:
Can I merge the branches? If not, what is missing?

orelgueta:
We need to resolve the column names point first. I still think those should be the axis titles.

RaulRPrado:
> We need to resolve the column names point first. I still think those should be the axis titles.

Ok. I'll change it back.
It'll be good if we manage to use the title key at some point, though.

orelgueta:
I will try to get it to work, but I didn't have time yet, so let's merge first and see later if we can.

RaulRPrado:
Ok. I'll also have a look if I have time this week.

RaulRPrado:
Running pytest ...

PR #18, opened 2020-05-15
Reports

orelgueta:
Add validate camera FoV into master.


simtools/model/camera.py
```diff
@@ -0,0 +1,670 @@
...
+                pixels['pixID'].append(int(pixInfo[1].strip()))
+                if len(pixInfo) > 9:
+                    if int(pixInfo[9].strip()) != 0:
+                        pixels['pixOn'].append(True)
+                    else:
+                        pixels['pixOn'].append(False)
+                else:
+                    pixels['pixOn'].append(True)
+
+        if pixels['diameter'] == 9999:
```
RaulRPrado, line 130: Maybe we could add an if to verify the funnelShape as well. I think the code will break in some other parts if funnelShape not 1, 2 or 3.

simtools/model/camera.py
```diff
@@ -0,0 +1,670 @@
...
+from simtools.model.model_parameters import TWO_MIRROR_TELS, CAMERA_ROTATE_ANGLE
+
+__all__ = ['Camera']
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class Camera:
+
```
RaulRPrado, line 21: I understood we should also add docstrings for the class itself, despite of the description in the init. The attibutes list should go here, I guess.

simtools/model/camera.py
```diff
@@ -0,0 +1,670 @@
...
+        -------
+        fov: float
+            The FOV of the camera in the degrees.
+        averageEdgeDistance: float
+            The average edge distance of the camera
+
+        Notes
+        -----
+        The x,y pixel positions and focal length are assumed to have the same unit (usually cm)
+
```
RaulRPrado, line 203: I don't use these blank lines in the end of docstrings. We should decide for using it or not.

simtools/model/camera.py
```diff
@@ -0,0 +1,670 @@
...
+            # Pixels in the same row/column can be 20% shifted from one another
+            # Inside find_adjacent_neighbour_pixels the distance is increased
+            # further for pixels in the same row/column to 1.68*diameter.
+            neighbours = self._findAdjacentNeighbourPixels(
+                pixels['x'],
+                pixels['y'],
+                1.4*pixels['diameter'],
+                0.2*pixels['diameter']
+            )
+
```
RaulRPrado, line 374: Here, for example, it will break if funnelShape not in [1, 2, 3]. So either we verify it after reading the camera files, or we add else here to deal with other options.

simtools/model/camera.py
```diff
@@ -0,0 +1,670 @@
...
+        yTitle = 'Az'
+        xPos, yPos = (xLeft + 0.15, 0.25)
+        axesPars = {'xTitle': xTitle, 'yTitle': yTitle, 'xPos': xPos, 'yPos': yPos,
+                    'rotateAngle': (3/2.)*np.pi, 'fc': 'red',
+                    'ec': 'red', 'invertYaxis': invertYaxis}
+        self._plotOneAxisDef(plt, **axesPars)
+
+        return
+
+    def _plotOneAxisDef(self, plt, **axesPars):
```
RaulRPrado, line 463: I think it's a good practice in python to always use the name kwargs for **kwargs. But that is only an impression, I'm not sure about it.

simtools/model/camera.py
```diff
@@ -0,0 +1,670 @@
...
+        legendObjects = [legH.pixelObject(), legH.edgePixelObject()]
+        legendLabels = ['Pixel', 'Edge pixel']
+        if (type(onPixels[0]) == mlp.patches.RegularPolygon):
+            legendHandlerMap = {legH.pixelObject: legH.hexPixelHandler(),
+                                legH.edgePixelObject: legH.hexEdgePixelHandler(),
+                                legH.offPixelObject: legH.hexOffPixelHandler()}
+        elif (type(onPixels[0]) == mlp.patches.Rectangle):
+            legendHandlerMap = {legH.pixelObject: legH.squarePixelHandler(),
+                                legH.edgePixelObject: legH.squareEdgePixelHandler(),
+                                legH.offPixelObject: legH.squareOffPixelHandler()}
```
RaulRPrado, line 616: I'd add a blank line here

simtools/util/legend_handlers.py
```diff
@@ -0,0 +1,181 @@
+import matplotlib.patches as mpatches
+from matplotlib import colors as mcolors
+import numpy as np
+
+
+class pixelObject(object):
```
RaulRPrado, line 6: Are there skeleton for future implementation?

RaulRPrado:
@orelgueta , can I merge this pull request?
I need to do another one.

RaulRPrado:
Well, I did mine anyway. I hope there won't be many conflicts.

orelgueta:
I would like to first make the corrections you asked for and only then merge. I assume you can do yours without issues, the changes are only to camera.py and I assume you didn't touch that.

RaulRPrado:
Ok, no problem

simtools/model/camera.py
orelgueta: Sure, added one and also raised a ValueError instead of defaulting to some random value for the pixel diameter.

simtools/model/camera.py
orelgueta: Yeah, I misunderstood the location. I now put the parameters in the init and the rest in the class definition. Not sure what to do with the general description, so I left it in both places. If you think it's OK, please resolve this conversation.

simtools/model/camera.py
orelgueta: Wasn't intentional, removed.

simtools/model/camera.py
orelgueta: Yeah, verified when reading the file.

simtools/model/camera.py
orelgueta: Shame, I really like the fact the dictionary provides information on what's inside. But OK, rules are rules, I changed it to kwargs.

simtools/model/camera.py
orelgueta: OK

orelgueta:
Hi @RaulRPrado, 
I made all changes requested. I kept a few things "unresolved", so if you agree with them, please resolve them and merge the pull request.
If you want to go through the code and fix brackets, please go ahead of course. I tried to write everything in the same style as you, but to me some things look better this way. I don't care too much though, so feel free.

simtools/model/camera.py
RaulRPrado: Great. I also don't know where the general description is suppose to go. I assume it should go to the class, not to the init.

simtools/util/legend_handlers.py
RaulRPrado: Hmm I see, no problem then. Thanks for the explanation.

RaulRPrado:
Ok, great.
I resolved all and did the style edits.
Will merge now.

PR #22, opened 2020-05-25
Add argparse to FoV application

orelgueta:
Simple one for now

### DIFF
--- a/applications/validate_camera_fov.py
+++ b/applications/validate_camera_fov.py
@@ -2,6 +2,7 @@
 
 import logging
 import matplotlib.pyplot as plt
+import argparse
 
 import simtools.config as cfg
 from simtools.model.telescope_model import TelescopeModel
@@ -12,24 +13,44 @@ logger.setLevel(logging.INFO)
 
 
 if __name__ == '__main__':
-    '''
-    This is an example application to calculate the FoV of a camera, in this case LST is used.
-    The application prints out the FoV and plots the camera.
-    '''
 
-    print('\nValidating the camera FoV of LST\n')
+    parser = argparse.ArgumentParser(
+        description=(
+            'Calculate the camera FoV of the telescope requested. '
+            'Plot the camera as well, as seen for an observer facing the camera.'
+        )
+    )
+    parser.add_argument(
+        '--tel_type',
+        help='Telescope type (e.g. mst-flashcam, lst)',
+        type=str,
+        required=True
+    )
+    parser.add_argument(
+        '--model_version',
+        help='Model version (default=prod4)',
+        type=str,
+        default='prod4'
+    )
+    parser.add_argument(
+        '--site',
+        help='Site (default=South)',
+        type=str,
+        default='south'
+    )
 
-    site = 'south'
-    telescope = 'lst'
-    version = 'prod4'
-    label = 'lst-test'
+    args = parser.parse_args()
+
+    label = '{}-test'.format(args.tel_type.lower())
 
     telModel = TelescopeModel(
-        telescopeType=telescope,
-        site=site,
-        version=version,
+        telescopeType=args.tel_type,
+        site=args.site,
+        version=args.model_version,
         label=label
     )
+
+    print('\nValidating the camera FoV of {}\n'.format(telModel.telescopeType))
 
     cameraConfigFile = telModel.getParameter('camera_config_file')
     focalLength = float(telModel.getParameter('effective_focal_length'))
@@ -47,5 +68,7 @@ if __name__ == '__main__':
 
     # Now plot the camera as well
     plt = camera.plotPixelLayout()
-    plt.savefig('pixelLayout-LST.pdf', format='pdf', bbox_inches='tight')
+    cameraPlotFile = 'pixelLayout-{}.pdf'.format(telModel.telescopeType)
+    plt.savefig(cameraPlotFile, bbox_inches='tight')
+    print('\nPlotted camera in {}\n'.format(cameraPlotFile))
     plt.clf()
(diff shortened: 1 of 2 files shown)


applications/validate_camera_fov.py
```diff
@@ -12,25 +13,45 @@
...
+        default='south'
+    )
+
+    args = parser.parse_args()
 
-    site = 'south'
-    telescope = 'lst'
-    version = 'prod4'
-    label = 'lst-test'
+    label = '{}-test'.format(args.tel_type.lower())
```
RaulRPrado, line 44: All the output files (e.g. from the telescope_model) already contain the telescope type on the name, so it's redundant to add it to the label. I would use a label that only refers to the application itself. For the derive_mirror_rnda, I used derive_rnda.

applications/validate_camera_fov.py
```diff
@@ -47,5 +68,7 @@
 
     # Now plot the camera as well
     plt = camera.plotPixelLayout()
-    plt.savefig('pixelLayout-LST.pdf', format='pdf', bbox_inches='tight')
+    cameraPlotFile = 'pixelLayout-{}.pdf'.format(telModel.telescopeType)
+    plt.savefig(cameraPlotFile, bbox_inches='tight')
```
RaulRPrado, line 72: Note for the future: We need to figure out a system to handle the application plots.

simtools/model/camera.py
```diff
@@ -4,7 +4,7 @@
 import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
 import matplotlib.colors as mcolors
-import scipy.spatial.cKDTree as KDTree
+from scipy.spatial import cKDTree as KDTree
```
RaulRPrado, line 7: That was the bug that I added and fixed later. It was fixed in the master, though.

applications/validate_camera_fov.py
orelgueta: Oh, OK, changed it now to `label = 'validate-FoV'`.

simtools/model/camera.py
orelgueta: Yes, but when I fixed it in my branch and then asked to pull it wanted me to commit first etc. and I was lazy, so I figure the merge would take care of it instead.

orelgueta:
I guess now we can merge?

applications/validate_camera_fov.py
orelgueta: Yes, we should find a solution. I don't know if the DB will provide a solution to this, it probably needs to be produced locally and optionally be pushed to the DB. Requires a bit more thinking I guess...

RaulRPrado:
Yep

applications/validate_camera_fov.py
GernotMaier: Can you explain to me what 'label' is doing / how it is used?

orelgueta: It is not relevant for this application. It is used when you compare one model to another. Most common use case is when you want to compare a modified model (lets say with modified random reflection angle) to one in the database (e.g., Prod4). In that case the label will be used in the plot comparing between the two (and in other places I guess, @RaulRPrado can expand I am sure).

RaulRPrado: Label is actually always relevant.
It's used to label all the output files, and the output directory structure. It's suppose to be used by all the modules that produce any kind of output.
There is the option of not giving any label (the date will be used for the output directory name) but it's not a good practice.

GernotMaier:
When you say output directory, you mean for what?

The FOV produced a pdf in the working directory.

> On 25. May 2020, at 13:48, RaulRPrado <notifications@github.com> wrote:
> 
> 
> @RaulRPrado commented on this pull request.
> 
> In applications/validate_camera_fov.py:
> 
> >  
> -    site = 'south'
> -    telescope = 'lst'
> -    version = 'prod4'
> -    label = 'lst-test'
> +    label = '{}-test'.format(args.tel_type.lower())
> 
> Label is actually always relevant.
> It's used to label all the output files, and the output directory structure. It's suppose to be used by all the modules that produce any kind of output.
> There is the option of not giving any label (the date will be used for the output directory name) but it's not a good practice.
> 
> —
> You are receiving this because you commented.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

RaulRPrado:
> When you say output directory, you mean for what? The FOV produced a pdf in the working directory.

You can ask the TelescopeModel to exportConfigFile, for example. Or you can ask the Camera to export a new camera config file with a certain change (not implemented, just an example).
This is a peculiar application that does not produce any output, but basically all the other ones do.

The output directory is split. If you run any ray tracing application, there will be model related outputs and ray tracing related outputs, for example.

orelgueta:
And in particular we haven't figured out yet what to do with output plots (as we mentioned at some point today in one of these many issues). We think they should be produced locally and then optionally be pushed into the database.

GernotMaier:
ok!

> On 25. May 2020, at 14:01, RaulRPrado <notifications@github.com> wrote:
> 
> 
> When you say output directory, you mean for what? The FOV produced a pdf in the working directory.
> 
> You can ask the TelescopeModel to exportConfigFile, for example. Or you can ask the Camera to export a new camera config file with a certain change (not implemented, just an example).
> This is a peculiar application that does not produce any output, but basically all the other ones do.
> 
> The output directory is split. If you run any ray tracing application, there will be model related outputs and ray tracing related outputs, for example.
> 
> —
> You are receiving this because you commented.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

PR #38, opened 2020-07-09
Application for the whole dish psf and eff area

RaulRPrado:
I prepared the application validate_optics which produces PSF and eff area vs off-axis angle using ray tracing simulations. (Tarek asked for the PST for sources at infinity and I decided to do the app at once.) 

@orelgueta, could you please have a look? It's very little code and not urgent.
I called it validate_optics but we can discuss a more specific name if that makes sense.

I did small changes on other apps in order to allow the label to be passed as command line arg.

### DIFF
--- a/applications/validate_optics.py
+++ b/applications/validate_optics.py
@@ -1,0 +1,84 @@
+#!/usr/bin/python3
+
+import logging
+import matplotlib.pyplot as plt
+import argparse
+
+import simtools.config as cfg
+import simtools.util.general as gen
+from simtools.model.telescope_model import TelescopeModel
+from simtools.model.camera import Camera
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        description=(
+            'Calculate the camera FoV of the telescope requested. '
+            'Plot the camera as well, as seen for an observer facing the camera.'
+        )
+    )
+    parser.add_argument(
+        '--tel_type',
+        help='Telescope type (e.g. mst-flashcam, lst)',
+        type=str,
+        required=True
+    )
+    parser.add_argument(
+        '--model_version',
+        help='Model version (default=prod4)',
+        type=str,
+        default='prod4'
+    )
+    parser.add_argument(
+        '--site',
+        help='Site (default=South)',
+        type=str,
+        default='south'
+    )
+    parser.add_argument(
+        '-v',
+        '--verbosity',
+        dest='logLevel',
+        action='store',
+        default='info',
+        help='Log level to print (default is INFO)'
+    )
+
+    args = parser.parse_args()
+
+    logger = logging.getLogger('validate_camera_fov')
+    logger.setLevel(gen.getLogLevelFromUser(args.logLevel))
+
+    label = 'validate-FoV'
+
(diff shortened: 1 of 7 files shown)


applications/validate_optics.py
```diff
@@ -0,0 +1,112 @@
...
+import simtools.config as cfg
+import simtools.util.general as gen
+from simtools.model.telescope_model import TelescopeModel
+from simtools.ray_tracing import RayTracing
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        description=(
```
orelgueta, line 19: Is this really the description or is it left over from previous code?

simtools/ray_tracing.py
```diff
@@ -350,6 +355,56 @@ def _isValidKeyToPlot(self, key):
...
+        ------
+        KeyError
+            If key is not among the valid options.
+        '''
+        if not self._isValidKeyToPlot(key):
+            msg = 'Invalid key to plot'
+            self._logger.error(msg)
+            raise KeyError(msg)
+
+        ylabel = {
```
orelgueta, line 379: There is duplication between this list and the list in the _isValidKeyToPlot. I would simply make the dictionary of ylables a class variable and replace the function _isValidKeyToPlot with a simple `if not in ylabel.keys()`.

simtools/ray_tracing.py
```diff
@@ -350,6 +355,56 @@ def _isValidKeyToPlot(self, key):
...
+        self._outputDirectory.joinpath('figures').mkdir(exist_ok=True)
+        plotFile = self._outputDirectory.joinpath('figures').joinpath(plotFileName)
+        self._logger.info('Plotting {} and saving it in {}'.format(key, plotFile))
+
+        plt.figure(figsize=(8, 6), tight_layout=True)
+        ax = plt.gca()
+        ax.set_xlabel('off-axis [deg]')
+        ax.set_ylabel(ylabel[key])
+
+        self.plot(key, **kwargs)
```
orelgueta, line 403: I can't comment within "plot" cause it wasn't added in this commit, but why not use "visualize" in order to keep all plots with a consistent look?

applications/validate_optics.py
RaulRPrado: Left over, sorry.

simtools/ray_tracing.py
RaulRPrado: Good idea.

simtools/ray_tracing.py
RaulRPrado: Yes, I agree. I just didn't learn yet to use the visualize, but it's a only matter of spending the time.
I will merge like this and create an issue to remind me to fix that later.

simtools/ray_tracing.py
```diff
@@ -353,7 +361,7 @@ def _isValidKeyToPlot(self, key):
         -------
         bool
         '''
-        return key in ['d80_cm', 'd80_deg', 'eff_area', 'eff_flen']
+        return key in self.YLABEL.keys()
```
orelgueta, line 364: I would remove this function completely, it's just one command =)

PR #40, opened 2020-07-17
Change ray-tracing to use Astropy tables

orelgueta:
Made the discussed changes. In addition, added units where I was certain what they were. 
I think the unit of the effective mirror area was wrong (said cm^2, but I think it is m^2).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ __pycache__
 
 # Sphinx
 docs/build
+
+# DB config gfile
+dbDetails.yml
--- a/clean_files
+++ b/clean_files
@@ -3,4 +3,5 @@
 rm telarray_rand.conf.used > /dev/null 2>&1
 rm -r simtools-files  > /dev/null 2>&1
 rm data/test-plots/*  > /dev/null 2>&1
+rm data/test-output/*  > /dev/null 2>&1
 rm *pdf  > /dev/null 2>&1
--- a/config_template.yml
+++ b/config_template.yml
@@ -2,10 +2,9 @@
 
 databaseLocation: /home/prado/Work/Projects/CTA_MC/gammasim-db
 
+useMongoDB: True
+
 modelFilesLocations: [
-  /home/prado/Work/Projects/CTA_MC/svn/Simulations/SimulationModel/ModelDescription/trunk/configReports,
-  /home/prado/Work/Projects/CTA_MC/svn/Simulations/SimulationModel/ModelDescription/trunk/datFiles,
-  /home/prado/Work/Projects/CTA_MC/svn/Simulations/SimulationModel/Verification/dataFromInstrumentTeams,
   /Users/ogueta/work/cta/Simulations/SimulationModel/ModelDescription/trunk/configReports,
   /Users/ogueta/work/cta/Simulations/SimulationModel/ModelDescription/trunk/datFiles,
   /Users/ogueta/work/cta/Simulations/SimulationModel/Verification/dataFromInstrumentTeams
--- a/config_template.yml
+++ b/config_template.yml
@@ -3,6 +3,7 @@
 databaseLocation: /home/prado/Work/Projects/CTA_MC/gammasim-db
 
 useMongoDB: True
+mongoDBConfigFile: ./dbDetails.yml
 
 modelFilesLocations: [
   /Users/ogueta/work/cta/Simulations/SimulationModel/ModelDescription/trunk/configReports,
--- a/environment.yml
+++ b/environment.yml
@@ -9,6 +9,7 @@ dependencies:
   - pytest
   - pytest-cov
   - scipy
+  - pymongo
 
 # cheatsheet
 
--- a/simtools/config.py
+++ b/simtools/config.py
@@ -76,7 +76,7 @@ def get(par):
 
 def getConfigArg(name, value):
     '''
-    Collect a config parameter if value is None. To be used to receive input arguments in classes.
+    Get a config parameter if value is None. To be used to receive input arguments in classes.
 
     Parameters
     ----------
--- a/simtools/config.py
+++ b/simtools/config.py
@@ -4,7 +4,7 @@ from pathlib import Path
 import logging
 import yaml
 
-__all__ = ['loadConfig', 'get', 'findFile']
+__all__ = ['loadConfig', 'get', 'findFile', 'change']
 
 logger = logging.getLogger(__name__)
 
@@ -50,6 +50,12 @@ def loadConfig(fileName=None):
 
     with open(thisFileName, 'r') as stream:
         config = yaml.load(stream, Loader=yaml.FullLoader)
+
+    # Running over the parameters set for change
+    if 'CONFIG_CHANGED_PARS' in globals():
+        for par, value in CONFIG_CHANGED_PARS.items():
+            config[par] = value
+
     return config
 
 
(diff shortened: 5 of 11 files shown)


orelgueta:
Oh, now I see that also the DB changes are included there unfortunately. I thought that if I change from one branch to the other the changes shouldn't be committed to this one...
This might break the master a bit if we merge. I think the only thing you would need is "useMongoDB: False" in your config.yml to work with the old yaml DB.

RaulRPrado:
Ok, this will take a while to review and test. 
I'll try to merge the master in here and make it work before we merge the PR.

orelgueta:
No problem, take your time, not urgent.
I should note though that to use the DB we need to change the naming system to the one we discussed before (South-LST-1). So until we change everywhere to the naming system, the two DBs are incompatible. Again, not urgent cause it's just a flag in the config file.

RaulRPrado:
@orelgueta, what/where is the file dbDetails.yml?

RaulRPrado:
Ok, I get what the file is now. What about we put this info in the config file?

orelgueta:
No, for security reasons that file should not be on Github. I will transfer it to you through DESY once it's needed.
For now it should be OK to work like this. Not sure how we will do it in the future, but we don't have to worry about it now.

RaulRPrado:
Ok. I need it now, otherwise I can't test anything.

RaulRPrado:
The test_db_handler is working fine with the useMongoDB on.
I'll test and maybe change the ray-tracing part with useMongoDB off. When its working, we can merge the request and then work on the names on a new branch.
I don't want to mix the names with this PR because it might be a big edit.

orelgueta:
I agree, that's a good idea. 
Already good that you confirm the MongoDB works, that's non-trivial...

Either way, none of this was the purpose of this PR. Github included my DB changes to this commit, but I didn't want it to. This PR was only about astropy tables.

RaulRPrado:
About the astropy.units in the internal variables, like d80_cm: I don't want to use units internally. I've been using units only in input parameters so that the user doesn't need to know  about the internal unit conventions. For any internal quantity, I prefer to keep the consistency and not use the explicit units. Main reason for that is to keep it simple. This is important because we don't have any control over the units that simtel uses, so we will have to hardcode some units anyway, so it's good to have a consistent unit system.
In the end, the astropy.units are not very useful anyway, because we only have few quantities that we calculate and write it into the table. We don't actually do many calculations with them.
I'd like to remove them if that is ok.

orelgueta:
Sorry, I don't follow the argument against astropy units.

> About the astropy.units in the internal variables, like d80_cm: I don't want to use units internally. I've been using units only in input parameters so that the user doesn't need to know about the internal unit conventions. For any internal quantity, I prefer to keep the consistency and not use the explicit units. 

But defining d80_cm is using units explicitly. All I did is also add the astropy units to it, to make it clearer in the code and because it could be useful in the future. In fact, the plotting tool also adds the unit to the axes automatically this way. 
Originally I thought it would be better to remove the cm/deg completely and simply use absolute/angular instead, but I didn't want to change your functionality. Now that it works with the units, I think it's fine.

> Main reason for that is to keep it simple. 

I agree astropy units add a bit of complexity, but not such a big one.

> This is important because we don't have any control over the units that simtel uses, so we will have to hardcode some units anyway, so it's good to have a consistent unit system.

Exactly because we don't have control over sim_telarray it is useful to use astropy units. That way, if sim_telarray changes a unit, all we will have to do is do the conversion once when we read from sim_telarray and the rest of the code can stay as is.

> In the end, the astropy.units are not very useful anyway, because we only have few quantities that we calculate and write it into the table. We don't actually do many calculations with them.

That may be true now, but will definitely not be true in the future. Especially when we consider the data we read in from various sources (calibration pipeline, observation conditions, instrument teams, etc.).
Also, if we export the results and keep them in the DB for the future, we should save them with the units in ECSV (for example).

> I'd like to remove them if that is ok.

I don't care that much, but I think it would be good to come up with a good reason to remove them before we do. Especially considering it is a DPPS recommendation to use them wherever possible.

RaulRPrado:
> Sorry, I don't follow the argument against astropy units.
> 
> > About the astropy.units in the internal variables, like d80_cm: I don't want to use units internally. I've been using units only in input parameters so that the user doesn't need to know about the internal unit conventions. For any internal quantity, I prefer to keep the consistency and not use the explicit units.
> 
> But defining d80_cm is using units explicitly. All I did is also add the astropy units to it, to make it clearer in the code and because it could be useful in the future. In fact, the plotting tool also adds the unit to the axes automatically this way.
> Originally I thought it would be better to remove the cm/deg completely and simply use absolute/angular instead, but I didn't want to change your functionality. Now that it works with the units, I think it's fine.

The d80_cm and d80_deg is just because we need to differentiate between this too. And we need to keep both because the conversion is not directly, but it depends on the focal length.
I forgot the fact that the plotting tool needs the unit. That is important.

> > This is important because we don't have any control over the units that simtel uses, so we will have to hardcode some units anyway, so it's good to have a consistent unit system.
> 
> Exactly because we don't have control over sim_telarray it is useful to use astropy units. That way, if sim_telarray changes a unit, all we will have to do is do the conversion once when we read from sim_telarray and the rest of the code can stay as is.

That is a different way to see it. My approach was to keep our internal unit system "consistent" with sim_telarray, which means actually consistent case-by-case, not globally.

> > In the end, the astropy.units are not very useful anyway, because we only have few quantities that we calculate and write it into the table. We don't actually do many calculations with them.
> 
> That may be true now, but will definitely not be true in the future. Especially when we consider the data we read in from various sources (calibration pipeline, observation conditions, instrument teams, etc.).
> Also, if we export the results and keep them in the DB for the future, we should save them with the units in ECSV (for example).

Yes, I agree. It'd be important if we were exporting in ecsv. 
 
> > I'd like to remove them if that is ok.
> 
> I don't care that much, but I think it would be good to come up with a good reason to remove them before we do. Especially considering it is a DPPS recommendation to use them wherever possible.

I think the recommendation is to use it for interfaces/inputs. I remember reading that a consistent internal unit system is also fine. But maybe I'm mixing up guidelines.

Ok, let's keep the units them.

orelgueta:
> That is a different way to see it. My approach was to keep our internal unit system "consistent" with sim_telarray, which means actually consistent case-by-case, not globally.

I think that is more or less how it is implemented now, no? 
That is, I added the units after reading from sim_telarray (without any conversions) or after all calculations, just before plotting/exporting. I didn't expect us to add units everywhere and in every single step (unless necessary).

> I think the recommendation is to use it for interfaces/inputs. I remember reading that a consistent internal unit system is also fine. But maybe I'm mixing up guidelines.

No, you are right. I simply thought of reading from sim_telarray and exporting/plotting as interfaces/inputs.

RaulRPrado:
> > That is a different way to see it. My approach was to keep our internal unit system "consistent" with sim_telarray, which means actually consistent case-by-case, not globally.
> 
> I think that is more or less how it is implemented now, no?
> That is, I added the units after reading from sim_telarray (without any conversions) or after all calculations, just before plotting/exporting. I didn't expect us to add units everywhere and in every single step (unless necessary).
> 
> > I think the recommendation is to use it for interfaces/inputs. I remember reading that a consistent internal unit system is also fine. But maybe I'm mixing up guidelines.
> 
> No, you are right. I simply thought of reading from sim_telarray and exporting/plotting as interfaces/inputs.

Ok, so let's keep them and try to be careful to not over complicate it.

orelgueta:
No problem

RaulRPrado:
I actually had to set the units only later, when filling the results table, because there was an issue with the units in thisOffset.
And I already changed the format of the exporting to ecsv, because we need to also read the units when the results already exist.

orelgueta:
So you set them after where I did?

RaulRPrado:
> So you set them after where I did?

Right after, when filling the tuple.

RaulRPrado:
Everything is fine for me now. Pytest working fine too.
May I merge?

orelgueta:
Yea, all looks OK!

PR #56, opened 2020-08-11
Replacing few key words in telType by its right form

RaulRPrado:
Now the telescope name validation will work fine for all current cases (flashcam, astri ...). We might need a better solution if the number of telescope types increases too much.

### DIFF
--- a/simtools/util/names.py
+++ b/simtools/util/names.py
@@ -197,6 +197,21 @@ def validateTelescopeName(name):
         Validated name.
     '''
     telSite, telClass, telType = splitTelescopeName(name)
+    telSite = validateSiteName(telSite)
+    telClass = validateName(telClass, allTelescopeClassNames)
+    if 'flashcam' in telType:
+        telType = telType.replace('flashcam', 'FlashCam')
+    if 'nectarcam' in telType:
+        telType = telType.replace('nectarcam', 'NectarCam')
+    if '1m' in telType:
+        telType = telType.replace('1m', '1M')
+    if 'gct' in telType:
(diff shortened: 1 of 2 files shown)


simtools/util/names.py
```diff
@@ -197,6 +197,21 @@ def validateTelescopeName(name):
...
+    if 'nectarcam' in telType:
+        telType = telType.replace('nectarcam', 'NectarCam')
+    if '1m' in telType:
+        telType = telType.replace('1m', '1M')
+    if 'gct' in telType:
+        telType = telType.replace('gct', 'GCT')
+    if 'astri' in telType:
+        telType = telType.replace('astri', 'ASTRI')
+    if '-d' in '-' + telType:
+        telType = telType.replace('d', 'D')
```
orelgueta, line 213: This one is a little dangerous, as it will change any "d" into a "D". Perhaps worth changing to

`replace('-d', '-D')`

Though I agree that if names become complicated, we will have to come up with smarter solutions (or give up on this generally).

RaulRPrado: That was my first attempt, but it happens that the telType does not contains the '-' anymore. That is why I searched for '-d' in '-' + telType.
If the name is *-d, it will work fine and replace by *-D. If it is like '-bla-d', it will also work fine. The only bad case is something like *-d-dla, which will become *-D-Dla.

I think it's fine for now, since we have full control over all the telescope names, which are not many. In the future we can redo this in a better way.

orelgueta: OK, sounds good.

PR #58, opened 2020-08-12
Turn db_handler to a class (and other small things)

orelgueta:
The main difference is the db_handler is now a class. Had to make some changes to the telescope model as well, but they were quite small.
The problem, as usual, is that I can't test those changes. The db_handler tests pass (both Yaml and MongoDB), but I can't run the telescope model part. I am trying to do it on the wgs, but that turned out to be a nightmare (ran out of quota in the middle of creating the conda environment and now my conda is completely screwed up). 
@RaulRPrado, can you please run the other tests before you approve the merge?

### DIFF
--- a/simtools/db_handler.py
+++ b/simtools/db_handler.py
@@ -167,12 +167,14 @@ class DatabaseHandler:
         Tunnel process handle.
         '''
 
-        tunnelCmd = 'ssh -N -L {localport}:{mongodbServer}:{remoteport} {user}@{tunnelServer}'.format(
-            localport=localport,
-            remoteport=remoteport,
-            user=user,
-            mongodbServer=mongodbServer,
-            tunnelServer=tunnelServer
+        tunnelCmd = (
+            'ssh -N -L {localport}:{mongodbServer}:{remoteport} {user}@{tunnelServer}'.format(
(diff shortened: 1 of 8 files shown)


orelgueta:
Fixed my conda installation on the server and will run the tests myself (already found issues). Will fix and update this PR.

RaulRPrado:
Looks good. I'll review it carefully soon, but probably not today.

RaulRPrado:
Ok, everything is fine. You may merge it.

simtools/db_handler.py
```diff
@@ -20,653 +22,879 @@
...
+    'writeModelFile',
+    'getModelFile',
+    'getFileMongoDB',
+    'writeFileFromMongoToDisk',
+    'copyTelescope',
+    'deleteQuery',
+    'updateParameter',
+    'addParameter',
+    'addNewParameter'
+]
```
RaulRPrado, line 38: I think we should keep only DatabaseHandler in this list. All these functions are class methods, so they should not be here.

orelgueta: Yea, that's a remnant from before I turned it into a class. Will fix now and merge.

PR #60, opened 2020-08-17
Removing the old array db system

RaulRPrado:
This is a short PR to clean up the old db-like system to read the array layout. @orelgueta, have a look if you want and merge it, please.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,201 +1,29 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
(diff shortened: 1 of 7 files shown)


RaulRPrado:
The LICENSE change went together by accident.

RaulRPrado:
Yes, it's only a temporary solution. I thought of adding the yml file to the data directory, but decided to keep it in the module because it was only a short dict.
The point here is just not to break the corsika_config meanwhile we don't implement a long term solution.

Issue #47, opened 2020-08-04
Write application to read parameters from DB


orelgueta:
Not sure how useful it would be at the moment, but we should probably have it as an example at least.

orelgueta:
Fail to see the usefulness of this application, especially since we have other ways to browse the DB and since we plan to write full configuration files which are easy to read with a text editor. 
I will keep the current example application for now, but won't continue developing it and close this issue.

PR #64, opened 2020-08-20
Get parameter application (not final)

orelgueta:
This partly answers #47, but it's not done (and generally not very useful, it's just there for the DB example). 
However, I need to merge this PR because of a problem I created in the environment a while back...

### DIFF
--- a/applications/compare_cumulative_psf.py
+++ b/applications/compare_cumulative_psf.py
@@ -35,14 +35,13 @@ def getData(**kwargs):
 
 if __name__ == '__main__':
     sourceDistance = 12 * u.km
-    site = 'south'
     version = 'prod4'
     label = 'lst_integral'
     zenithAngle = 20 * u.deg
     offAxisAngle = [0 * u.deg]
 
     tel = TelescopeModel(
-        telescopeName='north-lst-1',
+        telescopeType='north-lst-1',
         version=version,
         label=label,
         logger=logger.name
--- a/applications/get_parameter.py
+++ b/applications/get_parameter.py
@@ -1,0 +1,73 @@
+#!/usr/bin/python3
+
+import logging
+import argparse
+from pprint import pprint
+
+from simtools import db_handler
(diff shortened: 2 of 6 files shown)


orelgueta:
I am going to merge it, so I can keep playing with the docker, and we can discuss the full PR later.

applications/compare_cumulative_psf.py
```diff
@@ -35,14 +35,13 @@ def getData(**kwargs):
...
     sourceDistance = 12 * u.km
-    site = 'south'
     version = 'prod4'
     label = 'lst_integral'
     zenithAngle = 20 * u.deg
     offAxisAngle = [0 * u.deg]
 
     tel = TelescopeModel(
-        telescopeName='north-lst-1',
+        telescopeType='north-lst-1',
```
RaulRPrado, line 44: Why did you change it to telescopeType? It shouldn't work now. I changed it to telescopeName every where.

orelgueta: A mistake that was propagated accidentally because I had to merge in order to test the Docker.
Sorry about that.

RaulRPrado: No problem. I think we should fix it directly in the master, just to prevent propagating it. Can you do that or I do?

orelgueta: I don't think I am allowed to push directly to the master, at least it was rejected last time I tried...

RaulRPrado: That is weird. I'm allowed.
I'll fix it now and check what is going on in your case.

RaulRPrado: I learned how to set this kind of restriction but there is none set for this repository. Maybe you could try again someday jut to be sure.

orelgueta: OK, weird, maybe I misunderstood a previous error. Will try again in the future.

Issue #66, opened 2020-09-03
Define "current" version of model in db_handler


orelgueta:
Allow providing "current"/"latest" as the version to the db_handler. The "current" can be the latest official tag (e.g., prod5) and the latest can be the newest version of each parameter.

orelgueta:
@RaulRPrado, I did this just before leaving and it was working well I think. Can you confirm that you didn't run into additional problems and that I can close this issue?

orelgueta:
Fixed a long time ago with #70

PR #70, opened 2020-09-17
Add "Current" and "Latest" labels of the model

orelgueta:
This PR should answer #66. I essentially added a metadata collection to our DB to keep the information of what is the "Current" and what is the "Latest" Telescope Model tag. Currently the two are the same (Prod5), but in the future I expect the "latest" to include the latest changes, while the "Current" should be actual productions. Once construction is over, this will be replaced by a more sophisticated system of course.
I suggest that the default from now on is "Current". This is what I set it in "TelescopeModel".

### DIFF
--- a/simtools/db_handler.py
+++ b/simtools/db_handler.py
@@ -219,11 +219,15 @@ class DatabaseHandler:
         dict containing the parameters
         '''
 
+        _version = version
+        if version in ['Current', 'Latest']:
+            _version = self._getTaggedVersion(DatabaseHandler.DB_CTA_SIMULATION_MODEL, version)
+
         if cfg.get('useMongoDB'):
             _pars = self._getModelParametersMongoDB(
                 DatabaseHandler.DB_CTA_SIMULATION_MODEL,
                 telescopeName,
-                version,
+                _version,
                 runLocation,
                 onlyApplicable
             )
@@ -384,9 +388,13 @@ class DatabaseHandler:
         collection = DatabaseHandler.dbClient[dbName]['telescopes']
         _parameters = dict()
 
+        _version = version
+        if version in ['Current', 'Latest']:
+            _version = self._getTaggedVersion(dbName, version)
+
         query = {
             'Telescope': telescopeName,
-            'Version': version,
+            'Version': _version,
         }
         if onlyApplicable:
             query['Applicable'] = onlyApplicable
@@ -547,9 +555,13 @@ class DatabaseHandler:
         collection = DatabaseHandler.dbClient[dbName].sites
         _parameters = dict()
 
+        _version = version
+        if version in ['Current', 'Latest']:
+            _version = self._getTaggedVersion(dbName, version)
+
         query = {
             'Site': site,
-            'Version': version,
+            'Version': _version,
         }
         if onlyApplicable:
             query['Applicable'] = onlyApplicable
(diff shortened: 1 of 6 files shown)


RaulRPrado (commented):
Looks all fine.

I will change the default model in the application and in some other tests (if needed) and then you can merge it.

RaulRPrado:
Actually I'll not do that.

In the tests, it should stay prod4 and the applications are being changed in my docs branch.

RaulRPrado:
You may merge it!

orelgueta:
> In the tests, it should stay prod4 and the applications are being changed in my docs branch.

Why should it stay prod4? Why not the stable tag? I assume the tests will evolve with the tags as well and should be up-to-date with the model no? 
And what about prod5 instead of prod4?

RaulRPrado:
> > In the tests, it should stay prod4 and the applications are being changed in my docs branch.
> 
> Why should it stay prod4? Why not the stable tag? I assume the tests will evolve with the tags as well and should be up-to-date with the model no?
> And what about prod5 instead of prod4?

These tests should not test the model, but only the code. We should pick a model version which we know that will work so we can effectively test the code.
We might want to test the model itself somehow, but this should be done separately.

I assume it's going to happen often that the Latest model will be broken. In this situation, we should still be able to test any change on the code independently.

That is my position. What do you think?

orelgueta:
I agree completely, that's why I created two "defaults", "Current" and "Latest". The latest may be broken every now and then, but the "Current" should be stable, working and point only to production tags like prod4, prod5, etc. I think our tests should work for any stable model and not stay behind, checking only a specific old model like prod4.

RaulRPrado:
> I agree completely, that's why I created two "defaults", "Current" and "Latest". The latest may be broken every now and then, but the "Current" should be stable, working and point only to production tags like prod4, prod5, etc. I think our tests should work for any stable model and not stay behind, checking only a specific old model like prod4.

Ok ok. This means that the Current version will always be tested and stable. In this case make sense to use the Current one for the testing. I'll do that

RaulRPrado:
I did the changes on the tests.
The application I'll keep as they are by now because I'll change it on the other branch. If I change then here, I'll be creating conflicts  already.

orelgueta:
Thanks! I changed the "current" to "Current" cause that's how it appears in the DB. Maybe in the future I will implement case insensitive search, but for now it's easier this way.
I will merge now then.

RaulRPrado:
> Thanks! I changed the "current" to "Current" cause that's how it appears in the DB. Maybe in the future I will implement case insensitive search, but for now it's easier this way.
> I will merge now then.

I used "current" because the validateModelVersion do the job of converting it to "Current". It's a similar system as for the telescope names.

orelgueta:
Don't we have to add "current" to the "Current" list then?

https://github.com/gammasim/gammasim-tools/blob/d5bea7faac60d0bae3db3ef815439525db1aeeea/simtools/util/names.py#L332

RaulRPrado:
No, different capitalizations are automatically accepted. That is why we can write all the alternative names in low case.

orelgueta:
Ahh, OK now I see, very nice!
Should I change it back or is it OK to leave it like this?

RaulRPrado:
No, don't worry. It's not that important.

orelgueta:
OK, thanks for the review!

PR #110, opened 2021-03-31
ArrayModel to produce array config files 

RaulRPrado:
A first version of the array_model module was implemented.
An application called produce_array_config.py is provided as an example of usage of the ArrayModel class.

The sim_telarray config files produced were tested (for the given example) with sim_telarray using an external script. This kind of test will be implemented here in a future PR.


simtools/model/array_model.py
```diff
@@ -1,19 +1,473 @@
...
+                    msg = (
+                        'Key {} was not found in arrayConfigData '.format(key)
+                        + '- impossible to build array model'
+                    )
+                    self._logger.error(msg)
+                    raise InvalidArrayConfigData(msg)
+
+        runOverPars(['site', 'layoutName', 'default'], arrayConfigData)
+        runOverPars(['LST', 'MST'], arrayConfigData, parent='default')
+        if names.validateSiteName(arrayConfigData['site']) == 'South':
```
GernotMaier, line 169: I don't think we should hard wire the telescope types for North and South into the code. Please check if there isn't another good way to do this.

simtools/model/array_model.py
```diff
@@ -1,19 +1,473 @@
...
+                    )
+                    telModel = copy(tel)
+                    break
+
+            self._telescopeModel.append(telModel)
+
+        # Checking whether the size of the telescope list and the layout match
+        if len(self._telescopeModel) != len(self.layout):
+            self._logger.warning(
+                'Size of telModel does not match size of layout - something is wrong!'
```
GernotMaier, line 241: Still trying to understand what a telModel and a layout is in this language (maybe from this comment you can see that the error message is too cryptic?).

The comment to the class ArrayModel (line 22) says: "It contains the list of TelescopeModel's and an LayoutArray". So with size of telModel, you mean the length of the list of telescope models? Or actually the lenght of the telescope model? I would understand the first, but not the second point.

simtools/model/array_model.py
```diff
@@ -1,19 +1,473 @@
...
+                + '% ARRAY CONFIGURATION FILE\n'
+                + '% Site: {}\n'.format(self.site)
+                + '% LayoutName: {}\n'.format(self.layoutName)
+                + '% ModelVersion: {}\n'.format(self.modelVersion)
+                + ('% Label: {}\n'.format(self.label) if self.label is not None else '')
+                + '%{}\n\n'.format(50 * '=')
+            )
+            file.write(header)
+
+            # Be carefull with the formating - simtel is sensitive
```
GernotMaier, line 367: Super pedantic: do we call sim_telarray now always simtel?

GernotMaier (commented):
As a general comment on reviewing, as this is one of the first pull requests I review: on what level of detail do we expect the review? I am certainly not familiar enough with all the details of gammasim-tools to do a line-by-line review, also I don't know python enough to give you good feedback on the language level.

As a very general point - something we have discussed before but I think we should do again:
- is the code generic enough? I read a lot of 'SST', 'LST', and very specific names. This might hurt on the long term if CTA evolves or if we find other applications for gammasim-tools.
- sim_telarray dependence: I wonder if we should review the current implementation of putting sim_telarray-specific code into the model classes, or if it would be better to completely separate this and have something like a sim_telarray-writer class. On the long term, this might help if we want to use other simulation codes.

applications/produce_array_config.py
```diff
@@ -0,0 +1,133 @@
...
+    It must contain entries for 'LST' and 'MST' (and 'SST' in case of South site).
+
+    As optional data, specific telescope models can be assigned to individual telescopes. \
+    This is done by the entries with the name of the telescope (as used by the layout \
+    definition, ex. L-01, M-05, S-10).
+
+    Each telescope model can be set in two ways.
+
+    a) A single str with the name of telescope model.
+    Ex. 'M-05': 'NectarCan-D'
```
GernotMaier, line 30: Should be 'NectarCam-D'? What does does the 'D' stand for?

applications/produce_array_config.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/python3
+
+'''
+    Summary
+    -------
+    This application is a simple example of how to produce sim_telarray config \
```
GernotMaier, line 6: Remove 'simple' - what is 'simple' for you might be complex for others.

applications/produce_array_config.py
```diff
@@ -0,0 +1,133 @@
...
+    The required entries in the array_config file are:
+
+    'site': South or North
+
+    'layoutName': name of a valid layout array.
+
+    'modelVersion': name of a valid model version.
+
+    'default': telescope model names to be assigned to each telescope size by default. \
+    It must contain entries for 'LST' and 'MST' (and 'SST' in case of South site).
```
GernotMaier, line 21: Not sure if I understand correctly: 'it must contain entries for 'LST'' --> does this mean that arrays without LSTs are not possible to define?

applications/produce_array_config.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Example
+    -------
+    North - Prod5, simple example
+
+    Runtime < 1 min.
+
+    .. code-block:: console
+
+        python applications/produce_array_config.py --label test --array_config \
```
GernotMaier, line 73: See issue #111

applications/produce_array_config.py
```diff
@@ -0,0 +1,133 @@
...
+import argparse
+
+import simtools.util.general as gen
+from simtools.model.array_model import ArrayModel
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        description=(
```
GernotMaier, line 90: Is this description correct? 'calculate PSF'?

RaulRPrado:
> As a general comment on reviewing, as this is one of the first pull requests I review: on what level of detail do we expect the review? I am certainly not familiar enough with all the details of gammasim-tools to do a line-by-line review, also I don't know python enough to give you good feedback on the language level.

Ideally the review should be as deep as possible. But of course we cannot do that and need to find a good balance in order to optimize our time use. I think the most important items to review are: testing the applications, overview of the code to check for obvious problems and reviewing the application documentation. 

> As a very general point - something we have discussed before but I think we should do again:
> 
> * is the code generic enough? I read a lot of 'SST', 'LST', and very specific names. This might hurt on the long term if CTA evolves or if we find other applications for gammasim-tools.

This is something to be discussed in a call again. The code is being designed to be as modular and flexible as possible. That does not mean that it is completely generic. I honestly don't think it is realistic to expect a fully generic software. The goal should be to make it easily expandable, but also efficient and usable for our needs now (and soon).

> * sim_telarray dependence: I wonder if we should review the current implementation of putting sim_telarray-specific code into the model classes, or if it would be better to completely separate this and have something like a sim_telarray-writer class. On the long term, this might help if we want to use other simulation codes.

Do you consider the name of the parameters as sim_telarray specific code?
Apart from the name of the parameters, I'm trying to make specific signatures for sim_telarray related functions. We have, for example, exportSimtelConfigFile(). In the future we can expand by writing another function exportOtherSoftwareConfigFile().

We could add modules in between (interfaces) to avoid that. However, I think that is an unnecessary over complication, given the size of our software.

simtools/model/array_model.py
RaulRPrado: I mean the length of the list of telescope models does not match the number of telescopes in the LayoutArray object.
I expanded the comment to make it clearer.

simtools/model/array_model.py
RaulRPrado: We have been doing it in some contexts. If you say that this is a problem, we can change that.

applications/produce_array_config.py
RaulRPrado: Fixed!

D stand for "design". This comes from the convention we came up with for the naming of the telescope models.
Real telescopes will have single numbers (e.g. LST-1) and design models should preferably contain a D (e.g LST-D234).

applications/produce_array_config.py
RaulRPrado: Done!

applications/produce_array_config.py
RaulRPrado: The 'default' model is required.

The other option is checking whether all the telescopes were set individually and allow for that. I don't think this will be very useful in practice, but we could do that.

applications/produce_array_config.py
```diff
@@ -0,0 +1,133 @@
...
+    North - Prod5, simple example
+
+    Runtime < 1 min.
+
+    .. code-block:: console
+
+        python applications/produce_array_config.py --label test --array_config \
+        data/test-data/arrayConfigTest.yml -v DEBUG
+
+    All the model files can be found in simtools-output/test/model/
```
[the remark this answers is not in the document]
RaulRPrado: I changed it to 'produced model files'.

applications/produce_array_config.py
RaulRPrado: No, it is a leftover from the 'template' that I used, It is fixed now.

simtools/model/array_model.py
RaulRPrado: As I replied on another comment, we could actually check whether all the telescopes listed in the layout have a tel model given individually, and if not, we search for the default and raise an error when it is not there.

In this way, the default tel models won't be required anymore and we don't need to know which tel sizes apply for each site.

I will implement that later. It seems like a better option.

RaulRPrado: Done in https://github.com/gammasim/gammasim-tools/pull/110/commits/1bc0416bc8ec6114b8a95222dd14badbdf9a8ddc

applications/produce_array_config.py
RaulRPrado: Fixed in https://github.com/gammasim/gammasim-tools/pull/110/commits/1bc0416bc8ec6114b8a95222dd14badbdf9a8ddc

simtools/model/array_model.py
GernotMaier: OK!

simtools/model/array_model.py
GernotMaier: Huge problem (no, just kidding - but be prepared the KB might complain at some point)

applications/produce_array_config.py
GernotMaier: OK - forgot about the 'D' for design. All good!

GernotMaier:
Should discuss this, even if it adds another module: separating out the sim_telarray specific exporters would make sense in my opinion.

But also clear that we have an extremely strong dependency on sim_telarray. But let’s think hard if and where we can reduce it.

> On 7. Apr 2021, at 11:03, RaulRPrado ***@***.***> wrote:
> 
> 
> As a general comment on reviewing, as this is one of the first pull requests I review: on what level of detail do we expect the review? I am certainly not familiar enough with all the details of gammasim-tools to do a line-by-line review, also I don't know python enough to give you good feedback on the language level.
> 
> Ideally the review should be as deep as possible. But of course we cannot do that and need to find a good balance in order to optimize our time use. I think the most important items to review are: testing the applications, overview of the code to check for obvious problems and reviewing the application documentation.
> 
> As a very general point - something we have discussed before but I think we should do again:
> 
> 	• is the code generic enough? I read a lot of 'SST', 'LST', and very specific names. This might hurt on the long term if CTA evolves or if we find other applications for gammasim-tools.
> This is something to be discussed in a call again. The code is being designed to be as modular and flexible as possible. That does not mean that it is completely generic. I honestly don't think it is realistic to expect a fully generic software. The goal should be to make it easily expandable, but also efficient and usable for our needs now (and soon).
> 
> 	• sim_telarray dependence: I wonder if we should review the current implementation of putting sim_telarray-specific code into the model classes, or if it would be better to completely separate this and have something like a sim_telarray-writer class. On the long term, this might help if we want to use other simulation codes.
> Do you consider the name of the parameters as sim_telarray specific code?
> Apart from the name of the parameters, I'm trying to make specific signatures for sim_telarray related functions. We have, for example, exportSimtelConfigFile(). In the future we can expand by writing another function exportOtherSoftwareConfigFile().
> 
> We could add modules in between (interfaces) to avoid that. However, I think that is an unnecessary over complication, given the size of our software.
> 
> —
> You are receiving this because you were mentioned.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

simtools/model/array_model.py
```diff
@@ -1,19 +1,485 @@
...
+            k: v for (k, v) in arrayConfigData.items()
+            if k not in ['site', 'layoutName', 'modelVersion']
+        }
+    # End of _loadArrayData
+
+    def _validateArrayData(self, arrayConfigData):
+        '''
+        Validate arrayData by checking the existence of the relevant keys.
+        Searcing for the keys: 'site', 'array' and 'default'
+        And the sub-keys in 'default': 'LST', 'MST' and 'SST', in case of
```
GernotMaier, line 150: Is that comment still relevant?

GernotMaier:
On the definition of the array model as described in the header of produce_array_config.py: is there a possibility to disable any default settings and that each telescope model for each telescope has to be set explicitly? There are use cases where we want to do this, but at this point the system will fall back (unintentionally) to defaults if a telescope has been forgotten. Maybe just an option like 'setexplicitelyallModelParameters' (find a better name).

data/test-data/arrayConfigTest.yml
```diff
@@ -0,0 +1,10 @@
+site: North
+layoutName: Prod5
+modelVersion: Prod5
+default:
+    LST: '1'
```
GernotMaier, line 5: Took me a while to understand that the '1' is 'LST1' and not a numerical value (e.g., there should be 1 LST). Need to discuss names.

simtools/model/array_model.py
```diff
@@ -1,19 +1,485 @@
...
+            '''
+            Parameters
+            ----------
+            data: dict or str
+                Piece of the arrayConfigData for one specific telescope.
+            '''
+            if isinstance(data, dict):
+                # Case 0: data is dict
+                if 'name' not in data.keys():
+                    msg = 'ArrayConfig has no name for a telescope'
```
GernotMaier, line 283: What does this error message mean? That the telName is not found in the list of telescope models given in ArrayConfig? Don't understand 'no name'

simtools/model/array_model.py
```diff
@@ -1,19 +1,485 @@
...
+                value = self._siteParameters[par]['Value']
+                file.write(tab + '{} = {}\n'.format(par, value))
+            file.write('\n')
+
+            # Writing common parameters
+            file.write(tab + '% Common parameters\n')
+            self._writeCommonParameters(file)
+            file.write('\n')
+
+            # Maximum telescopes
```
GernotMaier, line 408: What is 'maximum telescope'? The last telescope in the list?

simtools/model/array_model.py
```diff
@@ -1,19 +1,485 @@
...
+            file.write(tab + 'maximum_telescopes = {}\n\n'.format(self.numberOfTelescopes))
+
+            # Default telescope - 0th tel in telescope list
+            telConfigFile = (
+                self._telescopeModel[0].getConfigFile(noExport=True).name
+            )
+            file.write('# include <{}>\n\n'.format(telConfigFile))
+
+            # Looping over telescopes - from 1 to ...
```
GernotMaier, line 417: to ...? Over all telescopes in the list? What is '...'?

simtools/model/array_model.py
```diff
@@ -1,19 +1,485 @@
...
+                telConfigFile = telModel.getConfigFile(noExport=True).name
+                file.write('%{}\n'.format(self.layout[count].name))
+                file.write('#elif TELESCOPE == {}\n\n'.format(count + 1))
+                file.write('# include <{}>\n\n'.format(telConfigFile))
+            file.write('#endif \n\n')
+    # END exportSimtelArrayConfigFile
+
+    def _writeCommonParameters(self, file):
+        # Common parameters taken from CTA-PROD4-common.cfg
+        # TODO: Store these somewhere else
```
GernotMaier, line 428: They should go into the Database, as some of them might change with time.

data/test-data/test_collect_dict_data.yml
```diff
@@ -0,0 +1,5 @@
+k3: {
```
GernotMaier, line 1: I don't understand this file...some test data for what? Would it make sense to add a comment to the yml file?

simtools/tests/test_corsika_config.py
```diff
@@ -3,85 +3,88 @@
 import logging
```
GernotMaier, line 3: I guess this is a skeleton for now. Fine!

simtools/tests/test_general.py
```diff
@@ -3,7 +3,8 @@
 import logging
```
GernotMaier, line 3: Fails for me:
```
# pytest simtools/tests/test_general.py
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.9.2, pytest-6.2.3, py-1.10.0, pluggy-0.13.1 -- /root/miniconda3/envs/gammasim-tools-dev/bin/python
cachedir: .pytest_cache
rootdir: /workdir/external/gammasim-tools, configfile: setup.cfg
plugins: cov-2.11.1
collected 2 items

simtools/tests/test_general.py::test_collect_args PASSED                                                                                                                              [ 50%]
simtools/tests/test_general.py::test_collect_dict_data FAILED                                                                                                                         [100%]

========================================================================================= FAILURES ==========================================================================================
__________________________________________________________________________________ test_collect_dict_data ___________________________________________________________________________________

    def test_collect_dict_data():
        inDict = {
            'k1': 2,
            'k2': 'bla'
        }
        inYaml = io.getTestDataFile('test_collect_dict_data.yml')

        d1 = collectDataFromYamlOrDict(None, inDict)
        assert 'k2' in d1.keys()
        assert d1['k1'] == 2

        d2 = collectDataFromYamlOrDict(inYaml, None)
        assert 'k3' in d2.keys()
        assert d2['k4'] == ['bla', 2]

>       d3 = collectDataFromYamlOrDict(inYaml, inDict)

simtools/tests/test_general.py:52:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

inYaml = PosixPath('data/test-data/test_collect_dict_data.yml'), inDict = {'k1': 2, 'k2': 'bla'}

    def collectDataFromYamlOrDict(inYaml, inDict):
        '''
        Collect input data that can be given either as a dict
        or as a yaml file.

        Parameters
        ----------
        inYaml: str
            Name of the Yaml file.
        inDict: dict
            Data as dict.

        Returns
        -------
        data: dict
            Data as dict.
        '''
        if inYaml is not None:
            if inDict is not None:
>               logger.warning('Both inDict inYaml were given - inYaml will be used')
E               NameError: name 'logger' is not defined

simtools/util/general.py:213: NameError
================================================================================== short test summary info ==================================================================================
FAILED simtools/tests/test_general.py::test_collect_dict_data - NameError: name 'logger' is not defined
================================================================================ 1 failed, 1 passed in 0.32s ================================================================================
```

simtools/util/names.py
```diff
@@ -343,7 +344,7 @@ def convertTelescopeNameToYaml(name):
     'Trigger': ['trigger']
 }
 
-allArrayNames = {
+allLayoutArrayNames = {
```
GernotMaier, line 347: Need to discuss if we want to have all these hardwired lists.

GernotMaier (approved):
Great work, thanks! Most comments are minor or beyond these pull request for general discussions. Please make sure that we keep track of them - I am otherwise approve the merge (maybe fix the test_general issue).

RaulRPrado:
> Should discuss this, even if it adds another module: separating out the sim_telarray specific exporters would make sense in my opinion. But also clear that we have an extremely strong dependency on sim_telarray. But let’s think hard if and where we can reduce it.

Yes, that is right. I created an issue and will work on this later.

simtools/model/array_model.py
RaulRPrado: It is not. Fixed!

RaulRPrado:
> On the definition of the array model as described in the header of produce_array_config.py: is there a possibility to disable any default settings and that each telescope model for each telescope has to be set explicitly? There are use cases where we want to do this, but at this point the system will fall back (unintentionally) to defaults if a telescope has been forgotten. Maybe just an option like 'setexplicitelyallModelParameters' (find a better name).

Since yesterday, you can set a telescope model for each telescope in the arrayConfigFile, you don't need to give a default.
You can then set a base telescope model and change any parameter you want.

If you mean setting a completely new telescope model, ignoring the base model, no, that is not possible.

You can play around with a completely new telescope model at a lower level, using the TelescopeModel. But at the level of the array_model, we always require a base model.
Note that creating a new telescope model from the grounds is not supposed to be a simple task and should never be done at this level. The ideal process would be creating the new model using a telescope_model module, uploading it to the DB and then, at the array_model level, building it from the DB.
We intentionally don't want to give this much freedom at the high level stage of the array_module.

simtools/util/names.py
RaulRPrado: Probably not. This should go into a global config file. Keeping these names here was a temporary solution that is very easy to fix.

simtools/tests/test_general.py
RaulRPrado: Fixed!

PR #117, opened 2021-04-16
shower_simulator and corsika_runner

RaulRPrado:
This PR includes a new version of corsika_config, first version of corsika_runner and shower_simulator and an application sim_showers_for_trigger_rates as a practical example.

@GernotMaier and @orelgueta, I suggest we have a call before you review this PR.

*corsika_config* is responsible mainly for creating and managing the CORSIKA input file, for a given set of parameters. The input parameters are divided into two groups. The first are called "user parameters", they should be set when instancing the CorsikaConfig class and they include the parameters that are expected to change in each simulation set or run. The user parameters have to be passed to corsika_config when instancing as a dict (corsikaConfigData) or as a yaml file (corsikaConfigFile). The information about the user parameters needed for validation (type, unit, len etc) is stored in the corsika_parameter.yml file.

The remaining parameters are expected to be fixed, and these are stored with values in the corsika_parameters.yml file.

The default corsika_parameter file can be found in data/corsika/corsika_parameters.yml. However, at the corsika_config level, a specific corsika_parameter file can be given as argument as well. Although this is only expected to be used for testing or very specific studies.

*corsika_runner* is responsible for producing and managing the scripts to be used to run CORSIKA simulations. It uses corsika_config to manage the input file. Its configuration is very similar to the corsika_config one (corsikaConfigData/FIle) but it also includes a corsikaDataDirectory entry, which is the directory where the CORSIKA data will be stored.

*shower_simulator* is a higher level module which is meant to deal with shower simulations (in oposition to the array_simulator, still to be implemented, which will deal with detector simulations). shower_simulator interfaces with the software-specific simulators (only corsika_runner at the moment).  It is responsible for running (locally or by submitting to a batch system) the scripts created by corsika_runner.

The application *sim_showers_for_trigger_rates* is a simple example of using shower_simulator. It produces simulations of CR initiated showers with regular arrays (single or 4 telescopes) for the trigger rates calculations. This application was tested at the DESY farm.

An extra entry was added to the config.yml file, called extraCommands. Please, see config_template.yml and update your local config.yml file.


data/corsika/corsika_parameters.yml
```diff
@@ -0,0 +1,105 @@
...
+    names: ['PHOTON']
+  ELECTRON:
+    number: 3
+    names: ['ELECTRONS']
+  PROTON:
+    number: 14
+    names: ['PROTONS']
+  HELIUM:
+    number: 402
+    names: []
```
GernotMaier, line 52: Would add here for consistency 'HELIUM"

data/corsika/corsika_parameters.yml
```diff
@@ -0,0 +1,105 @@
...
+
+SITE_PARAMETERS:
+  South:
+    OBSLEV: [2150.e2]
```
GernotMaier, line 56: Why here no units?

data/corsika/corsika_parameters.yml
```diff
@@ -0,0 +1,105 @@
...
+
+IACT_TUNING_PARAMETERS:
+  IACT: [
+    ['SPLIT_AUTO', '15M'],
+    ['IO_BUFFER', '800MB'],
+    ['MAX_BUNCHES', '1000000']
+  ]
+
+# Telescopes info
+TELESCOPE_Z: # Z coordinate in m
```
GernotMaier, line 96: Are you sure you want to use units of 'm' here? The CORSIKA input is in 'cm'

data/layout/telescope_positions-North-1LST.ecsv
```diff
@@ -52,4 +52,4 @@
 #       value: 3.0
 # schema: astropy-2.0
 telescope_name pos_x pos_y pos_z lat lon utm_east utm_north alt
-L-01 0.0 0.0 0.0 28.762166100000005 -17.8920302 217609.2270142641 3185067.278324084 2142.0
+L-01 0.0 0.0 0.0 28.762158000000007 -17.891571 217654.06411838497 3185065.2899643113 2142.0
```
GernotMaier, line 55: Not really relevant, as this is only an example. But L-01 position is different to what the CTAO list says:

https://gitlab.cta-observatory.org/cta-science/cta-science/-/blob/master/Array%20Element%20Positions/CTAN_ArrayElements_Positions.md

data/layout/telescope_positions-North-1SST.ecsv
```diff
@@ -52,4 +52,4 @@
 #       value: 3.0
 # schema: astropy-2.0
 telescope_name pos_x pos_y pos_z lat lon utm_east utm_north alt
-S-01 0.0 0.0 0.0 28.762166100000005 -17.8920302 217609.2270142641 3185067.278324084 2154.75
+S-01 0.0 0.0 0.0 28.762158000000007 -17.891571 217654.06411838497 3185065.2899643113 2154.75
```
GernotMaier, line 55: Having a small-sized telescopes in the north will raise questions.

data/layout/telescope_positions-North-4MST.ecsv
```diff
@@ -52,7 +52,7 @@
 #       value: 3.0
 # schema: astropy-2.0
 telescope_name pos_x pos_y pos_z lat lon utm_east utm_north alt
```
GernotMaier, line 54: Is the altitude correct? center_alt is at 2150m, and this is probably as well the altitude we put into corsika. So all telescopes should at or above center_alt.

data/test-data/corsikaConfigTest.yml
```diff
@@ -0,0 +1,44 @@
+# RUn number
```
GernotMaier, line 1: typo

data/test-data/corsikaConfigTest.yml
```diff
@@ -0,0 +1,44 @@
...
+# Energy range - [emin, emax]
+erange:
+- !astropy.units.Quantity
+  unit: !astropy.units.Unit {unit: GeV}
+  value: 0.01
+- !astropy.units.Quantity
+  unit: !astropy.units.Unit {unit: GeV}
+  value: 10.0
+
+# Energy slope
```
GernotMaier, line 19: is 'energy slope' the right expression? Better to use what is used in the corsika manual 'slope of  energy spectrum'

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/python3
+
+'''
+    Summary
+    -------
+    This application simulates showers to be used in trigger rate calculations.
+    Arrays with one (1MST) or four telescopes (4LST) can be used, in case of \
```
GernotMaier, line 7: Only 1 or 4 telescopes, or also 2 or 5?

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/python3
```
GernotMaier, line 1: In general, I would advocate for full names of what applications are doing (e.g., this one, I would call simulate_showers_for_trigger_rates), otherwise we might have very soon difficulties in understanding what the executables are doing (very minor comment)

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+
+    At the moment, the shower simulations are performed by CORSIKA, which requires \
+    the zstd package. Please, make sure that the command to set your zstd path is \
+    properly set by the extraCommands in config.yml.
+
+    Command line arguments
+    ----------------------
+    array (str, required)
+        Name of the array (1MST, 4LST ...).
+    site (str, required)
```
GernotMaier, line 24: All the array data files (e.g., data/layout/telescope_positions-North-1LST.ecsv) are for a given site. Why do need to give additional the site as an input parameter?

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+        Number of runs to be simulated (default=100).
+    nevents (int, optional)
+        Number of events simulated per run (default=100000).
+    zenith (float, optional)
+        Zenith angle in deg (default=20).
+    azimuth (float, optional)
+        Azimuth angle in deg (default=0).
+    output (str, optional)
+        Path of the directory to store the output simulations. By default, \
+        the standard output directory will be used.
```
GernotMaier, line 38: 'standard output directory' - that is the standard set where, in config.yml? And is it outputLocation? Maybe write this clearer.

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+        Log level to print (default=INFO).
+
+    Example
+    -------
+    Producing a set of proton showers for trigger rate simulations of LST.
+
+    .. code-block:: console
+
+        python applications/sim_showers_for_trigger_rates.py -a 4LST -s North \
+        --primary proton --nruns 100 --nevents 10000 --output {some dir for large files}
```
GernotMaier, line 52: Maybe put in brackets what a large file is in this context.

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+import simtools.io_handler as io
+import simtools.config as cfg
+import simtools.util.general as gen
+from simtools.shower_simulator import ShowerSimulator
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        description=(
```
GernotMaier, line 69: Is this description correct? Why PSF?

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+            'Calculate and plot the PSF and eff. mirror area as a function of off-axis angle '
+            'of the telescope requested.'
+        )
+    )
+    parser.add_argument(
+        '-a',
+        '--array',
+        help='Name of the array (e.g. 1MST, 4 LST ...)',
```
GernotMaier, line 77: 1MST (without space) but 4 LST (with space)?

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+    ----------------------
+    array (str, required)
+        Name of the array (1MST, 4LST ...).
+    site (str, required)
+        South or North.
+    primary (str, required)
+        Name of the primary particle (proton, helium ...).
+    nruns (int, optional)
+        Number of runs to be simulated (default=100).
+    nevents (int, optional)
```
GernotMaier, line 30: For this particular application, why does the user have to bother with runs and number of events per run? Would it make sense to introduce e.g., as goal an expected error on the derived trigger rates? I really wouldn't know what to add here as values for nruns and nevents - but I would know that I want to know the trigger rates on an accuracy of XX%

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+        'site': args.site,
+        'layoutName': args.array,
+        'runRange': [1, args.nruns + 1],
+        'nshow': args.nevents,
+        'primary': args.primary,
+        'erange': [100 * u.GeV, 300 * u.TeV],
+        'eslope': -2,
+        'zenith': args.zenith * u.deg,
+        'azimuth': args.azimuth * u.deg,
+        'viewcone': 10 * u.deg,
```
GernotMaier, line 158: Is viewcone always fixed? Does that depend on the FOV of the considered telescope type?

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+    outputDir = io.getApplicationOutputDirectory(cfg.get('outputLocation'), label)
+
+    showerConfigData = {
+        'corsikaDataDirectory': args.output,
+        'site': args.site,
+        'layoutName': args.array,
+        'runRange': [1, args.nruns + 1],
+        'nshow': args.nevents,
+        'primary': args.primary,
+        'erange': [100 * u.GeV, 300 * u.TeV],
```
GernotMaier, line 154: 100 GeV as lower energy?

applications/sim_showers_for_trigger_rates.py
```diff
@@ -0,0 +1,186 @@
...
+        'eslope': -2,
```
GernotMaier, line 155: (I think I've missed where the parameters are set; these are simply default values?)

-2 is obviously not a good slope for trigger estimation, or is there a reweighting in sim_telarray?

simtools/corsika/__init__.py
```diff
@@ -0,0 +1,3 @@
+import logging
+
+logging.basicConfig(format='%(levelname)s::%(module)s(l%(lineno)s)::%(funcName)s::%(message)s')
```
GernotMaier, line 3: (I don't understand - but it is me!)

simtools/corsika/corsika_config.py
```diff
@@ -0,0 +1,522 @@
...
+        'nrun': 1,
+        'zenith': 20 * u.deg,
+        'viewcone': 5 * u.deg,
+        'erange': [10 * u.GeV, 100 * u.TeV],
+        'eslope': -2,
+        'phi': 0 * u.deg,
+        'cscat': [10, 1500 * u.m, 0]
+    }
+
+    The remaining CORSIKA parameters can be set as a yaml file, using the argument \
```
GernotMaier, line 50: Discuss later again the difference between 'user parameters' and 'corsika parameters'. It seems the division is sometimes not entirely clear?

simtools/corsika/corsika_config.py
```diff
@@ -0,0 +1,522 @@
...
+    corsikaParametersFile. When not given, corsikaParameters will be loaded \
+    from data/corsika/corsika_parameters.yml.
```
GernotMaier, line 52: Need to discuss in general for gamma-simtools what should be done if there are no parameters or parameter files are given. Using default values and parameter files can be very useful, but sometimes 'dangerous'.

simtools/corsika/corsika_config.py
```diff
@@ -0,0 +1,522 @@
...
+
+    Attributes
+    ----------
+    site: str
+        North or South.
+    layoutName: str
+        Name of the layout.
+    layout: LayoutArray
+        Instance of LayoutArray.
```
GernotMaier, line 61: 'A LayoutArray is an Instance of a LayoutArray'...maybe find a better description? :-)

simtools/corsika/corsika_config.py
```diff
@@ -0,0 +1,522 @@
...
+            self.__class__.__name__,
+            self.site,
+            self.layoutName,
+            self.label
+        )
+        return text
+
+    def _loadCorsikaParametersFile(self, filename):
+        '''
+        Load CORSIKA parameters from a file, if given (filename not None),
```
GernotMaier, line 152: I think we are loading on both cases from a file, maybe write the commend a little bit different that this is clearer.

data/corsika/corsika_parameters.yml
RaulRPrado: But 'HELIUM' is already the main name. The other particles had the plural in the names.
BTW, this list is incomplete. I will complete it now and push it before merging.

data/corsika/corsika_parameters.yml
RaulRPrado: This info is actually being carried by the layout files. Both TELESCOPE_Z and and TELESCOPE_SPHERE_RADIUS are there. We should discuss where we want to keep this, in the corsika_parameters or in each of the layout files.

data/layout/telescope_positions-North-1LST.ecsv
RaulRPrado: Good to know. I just took the data from your telescope-layouts repo.
I will create an issue to validate the layout positions.

data/layout/telescope_positions-North-1SST.ecsv
RaulRPrado: I created all the possibilities of 1 and 4 telescope arrays, as is done in sim_telarray. Better remove the North-1SST then?

simtools/corsika/corsika_config.py
```diff
@@ -0,0 +1,522 @@
...
+                    raise InvalidCorsikaInput(msg)
+                else:
+                    valueArgsWithUnits.append(arg.to(unit).value)
+
+            return valueArgsWithUnits
+    # End of _validateAndConvertArgument
+
+    def _convertPrimaryInputAndStorePrimaryName(self, value):
+        '''
+        Convert a primary name into the right numberand store its name in self.primary attribute.
```
GernotMaier: 'right number' -> 'corsika particle ID'?

data/test-data/corsikaConfigTest.yml
RaulRPrado: Fixed.

data/test-data/corsikaConfigTest.yml
RaulRPrado: Fixed.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: We don't use anything other than 1 or 4 telescope arrays for trigger rate estimations. Actually we use only 1SST, 1MST and 4LST.
If this change in the future, we can build more layouts and change the documentation.

simtools/corsika/corsika_config.py
```diff
@@ -0,0 +1,522 @@
...
+        Generate and write seeds in the CORSIKA input file.
+
+        Parameters
+        ----------
+        file: stream
+            File where the telescope positions will be written.
+        '''
+        randomSeed = self._userParameters['PRMPAR'][0] + self._userParameters['RUNNR'][0]
+        random.seed(randomSeed)
+        corsikaSeeds = [int(random.uniform(0, 1e7)) for i in range(4)]
```
GernotMaier, line 482: This needs further clarification. Seeds are generated with a random generator?

applications/sim_showers_for_trigger_rates.py
RaulRPrado: In general, that is what we are doing. We are using long and clear names everywhere. In this case, I thought 'sim' was fine because we will have multiple applications of this kind, so this would be a clear pattern.
But I don't mind rename it. I can do it before merging if you want that.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: We need the site and the layout name to build a LayoutArray.  I don't understand the issue.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: It's not exactly the directory given in config.yml. We have a system to create the 'standard' output directories where the parent directory is outputLocation and the structure of the sub-directories follow a certain pattern.

The idea is that this pattern is basically uncoupled and can be changed in the future, if we find necessary. So, it doesn't make sense to write here exactly the path of the standard output because that may change.

This system for output directories is widely used over the package. It's expected that a familiar user will understand what that means.

By now I just replaced the text by 'standard output directory defined by config will be used'. I still have to figure out good terms to refer to the global config system and make all the documentation consistent.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: Is that needed here? You have right above the actual description of the arguments. This lines is just a simple example. Ideally all the arguments in the example would be set, but in this case that is not possible.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: Leftover from the application that I used as template. I will fix it.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: Fixed

applications/sim_showers_for_trigger_rates.py
RaulRPrado: My idea was to have default values that are reasonable and expected to be used in regular cases, but still leave the option to change it in case the user knows what he/she is doing.

I think this is fine. We could maybe write that in the description of the application to tell users that the default values are optimal choices for regular cases. 

The current default values are placeholders. I still need to perform this end-to-end simulations to figure out what the reasonable numbers are.

applications/sim_showers_for_trigger_rates.py
GernotMaier: That when the array is given, the site is known. Not a problem, but the same information is entered twice.

data/corsika/corsika_parameters.yml
GernotMaier: Ahh - didn't see that it is the plural for the others. All good.

data/layout/telescope_positions-North-1LST.ecsv
GernotMaier: OK. Not so important as I assume that this will all go into the database.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: I'm following what has been done by Konrad. You can see the file corsika-run/INPUTS_CTA-trg-test-prod3-LST4_proton, for example.
I don't actually know whether is a good idea to reduce the FoV or not. It's easy to do anyway.

The change wouldn't be very large anyway, right?

data/layout/telescope_positions-North-1SST.ecsv
GernotMaier: There will be no SSTs in the north, so that is not needed.

applications/sim_showers_for_trigger_rates.py
RaulRPrado: That is wrong, indeed. I will change to 3 GeV by now. However, different particles have different lower energy limit in CORSIKA. We have to figure out a way to take this into account.
We could have a safety system in corsika_config that changes the energy automatically in case the user has set a invalid value.

Example: the user set 1 GeV for helium but the minimum allowed is 10 GeV. Then we automatically change it to 10 GeV.

A warning should be printed, of course. But still, one could argue that is not a good idea to change the user setting on the background like this. What do you think?

applications/sim_showers_for_trigger_rates.py
RaulRPrado: These are default of the application. These are parameters that are not supposed to change when running this specific application.

Yes, there is a reweighting in sim_telarray and this can also be done when calculating the trigger rates.

simtools/corsika/__init__.py
RaulRPrado: I don't think this is needed. I will remove it.
I just copied this files from another directory.

simtools/corsika/corsika_config.py
RaulRPrado: I will add this as a topic for our call.

simtools/corsika/corsika_config.py
RaulRPrado: Yes, but we have to evaluate when it's dangerous and when it's necessary.
I don't think we should ask the user to set these corsika_parameters every time he/she wants to run an application.
The code has been designed to be flexible, meaning it's easy to change the defaults in case the user want to do so. However, it's bad for usability if we request him/her to do so every time.

simtools/corsika/corsika_config.py
RaulRPrado: Fixed

applications/sim_showers_for_trigger_rates.py
RaulRPrado: No, you need site and layoutName to define a LayoutArray.
The layoutName is '4LST', 'Prod5', ...
You can have a '4LST' array for both South and North.

Maybe the confusion comes from naming the argument 'array' and not 'array_name' or 'layout_name'.

simtools/corsika/corsika_config.py
RaulRPrado: I replaced it by a more detailed description: 'LayoutArray object containing the telescope positions.'

These docstrings are redundant many times, but that is inevitable. Some times the variable is perfectly described by the name but we still need to add a description.

simtools/corsika/corsika_config.py
RaulRPrado: That is actually written in the docstring.
I changed to '... from a given file ...' to make it clearer.

simtools/corsika/corsika_config.py
RaulRPrado: The seed for the random module is calculated deterministically. And then the CORSIKA seeds are generated with the random module. This is reproducible.

Do you think we should set the seeds directly?

simtools/corsika/corsika_config.py
GernotMaier: OK!

simtools/corsika/corsika_config.py
GernotMaier: For the prodN, we use the same run numbers several times (e.g., 180 deg and 0 deg files have the same run numbers, just a different MC az). Do I see it correctly that for the same runnumber, we get the same seeds in this case?

I would prefer to have a explicit random seed parameter - this make it clearer that we use a random generator, is easier to follow up, and does not depend on other parameters.

GernotMaier: Also note that the default random generator algorithm changed in python in the past, which completely messes up the idea of reproducibility even if you keep the seeds. 

I think we should somewhere make a note (an issue) to revisit how we want to use random generators in gammasim-tools. It is a central concept, and we might introduce a separate module where both seeds and algorithms are saved.

simtools/corsika/corsika_config.py
GernotMaier: Yes, this sounds much better. Sorry to be so pedantic with my comments - but obviously one should learn something from reading a description. Thanks!

simtools/corsika/corsika_config.py
RaulRPrado: > For the prodN, we use the same run numbers several times (e.g., 180 deg and 0 deg files have the same run numbers, just a different MC az). Do I see it correctly that for the same runnumber, we get the same seeds in this case?

Yes, but that is not a problem. We actually only need seeds that depends on the run number. If any other parameter is different (e.g. azimuth), we will not simulate the same showers.

The reproducibility issue goes beyond the python version. You can only reproduce the same showers if you use the same seeds and run CORSIKA in exactly the same system you ran it before. So using the same python version is only one of the requirements if one wants to simulate the same showers.

I created an issue about that.

applications/sim_showers_for_trigger_rates.py
GernotMaier: I think I am a bit lost here what happens:

- choosing 4LSTs + North means that the layout 'telescope_positions-North-4LST' is read (right now from a the file telescope_positions-North-4LST.ecsv, probably in future from the data base). What if I want to have another layout for 4 LSTs in the North? Are code changes necessary, or does the code builds e.g. telescope_positions-North-4LST-New.ecsv when I give array="4LST-New"?
- I think I am fine for the site parameter, as we are deriving other parameters (e.g., magnetic field) from the site parameter.

RaulRPrado: > * choosing 4LSTs + North means that the layout 'telescope_positions-North-4LST' is read (right now from a the file telescope_positions-North-4LST.ecsv, probably in future from the data base). What if I want to have another layout for 4 LSTs in the North? Are code changes necessary, or does the code builds e.g. telescope_positions-North-4LST-New.ecsv when I give array="4LST-New"?

The code doesn't build the new layout at this level. You have to use another application (or write your own [1]) to build a new layout and then come back to this application, once your new layout is pre-defined.
The name 4LST is just a name, it doesn't mean anything for the code. We could have used instead 'array-with-4-LST-to-simulate-trigger-rates'. So it's our convention naming the regular arrays like this (following the sim_telarray convention). 

The idea is that building a new layout is a 'serious' action and should be done separately from other actions. We won't be building new layouts all the time. Most of the time we will be using pre-defined layouts that have been built in the past.

[1] As it's now (using ecsv files to store the layouts), one can just take an existing layout file, change its name and change the values by hand in order to create a new layout. That would work too.

GernotMaier: (resolved, as I understand now better what is happening)

applications/sim_showers_for_trigger_rates.py
GernotMaier: Whenever you write 'large file' there is no information given. Is in 1 GB? 1 TB? What is large? Really not important, but comments should be 'maximal informational'.

applications/sim_showers_for_trigger_rates.py
GernotMaier: I've moved this into issue #123

applications/sim_showers_for_trigger_rates.py
GernotMaier: I am still trying to understand exactly: all parameters given as values in this dict are fixed and not changeable by the user?

Having a too large FOV is a waste of computing time. Why would one fix the energy range? 10 GeV is correct for LSTs, but definitely far too low for SSTs.

applications/sim_showers_for_trigger_rates.py
GernotMaier: Why is the minimum for helium set to 10 GeV?

Otherwise printing out a warning is fine.

applications/sim_showers_for_trigger_rates.py
GernotMaier: (see question earlier, so I resolve this one)

simtools/corsika/corsika_config.py
GernotMaier: (ok - still need to discuss this but should not prevent us from merging here)

simtools/corsika/corsika_config.py
GernotMaier: Moved into issue #124

applications/sim_showers_for_trigger_rates.py
RaulRPrado: > I am still trying to understand exactly: all parameters given as values in this dict are fixed and not changeable by the user?

Yes. But note that this is a specific application to produce simulations for trigger rate calculations.  

> Having a too large FOV is a waste of computing time. Why would one fix the energy range? 10 GeV is correct for LSTs, but definitely far too low for SSTs.

About the FoV, isn't better to be conservative for the trigger rate calculation? We want to simulate every event that may have any chance of triggering and there may be many of them slightly outside the nominal FoV.

Similar argument for the energy. Even low energy proton showers may contribute for the trigger rate of SSTs.

GernotMaier: Yes, it is better to be conservative. But what about the other way around? Tomorrow we upgrade telescopes with a larger FOV, will we notice that there is a hardwired parameter? My main point is really that we shouldn't have hardwired parameters.

RaulRPrado: It's not hardwired. This is an application, which means a specific tool that requires specific parameters.
We keep everything as flexible as possible along the library. The point of the application is to reduce all this flexibility to the minimum necessary for executing an specific task. 

And yes, if we upgrade our telescopes with a larger FoV, we will surely need to do some work to upgrade our software too.
And if the work on the simulation software means changing parameters at the application level, I consider this a win in terms of maintainability. 

On the other hand, if every time we want to calculate trigger rates we need to set FoV, energy range etc, I consider this a big loss in usability.

GernotMaier: Nono! I do not expect anything to be hardwired anywhere. All should be be in the configuration.

And as default configurations are provided (in the DB or through configuration files), one does not have to set the parameters each time you start it.

Whatever you hardwire now will be the reason for problems in the future.

RaulRPrado: I will add this is a topic for discussion, but I'd keep it as it is by now.
This issue actually relates to a couple of other application as well, not only this one.

applications/sim_showers_for_trigger_rates.py
orelgueta: Not sure I agree that changing the user's setting and just printing a warning is fine. I think we should never change an explicit user setting and continue running, if we think a setting is wrong we should always stop running (or allow a force mode of running).

Issue #114, opened 2021-04-08
Separating the sim_telarray specific exporters from the model modules


RaulRPrado:
A module simtel_config should be created and the functions that export sim_telarray config files in telescope_model and array_model should use it as an interface.

RaulRPrado:
Fixed in #119

PR #119, opened 2021-04-20
Separating the sim_telarray config writer from the model classes 

RaulRPrado:
This PR is mainly focus on fixing #114.
A module simtel_config_writer was created, with a class names SimtelConfigWriter.
This class is used by TelescopeModel and ArrayModel to write the sim_telarray config files. Its design is very specific for this context.

There are some data 'hardcoded' in SimtelConfigWriter, e.g. the list of parameters that should not be written.

There were a couple of TODO's in the TelescopeModel that were also fixed in this PR. They were related to the selection of parameters to write into the config file. This is needed because we added extra parameters to the DB that sim_telarray does not read.
Now, the TelescopeModel knows about all the parameters that exists in the DB and doesn't know anything about which ones sim_telarray can process or not. This information is managed by SimtelConfigWriter.
Moreover, the _parameters dict in TelescopeModel uses now the same structure that comes out of the DB. Meaning that it carries few keys for each parameters, like Value, Type, isFile etc.

A number of small improvements were made to the TelescopeModel class and its test module.


RaulRPrado:
Please, ignore the changes on the layout files.
There was a bug on the master that I fixed in the #117. Here I just checkout out the files from there to be able to run tests.

docs/source/library.rst
```diff
@@ -10,6 +10,6 @@ simtools Library
   mc_model
   ray_tracing
   camera_efficiency
-  simtel_runner
+  simtel
```
GernotMaier, line 13: KB will tell you that it is sim_telarray....

simtools/model/telescope_model.py
```diff
@@ -502,47 +515,47 @@ def exportSingleMirrorListFile(self, mirrorNumber, setFocalLengthToZero):
...
-            ))
+        # Using SimtelConfigWriter
+        self._loadSimtelConfigWriter()
+        self.simtelConfigWriter.writeSingleMirrorListFile(
+            mirrorNumber,
+            self.mirrors,
+            self._singleMirrorListFilePaths[mirrorNumber],
+            setFocalLengthToZero
+        )
     # END of exportSingleMirrorListFile
```
GernotMaier, line 535: Is this pythonic? Have to smile while seeing this comment....

simtools/model/telescope_model.py
```diff
@@ -574,7 +587,15 @@ def _loadCamera(self):
...
+    def _loadSimtelConfigWriter(self):
+        if not hasattr(self, 'simtelConfigWriter'):
+            self.simtelConfigWriter = SimtelConfigWriter(
+                site=self.site,
+                telescopeName=self.telescopeName,
+                modelVersion=self.modelVersion,
+                label=self.label
+            )
 
     def isASTRI(self):
```
GernotMaier, line 600: Why do we need isASTRI? I Think I've asked it before.

simtools/simtel/simtel_config_writer.py
```diff
@@ -0,0 +1,283 @@
...
+        Writes the sim_telarray config file for an array of telescopes.
+    writeSingleMirrorListFile(
+        mirrorNumber,
+        mirrors,
+        singleMirrorListFile,
+        setFocalLengthToZero=False
+    )
+        Writes the sim_telarray mirror list file for a single mirror.
+    '''
+    TAB = ' ' * 3
```
GernotMaier, line 28: What does this do? Three spaces?

simtools/simtel/simtel_config_writer.py
```diff
@@ -0,0 +1,283 @@
...
+    SITE_PARS = [
+        'altitude',
+        'atmospheric_transmission',
+        'ref_lat',
+        'ref_long',
+        'array_coordinates',
+        'atmospheric_profile',
```
GernotMaier, line 35: Does sim_telarray need the magnetic field?

simtools/model/telescope_model.py
RaulRPrado: This what? I don't get it. Which comment?

simtools/simtel/simtel_config_writer.py
RaulRPrado: This is the indentation that has to be used in sim_telarray config files. It's used when writing sim_telarray config files.

simtools/simtel/simtel_config_writer.py
RaulRPrado: It doesn't! That is why magnetic_field is on the list of PARS_NOT_TO_WRITE.

simtools/model/telescope_model.py
RaulRPrado: We need this to fix an incompatibility problem between the model and sim_telarray (testeff in particular).

The testeff program, used by our CameraEfficiency, does not deal with 2D camera_filters and mirror_reflectivity. So we have to change these files to an 1D version of them (averaged).

Ideally, this problem should be fixed in testeff, and then we can make our code cleaner.

docs/source/library.rst
RaulRPrado: We write 'sim_telarray' in the documentation. 'simtel' is just a term used in the code.

simtools/tests/test_simtel_config_writer.py
```diff
@@ -0,0 +1,58 @@
...
+import simtools.io_handler as io
+from simtools.simtel.simtel_config_writer import SimtelConfigWriter
+from simtools.model.telescope_model import TelescopeModel
+from simtools.layout.layout_array import LayoutArray
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+
+
+def file_has_text(file, text):
```
GernotMaier, line 15: This seems to be a very generic function, useful as for other configuration writers.

GernotMaier (approved):
None of the comments was really relevant.

docs/source/library.rst
GernotMaier: Always good to be consistent (but it really doesn't matter)

simtools/simtel/simtel_config_writer.py
GernotMaier: Ah!

simtools/tests/test_simtel_config_writer.py
RaulRPrado: Yes, you are right. I moved it to the utils.general

simtools/model/telescope_model.py
GernotMaier: I meant this one:
```
# END of exportSingleMirrorListFile 
```
Really not important, I though that the indentation was introduced so that this is not needed...

simtools/model/telescope_model.py
GernotMaier: OK, but both is handled by the optics module? And does the GCT camera will have a window where we want to use 2D efficiencies?

RaulRPrado: This is handled by the camera_efficiency module, because it's an specific issue of testeff.
At the moment only ASTRI (old ASTRI design and the new unified SST design) are affected by that.

The problem is that we have to replace the 2D files by an specific file with the average. E.g. 'transmission_astri_window_average.dat'.

If this is not fixed in sim_telarray soon, we can spend some time on writing a tool to turn a 2D file into a 1D average file.
I'd assume that this is feasable.

GernotMaier: OK - but wouldn't it the be more appropriate to call the function not isAStri() but something like 'has2DEfficiencyTables()' or so? 

In a couple of years, noone will remember what ASTRI was, as we all speak of CTA SSTs...

RaulRPrado: I already thought about this issue and I'm also not happy with the solution. It turns out it is a very specific issue. 

This is what is done in camera_efficiency:

        cameraFilterFile = self._telescopeModel.getParameter('camera_filter')
        if self._telescopeModel.isASTRI() and self._telescopeModel.isFile2D('camera_filter'):
            self._logger.warning(
                'Camera filter file is being replaced by transmission_astri_window_average.dat'
                ' because testeff does not support 2D camera filters.'
            )
            cameraFilterFile = 'transmission_astri_window_average.dat'

We replace the file by an ASTRI specific file, so in the end we can't solve the specificity of the code unless we solve the problem for real.

This means that if has2DEfficiencyTables() is True and it's not an ASTRI telescope, we still don't have a solution yet.

I'd prefer to keep it like this and open an issue requesting the real solution (either by in tesfeff or making the 2D files into 1D)

GernotMaier: OK. Could you please open an issue for this, so that we don't forget what is happening?

RaulRPrado: Done!

PR #131, opened 2021-04-26
Adding unit tests for the applications

RaulRPrado:
I added a unit test module (applications/tests/test_aplications.py) to test the applications with pytest.

The test assures that the execution of the application with a certain set of arguments does not raise any error.

Note that these are not functional nor integration tests. No physical result is being tested here.


orelgueta (approved):
Looks good!

PR #134, opened 2021-04-29
Changing the system to validate input config data

RaulRPrado:
We have a system in place to collect and validate input config data that uses kwargs and requires the information about the expected input to be stored in a class dict (called ALL_INPUTS).
This system is not ideal, and this PR intend to start changing it.

The plan is to implement a clearer system that reads a dict with the config data and uses the parameter information given from another dict. These parameters can be stored in yml files (or DB in the future).

I implemented something similar in corsika_config, and now I want to expand it and use it consistently all around the modules.

This PR only includes the implementation of the functions in util/general.py, its test module and an example of the parameter yml file. After we agree on these, I will start replacing the old system by this new one. This will clean significantly few modules and the general module.

I suggest to have a look at:
- the function validateConfigData in util/general
- the function test_validate_config_data in tests/test_general
- the example of the parameter file in data/test-data/test_parameters.yml

I tried to add enough comments to make it clear.

### DIFF
--- a/data/ray-tracing/ray-tracing_parameters.yml
+++ b/data/ray-tracing/ray-tracing_parameters.yml
@@ -1,0 +1,13 @@
+zenithAngle: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
+  default: 20
+  names: ['zenith', 'theta']
+offAxisAngle: 
+  len: null
+  unit: !astropy.units.Unit {unit: deg}
+  names: ['offaxis', 'offset']
+sourceDistance:
+  len: 1
+  unit': !astropy.units.Unit {unit: km}
+  default: 10
--- a/data/test-data/test_parameters.yml
+++ b/data/test-data/test_parameters.yml
@@ -1,0 +1,30 @@
+
+
+zenithAngle: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
+  default: 20
+  names: ['zenith', 'theta']
+# Parameters with undefined length must contain len: null
+# A single unit must be given in this case
+offAxisAngle: 
+  len: null
+  unit: !astropy.units.Unit {unit: deg}
+  names: ['offaxis', 'offset']
+
+sourceDistance:
+  len: 1
+  unit: !astropy.units.Unit {unit: km}
+  default: 10
+
+# Unit must be given as a list when len > 1
+cscat:
+  len: 3
+  unit: [null, !astropy.units.Unit {unit: m}, !astropy.units.Unit {unit: m}]
+  names: ['scat']
+
+# Names are case insensitive
+validatedName:
+  len: 1
+  unit: null
+  names: ['testname']
--- a/data/test-data/test_parameters.yml
+++ b/data/test-data/test_parameters.yml
@@ -1,6 +1,6 @@
 # This is an example of the parameter file to be used
 # to validate config data.
-# 
+#
 # Each parameter is listed by its main name as parent key.
 # A list of alternative names can be listed using the key 'names'.
 # If a config entry with one of the names listed are found (case insensitive),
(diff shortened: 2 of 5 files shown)


data/ray-tracing/ray-tracing_parameters.yml
```diff
@@ -0,0 +1,13 @@
+zenithAngle: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
+  default: 20
+  names: ['zenith', 'theta']
+offAxisAngle: 
+  len: null
```
GernotMaier, line 7: What is the meaning of the 'len' parameter? Why is it null for the offAxisAngle, but '1' for the other parameters?

GernotMaier (approved):
All good - can be merged from my point of view.

data/ray-tracing/ray-tracing_parameters.yml
RaulRPrado: The explanations can be found as comments in test_parameters.yml. I assume you found them.

GernotMaier:
No, I didn’t. Why is the length undefined for the offAxisAngle?

> On 3. May 2021, at 10:19, RaulRPrado ***@***.***> wrote:
> 
> 
> @RaulRPrado commented on this pull request.
> 
> In data/ray-tracing/ray-tracing_parameters.yml:
> 
> > @@ -0,0 +1,13 @@
> +zenithAngle: 
> +  len: 1
> +  unit: !astropy.units.Unit {unit: deg}
> +  default: 20
> +  names: ['zenith', 'theta']
> +offAxisAngle: 
> +  len: null
> 
> The explanations can be found as comments in test_parameters.yml. I assume you found them.
> 
> —
> You are receiving this because you commented.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

RaulRPrado:
# This is an example of the parameter file to be used
# to validate config data.
#
# Each parameter is listed by its main name as parent key.
# A list of alternative names can be listed using the key 'names'.
# If a config entry with one of the names listed are found (case insensitive),
# a validated config entry is set with the main name.
#
# The length of the entry must be given by the key 'len'.
# Lists with undefined length can be set by len: null.
#
# The 'unit' has to be a list of the same len as the entry.
# No units can be set by null or by not adding the 'unit' key.
#
# Config entries will be checked whether they are given
# with equivalents units or not. If so, the entry will
# be converted to the proper unit and only its value
# will be returned.
#
# A default value can be given by the key 'default'.
# Parameters without default key will raise error when missing.

Some parameters have undefined length. The offAxisAngle for ray tracing simulations is a good example. The user can give [0, 1, 2, 4] or [0, 0.2, 0.4, 0.6 ... 5]. In these cases, one have to set len: null.
Other parameters you want to specify the exact length of the parameter, then one has to give a number.

GernotMaier:
Ok - understood. I thought it is defined like the view cone parameter in CORSIKA, but this makes sense! Thanks.



> On 3. May 2021, at 10:40, RaulRPrado ***@***.***> wrote:
> 
> 
> This is an example of the parameter file to be used
> 
> to validate config data.
> 
> Each parameter is listed by its main name as parent key.
> 
> A list of alternative names can be listed using the key 'names'.
> 
> If a config entry with one of the names listed are found (case insensitive),
> 
> a validated config entry is set with the main name.
> 
> The length of the entry must be given by the key 'len'.
> 
> Lists with undefined length can be set by len: null.
> 
> The 'unit' has to be a list of the same len as the entry.
> 
> No units can be set by null or by not adding the 'unit' key.
> 
> Config entries will be checked whether they are given
> 
> with equivalents units or not. If so, the entry will
> 
> be converted to the proper unit and only its value
> 
> will be returned.
> 
> A default value can be given by the key 'default'.
> 
> Parameters without default key will raise error when missing.
> 
> Some parameters have undefined length. The offAxisAngle for ray tracing simulations is a good example. The user can give [0, 1, 2, 4] or [0, 0.2, 0.4, 0.6 ... 5]. In these cases, one have to set len: null.
> Other parameters you want to specify the exact length of the parameter, then one has to give a number.
> 
> —
> You are receiving this because you commented.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

RaulRPrado:
The large font is a mistake of formatting. I tried to edit it but it didn't work.

PR #138, opened 2021-05-04
Replacing system to validate configurable parameters

RaulRPrado:
Following up #134, I replaced the old system by the new one in all the modules.

I also changed the output of validateConfigData to a namedtuple, instead of a dict. 
Now the classes contain an attribute called config which contains the validated parameters as namedtuple.
Some classes need to manipulate these parameters, so in these cases we are turning them into private attributes.

All classes that deals with configurable parameters have now a classmethod called fromKwargs that can be used to instance by giving all the configurable parameters as kwargs, instead a dict or yaml file.

Each class that uses the system has a correspondent parameter file in data/parameters with the info about all the configurable parameters (len, unit, etc).

### DIFF
--- a/data/parameters/layout-array_parameters.yml
+++ b/data/parameters/layout-array_parameters.yml
@@ -1,0 +1,63 @@
+epsg: 
+  len: 1
+  names: []
+  default: null
+centerLongitude: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
+  names: ['longitude', 'long', 'center_longitude']
+  default: null
+centerLatitude: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
+  names: ['latitude', 'lat', 'center_latitude']
+  default: null
+centerNorthing: 
+  len: 1
+  unit: !astropy.units.Unit {unit: m}
+  names: ['northing', 'north', center_northing']
+  default: null
+centerEasting: 
+  len: 1
+  unit: !astropy.units.Unit {unit: m}
+  names: ['easting', 'east', center_easting']
+  default: null
+centerAltitude: 
+  len: 1
+  unit: !astropy.units.Unit {unit: m}
+  names: ['altitude', 'centeralt']
+  default: null
+corsikaObsLevel: 
+  len: 1
+  unit: !astropy.units.Unit {unit: m}
+  names: ['obslevel']
+  default: null
+corsikaSphereCenter: 
+  len: 3
+  unit: [!astropy.units.Unit {unit: m}, !astropy.units.Unit {unit: m}, !astropy.units.Unit {unit: m}]
+  names: ['spherecenter']
+  default: 
+    LST: !astropy.units.Quantity
+      unit: !astropy.units.Unit {unit: m}
+      value: 16.0
+    MST: !astropy.units.Quantity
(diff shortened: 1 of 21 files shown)


data/parameters/layout-array_parameters.yml
```diff
@@ -0,0 +1,63 @@
...
+  len: 1
+  unit: !astropy.units.Unit {unit: m}
+  names: ['obslevel']
+  default: null
+corsikaSphereCenter: 
+  len: 3
+  unit: [!astropy.units.Unit {unit: m}, !astropy.units.Unit {unit: m}, !astropy.units.Unit {unit: m}]
+  names: ['spherecenter']
+  default: 
+    LST: !astropy.units.Quantity
```
GernotMaier, line 40: Do we want to add the SCT here to have all CTA telescope types? I am a bit undecided, as it is not a telescope type for the early construction. But we have simulated it a few times and will continue to simulate it.

simtools/ray_tracing.py
```diff
@@ -108,38 +120,33 @@ def __init__(
...
-        if self._singleMirrorMode:
-            if self._mirrorNumbers == 'all':
+        # Loading relevant attributes in case of single mirror mode.
+        if self.config.singleMirrorMode:
+            # Recalculating source distance.
+            self._logger.debug(
+                'Single mirror mode is activate - source distance is being recalculated to 2 * flen'
+            )
+            mirFlen = self._telescopeModel.getParameterValue('mirror_focal_length')
+            self._sourceDistance = 2 * float(mirFlen) * u.cm.to(u.km)  # km
```
GernotMaier, line 141: Is that also correct for parabolic mirrors? I don't remember, but somehow triggers the thought that the definition of the light source at 2x the focal length is only used for Davies-Cotton optical design.

GernotMaier (approved):
Everything good from my side!

data/parameters/layout-array_parameters.yml
RaulRPrado: For this purpose, SCT is a MST. These are the CORSIKA sphere parameters for all the telescopes named 'M-XX'.
The CORSIKA sphere center is the same for MST and SCT, right?
And the radius would be just slighlty larger for MST, so no problem.

simtools/ray_tracing.py
RaulRPrado: That might be. This feature was written for the MST case.
What would be the source distance for parabolic mirrors then?

GernotMaier: OK. Not the most efficient way, as the SC MST is much small than the MST, but ok!

RaulRPrado: If you want to simulate an array with SCT instead of MST, you can reduce the sphere when build these layout files. No problem.

But if you want to simulate a mix array with MST and SCT, then you have to set the sphere to the larger radius.

GernotMaier: > That might be. This feature was written for the MST case.
> What would be the source distance for parabolic mirrors then?

I don't know...always worked with Davies Cotton telescopes. Isn't there an optics paper from MAGIC around? Ask someone there?

RaulRPrado: I will create an issue for that.

Issue #109, opened 2021-03-30
Separate reading parameters from writing files in the db_handler


RaulRPrado:
By now the function getModelParameters reads the list of parameters and, in the background, writes the files out.
We would better separate these two steps.
In the TelescopeModel, the files should be written out only when calling the exportConfigFile function, not whe building the list of parameters.

orelgueta:
I agree that it should be written out only when you call exportConfigFile. I am not sure how to implement it though. It means that TelescopeModel will have to keep the list of files (or rebuilt it by going over the parameters again) and provide to the configuration exporter. Then the exporter will have to call the db_handler. Probably that could be done.

Who should work on it depends on the urgency. I don't think I will be able to implement it in the next few weeks, if it can wait, then I can do it.

RaulRPrado:
> I agree that it should be written out only when you call exportConfigFile. I am not sure how to implement it though. It means that TelescopeModel will have to keep the list of files (or rebuilt it by going over the parameters again) and provide to the configuration exporter. Then the exporter will have to call the db_handler. Probably that could be done.
> 
> Who should work on it depends on the urgency. I don't think I will be able to implement it in the next few weeks, if it can wait, then I can do it.

I think the db_handler should provide a method to write a single file into a specific directory and then we can go over all parameters of TelescopeModel and write them out. The TelescopeModel is storing the parameters with the same structure as they come out of the DB, so we have the flags isFile and Applicable available.
This should all be done by the TelescopeModel, not by the SimtelConfigWriter.

I can probably work on that soon. But I wanted to give you the chance in case you have the time.

orelgueta:
Yes, this is similar to what I would like to do for the LST file missing in #107. I will try to do it at some point in the next couple of weeks, but if it is urgent and you want to do it yourself, no problem.

RaulRPrado:
Ok, I will work on this soon, while I want for the PR that I opened today.

PR #139, opened 2021-05-05
Separating getting parameters from the DB and exporting files

RaulRPrado:
Fixing #109.

The function exportModelParameters was added to DatabaseHandler and it is supposed to be used externally, like getModelParameters.
getSiteParameters only returns the parameters but it does not write them anymore. 

I also did some refactoring of db_handler and test_db_handler.


simtools/db_handler.py
```diff
@@ -219,33 +216,94 @@ def getModelParameters(
...
+                file = self._getFileMongoDB(
+                    DatabaseHandler.DB_CTA_SIMULATION_MODEL,
+                    info['Value']
+                )
+                self._writeFileFromMongoToDisk(DatabaseHandler.DB_CTA_SIMULATION_MODEL, dest, file)
+        else:
+            self._logger.debug('Exporting model files from local model file directories')
+            for par, value in parameters.items():
+
+                shallNotSearch = (
```
orelgueta, line 272: In order to check if a parameter is a file, I check either

`if any(ext in str(value) for ext in ['.dat', '.txt', '.lis']):`

or 

`if '.dat' in str(value) or '.txt' in str(value):`

I think we should make it a function in db_handler and use the same one everywhere. 
Either way, "shallNotSearch" is not the best name, I would call it "isFile".

simtools/db_handler.py
```diff
@@ -219,33 +216,94 @@ def getModelParameters(
...
+        destDir: str or Path
+            Path of the directory where the file will be written.
+        '''
+
+        destFile = Path(destDir).joinpath(fileName)
+        try:
+            file = cfg.findFile(fileName, cfg.get('modelFilesLocations'))
+        except FileNotFoundError:
+            if noFileOk:
+                self._logger.debug('File not found but noFileOk')
```
orelgueta, line 300: Add the name of the file to the message for easier debugging?

simtools/db_handler.py
```diff
@@ -590,53 +638,10 @@ def _getSiteParametersMongoDB(
             _parameters[parNow].pop('Parameter', None)
             _parameters[parNow].pop('Site', None)
             _parameters[parNow]['entryDate'] = ObjectId(post['_id']).generation_time
-            if _parameters[parNow]['File']:
```
orelgueta, line 593: Do you take care of this somewhere else? Notice that for the sites parameters' it's done separately so when you get the site parameters you also need to read the associated files.
Maybe you took care of it, It's hard to tell from the review since I only see the changes here (and finding it in the code would be hard).

simtools/model/telescope_model.py
```diff
@@ -457,21 +456,40 @@ def removeParameters(self, *args):
...
+        self._addedParameterFiles.append(parName)
         shutil.copy(filePath, self._configFileDirectory)
-        return
+
+    def exportModelFiles(self):
+        ''' Exports the model files into the config file directory. '''
+        db = db_handler.DatabaseHandler()
+
+        # Removing parameter files added manually (which are not in DB)
+        parsToExportFiles = copy(self._parameters)
```
orelgueta, line 480: This name is misleading (I think). These are essentially all the parameters, no? Then you simply remove the ones which are files added manually.
The list you give to `db.exportModelFiles` is of all of the parameters and in the `db_handler` you choose from them the file ones, right?
In that case, I would change the name of the variable to make it clearer.

orelgueta (approved):
Looks good, made just minor comments. Only the site point has the potential to break something.

simtools/db_handler.py
RaulRPrado: Done

simtools/db_handler.py
RaulRPrado: Where is this function that you use? I agree we should use the same here

RaulRPrado: I created the function _isFile and replaced the code.

simtools/db_handler.py
RaulRPrado: Yes. This is done externally. The TelescopeModel, for example, adds the site parameters to the list of files to export.

simtools/model/telescope_model.py
RaulRPrado: Yes, you are right. I renamed it to parsFromDB. It's ok, I think, because the scope of the variable is very small.

RaulRPrado:
I will merge it, but we can continue the discussions.

simtools/db_handler.py
orelgueta: I use it twice, once in 
https://github.com/gammasim/gammasim-tools/blob/9f6ab59429fbbeab8b2265660f3d1ac1ac35ebc7/simtools/db_handler.py#L931

and another time in "updateParameter", but that code is not pushed yet since I didn't finish working on it. Now that you made these changes, I will have to rewrite it a bit, but that's OK. If you change the line above, we can close the conversation and I will make sure to use the function in my next update.

RaulRPrado: What line above?
I created the _isFile function as you suggested.

orelgueta: Line 931 showed above. That one should be changed to use the new `_isFile` method.

PR #148, opened 2021-06-04
Adding array_simulator

RaulRPrado:
This PR adds the module array_simulator, which contains the class ArraySimulator.

It's a simple module, analogous to shower_simulator. It used simtel_runner_array to perform array simulator for given corsika files.

The complete set of tests will come next with the appalications to perform showers and array simulations.

### DIFF
--- a/data/parameters/array-simulator_parameters.yml
+++ b/data/parameters/array-simulator_parameters.yml
@@ -19,3 +19,7 @@ azimuthAngle:
     value: 0
     unit: !astropy.units.Unit {unit: deg}
   names: ['azimuth', 'phi']
+viewcone:
+  len: 2
+  unit: [!astropy.units.Unit {unit: deg}, !astropy.units.Unit {unit: deg}]
+  names: ['cone']
--- a/data/parameters/array-simulator_parameters.yml
+++ b/data/parameters/array-simulator_parameters.yml
@@ -1,0 +1,21 @@
+simtelDataDirectory:
+  len: 1
+  unit: null
+  default: null
+primary:
+  len: 1
+  unit: null
+zenithAngle: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
+  default: !astropy.units.Quantity
+    value: 20
+    unit: !astropy.units.Unit {unit: deg}
+  names: ['zenith', 'theta']
+azimuthAngle: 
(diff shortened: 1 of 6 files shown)


data/parameters/array-simulator_parameters.yml
```diff
@@ -0,0 +1,25 @@
+simtelDataDirectory:
+  len: 1
+  unit: null
+  default: null
+primary:
+  len: 1
+  unit: null
+zenithAngle: 
+  len: 1
+  unit: !astropy.units.Unit {unit: deg}
```
orelgueta, line 10: Are these yaml files supposed to be produced by the user? If so, can't the user just write "deg" in the unit and we complete it to full astropy syntax in the code later? It will be much more user friendly that way.

I am not sure if this was already discussed before, haven't really been keeping up, so sorry if it has.

simtools/array_simulator.py
```diff
@@ -0,0 +1,320 @@
...
+    It interfaces with simulation software-specific packages, like sim_telarray.
+
+    The configuration is set as a dict configData or a yaml \
+    file configFile. An example of configData can be found \
+    below.
+
+    .. code-block:: python
+
+    configData = {
+        'simtelDataDirectory': '.',
```
orelgueta, line 34: Is the data directory really the local one? Maybe it's better to give in the example something more realistic like `data/ `

orelgueta (approved):
All looks good to me. I am not the best reviewer these days though, I haven't looked at the code for a while now.

data/parameters/array-simulator_parameters.yml
RaulRPrado: No, these are pre-defined parameter file. The user is not supposed to touch it.
This way of writing the units is how the yaml module from astropy can read it.

simtools/array_simulator.py
RaulRPrado: Right. I will change it. Anyway, it's just an example.

data/parameters/array-simulator_parameters.yml
orelgueta: OK!

simtools/array_simulator.py
orelgueta: OK!

PR #151, opened 2021-10-04
First implementation of the production application

RaulRPrado:
This PR adds an application called production, which is meant to be a simple tool to produce shower and array level simulations.

A configuration file is required. An example can be found in data/test-data/prodConfigTest.yml


simtools/util/names.py
```diff
@@ -667,6 +667,8 @@ def corsikaConfigFileName(arrayName, site, zenith, viewCone, label=None):
         Array name.
     site: str
         Paranal or LaPalma.
+    primary: str
+        Primary particle (e.g gamma, proton etc).
```
orelgueta, line 671: Is there a way in the documentation to link to the yaml file with the options to make it clear for the user? Just curious, not urgent at all.

PR #161, opened 2021-11-02
Upload files to db

orelgueta:


applications/add_file_to_db.py
```diff
@@ -0,0 +1,145 @@
...
+    verbosity (str, optional)
+        Log level to print (default=INFO).
+
+    Example
+    -------
+    uploading a dummy file.
+
+    .. code-block:: console
+
+        python applications/add_file_to_db.py -f data/data-to-upload/test-data.dat
```
RaulRPrado, line 34: Does this file exist? I mean, is there a file that we can use to run tests?
If yes, I'd add it to data/test-data

applications/add_file_to_db.py
```diff
@@ -0,0 +1,145 @@
...
+        help='Log level to print (default is INFO)'
+    )
+
+    args = parser.parse_args()
+
+    logger = logging.getLogger()
+    logger.setLevel(gen.getLogLevelFromUser(args.logLevel))
+
+    if args.fileName is not None:
+        filesToInsert = args.fileName
```
RaulRPrado, line 120: What about we check for the allowed extension in case of a single file?

simtools/db_handler.py
```diff
@@ -975,3 +1045,54 @@ def _getTaggedVersion(self, dbName, version='Current'):
         tags = collection.find(query).sort('_id', pymongo.DESCENDING)[0]
 
         return tags['Tags'][version]['Value']
+
+    def insertFileToDB(self, file, dbName=DB_CTA_SIMULATION_MODEL, **kwargs):
```
RaulRPrado, line 1049: I think we should add a description of kwargs too

RaulRPrado (approved):
Looks all good, small comments only

RaulRPrado:
Can we merge this and deal with the comments later? I'm kind of stuck while this is not merged.

orelgueta:
Sure, go ahead and merge and I will deal with the comments in a second PR.

applications/add_file_to_db.py
orelgueta: No, this was just a documentation example. 

In order to run the test in the unit test I actually create a file on the fly, upload it to the DB, make sure it is there and with the correct ID and then delete it from disk and the DB. See the snippet below

https://github.com/gammasim/gammasim-tools/blob/47061d495df314b9a348dc9709e69e32344d9277/simtools/tests/test_db_handler.py#L285-L296

applications/add_file_to_db.py
orelgueta: Good idea, will add this check now.

simtools/db_handler.py
orelgueta: Added (linking to the official MongoDB documentation)

PR #167, opened 2021-11-05
Fixing environment

RaulRPrado:
When trying to use the package at the wgs, I found out that eventio cannot be compiled with python 3.9.
It seems to be related to this https://github.com/ContinuumIO/anaconda-issues/issues/12319

So I added the python<=3.8 to the env file and it works now.

I'm PRing to let you know about it.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,6 +1,7 @@
 # A conda environment with all useful package for ctamclib developers
 name: gammasim-tools-dev
 dependencies:
+  - python<=3.8
   - pyyaml
   - numpy
   - astropy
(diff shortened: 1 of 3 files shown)


orelgueta:
Did you open an issue on the eventio github? 
I guess we anyway will always require a python version for the package, but if eventio is fixed and there are no other issues, I would remove this requirement and put it only when it becomes an issue again.

RaulRPrado:
No, I did not open an issue.

I assume this will be fixed at some point. Python 3.9 is still quite fresh.

I would leave like that until 3.9 is safer and we actually want to allow for it.

orelgueta:
OK. Again, we will probably always require some python version, so I don't see it as an issue.

PR #168, opened 2021-11-09
Adding SimtelEvents

RaulRPrado:
This PR adds the module simtel_events, which is meant to handle events from sim_telarray standard output files.
By now it is very simple, containing only functions to calculate number of triggered and simulated events.
These functions will be useful next to build the trigger simulation modules.

The simtel_events module should be expanded as we need more functionalities.


RaulRPrado:
I juts noticed that I commited the new files to the master by accident in the beginning.
So github is showing differences but it should be showing them as new files.

orelgueta:
So sim_televents.py and its test file are completely new? It looks very weird with the changes, but if I ignore it and just look at the files themselves I can do the review.

RaulRPrado:
Yes, they are completely new. I tryied to fixe it now by removing them from the master, but it didn't seem to work.

orelgueta:
No worries, I will simply go over the files themselves.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
+
+        # Calculating number of events
+        self._mcHeader['n_events'] = (
+            self._mcHeader['n_use'] * self._mcHeader['n_showers'] * self._numberOfFiles
+        )
+        self._mcHeader['n_triggered'] = numberOfTriggeredEvents
+        return
+
+    @u.quantity_input(coreMax=u.m)
+    def countTriggeredEvents(self, energyRange=None, coreMax=None):
```
orelgueta, line 135: This is just a thought, but would it be better or easier to understand if the energy and code-distances arrays be given as input as well?

I only ask cause while reading the code it took me a second to understand. I don't think it makes a difference.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
-            return bins[non_zero[0][0]], bins[non_zero[0][-1] + 1]
+        selectedEvents = list()
+        for file in self.inputFiles:
+            with SimTelFile(file) as f:
 
-        pdfPages = PdfPages(figName)
-        for hist in combinedHists:
+                for event in f:
+                    energy = event['mc_shower']['energy']
+                    x_core = event['mc_event']['xcore']
```
orelgueta, line 190: It could be a little more efficient if you do the second calculation after performing the first test (you can then avoid unnecessary calculations). Won't make a big difference though I think.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
 
-            if hist['n_bins_y'] > 0:
-                # 2D histogram
+                    selectedEvents.append(event)
+        return selectedEvents
 
-                xlim = _get_ax_lim(hist, axis=0)
-                ylim = _get_ax_lim(hist, axis=1)
+    @u.quantity_input(coreMax=u.m)
+    def countSimulatedEvents(self, energyRange=None, coreMax=None):
```
orelgueta, line 204: This function assumes a straightforward powerlaw simulation. You do not allow weights, simulation in bins of energy, different functions other than a powerlaw, etc. 

That is also true for the code radius. Now that we have the option of simulating events based on a distribution which depends on the energy, this function would give the wrong number.

It is very efficient, much better than actually counting events, but I would make it clear in the documentation and maybe even in the name of the function.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
-                    ax.text(
-                        0.5,
-                        0.5,
-                        'EMPTY',
-                        horizontalalignment='center',
-                        verticalalignment='center',
-                        transform=ax.transAxes
-                    )
-                    continue
+    def _validateEnergyRange(self, energyRange):
```
orelgueta, line 236: You are missing documentation for this function.

Also, I would not expect that a function called "validate" would return the full energy range in case I did not supply one.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
+        try:
+            return (energyRange[0].to(u.TeV).value, energyRange[1].to(u.TeV).value)
+        except u.core.UnitConversionError:
+            msg = 'energyRange must be in units of energy'
+            self._logger.error(msg)
+            raise TypeError(msg)
 
-        pdfPages.close()
-    # End
+    def _validateCoreMax(self, coreMax):
```
orelgueta, line 252: Same comment as for the previous function. In addition, in the previous function you made tests and exceptions about the units and here you assume they are OK. Maybe it would be OK in this case if you had had the `@u.quantity_input(coreMax=u.m)` decorator?

orelgueta (changes requested):

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
-                        self._logger.warning('Problematic file {}'.format(file))
-                        countFile = False
-                        continue
+        if not hasattr(self, 'inputFiles'):
+            self.inputFiles = list()
 
-                    if len(combinedHists) == 0:
-                        # First file
-                        combinedHists = copy.copy(hists)
+        if files is None:
```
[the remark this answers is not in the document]
RaulRPrado: Right. I fixed it

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
+        Read MC header from sim_telarray files and store it into _mcHeader.
+        Also fills summaryEvents with energy and core radius of triggered events.
+        '''
 
-        return combinedHists
+        self._numberOfFiles = len(self.inputFiles)
+        keysToGrab = ['obsheight', 'n_showers', 'n_use', 'core_range', 'diffuse', 'viewcone',
+                      'E_range', 'spectral_index', 'B_total']
+        # Dict has to have its keys defined and filled beforehands
+        self._mcHeader = {k: 0 for k in keysToGrab}
```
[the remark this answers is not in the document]
RaulRPrado: Ok, that is trick.

I started by doing what you suggested but then I ran out on some problems. Things that are learned (but did not understand) are: a) I have to create the dict beforehand and b) it has be initiated with an integer (not None).

When I didn't do those things, I ran on very weird problems. In some cases I could print the value of a certain key but when I tried to operate on it (to calculate n_events) I'd get a KeyError, which makes zero sense.

When I initiated the dict with None, I found out that the new values (int) would be printed correctly but somehow it's type would still be None. So I'd get an error when trying to calculate n_events.

These issues are quite mysterious to me, I spend a good amount of time on this and in the end decided to go for this non-ideal solution. There is nothing wrong though.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,253 @@
...
+
+                for event in f:
+                    en = event['mc_shower']['energy']
+                    rc = math.sqrt(
+                        math.pow(event['mc_event']['xcore'], 2)
+                        + math.pow(event['mc_event']['ycore'], 2)
+                    )
 
-    def _plotCombinedHistograms(self, combinedHists, figName):
+                    self.summaryEvents['energy'] = np.append(self.summaryEvents['energy'], en)
```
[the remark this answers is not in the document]
RaulRPrado: Ok, good. I changed it to lists now.

simtools/simtel/simtel_events.py
RaulRPrado: Yep, good catch. I changed it now

simtools/simtel/simtel_events.py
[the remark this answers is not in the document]
orelgueta: That sounds very strange...

Can I try playing around to see if I can figure it out? (I need the distraction...)

simtools/simtel/simtel_events.py
RaulRPrado: I don't think now it's the time for us to think about it.
I can come back to this in the future and fix it. I need to understand how sim_telarray stores the info about the simulated spectrum.

simtools/simtel/simtel_events.py
RaulRPrado: I added some docstrings.

Well, it returns the default in case energyRange=None is provided.
I agree it is not the perfect name but I think it makes sense in this context. Note that it is a private/internal function.

simtools/simtel/simtel_events.py
orelgueta: I definitely don't want to write a function that deals with everything I mentioned. Just make it clear in the documentation that those assumptions are made. Otherwise in the future it would be hard for us (or someone else) to find this as a source of misinformation. 

I would even change the name of the function to be less general, to make it even clearer, but that is less critical.

simtools/simtel/simtel_events.py
orelgueta: Yea, the name isn't a big deal in this case since it's an internal function, as long as the docstring makes it clear.

simtools/simtel/simtel_events.py
```diff
@@ -1,194 +1,261 @@
...
-        histogramFiles: list
-            List of sim_telarray histogram files (str of Path).
-
+        inputFiles: list
+            List of sim_telarray output files (str of Path).
         '''
         self._logger = logging.getLogger(__name__)
-        self._histogramFiles = histogramFiles
+        self.loadInputFiles(inputFiles)
+        self.loadHeaderAndSummary()
```
orelgueta, line 48: This should be called only if `inputFiles` is not `None` or its length is longer than 1.

simtools/simtel/simtel_events.py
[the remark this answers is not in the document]
orelgueta: My tests were successful out of the box (almost). Will commit now.

simtools/simtel/simtel_events.py
orelgueta: I made that change now, but check if you prefer to check length instead.

simtools/simtel/simtel_events.py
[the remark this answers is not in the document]
RaulRPrado: Perfect. Thanks.
(No idea what went wrong before)

simtools/simtel/simtel_events.py
RaulRPrado: Do you mean adding both to the quantity_input decorator?
That would be ideal but this decorator only accepts scalers, not arrays.

I thought of using energyMin and energyMax to fix that, but I think it's conceptually cleaner to keep energyRange and check the units by ourselves.

simtools/simtel/simtel_events.py
RaulRPrado: I agree with adding something to the documentation (will do that).
But I prefer to keep the signature generic as it is. In the future we should deal with these possible non-power law spectra inside the function and the user won't need to know anything it.

simtools/simtel/simtel_events.py
RaulRPrado: Yes, for the coreMax we have the decorators to assure the units is fine.
We don't use it for the energyRange because it only works with scalers, not with arrays.

I added the docstrings.

simtools/simtel/simtel_events.py
orelgueta: OK, no problem, documentation is enough then.

simtools/simtel/simtel_events.py
RaulRPrado: Perfect.

I actuallly did something a bit fancier. I added a property numberOfFiles and I check if numberOfFiles > 0.

simtools/simtel/simtel_events.py
orelgueta: And you don't need the decorator also in this function? Or you trust that you always get it "validated" in the calling function?

RaulRPrado: I trust.
The argument should be checked by the public method. This private method is meant to be called by the public ones, so the argument should be validated already.

simtools/simtel/simtel_events.py
orelgueta: No, I meant providing as input the energy and core distance arrays. Those which are used in this function with `self.summaryEvents['energy']` and `self.summaryEvents['r_core']`. Now that I think about it again, it probably doesn't make a lot of sense since you want to count for this specific instance of the class. So we can probably safely ignore my comment.

simtools/simtel/simtel_events.py
orelgueta: OK!

orelgueta (approved):

PR #175, opened 2022-02-04
DRAFT:Reformatting using black tool

GernotMaier:
Run black without options on all files.

For illustration, not for merging!


RaulRPrado:
I will do a pytest to ensure it didn't break anything.

From a quick look, it seems fine.

It is replacing all single quotes by double one, interesting. I didn't know black says anything about this.

GernotMaier:
It really should only do reformatting.
(and I don't understand the difference in python between single and double quotes. It doesn't make sense to have both of them).

RaulRPrado:
> It really should only do reformatting. (and I don't understand the difference in python between single and double quotes. It doesn't make sense to have both of them).

There is no difference. When we started, I found that single quotes were more common.

RaulRPrado:
pytest is all good.

I went quickly through the file changes and I only found:
- singles quotes -> double quotes,
- 0. -> 0.0
- comas added after the last entry of a tuple
- line breaks (probably because we were using 100 char and the black standard must be 80)

So in general our code was quite well formatted.

I think it's safe to merge this branch, if you want to.

RaulRPrado:
I can work on the conflicts actually. Just let me know if we want to merge it.

GernotMaier:
Yes, I have seen that the code was already in a very good state and updates are minor. This is very nice to see!

I would suggest to merge, so that we have the possibility to use black as automated formatting / linting in future. I still have to play around and understand exactly how those pre-commit hooks work.

For the conflicts, we would always accept the version of the branch? Should be fast to solve, please go ahead.

(and just discovered that github has a button 'Draft pullrequest', while gitlab requires to put 'DRAFT:' in front of the pull request title).

RaulRPrado:
Ready to merge

GernotMaier:
Could you check https://github.com/gammasim/gammasim-tools/runs/5067013375?check_suite_focus=true

RaulRPrado:
Ok, NOW it's ready to merge.

GernotMaier:
Very nice, thank you!

orelgueta:
> * line breaks (probably because we were using 100 char and the black standard must be 80)

It uses 88 characters. We should change our local linter to do that as well if we move to this standard. Glad to see our code was already good prior to this automatic formatting.

GernotMaier:
I've seen many organisations (e.g. ctapipe) extending this value. We could also tell the linter to allow for 100 characters.

orelgueta:
I would vote for that.

GernotMaier:
Should we leave it at 88?

The merged reformatting of the code used 88. If we use now 'black --line-length 100', we have to do it again. Maybe we leave it (I am getting older and probably have to zoom in very soon, 100 might be too long then).

orelgueta:
I don't mind. I like 100 cause it avoids too many line breaks, but I don't feel strongly about it.

PR #182, opened 2022-02-07
Fixing bug with empty plot in compare_cumulative_psf

RaulRPrado:
There was a bug in the use of visualize.plotHist2D() in the application. The functions plot(...) return a figure, and this figure should be used by the application.

### DIFF
--- a/applications/compare_cumulative_psf.py
+++ b/applications/compare_cumulative_psf.py
@@ -201,7 +201,6 @@ if __name__ == "__main__":
 
     # Plotting image
     dataToPlot = im.getImageData()
-    print(dataToPlot)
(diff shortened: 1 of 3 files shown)


applications/compare_cumulative_psf.py
```diff
@@ -201,12 +201,14 @@ def loadData(datafile):
 
     # Plotting image
     dataToPlot = im.getImageData()
-    visualize.plotHist2D(dataToPlot, bins=80)
+    print(dataToPlot)
```
orelgueta, line 204: This debug line is probably not necessary

RaulRPrado: Right. Just removed it

PR #183, opened 2022-02-07
Removing usetex from visualize

RaulRPrado:
One-liner change in visualize to remove usetex=True


orelgueta:
Did you also check that we don't have Latex commands anywhere that may fail without this line?

RaulRPrado:
> Did you also check that we don't have Latex commands anywhere that may fail without this line?

Not yet. I will do that.

RaulRPrado:
I checked and it is all fine. Can I merge?

PR #186, opened 2022-02-15
Typos fixes

GernotMaier:
Typo fixes in comments found while browsing through the code.

Should not change anything (apart from being with less typos)

### DIFF
--- a/simtools/util/names.py
+++ b/simtools/util/names.py
@@ -743,7 +743,7 @@ def corsikaOutputFileName(run, primary, arrayName, site, zenith, azimuth, label=
 
     Warning
     -------
-        zst extension is hardcoded here.
+        zst extension is hard coded here.
 
     Parameters
     ----------
@@ -839,7 +839,7 @@ def simtelOutputFileName(run, primary, arrayName, site, zenith, azimuth, label=N
 
     Warning
     -------
-        zst extension is hardcoded here.
+        zst extension is hard coded here.
 
     Parameters
     ----------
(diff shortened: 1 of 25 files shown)


simtools/util/names.py
```diff
@@ -743,7 +743,7 @@ def corsikaOutputFileName(run, primary, arrayName, site, zenith, azimuth, label=
 
     Warning
     -------
-        zst extension is hardcoded here.
+        zst extension is hard coded here.
```
orelgueta, line 746: One word is actually OK (in particular in this context).

simtools/util/names.py
```diff
@@ -839,7 +839,7 @@ def simtelOutputFileName(run, primary, arrayName, site, zenith, azimuth, label=N
...
```
orelgueta, line 842: One word is actually OK (in particular in this context).

simtools/util/names.py
```diff
@@ -875,7 +875,7 @@ def simtelHistogramFileName(run, primary, arrayName, site, zenith, azimuth, labe
...
```
orelgueta, line 878: One word is actually OK (in particular in this context).

simtools/util/names.py
```diff
@@ -911,7 +911,7 @@ def simtelLogFileName(run, primary, arrayName, site, zenith, azimuth, label=None
...
```
orelgueta, line 914: One word is actually OK (in particular in this context).

orelgueta (commented):
Looks good except the change of `hardcoded` to `hard coded` which is unnecessary.

simtools/util/names.py
GernotMaier: OK - reversed that. Although technically wrong, one would probably search the code for hardcoded.

PR #188, opened 2022-02-16
Deploying gh-pages with github actions

RaulRPrado:
Testing using github actions to deploy gh-pages.


RaulRPrado:
gh-pages deployment using the CI workflow seems to be working fine.

I added two more steps to the CI workflow: one for building the docs (which will be done every PR) and one for deploying the docs to gh-pages (which should be triggered only on pushing to master, but it wasn't tested yet). 
I am using the same CI in order to avoid resetting the conda environment.

I learned a few things during these tests. The most important is that we need the line 'shell: bash -l {0}' to load the bash profile that was set previously. Otherwise, the step will use a new shell process.
I changed the lines in the Tests step accordingly.

@GernotMaier have a look and merge it when you are done, pls.
After the merging, I will check whether the gh-pages deployment worked properly.

GernotMaier (approved):

Issue #187, opened 2022-02-16
derive_rnda_mirror_angle reads the DB for each mirror


RaulRPrado:
As discussed in #176, derive_rnda_mirror_angle could read the DB only once.

RaulRPrado:
I had a look at the application and it turns out that we read the DB only once. See:

https://github.com/gammasim/gammasim-tools/blob/3990ca55098da48d59d8aec083071eb86532ebc2/applications/derive_mirror_rnda.py#L219-L224

For every mirror and new value of rnda_mirror_angle, we only replace the parameters using the same instance of TelecopeModel.

If there is a problem with the DB connection being open for too long, we could consider a public function in DatabaseHandler to close the connection.

orelgueta:
That is true in the application itself. However, every time we make the run command we call the getConfigFile()

https://github.com/gammasim/gammasim-tools/blob/3990ca55098da48d59d8aec083071eb86532ebc2/simtools/simtel/simtel_runner_ray_tracing.py#L186

that in turn exports the config file if it is updated (which I assume it is for each mirror)

https://github.com/gammasim/gammasim-tools/blob/3990ca55098da48d59d8aec083071eb86532ebc2/simtools/model/telescope_model.py#L510-L525

and that calls the DB

https://github.com/gammasim/gammasim-tools/blob/3990ca55098da48d59d8aec083071eb86532ebc2/simtools/model/telescope_model.py#L486-L496

I am fairly certain we are reading from the DB constantly because if I look at the number of operations per-second while running this application I get the plot below

![image](https://user-images.githubusercontent.com/22728856/154442708-124d73cc-af04-49da-b04c-c94874594eae.png)

RaulRPrado:
Ok, that is quite nested.
We connect again to the DB to export the model files, not to read the parameters.

I have then to split the self._isConfigFileUpdated flag into two: one for the model files and another for the parameters.

orelgueta:
Why do we need to export the model files again for each run?

RaulRPrado:
> Why do we need to export the model files again for each run?

I don't, that is the problem. I only need to export if any parameter that isFile was changed/added.

orelgueta:
Ahh OK, now I see what you mean.

RaulRPrado:
Fixed by #193

PR #193, opened 2022-02-17
Preventing TelescopeModel to read the DB when not necessary 

RaulRPrado:
Motivated by #187.

A flag was added to TelescopeModel to prevent reading the DB when there is no new parameter that is a file in the model.

### DIFF
--- a/simtools/tests/test_telescope_model.py
+++ b/simtools/tests/test_telescope_model.py
@@ -64,12 +64,13 @@ class TestTelescopeModel(unittest.TestCase):
         tel.exportConfigFile()
 
     def test_updating_export_model_files(self):
-        logger.info("Changing mirror_reflection_random_angle")
+        logger.info("Changing a parameter that is not a file - mirror_reflection_random_angle")
         new_mrra = "0.0080 0 0"
         self.telModel.changeParameter("mirror_reflection_random_angle", new_mrra)
 
-        self.assertTrue(self.telModel._areModelFilesUpdated)
+        self.assertTrue(self.telModel._isModelFilesUpdated)
 
+        # This should not connect to the DB
         self.telModel.exportConfigFile()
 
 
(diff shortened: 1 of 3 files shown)


simtools/tests/test_telescope_model.py
```diff
@@ -63,6 +63,70 @@ def test_cfg_file(self):
...
+        # Changing a file parameter
+        logger.debug(
+            "Changing a parameter that IS a file - camera_config_file"
+        )
+        tel.changeParameter(
+            "camera_config_file",
+            tel.getParameterValue("camera_config_file")
+        )
+        logger.debug(
+            "tel._isExportedModelFiles should be False because a file parameter "
```
orelgueta, line 121: Maybe make this sentence clearer like "a model file was changed" or "a parameter which is a file was changed".

I only make this comment because it was hard for me to understand until I read the test (see general comment).

RaulRPrado:
> I think it's all OK, but it took me quite a while to understand the logic. If not for the unit test, I wouldn't have understood it.
> 
> Basically, if I understand correctly, the flags `_isConfigFileUpdated` and `_isExportedModelFilesUpdated` are there to signal that the model is **up to date**, not that it has **been updated**. Both the word used `Updated` confused me and also the logic. Usually it is done the other way around (we keep a flag that something has changed, not that everything is up to date). I don't mind using this kind of logic, but then please change the flag name to make it easier to understand.

That makes a lot of sense. I will rename the flags to UpToDate.

RaulRPrado:
I renamed the flags and improved the comments. I will merge it now

orelgueta:
Looks good. I will run a long test on the laptop and on the WGS once it's merged.

orelgueta:
For the completeness, this made running derive_rnda_mirror_angle about 3 times faster on the WGS. It now finishes within 2-3 minutes. It is also significantly faster on the container. It takes ~20 minutes instead of ~2 hours (or crashing as @GernotMaier showed).

This also makes me want to try running in the container on the WGS. The difference is speed is very significant and I would like to know if it has to do with the container or the difference between the laptop and the WGS.

GernotMaier:
Just checked it on my laptop: works now! Much faster, takes 9 minutes on my laptop. Great!

GernotMaier:
Such a difference? I have a  MacBook Pro 2020 2,33 GHz Quad-Core Intel Core i7.

orelgueta:
Yeah, that might lead to a factor 2-3 difference (MacBook Pro 2016, 2.9 GHz Dual-Core Intel Core i5).

PR #195, opened 2022-02-21
New strategy for pytest - separating unit and integration tests

RaulRPrado:
Currently the tests run by pytest are strongly dependent on the DB connection and sim_telarray installation (which makes it hard to add it to a CI workflow). Ideally we should have "unit tests" independent of DB and simtel and "integration tests" specifics to each of them.

This PR presents a possible solution for this issue.

The general idea here is to mark the tests that require DB connection or sim_telarray installation and skip them, in case these are not available. This means that unit and integration tests will not be explicitly separated in the test modules, but they will be called separately depending on the availability of the DB and sim_telarray.

Why not separate all the test modules into unit and integration tests? First, this would mean rewriting all the tests modules and redesign them. Secondly, given the nature of our package (strong dependence with DB and sim_telarray), this separation can be quite difficult.

Now the technicalities.

Marking the tests in order to skip them is straightforward when using the the pytest decorator @pytest.mark.skipif. The more complicated step for us is to automatically identify whether the DB connection and sim_telarray installation are available.
I decided to do that by setting up a conftest.py file, that is loaded in the beginning of any pytest session. In this file I used the simtools config module to flag three things: a) the existence of the config file, b) the availability of the DB connection and c) the sim_telarray installation. 

a) is very simple. b) is also simple because we have a config entry called useMongoDB.
For c), I am looking at the sim_telarray path given in the config file and checking for the existence of the sim_telarray binary file.

The trick part was to pass these flags to the test modules. I tried to used pytest fixtures but it is not designed for that.
In the end I found a solution by using environment variables that are set in conftest.py and read by the modules. To encapsulate the reading of these variables I wrote some function in utils/tests.py.

Everything is working fine in the end. I have changed test_telescope_model and test_ray_tracing to exemplify how the test modules should me modified.

The next step is to add the skipif decorator to all test modules and also redesign some test modules in order to remove the dependence with the DB and simtel.

In summary, this is how our test setup will look like if this solution is accepted:

- The test modules run by pytest will contain unit tests and integration tests (DB and sim_telarray).
- If pytest is run in a environment without DB connection and sim_telarray installation (e.g. a CI environment), only the unit tests will be performed.
- Note that DB connection and sim_telarray are independent, so an environment may have only one of them available and only the respective integration tests will be run.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,7 +2,7 @@ import os
 import logging
 
 import simtools.config as cfg
-import simtools.db_handler as db
+from simtools import db_handler
 
 logger = logging.getLogger()
 
@@ -33,11 +33,16 @@ else:
         os.environ["HAS_DB_CONNECTION"] = "0"
     else:
         # Trying to connect to the DB
-        db.DatabaseHandler()
-        os.environ["HAS_DB_CONNECTION"] = "1"
-        print('DB CONNECTEDDDDDDDDDDDDDD')
-
-    logger.debug("Setting HAS_DB_CONNECTION = {}".format(os.environ["HAS_DB_CONNECTION"]))
+        db = db_handler.DatabaseHandler()
+        try:
+            db.getModelParameters("north", "lst-1", "Current")
+            os.environ["HAS_DB_CONNECTION"] = "1"
+            logger.debug("DB connection is available")
+            logger.debug("Setting HAS_DB_CONNECTION = 1")
+        except Exception:
+            os.environ["HAS_DB_CONNECTION"] = "0"
+            logger.debug("DB connection is NOT available")
+            logger.debug("Setting HAS_DB_CONNECTION = 0")
 
 
 def pytest_sessionfinish(session, exitstatus):
--- a/conftest.py
+++ b/conftest.py
@@ -21,3 +21,7 @@ else:
     # Checking whether there is DB connection
     useMongoDB = cfg.get("useMongoDB")
     os.environ["HAS_DB_CONNECTION"] = "1" if useMongoDB else "0"
+
+
+def pytest_sessionfinish(session, exitstatus):
+    os.system('./clean_files')
--- a/conftest.py
+++ b/conftest.py
@@ -47,4 +47,4 @@ else:
 
 def pytest_sessionfinish(session, exitstatus):
     """ Cleaning up output files before ending the pytest session. """
-    os.system('./clean_files')
+    os.system("./clean_files")
--- a/conftest.py
+++ b/conftest.py
@@ -1,27 +1,45 @@
 import os
+import logging
 
 import simtools.config as cfg
+import simtools.db_handler as db
+
+logger = logging.getLogger()
 
 
 try:
     cfg.loadConfig()
 except FileNotFoundError:
+    logger.debug("simtools configuration file was NOT found")
+    logger.debug("Setting HAS_CONFIG_FILE = 0")
+    logger.debug("Setting SIMTEL_INSTALLED = 0")
+    logger.debug("Setting HAS_DB_CONNECTION = 0")
     os.environ["HAS_CONFIG_FILE"] = "0"
     os.environ["SIMTEL_INSTALLED"] = "0"
     os.environ["HAS_DB_CONNECTION"] = "0"
 else:
-
     os.environ["HAS_CONFIG_FILE"] = "1"
+    logger.debug("simtools configuration found WAS found")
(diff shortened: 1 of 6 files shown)


conftest.py
```diff
@@ -0,0 +1,27 @@
...
+
+    os.environ["HAS_CONFIG_FILE"] = "1"
+
+    # Checking whether sim_telarray is properly installed
+    simtelPath = cfg.get("simtelPath")
+    simtelBinPath = simtelPath + "/sim_telarray/bin/sim_telarray"
+    os.environ["SIMTEL_INSTALLED"] = "1" if os.path.exists(simtelBinPath) else "0"
+
+    # Checking whether there is DB connection
+    useMongoDB = cfg.get("useMongoDB")
```
orelgueta, line 22: I think this is not a good enough test. I think we should actually test if we can read from the DB instead. Maybe we can use one of the simple applications to get a value of a parameters (like one of the unit tests)? If we put it in a try/except block it should do the trick, no?

data/test-data/CTA-North-LST-1-Current_test-telescope-model.cfg
```diff
@@ -0,0 +1,107 @@
+%==================================================
```
orelgueta, line 1: Is this what you referred to as "mock test data"? Just wondering what it is.

simtools/util/tests.py
```diff
@@ -0,0 +1,25 @@
+import os
+
+
+CONFIG_FILE_MSG = "Configuration file is not available"
```
orelgueta, line 4: Change to "The gammasim-tool configuration file, config.yml, is not available" to be cleater.

simtools/util/tests.py
```diff
@@ -0,0 +1,25 @@
...
+
+DB_CONNECTION_MSG = "There is not connection with DB"
```
orelgueta, line 6: Change to "Connection with the DB is not available".

simtools/util/tests.py
```diff
@@ -0,0 +1,25 @@
...
+
+SIMTEL_MSG = "sim_telarray is not installed"
+
+
+def _collect_conftest_flag(name):
```
orelgueta, line 11: I suggest to write explicitly "config_test" instead of "conftest" everywhere. Also in the name of "conftest.py".

orelgueta (approved):
The idea looks good to me. As I mentioned, I would change the test of the DB to be explicit and a few other small stuff, but the concept is good.

conftest.py
RaulRPrado: Isn't just creating a DatabaseHandler instance enough? It should crash if there is no DB connection, right?

RaulRPrado: No, it is not enough. We can create it successfully even with a broken connection.

RaulRPrado: I committed some changes. Now I use db.getModeParameters("north", "lst-1", "Current") to test whether the DB connection is fine.

data/test-data/CTA-North-LST-1-Current_test-telescope-model.cfg
RaulRPrado: Not exactly. In this case I am just using an existing config file to create a telescope_model, instead of reading from the DB. The mock data framework provided by pytest is fancier than that. But I don't know much details.

simtools/util/tests.py
RaulRPrado: Done

simtools/util/tests.py
RaulRPrado: Done

simtools/util/tests.py
RaulRPrado: Not possible, the name conftest.py is hardcoded into pytest. Since this has to be the name of the file, I would keep in other places for consistency.

RaulRPrado:
Thanks @orelgueta.
Please have a look at the changes and replies before I merge it.

I will merge this one and then work on a next PR adding the skipif to all the other test modules.

simtools/util/tests.py
orelgueta: really? how weird. Makes sense to keep consistent then and keep it everywhere.

conftest.py
orelgueta: OK, looks good. Need to test creating the DatabaseHandler in various environments probably, but we can fix it later if we see issues.

orelgueta:
It looks good, you can merge.

PR #197, opened 2022-02-22
Implementing pytest new strategy 

RaulRPrado:
Following up on #195, I implemented all the pytest.mark.skipif in all test modules.

### DIFF
--- a/simtools/tests/test_array_model.py
+++ b/simtools/tests/test_array_model.py
@@ -4,13 +4,18 @@ import pytest
 import logging
 
 from simtools.model.array_model import ArrayModel
-from simtools.util.tests import has_db_connection, DB_CONNECTION_MSG
-
+from simtools.util.tests import (
+    has_db_connection,
+    has_config_file,
+    DB_CONNECTION_MSG,
+    CONFIG_FILE_MSG,
+)
 
 logger = logging.getLogger()
 logger.setLevel(logging.DEBUG)
 
 
+@pytest.mark.skipif(not has_config_file(), reason=CONFIG_FILE_MSG)
 def test_input_validation():
     arrayConfigData = {
         "site": "North",
--- a/simtools/tests/test_array_model.py
+++ b/simtools/tests/test_array_model.py
@@ -1,8 +1,11 @@
 #!/usr/bin/python3
 
+import pytest
 import logging
 
 from simtools.model.array_model import ArrayModel
+from simtools.util.tests import has_db_connection, DB_CONNECTION_MSG
+
 
 logger = logging.getLogger()
 logger.setLevel(logging.DEBUG)
@@ -21,6 +24,7 @@ def test_input_validation():
     am.printTelescopeList()
 
 
+@pytest.mark.skipif(not has_db_connection(), reason=DB_CONNECTION_MSG)
 def test_exporting_config_files():
     arrayConfigData = {
         "site": "North",
@@ -37,8 +41,3 @@ def test_exporting_config_files():
 
     am.exportSimtelTelescopeConfigFiles()
     am.exportSimtelArrayConfigFile()
-
-
-if __name__ == "__main__":
-    # test_input_validation()
-    test_exporting_config_files()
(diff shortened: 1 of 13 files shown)


simtools/tests/test_array_model.py
```diff
@@ -1,13 +1,21 @@
...
+    has_config_file,
+    DB_CONNECTION_MSG,
+    CONFIG_FILE_MSG,
+)
 
 logger = logging.getLogger()
 logger.setLevel(logging.DEBUG)
 
 
+@pytest.mark.skipif(not has_config_file(), reason=CONFIG_FILE_MSG)
```
GernotMaier, line 18: I guess there is nothing significantly different between North and South that we want to run the tests for both sites.

simtools/tests/test_ray_tracing.py
```diff
@@ -1,21 +1,27 @@
 #!/usr/bin/python3
 
+import pytest
```
GernotMaier, line 3: Suprised that the linter allows you to break alphabetical order of imports.

GernotMaier (approved):
All looks good (two very minor comments).

Very elegant solution, but I think we should on the medium term consider making it obsolete: tests are most useful of they are most complete and running in the background.

GernotMaier:
Nice to see the messages now: 

simtools/tests/test_db_handler.py::TestDBHandler::test_reading_db_mst_fc SKIPPED (Connection with the DB is not available)

GernotMaier:
Was too fast, not all tests pass for me. I did this on a completely new clone / checkout of this branch without modify any configuration parameters just doing `pytest`. See attachment.
[gm_pytest_output.log](https://github.com/gammasim/gammasim-tools/files/8117492/gm_pytest_output.log)

simtools/tests/test_array_model.py
RaulRPrado: No, we don't. For the sake of unit tests, the specific site is not relevant.

RaulRPrado:
> simtools/tests/test_simtel_histograms.py::test_histograms takes notable longer than all other tests. Don't really see why, is it downloading that test file? run2_gamma_za20deg_azm0deg-North-Prod5_test-production-5.hdata.zst

I also noticed that.
It is probably because we have to read a large number of simtel histograms. I will try to make this test faster by selecting a subset of the histograms.

RaulRPrado:
> Do we want to switch on also the pytest part in the CI?
> 
> https://github.com/gammasim/gammasim-tools/blob/master/.github/workflows/CI-tests.yml

Yes. Done.

RaulRPrado:
> Was too fast, not all tests pass for me. I did this on a completely new clone / checkout of this branch without modify any configuration parameters just doing `pytest`. See attachment. [gm_pytest_output.log](https://github.com/gammasim/gammasim-tools/files/8117492/gm_pytest_output.log)

Did you run it twice?

In case there is no config.yml file, we have to create a dummy one to execute the tests. If this file is not removed after the session and you try to run the tests again, these errors will show up.
This is an issue. I will make the changes to remove the dummy config file after each session.

RaulRPrado:
It should be working now, @GernotMaier. You can run it multiple times, just be sure to remove any config.yml that were left from the previous tests.
If it works, please merge it.

GernotMaier:
All tests passed with sim_telarray installed and without DB - very nice.

Only issue is the warning

```
simtools/tests/test_layout_array.py: 42 warnings
simtools/tests/test_telescope_position.py: 2 warnings
  /workdir/external/gammasim-tools/simtools/layout/telescope_position.py:358: DeprecationWarning: This function is deprecated. See: https://pyproj4.github.io/pyproj/stable/gotchas.html#upgrading-to-pyproj-2-from-pyproj-1
    self._latitude, self._longitude = pyproj.transform(
```

It is probably me who should look after that. A quick search shows that I don't understand what is changing...

Also nice to see the CI in the github actions running.

GernotMaier (approved):

PR #200, opened 2022-02-23
Some trivial tests on making sim_telarray available for unit tests

GernotMaier:


RaulRPrado:
I changedit from draft to ready to review because I needed to see the results of the CI.

GernotMaier:
I am fine with merging. Do you want to add the DB access in a second step?

I am actually pleased by this solution. The tests take 7-10 minutes, was initially worried that it would take much longer.

PR #202, opened 2022-02-24
Remove the SSH tunnel and connect to the DB directly

orelgueta:
Removed all mentions of the SSH tunnel. Please notice that you will have to locally modify your dbDetails.yml file as shown in the template.


RaulRPrado:
Looks great, much simpler now.

pytest with DB and sim_telarray is all fine.

@orelgueta you may merge it.

PR #204, opened 2022-02-28
Adding tune_psf application

RaulRPrado:
This PR adds the tune_psf application, that tunes mirror_reflection_random_angle, mirror_align_random_horizontal  and mirror_align_random_vertical using cumulative PSF measurement.

@GernotMaier, please test it using MST data. I have done all tests using the LST-1 measurements.


GernotMaier:
Very nice!!

I've started first testing the LST psf tuning with

```
python applications/tune_psf.py --site North --telescope LST-1 --model_version prod5 --data PSFcurve_data_v2.txt --plot_all
```

Attached my results:
- are they similar to your's, @RaulRPrado ?
- what is again the tuning metric? The last page does not have the smallest RMSD (e.g., page 42 is smaller)
- parameter space for searches is quite large (ok for now), e.g., it seems unnecessary to look at the parameters on page 49

[tune_psf_LST-1.pdf](https://github.com/gammasim/gammasim-tools/files/8153357/tune_psf_LST-1.pdf)

RaulRPrado:
The intervals in which we were drawing the random values were wrong. I fixed it.

The problem with the best parameters not having the min RMSD is a bit trick.
We collect the best parameters properly but then we resimulate the ray tracing to produce the last plot. So it may be that when using this resimulated sample of photons, we get a different results, because of fluctuations.
This is only a problem when using the test flag that simulates a small number of photons.

Without the test flag on, I get consistent results, in which the last plot indeed shows the smaller RMSD. I still don't get exaclty the same number (0.0032 vs 0.0036), but they are close enough.

These are my results 
[tune_psf_LST-1.pdf](https://github.com/gammasim/gammasim-tools/files/8153514/tune_psf_LST-1.pdf)

GernotMaier:
OK, looks much nicer now!
[tune_psf_LST-1.pdf](https://github.com/gammasim/gammasim-tools/files/8153797/tune_psf_LST-1.pdf)

Do we have the PSF data for the MST as function of radius? I thought we have D80 values only.

RaulRPrado:
> OK, looks much nicer now! [tune_psf_LST-1.pdf](https://github.com/gammasim/gammasim-tools/files/8153797/tune_psf_LST-1.pdf)
> 
> Do we have the PSF data for the MST as function of radius? I thought we have D80 values only.

I don't know. I thought you said before that we had the cumulative PSF.

RaulRPrado:
> One should remember that the first parameter of the mirror_reflection_random_angle should be kept fixed and not used in the tuning (in cases where we have single-mirror PSF data, like in the LST case).

Right. I will add a command line argument to keep it fixed.

RaulRPrado:
> > One should remember that the first parameter of the mirror_reflection_random_angle should be kept fixed and not used in the tuning (in cases where we have single-mirror PSF data, like in the LST case).
> 
> Right. I will add a command line argument to keep it fixed.

Done.

I would like to merge this if everyone agrees.

orelgueta:
I thought you wanted to test with another telescope first. Also, you don't think we should run the tuning once more with the first parameter fixed to make sure it works also that way?

We have cumulative PSFs from SCT (see [here](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/verification/data-from-instrument-teams/sct/optics/psf)) which we can test with. The MST one we will get soon.

RaulRPrado:
> I thought you wanted to test with another telescope first. Also, you don't think we should run the tuning once more with the first parameter fixed to make sure it works also that way?
> 
> No, we don't have cumulative PSFs from MST. I wrote originally that we have from SCT but only later realised that those are vs off-axis, not cumulative. I will try to find if we have from a different team.

Yes, it would be good to test it in case we have cumulative PSF for another telescope. 

I did the test with the first parameter fixed. See plot attached. It seems all fine.
[tune_psf_LST-1.pdf](https://github.com/gammasim/gammasim-tools/files/8160335/tune_psf_LST-1.pdf)

orelgueta:
Couldn't find another cumulative PSF, so for now we will have to settle for just this one. We are supposed to get one from MST soon, so we will re-do the test then, but for now just this one is fine.

It all looks good, so I am happy to merge. Just one small question, in your latest results, the last page is not the best RMSD. On page 27 the RMSD is slightly better (though the parameters are the same). Why did that happen?

Obviously it is meaningless since the difference between them is negligible. I just want to make sure there are no bugs or anything like that.

RaulRPrado:
> Couldn't find another cumulative PSF, so for now we will have to settle for just this one. We are supposed to get one from MST soon, so we will re-do the test then, but for now just this one is fine.

Ok, good.

> It all looks good, so I am happy to merge. Just one small question, in your latest results, the last page is not the best RMSD. On page 27 the RMSD is slightly better (though the parameters are the same). Why did that happen?
> 
> Obviously it is meaningless since the difference between them is negligible. I just want to make sure there are no bugs or anything like that.

This is related to what I explained above.
I find the best parameters and then I re-simulate the ray tracing to produce the final plot. So small statistical fluctuations are expected, since these are not the same photon sample.
This could be a "problem" only when using the test flag, when we simulate much fewer photons and the fluctuations can be large.

orelgueta:
Oh, sorry, I should have reread the comments above, forgot about what you explained before.

All good from my side then, we can merge.

PR #205, opened 2022-03-01
Interpreting environment variables in config file using os.path.expandvars

### LINKED ISSUE #203: Set the dataLocation using absolute path in config_template.yml
Change the config_template.yml so it is possible to run it out-of-the-box from a different directory.

RaulRPrado:
Simple PR to allow the use of complex environment variables in config file.
Fixes #203.


RaulRPrado:
It is a very small PR. Please, any of you have a look and merge it.

Issue #198, opened 2022-02-23
test_simtel_histograms.py::test_histograms takes too long


RaulRPrado:
simtools/tests/test_simtel_histograms.py::test_histograms takes notable longer than all other tests. Don't really see why, is it downloading that test file? run2_gamma_za20deg_azm0deg-North-Prod5_test-production-5.hdata.zst

_Originally posted by @GernotMaier in https://github.com/gammasim/gammasim-tools/issues/197#issuecomment-1047860764_

PR #207, opened 2022-03-03
Adding test flag to simtel_histograms in order to make tests faster

RaulRPrado:
Simple change to add a test flag to simtel_histograms to make tests faster. When test=True, only a fraction (1/5) of the histograms are processed.


RaulRPrado:
Fixing #198

PR #212, opened 2022-03-10
Allowing the 

RaulRPrado:


orelgueta:
Why did you choose to do it through env variables rather than directly in the dbDetails.yml file? Is the latter option not possible?

RaulRPrado:
It is finally working with the api user. @GernotMaier please have a look and merge it if it is fine by you.

We have 95/101 tests passed and 6/101 skipped. The skipped ones require the model files that we are  keeping on the gitlab.

PR #216, opened 2022-03-15
Add image plotting to validate optics (was fix effective focal length)

RaulRPrado:
I started investigating this issue by plotting the images. I added a argument to the validate_optics to allow for that.

I then ran the app with LST-1 and got the images attached. I noticed that there is a small shift of the centroid in the Y direction as well. So I decided to calculate the eff focal length with the radius of the centroid instead and it seems to have fixed the calculation. See attached the new eff focal length plot. 
[validate_optics_LST-1_images.pdf](https://github.com/gammasim/gammasim-tools/files/8251883/validate_optics_LST-1_images.pdf)
[validate_optics_LST-1_eff_flen.pdf](https://github.com/gammasim/gammasim-tools/files/8251884/validate_optics_LST-1_eff_flen.pdf)

Now the question is where this shift in Y comes from and whether it is supposed to exist or not. We should continue digging to understand that.

### DIFF
--- a/applications/validate_optics.py
+++ b/applications/validate_optics.py
@@ -39,6 +39,8 @@
         Zenith angle in deg (default=20).
     max_offset (float, optional)
         Maximum offset angle in deg (default=4).
+    plot_images (activation mode, optional)
+        Produce a multiple pages pdf file with the image plots.
     test (activation mode, optional)
         If activated, application will be faster by simulating fewer photons.
     verbosity (str, optional)
@@ -53,10 +55,6 @@
     .. code-block:: console
 
         python applications/validate_optics.py --site North --telescope LST-1 --max_offset 5.0
-
-    .. todo::
-
-        * Change default model to default (after this feature is implemented in db_handler)
 """
 
 import logging
@@ -99,7 +97,7 @@ if __name__ == "__main__":
         "--model_version",
         help="Model version (default=prod4)",
         type=str,
-        default="prod4",
+        default="Current",
     )
     parser.add_argument(
         "--src_distance",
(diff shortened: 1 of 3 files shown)


orelgueta:
No, we only have one on axis image (from Vega if I am not mistaken). We can ask if they have images for us.

orelgueta:
I reviewed the code and it all looks fine, but it would be good to confirm first that this asymmetry is the result of the asymmetric mirror.

RaulRPrado:
> BTW, since I am running the MST simulation, I also figured I will check the grid issue. I understand why you said you can't plot it underneath. The best I could produce without wasting time is shown below, which version do you think is better?
> 
> image

It looks much worse with the LST because there were more 'empty' bins, so the white background was larger.
I personally think the simple lines at 0 are good enough.

orelgueta:
Bad news, the shift is probably due to the camera rotation. Repeating the calculation for MST, this time with NectarCam, which has the same camera rotation as LST, we get a shift of the image in the y direction. Since we do not see it for FlashCam, it is most likely due to the camera rotation angle (even though the camera is not actually used in the simulations). Need to dig further to get to the bottom of this.

Below is the image of a star at 3.5 off-axis angle with the NectarCam (compare to the FlashCam one above).

image

orelgueta:
OK the issue we see is definitely due to the camera rotation angle. I repeated the NectarCam and LST simulations after manually changing the rotation angle to zero and we get the images below.
Konrad runs these simulations with a single-pixel camera as far as I can tell. We should probably do the same. 

image

image

orelgueta:
@GernotMaier I removed the fix Raul added but kept the plotting addition. Can you take a quick look and merge if you are happy?

PR #217, opened 2022-03-15
Guidelines for developers in documentation

RaulRPrado:
This is the first version of the guidelines for developers in the documentation. I will write more but I wanted to show how it is going and maybe get some feedback.


orelgueta:
Is there a way I can read a rendered version of these new changes? Just wondering as it would be easier to read.

RaulRPrado:
> Is there a way I can read a rendered version of these new changes? Just wondering as it would be easier to read.

You can go the docs folder, 'make html' and open the docs/build/index.html with a browser.

docs/source/guidelines.rst
```diff
@@ -3,4 +3,193 @@
...
+The configurable input must be passed to classes through a dictionary or a yaml
+file. In the case of a dictionary the parameter is called configData, and in the
+case of a yaml file, configFile. See the ray_tracing module for an example.
+
+The function gen.collectDataFromYamlOrDict(configData, configFile, allowEmpty=False)
+must be used to read these arguments. It identifies which case was given and
+reads it accordingly, returning a dictionary. It also raises an exception in case none are
+given and not allowEmpty.
+
+The validation of the input is done by the function gen.validateConfigData, which
```
GernotMaier, line 108: I don't want to produce a lot of work and don't do it if this takes a lot of time: ideally one would link gen.validateConfigData to the code. If you tell me how to do that, I can add it.

GernotMaier (approved):
Very nice. Obviously will only notice the missing parts when start working with the code.

docs/source/guidelines.rst
RaulRPrado: Right. I didn't spend time on this because I wanted to get some text written first.

PR #219, opened 2022-03-16
Add functionality to change non-value parameters in the DB

orelgueta:
The functionality added to db_handler could be useful in general. The application on the other hand should never be used after today, so I did not write detailed documentation. However, I did decide to push it cause I might want to have it as a reference for the future.


RaulRPrado (approved):
Looks good to me.

PR #221, opened 2022-03-21
Prototype for setting workflow

GernotMaier:
Prototype for a setting workflow which should allow to:

- submit data and metadata using a yaml / ecsv file
- test metadata schema against a reference scheme, as seen e.g. [here](https://gitlab.cta-observatory.org/cta-computing/dpps/simulationpipeline/prototypes/notes/simulation-project-metadata-and-formats/-/blob/setting_workflow/SetParameterFromExternal.userinput-schema.yml)
- input data validation and if necessary transformation

Implemented on the example of `set_quantum_efficiency_from_external.py` (very likely could be a generic setting workflow; example of qe meta is [here](https://gitlab.cta-observatory.org/cta-computing/dpps/simulationpipeline/prototypes/notes/simulation-project-metadata-and-formats/-/blob/setting_workflow/quantum_efficiency/qe_R12992-100-05b.usermeta.yml)).

Will need a lot of refinements.

### DIFF
--- a/applications/set_quantum_efficiency_from_external.py
+++ b/applications/set_quantum_efficiency_from_external.py
@@ -23,8 +23,8 @@
     .. code-block:: console
 
         python ./set_quantum_efficiency_from_external.py \
-            --workflow_schema set_quantum_efficiency_from_external.yml \
-            --input_meta qe_R12992-100-05b.usermeta.yml \
+            --workflow_schema_file set_quantum_efficiency_from_external.yml \
+            --input_meta_file qe_R12992-100-05b.usermeta.yml \
 
 
 """
@@ -35,6 +35,33 @@ import logging
 import simtools.config as cfg
 import simtools.io_handler as io
 import simtools.util.general as gen
+import simtools.util.input_validation as validator
+
+def transformInput(
+    workflow_schema_file,
+    input_meta_file):
+    """
+    data transformation for simulation model data
+
+    includes:
+    -
+    - schema validation
+    - data validation
+    - data cleaning
+    - data conversion to standard units
+    - metadata enrichment
+    """
+
+    input_meta = gen.collectDataFromYamlOrDict(
+        input_meta_file,
+        None)
+    _schema_validator = validator.SchemaValidator(
+        workflow_schema_file,
+        input_meta)
+    _schema_validator.validate()
+
+    return None, None
+
 
 if __name__ == "__main__":
 
@@ -45,14 +72,14 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "-w",
-        "--workflow_schema",
+        "--workflow_schema_file",
         help="workflow description",
         type=str,
         required=True,
     )
     parser.add_argument(
         "-m",
-        "--input_meta",
+        "--input_meta_file",
         help="User-provided meta data file (yml)",
         type=str,
         required=True,
@@ -76,9 +103,9 @@ if __name__ == "__main__":
     logger.info("Outputdirectory {}".format(outputDir))
 
     # validate, transform, clean, enrich user metadata and data
-#    output_meta, output_data = transformInput(
-#        args.workflow_schema,
-#        args.input_meta)
+    output_meta, output_data = transformInput(
+        args.workflow_schema_file,
+        args.input_meta_file)
 
     # write model data in the format expected
 #    writeModelData(
--- a/applications/set_quantum_efficiency_from_external.py
+++ b/applications/set_quantum_efficiency_from_external.py
@@ -1,0 +1,97 @@
+#!/usr/bin/python3
+
+"""
+    Summary
+    -------
+    Setting workflow for photodetector / quantum efficiency
+
+
+    Command line arguments
+    ----------------------
+    workflow_schema (str, required)
+        Workflow description (yml format)
+    input_meta (str, required)
+        User-provided meta data file (yml format)
+    input_data (str, required)
+        User-provided data file (ecsv format)
+    verbosity (str, optional)
+        Log level to print (default=INFO).
+
+    Example
+    -------
+
+    Runtime 1 min
+
+    .. code-block:: console
+
+        python ./set_quantum_efficiency_from_external.py \
+            --workflow_schema set_quantum_efficiency_from_external.yml \
+            --input_meta qe_R12992-100-05b.meta.yml \
+            --input_data qe_R12992-100-05b.meta.ecsv
+
+
+"""
+
+import argparse
+import logging
+
+import simtools.config as cfg
+import simtools.io_handler as io
+import simtools.util.general as gen
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(
+        description=(
+            "Setting workflow for photodetector / quantum efficiency. "
+        )
+    )
+    parser.add_argument(
+        "-w",
+        "--workflow_schema",
+        help="workflow description",
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        "-m",
+        "--input_meta",
+        help="User-provided meta data file (yml)",
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        "-d",
+        "--input_data",
+        help="User-provided data file (ecsv)",
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        "-v",
+        "--verbosity",
+        dest="logLevel",
+        action="store",
+        default="info",
+        help="Log level to print (default is INFO)",
+    )
+
+    args = parser.parse_args()
+    label = "set_quantum_efficiency_from_external"
+
+    logger = logging.getLogger()
+    logger.setLevel(gen.getLogLevelFromUser(args.logLevel))
+
+    outputDir = io.getApplicationOutputDirectory(cfg.get("outputLocation"), label)
+    logger.info("Outputdirectory {}".format(outputDir))
+
+    # validate, transform, clean, enrich user metadata and data
+#    output_meta, output_data = transformInput(
+#        args.workflow_schema,
+#        args.input_meta,
+#        args.input_data)
+
+    # write model data in the format expected 
+#    writeModelData(
+#        output_meta,
+#        output_data)
--- a/applications/set_quantum_efficiency_from_external.py
+++ b/applications/set_quantum_efficiency_from_external.py
@@ -12,8 +12,6 @@
         Workflow description (yml format)
     input_meta (str, required)
         User-provided meta data file (yml format)
-    input_data (str, required)
-        User-provided data file (ecsv format)
     verbosity (str, optional)
         Log level to print (default=INFO).
 
@@ -26,8 +24,7 @@
 
         python ./set_quantum_efficiency_from_external.py \
             --workflow_schema set_quantum_efficiency_from_external.yml \
-            --input_meta qe_R12992-100-05b.meta.yml \
-            --input_data qe_R12992-100-05b.meta.ecsv
+            --input_meta qe_R12992-100-05b.usermeta.yml \
 
 
 """
@@ -61,13 +58,6 @@ if __name__ == "__main__":
         required=True,
     )
     parser.add_argument(
-        "-d",
-        "--input_data",
-        help="User-provided data file (ecsv)",
-        type=str,
-        required=True,
-    )
-    parser.add_argument(
         "-v",
         "--verbosity",
         dest="logLevel",
@@ -88,10 +78,9 @@ if __name__ == "__main__":
     # validate, transform, clean, enrich user metadata and data
 #    output_meta, output_data = transformInput(
 #        args.workflow_schema,
-#        args.input_meta,
-#        args.input_data)
+#        args.input_meta)
 
-    # write model data in the format expected 
+    # write model data in the format expected
 #    writeModelData(
 #        output_meta,
 #        output_data)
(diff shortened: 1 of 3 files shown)


GernotMaier:
@Orel - this is really a very first and early draft. Could you have quick <5 min look and give feedback if this makes sense.

At this point, all what happens is that we test that
- all required fields are available
- type check for datetime, email, str (latter should work for all generic python types)

Next step would be to add the checks on the data:
- are required data columns there
- are units and value ranges correct

applications/set_quantum_efficiency_from_external.py
```diff
@@ -0,0 +1,113 @@
...
+
+    input_meta = gen.collectDataFromYamlOrDict(
+        input_meta_file,
+        None)
+    _schema_validator = validator.SchemaValidator(
+        workflow_schema_file,
+        input_meta)
+    _schema_validator.validate()
+
+    return None, None
```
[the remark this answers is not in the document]
orelgueta: Ahh it's temporary I guess. Never mind.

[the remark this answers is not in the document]
GernotMaier: Temporary. The idea is to return here an updated metadata version / data version including the additions made by the workflow

simtools/util/input_validation.py
```diff
@@ -0,0 +1,114 @@
...
+            if data_dict and key in data_dict:
+                _this_data = data_dict[key]
+            else:
+                self._check_if_field_is_optional(key, value)
+                continue
+            if isinstance(value, dict) and 'type' in value:
+                self._validate_data_type(
+                    value, key, _this_data)
+            else:
+                self._validate_schema(value, _this_data)
```
orelgueta, line 114: Shouldn't this line be also under the `if isinstance(value, dict)`? When entering `_validate_schema` you assume the ref_schema is a dict. If it isn't the function will fail. It never happens now because your schema is only dicts. but it probably shouldn't be dependent on it.

simtools/util/input_validation.py
```diff
@@ -0,0 +1,114 @@
...
+                        format, data_field))
+
+        elif schema['type'] == 'email':
+            regex = r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'
+            if not re.fullmatch(regex, data_field):
+                raise ValueError(
+                    'invalid email format in field {}: {}'.format(
+                        key, data_field))
+
+        elif type(data_field).__name__ != schema['type']:
```
orelgueta, line 69: We could be a little more forgiving here and check first if we can convert to the correct type. It happens that a number is read as a string in yaml and a simple `float(entry)` works. I am not sure we should be forgiving, it opens up possible mistakes, but maybe you want to take it into account.

simtools/tests/test_input_validation.py
```diff
@@ -0,0 +1,49 @@
...
+
+    date_key = 'CREATION_TIME'
+    date_schema1 = {'type': 'datetime'}
+
+    # tests should succeed
+    date_validator._validate_data_type(
+        date_schema1, date_key, '2018-03-01 12:00:00')
+
+    # tests should fail
+    date_data1 = '2018-03-01 12:00'
```
orelgueta, line 24: What happens if you test something like `2018-13-01 12:00`? I assume it fails, but it should be worth a test because that would be the most common mistake I assume.

simtools/util/input_validation.py
GernotMaier: Agree - fixed. 07b23da

simtools/tests/test_input_validation.py
GernotMaier: Done. 

(at this point all exceptions are issued as format error; probably something to improve in the future).

simtools/util/input_validation.py
orelgueta: It won't be pretty, but we can define a converter dict like 

`convert = {'str': type('str'), 'float': type(1.0), 'int': type(0), 'bool': type(True)}`

and use a `try` and `except` like in the code below

```
elif type(data_field).__name__ != schema['type']:
    try:
        _value = convert[schema['type']](data_field)
    except ValueError:
        self._logger.error('invalid data type for key {}. Expected: {}, Found: {}'.format(
                    key, schema['type'], type(data_field).__name__))
        raise
```

I don't think it's necessarily good. Just a thought if we want to be forgiving.

GernotMaier: Implemented this.

But not entirely convinced: while converting int to floats is fine, I am hesitating about the opposite: converting 2.3 to 2 works, but maybe it is not what we want: when users give a value as a float, we should note convert it to an integer.

orelgueta: Absolutely we shouldn't. I only meant converting from string to float, int or boolean, nothing else. Should have written that explicitly.

simtools/util/input_validation.py
```diff
@@ -0,0 +1,114 @@
...
+        Assume that any field which is not label as required
+        is optional
+
+        Raises value error if required field is missing
+
+        """
+
+        if isinstance(value, dict) \
+                and 'required' in value \
+                and value['required']:
```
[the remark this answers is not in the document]
orelgueta: OK!

simtools/tests/test_input_validation.py
```diff
@@ -50,5 +50,57 @@ def test__validate_data_type_email():
         match=r"invalid email format in field EMAIL: me-blabla.de"):
             date_validator._validate_data_type(email_schema1, email_key, 'me-blabla.de')
 
+def test__validate_data_type_othertypes():
+
+    date_validator = validator.SchemaValidator(None, None)
+    test_key = 'SUBTYPE'
+    test_schema_1 = {'type': 'str'}
```
orelgueta, line 57: Could these be changed to be `test_schema_str`, `test_schema_float`, etc.? Would be much easier to read.

GernotMaier: Yes! Done.

simtools/util/input_validation.py
GernotMaier: I do think we need to be explicit and allow conversion from string or int, but not from e.g. float to int. I am allowing a bit of forgiveness now only for string and int.

orelgueta: OK!

orelgueta:
One other petty thing I mentioned before but forgot during the call, don't forget to document also the expected input and output variables.

orelgueta:
One more check we should add to the data verification workflow is the order of the entries in the table. For example, for a table with wavelengths, we should make sure the column is rising monotonically. We just received a table with the wavelengths in reverse order which sim_telarray does not like (IIRC).

GernotMaier:
Note the addition of column attributes for sorting and removing duplicates:

```
  DATA_COLUMNS:
    wavelength:
      description: |-
        wavelength
      required_column: true
      unit: nm
      type: float32
      required_range: {
        unit: nm,
        min: 300,
        max: 700
      }
      attribute:
        - remove_duplicates
        - sort
```

simtools/util/validate_data.py
```diff
@@ -96,6 +109,50 @@ def _check_required_columns(self):
...
+        columns have different values). Expectation is that
+        there is one column only with column attributes)
+
+        """
+
+        for key, value in self._reference_data_columns.items():
+            if 'attribute' in value and 'no_dublicates' in value['attribute']:
+                self._logger.debug("Removing duplicates for column '{}'".format(
+                    key))
+                self.data_table = unique(self.data_table)
```
orelgueta, line 154: I haven't tried it, so I am not sure about it, but I think that if the wavelength is the same, but the QE is different, this will not remove that row. I also am not sure if we want simply remove that row in that case. We should probably raise an error and stop in this case.
Also, in the case that you do find duplicate wavelengths, we should at least raise a warning so the user is aware and can make sure it is OK to simply remove it.

simtools/util/validate_data.py
```diff
@@ -52,9 +53,21 @@ def validate_and_transform(self):
 
         return self.data_table
 
+    def transform(self):
+        """
+        Apply any requested transformations to the
```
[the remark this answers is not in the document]
GernotMaier: Agree. Done with 1c3af65

simtools/util/validate_data.py
```diff
@@ -96,6 +109,50 @@ def _check_required_columns(self):
...
+        (if required by any column attribute)
+
+        Data is either sorted or reverse sorted
+        """
+
+        _columns_to_sort = []
+        _columns_to_reverse_sort = []
+        for key, value in self._reference_data_columns.items():
+            if 'attribute' in value:
+                if 'sort' in value['attribute']:
```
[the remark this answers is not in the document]
GernotMaier: Not sure, this is a list search:
```
value=['reversesort']
if 'sort' in value:
	print('found sort')
else:
	print('sort not found')
```
results in 'sort not found'. So that seems to be fine.

I've tested this by adding a unit test to reverse sorting (commit 0590a3a)

simtools/util/validate_data.py
```diff
@@ -96,6 +109,50 @@ def _check_required_columns(self):
...
 
+    def _sort_data(self):
+        """
+        Sort data according to one data column
+        (if required by any column attribute)
+
+        Data is either sorted or reverse sorted
+        """
+
+        _columns_to_sort = []
```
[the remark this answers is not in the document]
GernotMaier: Done (commit 2737f6a)

simtools/util/validate_data.py
GernotMaier: Improved this function now:
- test that if e.g.,  we remove duplicated wavelength values, the corresponding data columns are the same. Otherwise raise a ValueError
- allow to test for uniqueness of several columns

Added also a corresponding test.

Commit a73681e

simtools/util/validate_data.py
```diff
@@ -96,6 +109,50 @@ def _check_required_columns(self):
...
+
+        """
+
+        for key, value in self._reference_data_columns.items():
+            if 'attribute' in value and 'no_dublicates' in value['attribute']:
+                self._logger.debug("Removing duplicates for column '{}'".format(
+                    key))
+                self.data_table = unique(self.data_table)
+
     def _get_reference_unit(self, key):
```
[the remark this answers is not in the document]
GernotMaier: Renamed `key` to `_column_with_unique_requirement`

GernotMaier:
Open question: what kind of Exceptions to be raised. Very often I raise a ValueError, as this seems to be the broadest definition.

GernotMaier:
@orelgueta - when you are back: unit tests are failing and I don't understand why.

It is this test:
https://github.com/gammasim/gammasim-tools/blob/50f84daf6a2b2d9c9638b209eac63fca1265d1bd/simtools/tests/test_validate_schema.py#L131

And I don't understand why the try..except in the following line doesn't catch the UnboundLocalError:
https://github.com/gammasim/gammasim-tools/blob/50f84daf6a2b2d9c9638b209eac63fca1265d1bd/simtools/util/validate_schema.py#L114

(either a simple mistake or something fundamental, which I didn't get in the python error handling)

GernotMaier:
> @orelgueta - when you are back: unit tests are failing and I don't understand why.
> 
> It is this test:
> 
> https://github.com/gammasim/gammasim-tools/blob/50f84daf6a2b2d9c9638b209eac63fca1265d1bd/simtools/tests/test_validate_schema.py#L131
> 
> And I don't understand why the try..except in the following line doesn't catch the UnboundLocalError:
> 
> https://github.com/gammasim/gammasim-tools/blob/50f84daf6a2b2d9c9638b209eac63fca1265d1bd/simtools/util/validate_schema.py#L114
> 
> (either a simple mistake or something fundamental, which I didn't get in the python error handling)

Good point here - somehow missed the check when a required field is missing.

Fixed with 56e76b2

(CI will fail today due to cloud storage offline)

GernotMaier:
@orelgueta - ok with merging this (when the tests are successful)?

It is getting far too big, PRs shouldn't be so large.

orelgueta (approved):
I agree this PR became very large, such that it's hard to do a thorough review. However, going through it, it looks like all of my previous comments were addressed and I couldn't find any other issues, except the one comment I made. It's not critical though, so if you want we can merge this already and open a separate one for the Path point.

orelgueta:
> Open question: what kind of Exceptions to be raised. Very often I raise a ValueError, as this seems to be the broadest definition.

I see you actually improved this in many places. However, for the future, we sometimes defined our own exceptions to make it clearer in the log. It might be appropriate also in some of your cases.

GernotMaier:
> > Open question: what kind of Exceptions to be raised. Very often I raise a ValueError, as this seems to be the broadest definition.
> 
> I see you actually improved this in many places. However, for the future, we sometimes defined our own exceptions to make it clearer in the log. It might be appropriate also in some of your cases.

--> Issue #235

simtools/util/write_model_data.py
```diff
@@ -209,37 +283,61 @@ def _read_data_file_name(self, suffix=None):
...
+        except KeyError:
+            self._logger.error(
+                "Missing CTA:PRODUCT:ID in metadata")
+            raise
+
         if not suffix:
             suffix = '.' + self._read_data_file_format()
 
-        return _directory+'/'+_filename+suffix
+        return _directory + '/' + _filename+suffix
```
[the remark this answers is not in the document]
GernotMaier: OK - I merge now and add this in a follow-up merge request.

I need to check again what kind of generic functionality there is and if something is now duplicated --> issue #236

PR #227, opened 2022-03-31
Remove requirement for python <=3.8 and increase test matrix to newer versions

GernotMaier:
Added python 3.9 and 3.10 to test matrix.

Let's see if tests fail or if we really need the original requirement of python <=3.8

### DIFF
--- a/.github/workflows/CI-tests.yml
+++ b/.github/workflows/CI-tests.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
 
     steps:
       - name: checkout
--- a/environment.yml
+++ b/environment.yml
@@ -1,7 +1,7 @@
 # A conda environment with all useful package for ctamclib developers
 name: gammasim-tools-dev
 dependencies:
-  - python<=3.8
+  - python
   - pyyaml
   - numpy
   - astropy


GernotMaier:
In Summary:
- adding 3.9 and 3.10.4 to test matrix (3.10 is read as 3.1 by Github action and fails therefore)
- removed requirement on python<=3.8 from environment.yml file 

Do we need to have a lower limit on the python version?

orelgueta:
> Do we need to have a lower limit on the python version?

I don't know. I think we probably should have some lower limit, yes, but I don't have a well informed reason for this thought. I think it might get complicated otherwise with old versions. We might even decide to focus only on one python version like the ctapipe code does, but for now let's start with a lower limit.

> Note issue #228 (CI-docs fails with 3.10.4).

Then why did we add this to the tests? 3.10 is quite new, we don't have to keep up so closely with the latest versions.

GernotMaier:
> > Do we need to have a lower limit on the python version?
> 
> I don't know. I think we probably should have some lower limit, yes, but I don't have a well informed reason for this thought. I think it might get complicated otherwise with old versions. We might even decide to focus only on one python version like the ctapipe code does, but for now let's start with a lower limit.

My approach is that versions should not be an issue. If it is, than we should simplify the code.

Of course a mess with all the dependencies.

> 
> > Note issue #228 (CI-docs fails with 3.10.4).
> 
> Then why did we add this to the tests? 3.10 is quite new, we don't have to keep up so closely with the latest versions.

CI-docs is a different CI than the unit test ([CI-docs.yml](https://github.com/gammasim/gammasim-tools/blob/master/.github/workflows/CI-docs.yml) vs [CI-tests.yml](https://github.com/gammasim/gammasim-tools/blob/master/.github/workflows/CI-tests.yml)). CI-docs is running with 3.8.

We should solve the issues for 3.10 (therefore the issue)

GernotMaier:
@kosack - great, wasn't aware of the astropy requirement. I removed 3.6 and 3.7.

GernotMaier:
@orelgueta  - any reasons why we shouldn't merge this PR?

orelgueta:
No, no reason at all. I just never had a chance to look at it after making a comment a while back. Merging.

PR #230, opened 2022-03-31
Remove config.yml from gitignore

GernotMaier:
Motivation: very often local changes in config.yml lead to execution failures. As changes do not appear when running `git status`, there are often over looked and lead to long error searches.

(note the wrong name of the branch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# Local config.yml
-config.yml
-
 # Produced files
 simtools-files
 telarray_rand.conf.used


.gitignore
```diff
@@ -20,4 +17,4 @@ __pycache__
 docs/build
 
 # DB config gfile
-dbDetails.yml
\ No newline at end of file
+dbDetails.yml
```
orelgueta, line 20: I assume dbDetails is still in the gitignore and all you added is a new line, right?

orelgueta (approved):
All good, just added a question to be sure about something.

.gitignore
GernotMaier: No - I removed `config.yml` from .gitignore. I made a mistake in naming this branch, sorry for the confusion!

PR #232, opened 2022-04-07
Resolve HTML warning for sphinx documentation

GernotMaier:
Resolve errors and warnings reported by sphinx when making the documentation.

Most warnings are removed by commenting out [this line](https://github.com/gammasim/gammasim-tools/blob/fcd27a8dc4843fb972e580e7b5c0904b855108ff/docs/source/conf.py#L52) (after reading about it on [stackoverflow](https://stackoverflow.com/questions/62631362/get-rid-of-duplicate-label-warning-in-sphinx).


orelgueta (approved):
No issue. This system does look like a lot of duplication of work though...

GernotMaier:
Agree on the duplication. My suggestion to leave it for now. We anyway might agree in DPPS on a different documentation system (my assumption is that all subsystem would use the same)

orelgueta:
Sure. I hope I will remember to update these files as well in the future.

PR #237, opened 2022-04-28
Use joinpath in setting workflow for output file name

GernotMaier:


orelgueta:
Looks good, thanks!
(I ignore the failed test since it is the DB's fault.)

PR #239, opened 2022-04-28
Setting error messages

GernotMaier:
Very minor updates to logger messages: for exceptions use consistently `_logger.error` and not `_logger.debug`.


orelgueta:
Looks good, notice that you have one commit here which actually belonged to the other PR (#237), but I guess it doesn't matter. I will wait for the tests to complete and then merge.

PR #242, opened 2022-04-28
Small changes after using validate optics

orelgueta:
Most changes are not necessarily connected to each other, but I think all are good to include.


GernotMaier (approved):
All good - ready to merge!

PR #245, opened 2022-05-04
Add configurable containment radius to derive_mirror_rnda; setting workflow with association IDs

GernotMaier:
This is an 'intermediate' PR adding items (sorry, quite different):

1. allow to config the containment_fraction for the spot size determination on the command line:
e.g., the 80% containment radius is set by:
```
--containment_fraction 0.8
```
The variable `d80` has been renamed to containment_radius in the code to reflect this change.

Further changes are planned for the plotting part, so please do not worry about d80 appearing there. 

2. changed the default telescope model to `Current`
3. for settings workflows, do not apply numerical checks for columns of type 'string' (e.g. columns with the mirror panel IDs). This is done in a function called `_column_status`.

### DIFF
--- a/applications/derive_mirror_rnda.py
+++ b/applications/derive_mirror_rnda.py
@@ -138,6 +138,19 @@ def plotMeasuredDistribution(file, **kwargs):
     ax.hist(data, **kwargs)
 
 
+def efficiency_interval(value):
+    """
+    Argument parser check that value is an efficiency in the interval [0,1]
+
+    """
+    fvalue = float(value)
+    if fvalue < 0. or fvalue > 1.:
+        raise argparse.ArgumentTypeError(
+            "{} outside of allowed [0,1] interval".format(value))
+
+    return fvalue
+
+
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser()
@@ -169,7 +182,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--containment_fraction",
         help="Containment fraction for diameter calculation (in interval 0,1)",
-        type=float, required=False,
+        type=efficiency_interval, required=False,
         default=0.8,
     )
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


applications/derive_mirror_rnda.py
```diff
@@ -142,15 +152,25 @@ def plotMeasuredDistribution(file, **kwargs):
...
     parser.add_argument(
-        "--mean_d80", help="Mean of measured D80 [cm]", type=float, required=True
+        "--containment_sigma",
+        help="Std dev of measured containment diameter [cm]",
+        type=float, required=False
     )
     parser.add_argument(
-        "--sig_d80", help="Std dev of measured D80 [cm]", type=float, required=False
+        "--containment_fraction",
+        help="Containment fraction for diameter calculation (in interval 0,1)",
```
orelgueta, line 171: Suggest to add [], i.e., "in interval [0, 1]".
See comment below about another option instead.

applications/derive_mirror_rnda.py
```diff
@@ -210,6 +230,7 @@ def plotMeasuredDistribution(file, **kwargs):
     )
 
     args = parser.parse_args()
+    containment_fraction_percent = int(args.containment_fraction*100)
```
orelgueta, line 233: You need to make sure the containment radius is actually in the right range for this (and for other things).

I suggest you use the built in functionality of argparse and simply add `choices=range(0,1)` to the that input. You can even add then `metavar="[0-1]"` and then my comment above about the interval might not be necessary.

orelgueta (changes requested):
All good, just small non-critical changes requested.

applications/derive_mirror_rnda.py
GernotMaier: I've added now a method to the type definition for containment_fraction to check for a value inside the [0., 1.] interval.

As discussed, I will add a wrapper for the argparser where this can be included for general use.

orelgueta (approved):

PR #247, opened 2022-05-05
Argparser implementation

GernotMaier:
First 'suggestion' of an expansion of the argparser: @orelgueta - please have a very quick look at the general idea and give me feedback. There are still a lot of docstrings missing, but I would like to get some feedback.

Updates:
- new CommandLineParser as argparse.ArgumentParser Child
- allows to define default command line arguments which are the same for all applications (e.g., -v) without duplicating too much code (and therefore avoiding that the same functionality requires different command line options in different applications)
- allows generic implementation of functionality like the [0,1] interval checking (efficiency_interval)

Really a minor updates with minimum code changes.

(and please ignore the `--2f_measurement` option and the workflow configuration class, I mixed up again two branches)

### DIFF
--- a/simtools/util/commandline_parser.py
+++ b/simtools/util/commandline_parser.py
@@ -4,12 +4,15 @@ import argparse
 class CommandLineParser(argparse.ArgumentParser):
     """
     Command line parser for application and workflows
+    Wrapper around standard python argparse.ArgumentParser
 
     """
 
     def initialize_default_arguments(self):
         """
         Initialize default arguments used for all applications
+        (e.g., verbosity or test flag)
+
 
         """
 
@@ -32,6 +35,7 @@ class CommandLineParser(argparse.ArgumentParser):
     def initialize_workflow_arguments(self):
         """
         Initialize default arguments for workflow configuration parameters
+        and product data model
 
         """
 
@@ -51,9 +55,8 @@ class CommandLineParser(argparse.ArgumentParser):
             required=False,
         )
         self.add_argument(
-            "-r",
-            "--reference_schema_directory",
-            help="Directory with reference schema",
+            "--toplevel_metadata_schema",
+            help="Toplevel metadata reference schema",
             type=str,
             default=None,
             required=False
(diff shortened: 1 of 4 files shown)


simtools/util/commandline_parser.py
```diff
@@ -0,0 +1,101 @@
...
+            default="Current",
+        )
+
+    @staticmethod
+    def efficiency_interval(value):
+        """
+        Argument parser check that value is an efficiency in the interval [0,1]
+
+        """
+        fvalue = float(value)
```
orelgueta, line 96: Do you want to add a `try` and `except` here to provide an easier to interpret error or do you think it will be clear enough with the generic one?

orelgueta (approved):
Looks perfectly fine and will save us some coding in the future for sure.
I made a small comment, but it's not critical.

simtools/util/commandline_parser.py
GernotMaier: Not entirely sure how I would write this part as try and except, can you give me an example?

GernotMaier:
Let me know if I can improve `efficiency_interval` method. Otherwise ready from my side to review / merge.

simtools/util/commandline_parser.py
orelgueta: Forget it. My original thought was simply to have a slightly more informative error message which we could get if we use the code below (changing the print to the logger perhaps). However, it's adding unnecessary complications and is not how the try and  except are meant to be used.

```
try:
    fvalue = float(value)
except ValueError:
    print("The value provided for the efficiency is not a number")
    raise
```

orelgueta:
No need to improve the efficiency interval, merging.

orelgueta:
So we need to open an issue to change the code to use this new feature or do you plan to do it anyway?

GernotMaier:
We should open an issue - I am planning it anyway but first want to finish the output writing for derive_mirror_rnda

> On 6. May 2022, at 11:08, orelgueta ***@***.***> wrote:
> 
> 
> Do we need to open an issue to change the code to use this new feature or do you plan to do it anyway?
> 
> —
> You are receiving this because you were assigned.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

PR #252, opened 2022-06-05
Add coverage

GernotMaier:


orelgueta:
Wow, I seem to remember that Raul checked once and we had ~70% coverage. Either the test is different or we added a lot of untested code in the last year...

GernotMaier:
This is the true number - we had to wait until one full run of tests is completed in the github action (I've picked a single test to connect to repo to codecov, picket a bad one).

orelgueta:
Ahh OK, that looks more like it. Quite good considering we started checking this very late.

PR #262, opened 2022-06-13
Tune psf faster

orelgueta:
Change the algorithm to find the number of photons within a radius to be faster. For some reason also the changes to allow plotting the difference were included in this commit as well. Not sure why/how, but I guess this PR is not that large anyway so I keep as is.

### DIFF
--- a/applications/tune_psf.py
+++ b/applications/tune_psf.py
@@ -169,7 +169,7 @@ if __name__ == "__main__":
         modelVersion=args.model_version,
         label=label,
     )
-    # If we want to start from different values then the one in the model:
+    # If we want to start from values different than the ones currently in the model:
     # align = 0.0046
     # parsToChange = {
     #     'mirror_reflection_random_angle': '0.0075 0.125 0.0037',


applications/tune_psf.py
```diff
@@ -169,6 +169,14 @@ def loadData(datafile):
         modelVersion=args.model_version,
         label=label,
     )
+    # If we want to start from different values then the one in the model:
```
GernotMaier, line 172: Seems to be an incomplete sentence, what exactly do you mean here?

GernotMaier (approved):
I don't get that comment; after that good to merge.

applications/tune_psf.py
orelgueta: I guess a better sentence would be `If we want to start from values different than the ones currently in the model`. Would that be clear (I don't explain further on purpose)?

Either way the test failed for so I will fix it first.

GernotMaier: Good.

orelgueta:
I am merging even though the test has failed because I am sure that once merged the tests will pass. I made a mess with the various commits in different branches and different pull requests and I don't want to waste time trying to figure it out. If it fails after merging I will open a new PR.

PR #263, opened 2022-06-14
Difference plotting

orelgueta:
Allow to plot the difference as well, not just the ratio.


simtools/visualize.py
```diff
@@ -364,14 +373,14 @@ def plot1D(data, **kwargs):
         plt.title(title, y=1.02)
     if "_default" not in list(dataDict.keys()) and not noLegend:
         plt.legend()
-    if not plotRatio:
+    if not plotRatio or plotDifference:
```
GernotMaier, line 376: I am not entirely sure what this tests: is it `not plotRatio` or `not plotDifference` or `(not plotRatio) or plotDifference`? 
(I know that python is probably clear on this, but it is hard to read.

GernotMaier (approved):
Fine with me - have a look at the one line where I think readability could be improved.

simtools/visualize.py
orelgueta: This is a mistake, it should be `if not (plotRatio or plotDifference):`

PR #268, opened 2022-06-16
Changes triggered by codacy issues

GernotMaier:
Mostly minor improvements; also some test improvements.

### DIFF
--- a/simtools/db_handler.py
+++ b/simtools/db_handler.py
@@ -190,7 +190,7 @@ class DatabaseHandler:
             self._logger.debug(
                 "Exporting model files from local model file directories"
             )
-            for par, value in parameters.items():
+            for value in parameters.values():
 
                 if not self._isFile(value):
                     continue
--- a/simtools/db_handler.py
+++ b/simtools/db_handler.py
@@ -113,7 +113,8 @@ class DatabaseHandler:
 
         return _dbClient
 
-    def _getTelescopeModelNameForDB(self, site, telescopeModelName):
+    @staticmethod
+    def _getTelescopeModelNameForDB(site, telescopeModelName):
         """Make telescope name as the DB needs from site and telescopeModelName."""
         return site + "-" + telescopeModelName
 
@@ -176,7 +177,7 @@ class DatabaseHandler:
 
         if cfg.get("useMongoDB"):
             self._logger.debug("Exporting model files from MongoDB")
-            for par, info in parameters.items():
+            for _, info in parameters.items():
                 if not info["File"]:
                     continue
                 file = self._getFileMongoDB(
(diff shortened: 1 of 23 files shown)


simtools/db_handler.py
```diff
@@ -176,7 +178,7 @@ def exportModelFiles(self, parameters, dest):
 
         if cfg.get("useMongoDB"):
             self._logger.debug("Exporting model files from MongoDB")
-            for par, info in parameters.items():
+            for _, info in parameters.items():
```
orelgueta, line 181: In the case you changed below, you changed it instead to iterate only on the values. It's worth being consistent I think and the other solution is nicer so I suggest to implement it here as well.

simtools/model/camera.py
```diff
@@ -505,7 +507,7 @@ def _calcEdgePixels(self, pixels, neighbours):
 
         edgePixelIndices = list()
 
-        for i_pix, xyPixPos in enumerate(zip(pixels["x"], pixels["y"])):
+        for i_pix, _ in enumerate(zip(pixels["x"], pixels["y"])):
```
orelgueta, line 510: Since the positions are not used in this for loop (this is an excerpt from a larger code), the for loop should change to be `for i_pix in range(len(pixels["x"])):`

simtools/simtel/simtel_histograms.py
```diff
@@ -97,7 +97,7 @@ def _combineHistogramFiles(self):
             countFile = True
             with EventIOFile(file) as f:
 
-                for i, o in enumerate(yield_toplevel_of_type(f, Histograms)):
+                for _, o in enumerate(yield_toplevel_of_type(f, Histograms)):
```
orelgueta, line 100: We can remove the enumerate if the `i` is not used anywhere

simtools/visualize.py
```diff
@@ -466,7 +466,7 @@ def plotTable(table, yTitle, **kwargs):
 
     xAxis = table.keys()[0]
     dataDict = OrderedDict()
-    for i_col, column in enumerate(table.keys()[1:]):
```
orelgueta, line 469: Same, we can remove the enumerate since i_col is not used.

simtools/db_handler.py
GernotMaier: Agree. 0f918a3

simtools/model/camera.py
GernotMaier: OK - aa735c8

simtools/simtel/simtel_histograms.py
GernotMaier: OK. 35c61af

orelgueta:
Looks good, merging.

BTW, the commit numbers you quote in your answers are not converted to links unfortunately. Therefore, you can either skip mentioning them (since i assume it takes a bit of effort) or you take a quick look why it doesnt.

PR #270, opened 2022-06-20
Ci unittests experimental

GernotMaier:
(intermediate merge; without relevance)


simtools/config.py
```diff
@@ -44,6 +44,8 @@ def loadConfig(fileName=None):
     ----------
     fileName: str, optional
         Config file name.
+    ignoreGlobals: bool
```
orelgueta, line 47: Should probably be `useGlobals` instead.

tests/integration_tests/test_ray_tracing.py
```diff
@@ -64,9 +55,8 @@ def test_ssts(telescopeModelName):
     ray.analyze(force=True)
 
 
-@pytest.mark.skipif(not has_db_connection(), reason=DB_CONNECTION_MSG)
-@pytest.mark.skipif(not simtel_installed(), reason=SIMTEL_MSG)
-def test_rx():
+@pytest.mark.skip(reason='ignore unused rx method')
```
orelgueta, line 58: It's not always used, true, but should it really be skipped in general? I know it fails now (I have an issue to fix it), but I don't know if we should skip it.

orelgueta (commented):
This PR is not so small, maybe we can discuss it quickly on Monday/Tuesday?

PR #278, opened 2022-07-04
Add option to require workflow config as argument

GernotMaier:
This is a temporary solution, as we probably require later on that all applications require a workflow configuration file (I expect that this will finally replace the config.yml files). 

For now, this small changes helps when running this application.


orelgueta (approved):
All good.

PR #284, opened 2022-07-06
Init a new DB for test data

orelgueta:


simtools/db_handler.py
```diff
@@ -56,7 +56,7 @@ class DatabaseHandler:
     DB_CTA_SIMULATION_MODEL = "CTA-Simulation-Model"
     DB_CTA_SIMULATION_MODEL_DESCRIPTIONS = "CTA-Simulation-Model-Descriptions"
 
-    ALLOWED_FILE_EXTENSIONS = [".dat", ".txt", ".lis"]
+    ALLOWED_FILE_EXTENSIONS = [".dat", ".txt", ".lis", ".cfg"]
```
GernotMaier, line 59: These are file extensions for data tables submitted to the DB? Wondering, as (e)csv is not included in the list.

GernotMaier (approved):

simtools/db_handler.py
orelgueta: Only because we don't have any ecsv files in the DB yet. At the time I decided to build it based on need. However, I can add it preemptively to this PR if you wish.

GernotMaier: No worries - go ahead, we add it when it is required.

orelgueta: OK, merging for now. Can come back to this when we need other types of files.

Issue #275, opened 2022-06-28
Data and configuration files required for unit and integration tests


GernotMaier:
Some of the files used in unit and integrations tests can be retrieved from the databases and are not required to be in the tests/resources directory.

## Files required for unittests:

These files are probably not in the DB:

- [ ]  `tests/unit_tests/test_psf_analysis.py` : `tests/resources/photons-LST-d12.0-za20.0-off0.000_lst_integral.lis`
- [ ]  corsikaConfigTest.yml
- [ ]  (all sim_telarray files *.zstd)

These files are in the DB?

- [ ] tests/unit_tests/test_ray_tracing.py::test_config_data
	- parValues-Sites.yml could not be found
	- parValues-MST-optics.yml could not be found
	- parValues-MST-FlashCam.yml could not be found
	- CTA-North-LST-1-Current_test-telescope-model.cfg
- [ ] tests/unit_tests/test_visualize.py::test_plot_1D - OSError: tests/resources/ref_200_1100_190211a.dat not found
- [ ] tests/unit_tests/test_visualize.py - resources/Transmission_Spectrum_PlexiGlass.dat not found
- [ ] tests/unit_tests/model/test_mirrors.py::test_read_list - tests/resources/mirror_CTA-LST-flen_grouped.dat not found

Not sure what this file does:

- [ ] test_collect_dict_data.yml 

## Files required for integration tests:

- [ ] tests/integration_tests/test_applications.py::test_applications[tune_psf] PSFcurve_data_v2.txt
- [ ] tests/integration_tests/test_applications.py::test_applications[submit_data_from_external::submit] MLTdata-preproduction.usermeta.yml and MLTdata-preproduction.ecsv (are these in DB? Otherwise I would replace them by two 'dummy' files of similar type)
- [ ] test_applications[submit_data_from_external::submit] - set_MST_mirror_2f_measurements_from_external.config.yml (generic configuration file, we will have them at some point in the DB, but probably not now?)
- [ ] tests/integration_tests/test_applications.py::test_applications[produce_array_config]: arrayConfigTest.yml (generic configuration file, ok?)
- [ ] tests/integration_tests/test_applications.py::test_applications[production]: prodConfigTest.yml (generic configuration file, ok?)

GernotMaier:
[...]

PR #288, opened 2022-07-07
Read file from test data db

orelgueta:
This PR partially takes care of #275


simtools/db_handler.py
```diff
@@ -174,7 +176,35 @@ def getModelParameters(
...
+            Location where to write the file to.
+        fileName: str
+            Name of the file to get.
+        """
+
+        self._logger.debug(f"Getting {fileName} and writing it to {dest}")
+        file = self._getFileMongoDB(dbName, fileName)
+        self._writeFileFromMongoToDisk(dbName, dest, file)
+
     def exportModelFiles(self, parameters, dest):
```
GernotMaier, line 197: I comment on the different naming:
- `getFileDB` is getting something from the DB and writing to disk
- `exportModelFiles` is getting something from the DB and writing to disk

GernotMaier (approved):
All good - the comment on the method name is not critical.

simtools/db_handler.py
orelgueta: Yes, this is the result of two people with two different opinions =)
Which name would you prefer? You will be the tie break.

GernotMaier: Whatever - the opinion of the active developer counts.

GernotMaier (approved):

simtools/db_handler.py
orelgueta: Fixed.

orelgueta:
Since I did not merge before, I continued to push to this same branch. Therefore, may I ask you to review this PR again before merging? 

The conflict is strange, I don't understand how that's a conflict considering it's a deleted file and locally I do not get this error. I will try to figure it out in the meantime.

Also, why doesn't this PR run the full unit test? Is it cause it's not merged to master? Can we force it somehow?

GernotMaier (approved):
All good!

PR #296, opened 2022-07-12
Remove pulse shape mention

### LINKED ISSUE #295: Can probably remove pulse shape lines
https://github.com/gammasim/gammasim-tools/blob/ade5a2579f5143c8ecea1a4eb4fd9555ee3776a7/simtools/model/tests/test_array_model.py#L41

orelgueta:
Remove mention of pulse shape which was due to an old issue, but add a different parameter change in order to actually create that telescope. Fixes #295.


GernotMaier (approved):

GernotMaier:
Nice and short...

PR #297, opened 2022-07-12
Remove local calls to the LST config file and remove that file.

### LINKED ISSUE #291: Use CTA-North-LST-1-Current_test-telescope-model.cfg from the DB

orelgueta:
The `test_run_no_db` was removed cause we decided not to allow usage of gammasim-tools without the DB so the test is irrelevant now. Fixes #291.


GernotMaier (approved):
All good, thanks!

PR #298, opened 2022-07-12
Use layout file from DB.

### LINKED ISSUE #294: File ./tests/resources/telescope_positions_prod5_north.ecsv should go into test DB
https://github.com/gammasim/gammasim-tools/blob/b474bcf0b610e5a6967174fe4581bcb8d58ec96f/tests/unit_tests/layout/test_layout_array.py#L29

orelgueta:
Fixes #294


victorbmartin:
Hi Orel. The new function `def testTelescopeFile(db):` has no underline '_' after the 'test'. Will pytest run this test anyways?

orelgueta:
This function is not a test, it's a fixture which is called and run when a different function needs to `testTelescopeFile`. However, I agree it's not a good name cause of this convention. Therefore, I just changed the name to be `telescopeTestFile`. You can see the changes in https://github.com/gammasim/gammasim-tools/pull/298/commits/7eb3d32180d212ab17aff7c136ebf6779bf902ea

VictorBarbosaMartins (approved):
Good to go..

PR #301, opened 2022-07-13
Improve DB tests to use assert or similar

### LINKED ISSUE #292: Check if a real assert can be used in some DB testing

orelgueta:
Fixes #292


orelgueta:
We can ignore the pyflakes error for now, it is fixed in the next PR that will come tomorrow.

GernotMaier (approved):
All good!

PR #303, opened 2022-07-14
Change to return a diameter.

### LINKED ISSUE #243: Change to diameter
https://github.com/gammasim/gammasim-tools/blob/ba8a81d9172f3151afa4a109795c25ea522a0ed3/simtools/psf_analysis.py#L297

This line returns a radius instead of a diameter like all other options. Change to diameter.

orelgueta:
Fixes #243

### DIFF
--- a/simtools/psf_analysis.py
+++ b/simtools/psf_analysis.py
@@ -300,7 +300,7 @@ class PSFImage:
             return 2 * currentRadius
         else:
             self._logger.warning("Could not find PSF efficiently - trying by scanning")
-            return self._findRadiusByScanning(targetNumber, radiusSig)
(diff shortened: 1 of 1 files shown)


VictorBarbosaMartins:
I checked through the code the implications and it seems it is alright. I searched specifically for places where we might be using radius instead of diameter but it seems it is diameter everywhere. Please double check. Thanks!

orelgueta:
I searched before opening the PR, it is used only in one place and there we expect a diameter. If you are happy with it please approve the PR (and if you want merge and delete the branch).

VictorBarbosaMartins (approved):

PR #304, opened 2022-07-14
Updated fileHasText function in simtools/util/general.py to include the mmap module and the re.compile function. Fixes #293

### LINKED ISSUE #293: Use mmaps for simtools/util/general.py fileHasText()
see [reports](https://gitlab.cta-observatory.org/cta-consortium/aswg/simulations/simulation-model/simulation-model-description/-/blob/master/produceReport.py#L2660) or examples on Stackoverflow.

VictorBarbosaMartins:
Updated fileHasText function in simtools/util/general.py to include the mmap module and the re.compile function. Fixes #293

### DIFF
--- a/simtools/util/general.py
+++ b/simtools/util/general.py
@@ -1,6 +1,8 @@
 import logging
 import copy
 from collections import namedtuple
+import re
+import mmap
 
 import astropy.units as u
 from astropy.io.misc import yaml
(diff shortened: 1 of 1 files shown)


simtools/util/general.py
```diff
@@ -48,12 +50,14 @@ def fileHasText(file, text):
     -------
     bool
     """
-    with open(file, "r") as ff:
-        for ll in ff:
-            if text in ll:
+    with open(file, "rb",0) as stringFile, \
+        mmap.mmap(stringFile.fileno(), 0, access=mmap.ACCESS_READ) as textFileInput:
+            re_search_1 = re.compile(br'.*(%s).*'%text.encode())
```
orelgueta, line 55: Do you really need the `.*` around the text?

In addition, we should discuss style tomorrow, but in general use use either the `format` convention to build strings or `f-string` (the latter is preferred).

There's also a question of style of how to break the previous line, but we need to test it cause breaking `with` is always annoying.

orelgueta (commented):
It works, which is what's important, but I made some style comments. Let's discuss tomorrow.

VictorBarbosaMartins:
Thank you for the comments. I updated to include now the f-string format.

PR #305, opened 2022-07-15
Simplify setup

orelgueta:
Remove the need to add simtools to the path each time.


GernotMaier (approved):
Very nice - this was really needed.

Issue #135, opened 2021-04-30
Improve job submission options and add HTCondor / local submission


GernotMaier:
Improve the job submission for corsika / sim_telarray simulations and allow for a variety of job submission system (especially as DESY is moving towards HTCondor.

Goals:
- review current submission process using qsub
- add job submission using HTCondor (best using the python interface as described [here])(https://htcondor.readthedocs.io/en/latest/apis/python-bindings/index.html)
- allow also to run jobs simply in the background without any job submission system (one job at a time; not in parallel)

# HTCondor

Require for now a manual installation of the HTCondor python package (might do this automatically later)

# qsub

Add a one-line check that 'qsub' exists; e.g., when running

python applications/sim_showers_for_trigger_rates.py -a 4LST -s North         --primary proton --nruns 100 --nevents 10000 --output /workdir/

(if it doesn't, the application needs to be ctrl-c to stop)

PR #308, opened 2022-08-19
Improve job submission options

GernotMaier:
This PR adds additional options for running small productions on local workload system (see also issue #135) and removes duplicated code.

New class [simtools/job_submission/job_manager.py](simtools/job_submission/job_manager.py), e.g.:
```
subCmd="qsub -o log_out -e log_err"
# (could also be subCmd="condor_submit" when using HTCondor)
job_manager = JobManager(submitCommand=subCmd)
job_manager.submit(runScript, run_out_file)
```

I've put this classes into a new directory, in the expectation that we might expand this functionality in future and separate it into different classes (e.g., by getting more status information from the workload managers).

To use HTCondor, set 
```
submissionCommand: 'condor_submit '
```
in `config.yml`. 
Almost everything is identical compared to using gridengine. HTCondor provides as well a job information file (ending sublogfile.job with information about the job host and resources required. 

When using HTcondor, use `condor_q` to see your list of running jobs. 

To run a production without workload management system locally, set
```
submissionCommand: 'seriell_script '
```

### DIFF
--- a/applications/production.py
+++ b/applications/production.py
@@ -7,10 +7,16 @@
 
     The simulations are split into two stages: showers and array.
     Shower simulations are performed with CORSIKA and array simulations \
-    with sim_telarray.
+    with sim_telarray. Note that either shower or array simulations are \
+    submitted (so typically you first run shower simulations, and then the \
+    array simulations).
 
     A configuration file is required. See tests/resources/prodConfigTest.yml \
     for an example.
+
+    The workload management system used is given in the configuration file. \
+    Allowed systems are qsub (using gridengine), condor_submit \
+    (using HTcondor), and seriell_script (running the script locally).
 
     Command line arguments
     ----------------------
@@ -41,6 +47,8 @@
     .. code-block:: console
 
         python applications/production.py -t simulate -p tests/resources/prodConfigTest.yml --test
+
+    Running shower simulations.
 """
 
 import logging
--- a/applications/production.py
+++ b/applications/production.py
@@ -244,8 +244,7 @@ def main():
     if not args.showers_only:
         arraySimulators = dict()
         for primary, configData in arrayConfigs.items():
-            aa = ArraySimulator(label=label, configData=configData)
-            arraySimulators[primary] = aa
+            arraySimulators[primary] = ArraySimulator(label=label, configData=configData)
         # Running Arrays
         for primary, array in arraySimulators.items():
 
--- a/applications/production.py
+++ b/applications/production.py
@@ -228,7 +228,7 @@ def main():
 
             if args.task == "simulate":
                 print("Running ShowerSimulator for primary {}".format(primary))
-                shower.submit()
+                shower.submit(test=args.test)
 
             elif args.task == "list":
                 print(
@@ -251,7 +251,7 @@ def main():
             inputList = showerSimulators[primary].getListOfOutputFiles()
             if args.task == "simulate":
                 print("Running ArraySimulator for primary {}".format(primary))
-                array.submit(inputFileList=inputList)
+                array.submit(inputFileList=inputList, test=args.test)
 
             elif args.task == "lists":
                 print(
--- a/applications/production.py
+++ b/applications/production.py
@@ -205,7 +205,7 @@ def proccessSimulationConfigFile(configFile, primaryConfig, logger):
     return label, configShowers, configArrays
 
 
-if __name__ == "__main__":
+def main():
 
     args = parse(description=("Air shower and array simulations"))
 
@@ -273,3 +273,6 @@ if __name__ == "__main__":
             elif args.task == 'resources':
                 print('Printing computing resources report for primary {}'.format(primary))
                 array.printResourcesReport(inputList)
+
+if __name__ == "__main__":
+    main()
--- a/applications/production.py
+++ b/applications/production.py
@@ -274,5 +274,6 @@ def main():
                 print('Printing computing resources report for primary {}'.format(primary))
                 array.printResourcesReport(inputList)
 
+
 if __name__ == "__main__":
     main()
--- a/applications/production.py
+++ b/applications/production.py
@@ -98,7 +98,7 @@ def parse(description=None):
             "iron",
         ],
     )
-    group = parser.add_mutually_exclusive_group()
+    group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument(
         "--array_only",
         help="Simulates only array detection, no showers",
--- a/applications/production.py
+++ b/applications/production.py
@@ -16,7 +16,7 @@
 
     The workload management system used is given in the configuration file. \
     Allowed systems are qsub (using gridengine), condor_submit \
-    (using HTcondor), and seriell_script (running the script locally).
+    (using HTcondor), and local (running the script locally).
 
     Command line arguments
     ----------------------
@@ -217,8 +217,6 @@ def main():
     label, showerConfigs, arrayConfigs = proccessSimulationConfigFile(
         args.productionconfig, args.primary, logger)
 
-    submitCommand = "more " if args.test else None
-
     # ShowerSimulators
     showerSimulators = dict()
     for primary, configData in showerConfigs.items():
@@ -230,7 +228,7 @@ def main():
 
             if args.task == "simulate":
                 print("Running ShowerSimulator for primary {}".format(primary))
-                shower.submit(submitCommand=submitCommand)
+                shower.submit()
 
             elif args.task == "list":
                 print(
@@ -243,19 +241,18 @@ def main():
                 shower.printResourcesReport()
 
     # ArraySimulators
-    arraySimulators = dict()
-    for primary, configData in arrayConfigs.items():
-        aa = ArraySimulator(label=label, configData=configData)
-        arraySimulators[primary] = aa
-
     if not args.showers_only:
+        arraySimulators = dict()
+        for primary, configData in arrayConfigs.items():
+            aa = ArraySimulator(label=label, configData=configData)
+            arraySimulators[primary] = aa
         # Running Arrays
         for primary, array in arraySimulators.items():
 
             inputList = showerSimulators[primary].getListOfOutputFiles()
             if args.task == "simulate":
                 print("Running ArraySimulator for primary {}".format(primary))
-                array.submit(inputFileList=inputList, submitCommand=submitCommand)
+                array.submit(inputFileList=inputList)
 
             elif args.task == "lists":
                 print(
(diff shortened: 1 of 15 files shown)


applications/production.py
```diff
@@ -7,11 +7,17 @@
...
+    with sim_telarray. Note that either shower or array simulations are \
+    submitted (so typically you first run shower simulations, and then the \
+    array simulations).
 
     A configuration file is required. See tests/resources/prodConfigTest.yml \
     for an example.
 
+    The workload management system used is given in the configuration file. \
+    Allowed systems are qsub (using gridengine), condor_submit \
+    (using HTcondor), and seriell_script (running the script locally).
```
orelgueta, line 19: What is seriell? Should it be serial? Or perhaps clearer, "local"?

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+    ----------
+    submitCommand: str
+       job submission command (allowed is qsub, condor_submit, seriell_script
+    test: bool
+       testing mode without sub submission
+
+    Methods
+    -------
+    test_submission_system
+       Check that the requested workload manager exis
```
orelgueta, line 27: exists

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+
+class JobManager:
+    """
+    JobManager provides an interface to workload managers
+    like gridengine or HTCondor.
+
+    Attributes
+    ----------
+    submitCommand: str
+       job submission command (allowed is qsub, condor_submit, seriell_script
```
orelgueta, line 20: The ``more`` option in the code is not described here (also I am not entirely sure what it achieves, see below).

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        Print submit script to screen for testing
+
+        """
+
+        self._logger.info('Testing script')
+
+        shellCommand = 'more ' + self.run_script
+
+        if not self.test:
```
orelgueta, line 137: I don't understand this, if ``more`` is for testing, why does it run only when ``test`` is false? 
The use of ``more`` to show the user the submission script is good, but I would call that option ``test`` (since I can't figure out the different between ``more`` and ``test`` so far).

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        Run a job script on the command line
+        (no submission to a workload manager)
+
+        """
+
+        self._logger.info('Running script locally')
+
+        shellCommand = self.run_script + \
+            " > " + self.run_out_file + ".out"
```
orelgueta, line 152: Don't we need here to direct the error stream to the ``.err`` file as well? I am not sure about the log (something I always wanted to handle internally actually).

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        except FileNotFoundError:
+            self._logger.error(
+                "Failed creating condor submission file {}".format(
+                    _condor_file))
+
+        shellCommand = self.submitCommand + " " + _condor_file
+        if not self.test:
+            os.system(shellCommand)
+        else:
+            self._logger.info('Testing (HTcondor')
```
orelgueta, line 184: Missing `)`. Why here for example we don't ``more`` the output file?

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+        self,
+        submitCommand=None,
+        test=False
+    ):
+        """
+        JobManager init
+
+        Parameters
+        ----------
+        submitCommand: str
```
orelgueta, line 43: What is expected as the submit command here for example? A full qsub command with ``log_out`` etc.? I think that's what I understood from below, but I am not sure. If so, shouldn't we build it here like it's done for hcondor?

simtools/job_submission/job_manager.py
```diff
@@ -0,0 +1,207 @@
...
+    def _submit_gridengine(self):
+        """
+        Submit a job described by a shell script to gridengine
+
+        """
+
+        thisSubCmd = copy(self.submitCommand)
+        if 'log_out' in thisSubCmd:
+            thisSubCmd = thisSubCmd.replace(
+                'log_out', self.run_out_file + ".out")
```
orelgueta, line 195: I can't find where we inserted ``log_out`` and ``log_err`` in the submitCommand. If we trust the user it can break easily (also I don't see why we need to trust the user). 
Either way, I think we can create the submit command directly instead of this `replace`

simtools/util/names.py
```diff
@@ -949,7 +949,9 @@ def corsikaSubLogFileName(arrayName, site, primary, run, mode, label=None):
     """
     name = 'log-sub-corsika-run{}-{}-{}-{}'.format(run, arrayName, site, primary)
     name += '_{}'.format(label) if label is not None else ''
-    name += '-' + mode + '.log'
+    name += '.log'
+    if len(mode)>0:
+        name += "." + mode
```
orelgueta, line 954: Nowhere in the code did we use `.` to separate info in file names, we always used `-` or `_`. It's probably best not to mix a new convention, no?

orelgueta (commented):
Generally very simple and looks good.

I was wondering though, should we make it a little more automatic? For example, should we simply ask the user to say if they want `batch` simulations (or decide based on the number of jobs requested) and then decide automatically which system to use for submission based on priorities and the available systems (condor --> qsub)? This would allow us to avoid using `config.yml` for this.

Also, what about the requested resources? I couldn't see them in these specific changes. They can also be set automatically based on the compiled sim_telarray version and the number of events per job I guess.

applications/production.py
GernotMaier: Good suggestion - wasn't happy with serial - use local now.

simtools/job_submission/job_manager.py
GernotMaier: We can actually drop the `more` option, as we have the `--test` option (from the command line). This does not submit anything, but prints all the file names to the screen. One can easily check than that e.g., the sim_telarray command is correct and that the model has been exported correctly. 

I've remove everything related to `more`

simtools/job_submission/job_manager.py
GernotMaier: See previous comment.

simtools/job_submission/job_manager.py
GernotMaier: Agree - forgot the error stream. 

What exactly do you want to mean  with `I am not sure about the log...`?

simtools/job_submission/job_manager.py
GernotMaier: Agree - changed it to use the keyword `qsub` only and write always to out and err. Not sure why Raul had the option of specifying this in the config file. I think having a consistent way of doing things between qsub and HTcondor is good.

simtools/job_submission/job_manager.py
GernotMaier: This has been removed.

simtools/util/names.py
GernotMaier: OK - although we are writing so many files to disk with quite a lot of differences in style (e.g. `-` vs `_`).

GernotMaier:
I think I've address all points raised - so read for a second review.

On your comments at the beginning of the review: I think we need to discuss what production.py should do and what not and add some additional functionality in future updates (and not in this PR).

simtools/job_submission/job_manager.py
orelgueta: When submitting to the batch farm we also set a log file to which the logging goes. When running locally, the log goes to stdout. Not sure if through a unix command we can pipe the log (not the stdout) to a log file. 

I know it's possible in the logging package of python, which is something I have been wanting to do for a while now.

applications/production.py
```diff
@@ -243,19 +241,18 @@ def main():
...
     # ArraySimulators
-    arraySimulators = dict()
-    for primary, configData in arrayConfigs.items():
-        aa = ArraySimulator(label=label, configData=configData)
-        arraySimulators[primary] = aa
-
     if not args.showers_only:
+        arraySimulators = dict()
+        for primary, configData in arrayConfigs.items():
+            aa = ArraySimulator(label=label, configData=configData)
```
orelgueta, line 247: Why is this split into two lines? I actually saw you fixed this somewhere, just can't remember if it was here or somewhere else.

simtools/job_submission/job_manager.py
```diff
@@ -104,7 +104,7 @@ def submit(
 
         """
         self.run_script = str(run_script)
-        self.run_out_file = str(run_out_file)
+        self.run_out_file = str(run_out_file).replace(".log", "-log")
```
orelgueta, line 107: Wait, but isn't `log` here the actual extension and therefore should be `.log`? Sorry if I annoy with this point, just want to make sure.

orelgueta: OK now that I read below I actually see that there's another extension which is `.out`. We can chat about it tomorrow quickly, but in that case we probably don't need `-log` at all.

simtools/util/names.py
```diff
@@ -1097,7 +1098,8 @@ def simtelSubLogFileName(run, primary, arrayName, site, zenith, azimuth, mode, l
         arrayName
     )
     name += '_{}'.format(label) if label is not None else ''
-    name += '.log'
-    if len(mode)>0:
-        name += "." + mode
+    if len(mode) > 0:
+        name += "-log." + mode
```
orelgueta, line 1102: See my comment above, we can chat about it quickly tomorrow.

orelgueta (commented):
Made a couple of last comments. I think we are generally done, but thought I would mention anyway.

applications/production.py
GernotMaier: No idea - ` it wasn't me`

orelgueta: OK but this can definitely be made into one line then.

simtools/job_submission/job_manager.py
GernotMaier: OK - removed the log for out/err.

GernotMaier:
Hmm, integration tests are failing for applications not touched, e.g. running tests locally with:
```
python applications/tune_psf.py -s North -t LST-1 --model_version prod4 --data /tmp/pytest-of-maierg/pytest-35/test-data0/output/model/PSFcurve_data_v2.txt --zenith 20 --test --configFile /tmp/pytest-of-maierg/pytest-35/test-data0/config-simtools-test.yml
```
fails with
```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, minimal, minimalegl, offscreen, vnc, webgl, xcb
```

PR #311, opened 2022-08-23
Pre commit hook testv0

GernotMaier:
This is a test of the pre-config config file  https://github.com/gammasim/gammasim-tools/blob/pre-commit-hook-testv0/.pre-commit-config.yaml


GernotMaier:
As a side note here: coday is picking up all the applications without a main()... maybe also worth changing.

GernotMaier:
Closing this and redo it properly with another branch / PR.

PR #313, opened 2022-08-24
Adding a pre-commit cook

GernotMaier:
This adds a pre-commit hook for code formatting and white spaces. 

Initial commit with many changes, as `black` wasn't used for a long time.


GernotMaier:
Note that the review should be on general things: any changes in this PR are from the automatic / requested procedures implemented in the pre-commit hook (e.g., the black formatting, sorting of import, etc). So no comment necessary on individual line changes.

GernotMaier:
On the codacy issues - this is always of type `
variable 'allStructureNames' in global scope should not be mixedCase (N816) `.

I don't understand why the naming convention is different for global and local variables - any ideas? What should we do? Easiest way obviously would be to fix the 8 occurrences in the code. 

I can switch this as well off in codacy - but that would be more a workaround.

(see attachment for all occurrences)
[codacy.pdf](https://github.com/gammasim/gammasim-tools/files/9418611/codacy.pdf)

orelgueta:
I think that global variables should be in all capitals like in https://github.com/gammasim/gammasim-tools/blob/281f60a506aa071e77c06a8c5fbde2ac82470cb8/simtools/db_handler.py#L57-L59

So I suggest we fix these occurrences and try to keep this convention from now on.

orelgueta (approved):
It all looks good. We should all install this pre-commit-hook and instructions for installation should be provided in the README for developers.

GernotMaier:
Agree - I've added sentences to the README and added pre-commit also to the gamma-dev environment.

PR #315, opened 2022-09-05
Fix instructions

orelgueta:


orelgueta:
The test that failed here has nothing to do with PR and is the result of the change of files in the DB (to the correct one). If the change itself is OK, I suggest we merge this.

GernotMaier (approved):

PR #316, opened 2022-09-05
Calc nsb camera efficiency

### LINKED ISSUE #282: NSB camera efficiency calculation
Branch calc-nsb-camera-efficiency includes a new function calcNSBRate():

https://github.com/gammasim/gammasim-tools/compare/calc-nsb-camera-efficiency

I am not entirely sure if it works as intended and what it is need for. Any suggestion what to do with this branch, or delete it?

orelgueta:
Fixes #282.

Missing still the integration test, but probably worth merging first and then implementing the necessary integration test (requires a little bit of thinking).


simtools/db_handler.py
```diff
@@ -568,7 +570,55 @@ def _getSiteParametersMongoDB(self, dbName, site, modelVersion, onlyApplicable=F
         )
 
         query = {
-            "Site": site,
+            "Site": _siteValidated,
+            "Version": _modelVersion,
+        }
+        if onlyApplicable:
+            query["Applicable"] = onlyApplicable
```
GernotMaier, line 577: Why don't you set here `quere["Applicable"] = True`, would be clearer? Or remove the `if onlyApplicable`?

simtools/io_handler.py
```diff
@@ -276,6 +294,21 @@ def getTestModelFile(fileName):
     return directory.joinpath(fileName)
 
 
+def getTestDerivedDirectory():
```
GernotMaier, line 297: Is there a motivation why you hear do the path mkdir in this function, while in getDerivedOutputDirectory, `_getOutputDirectory(filesLocation, label, "derived")` is called?

simtools/model/camera.py
```diff
@@ -223,6 +223,17 @@ def getPixelDiameter(self):
...
         return self._pixels["pixel_diameter"]
 
+    def getPixelActiveSolidAngle(self):
+        """
+        Get the active solid angle of a pixel in sr.
+        """
+
+        pixelArea = self.getPixelDiameter() ** 2
+        # In case we have hexagonal pixels:
+        if self.getPixelShape() == 1 or self.getPixelShape() == 3:
```
GernotMaier, line 233: Not related to this function, but I think it is not a good solution to encode in the pixel shape the pixel rotation angle (I was initially confused that we have two different pixel shapes for hexagonal pixels)

Nothing to fix, just a remark.

simtools/model/telescope_model.py
```diff
@@ -502,14 +519,35 @@ def exportConfigFile(self):
...
             configFilePath=self._configFilePath, parameters=self._parameters
         )
 
+    def exportDerivedFiles(self, fileNames):
+        """
+        Write to disk a file from the derived values DB.
+
+        Parameters
+        ----------
+        dest: str or Path
```
GernotMaier, line 534: Is `dest` a parameter?

simtools/model/telescope_model.py
```diff
@@ -538,6 +576,16 @@ def getConfigDirectory(self):
         """
         return self._configFileDirectory
 
+    def getDerivedDirectory(self):
+        """
+        Get the path where all the files with derived values for are written to.
+
+        Returns
+        -------
+        Path where all the files with derived values for are written to.
```
GernotMaier, line 585: Could remove `for` in this sentence?

simtools/model/telescope_model.py
```diff
@@ -619,6 +667,22 @@ def _loadMirrors(self):
...
 
+    def _loadReferenceData(self):
+        """Load the reference data for this telescope from the DB."""
+        self._logger.debug("Reading reference data from DB")
+        db = db_handler.DatabaseHandler()
+        self._referenceData = db.getReferenceData(self.site, self.modelVersion, onlyApplicable=True)
+
+    def _loadDerivedValues(self):
+        """Load the derived values for this telescope from the DB."""
+        self._logger.debug("Reading reference data from DB")
```
GernotMaier, line 678: I think this should be `Reading derived data from DB`

tests/unit_tests/test_db_handler.py
```diff
@@ -76,6 +76,22 @@ def test_reading_db_sst(db):
         assert pars["camera_pixels"] == 2048
 
 
+def test_get_reference_data(db):
+
+    logger.info("----Testing reading reference data-----")
+    pars = db.getReferenceData("south", "Current")
```
GernotMaier, line 82: Is it by design that this test should fail when the `Current` model changes?

GernotMaier (approved):
Minor things and questions - after that ready to merge.

simtools/db_handler.py
orelgueta: Changed to `query["Applicable"] = True`

simtools/io_handler.py
orelgueta: I copied the way you wrote the other `getTest*Directory()` functions. I think it is because it is because it makes the tests simpler since we do not need to call `cfg.get("outputLocation")`. However, I must admit this is a question I should have asked you when you added those functions.

simtools/model/camera.py
orelgueta: You are right, but as "usual" this comes directly from sim_telarray. Not much we could do about it at the moment.

simtools/model/telescope_model.py
orelgueta: No, relic from previous version of the code.

simtools/model/telescope_model.py
orelgueta: Done.

tests/unit_tests/test_db_handler.py
orelgueta: No, for the others it is but in this case it I changed it to Prod5.

simtools/model/telescope_model.py
orelgueta: True.

orelgueta:
Took care of everything and replied to one question with a question. Can probably merge then but I will let you check and approve first.

simtools/db_handler.py
```diff
@@ -427,7 +427,7 @@ def readMongoDB(
 
         self._logger.debug("Trying the following query: {}".format(query))
         if onlyApplicable:
-            query["Applicable"] = onlyApplicable
+            query["Applicable"] = True
```
GernotMaier, line 430: Still interested why you can't write `query["Applicable"] = onlyApplicable`. Would that not work?

tests/integration_tests/test_applications.py
```diff
@@ -29,7 +29,7 @@
             "-t",
             "LST-1",
             "--model_version",
-            "prod5",
+            "prod4",
```
GernotMaier, line 32: Moving back to prod4?

GernotMaier (approved):
All good - two minor questions, which are independent of the merge request.

simtools/db_handler.py
orelgueta: Because then you will query either those which are False or those which are True. In our case we either want all parameters (i.e., not including that parameter in the query at all) or those which are True.

tests/integration_tests/test_applications.py
orelgueta: No, I reverted the wrong commit apparently. I committed other changes to a different branch that I didn't want to include in this PR and got confused with the commit strings. I am still not good enough with git commands and stuff.

I will fix this, make sure all the tests pass and then merge this.

PR #317, opened 2022-09-07
Round float

orelgueta:


GernotMaier (approved):
I guess you have a good reason to choose 5 decimals (and not 2 or 20).

Otherwise good.

orelgueta:
Sort of. The issue is usually in the last digit of the 64 bit float (i.e. 16th digit), so even 14 decimals would have worked for this purpose. However, for off-axis angles it doesn't make sense to discuss anything below 0.001 degree so I rounded at 5 decimals point to allow going even further. If we want to be safer though, I can change to 14, no problem.

PR #318, opened 2022-09-08
Merge array_simulator and shower_simulator

GernotMaier:
Both modules are very similar and do almost the same thing - merge them to a new `Simulator` class.

Code has been generalised a bit to allow addition of other simulation software in future.

Some changes to the modules in `simtools/corsika` and `simtools/simtel` have been applied to unify the API of the runners (e.g., getting a log file is now `getLogFile` and not `getCorsikaLogFile`, also `run` is always `runNumber`).


simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+
+    configData = {
+        'dataDirectory': '(..)/data',
+        'primary': 'gamma',
+        'zenith': 20 * u.deg,
+        'azimuth': 0 * u.deg,
+        'viewcone': 0 * u.deg,
+        # ArrayModel
+        'site': 'North',
+        'layoutName': '1LST',
```
orelgueta, line 67: Is this really a layout name? Or is it a name the user provides and isn't checked?

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+        Parameters
+        ----------
+        label: str
+            Instance label.
+        simulator: str
+            Type of simulator called (implemented is simtel and corsika).
+        filesLocation: str or Path.
+            Location of the output files. If not given, it will be set from \
+            the config.yml file.
+        simtelSourcePath: str or Path
```
orelgueta, line 128: See comment above about parameter name

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+        self._setSimulationRunner()
+
+    def _set_simulator(self, simulator):
+        """
+        Set and test simulator type
+
+        Parameters
+        ----------
+        simulator: str
+            ype of simulator called (implemented is simtel and corsika).
```
orelgueta, line 155: Typo and comment above about this docstring

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+            the config.yml file.
+        simtelSourcePath: str or Path
+            Location of source of the sim_telarray/CORSIKA package.
+
+        """
+
+        self._simtelSourcePath = Path(cfg.getConfigArg("simtelPath", simtelSourcePath))
+        self._filesLocation = Path(cfg.getConfigArg("outputLocation", filesLocation))
+        try:
+            self._outputDirectory = io.getOutputDirectory(
```
orelgueta, line 184: I am a bit surprised you decided to do it this way. 

So far all of these settings of output locations were done in `io_handler.py`. Including the CORSIKA and array-simulator which you left intact above. Instead, in this case, you decided to put this code here and call the function directly. I don't necessarily mind it, it's still fairly short, but then we are inconsistent with our use of `io_handler`. In particular, you might want to remove the special functions for CORSIKA and array-simulator.

Also, the exception for the FileNotFound is already done in `getOutputDirectoy`, not sure why you need it here as well.

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+            Dict with simulator configuration data.
+
+        """
+        _arrayModelConfig, _restConfig = self._collectArrayModelParameters(configData)
+
+        _parameterFile = io.getDataFile("parameters", "array-simulator_parameters.yml")
+        _parameters = gen.collectDataFromYamlOrDict(_parameterFile, None)
+        self.config = gen.validateConfigData(_restConfig, _parameters)
+
+        self.arrayModel = ArrayModel(label=self.label, arrayConfigData=_arrayModelConfig)
```
orelgueta, line 257: I think this should be done in a separate function since this does not load the config, it sets up the model. I know it might require an additional `if` somewhere, but it might be cleaner.

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+
+    def _validateRunListAndRange(self, runList, runRange):
```
orelgueta, line 259: This is a really nice function

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+
+        """
+
+        _runs_and_files = {}
+
+        if self.simulator == 'simtel':
+            _file_list = self._enforceListType(inputFileList)
+            for file in _file_list:
+                _runs_and_files[self._guessRunFromFile(file)] = file
+        else:
```
orelgueta, line 462: Should check if it's CORSIKA here in case in the future we have something else (like CARE or SST-CAM simulator).

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+
+        Returns
+        -------
+        path
+            Path of the pdf file.
+        """
+
+        figName = None
+
+        if self.simulator == 'simtel':
```
orelgueta, line 542: Should we print a warning if this function is called for a different simulator? Will help avoiding mistakes when writing applications for example.

tests/unit_tests/test_simulator.py
```diff
@@ -0,0 +1,343 @@
...
+    assert array_simulator.simulator == 'simtel'
+
+    shower_simulator._set_simulator('corsika')
+    assert shower_simulator.simulator == 'corsika'
+
+    with pytest.raises(gen.InvalidConfigData):
+        shower_simulator._set_simulator('this_simulator_is_not_there')
+
+
+def test_loadConfigData(array_simulator):
```
orelgueta, line 129: The names of the test functions are inconsistent. Everywhere else in the tests (and in some of the tests here) we only used snake_case for function names. Here you mixed snake_case with camelCase. I know it's annoying, but I suggest to keep our convention.

tests/unit_tests/test_simulator.py
```diff
@@ -0,0 +1,343 @@
...
+
+    assert shower_simulator._validateRunListAndRange(
+        runList=run_list, runRange=None) == list(set(run_list))
+
+    with pytest.raises(InvalidRunsToSimulate):
+        shower_simulator._validateRunListAndRange(
+            runList=[1, 'a', 4], runRange=None)
+
+    assert shower_simulator._validateRunListAndRange(
+        runList=None, runRange=[3, 6]) == list(set([3,4,5,6]))
```
orelgueta, line 170: Why is it a list of a set of a list? Instead of just `[3,4,5,6]`?

tests/unit_tests/test_simulator.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+def test_validateRunListAndRange(shower_simulator):
+
+    assert not shower_simulator._validateRunListAndRange(None, None)
+
+    run_list = [1, 24, 3]
+
+    assert shower_simulator._validateRunListAndRange(
+        runList=run_list, runRange=None) == list(set(run_list))
```
orelgueta, line 163: Why do you have to do `list(set(run_list))`?

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+        simulator: str
+            ype of simulator called (implemented is simtel and corsika).
+
+        Raises
+        ------
+        gen.InvalidConfigData
+
+        """
+
+        self.simulator = simulator.lower()
```
orelgueta, line 163: Shouldn't the order of these two lines be reversed? I.e., do not set the simulator before checking its validity?

simtools/corsika/corsika_runner.py
```diff
@@ -197,14 +197,18 @@ def _loadCorsikaDataDirectories(self):
...
 
-    def getRunScriptFile(self, runNumber=None, extraCommands=None):
+    def getRunScript(self, inputFile=None, runNumber=None, extraCommands=None):
         """
         Get the full path of the run script file for a given run number.
 
         Parameters
         ----------
+        inputFile: str or Path
+            currently not used (kept for symmetry with simtel_runner)
```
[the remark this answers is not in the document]
orelgueta: OK, now that I read everything, I think what you should do is define this function without the extra unused parameter. Then define a new function in the simulator class that will call this function or the sim_telarray equivalent one with the correct arguments. It adds the complexity of an additional function and an `if`, but I think it's more correct than adding a redundant argument.

The other option is to add `*args` to all of these functions. Then you can use it in the sim_telarray functions and ignore it here without having to justify it. You won't be able to set a default value in assignment, but if necessary you could do it within the function.

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+            if thisRuntime:
+                runtime.append(thisRuntime)
+
+        meanRuntime = np.mean(runtime)
+
+        resources = dict()
+        resources["Runtime/run [sec]"] = meanRuntime
+        if nEvents and nEvents > 0:
+            resources["#events/run"] = nEvents
+            resources["Runtime/1000 events [sec]"] = meanRuntime * 1000 / nEvents
```
orelgueta, line 673: You should probably mention here that this is wall time

applications/sim_showers_for_trigger_rates.py
```diff
@@ -121,14 +121,15 @@ def main():
         "cscat": [20, 1500 * u.m, 0],
     }
 
-    showerSimulator = ShowerSimulator(label=label, showerConfigData=showerConfigData)
+    showerSimulator = Simulator(
+        label=label, simulator='corsika', configData=showerConfigData)
```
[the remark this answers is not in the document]
GernotMaier: I think I was commiting from an environment without precommit installed - my fault. Fixed now.

simtools/corsika/corsika_runner.py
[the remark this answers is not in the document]
GernotMaier: Good, you convinced me. I've added now `kwargs` to the getRunScript function, as having a dict for the parameter makes the interface much clearer.

simtools/simtel/simtel_runner.py
```diff
@@ -116,14 +116,14 @@ def getRunScript(self, test=False, inputFile=None, run=None, extraCommands=None)
         self._scriptDir = self._baseDirectory.joinpath("scripts")
         self._scriptDir.mkdir(parents=True, exist_ok=True)
         self._scriptFile = self._scriptDir.joinpath(
-            "run{}-simtel".format(run if run is not None else "")
+            "run{}-simtel".format(runNumber if runNumber is not None else "")
```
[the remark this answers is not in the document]
GernotMaier: But using 1 or a random number wouldn't make it clear that there is no runNumber provided. I think this way it is clear that we don't have one and I would prefer it this way.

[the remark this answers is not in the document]
orelgueta: OK!

simtools/corsika/corsika_runner.py
```diff
@@ -215,7 +213,13 @@ def getRunScript(self, inputFile=None, runNumber=None, extraCommands=None):
         Path:
             Full path of the run script file.
         """
-        runNumber = self._validateRunNumber(runNumber)
+        kwargs = {
+            "runNumber": None,
+            **kwargs,
+            "extraCommands": None,
+            **kwargs,
```
orelgueta, line 220: Twice?

simtools/simtel/simtel_runner_array.py
```diff
@@ -210,10 +210,10 @@ def hasSubLogFile(self, run, mode="out"):
 
         """
 
-        runSubFile = self.getSubLogFile(run=run, mode=mode)
+        runSubFile = self.getSubLogFile(runNumber=runNumber, mode=mode)
         return Path(runSubFile).is_file()
 
-    def getResources(self, run):
+    def getResources(self, runNumber):
```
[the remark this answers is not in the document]
GernotMaier: `getResources` is implemented both in the simtel_runners and in the corsika_runners. This was done for symmetry reasons.

I changed it now to return a dict, with either only the `runTime` or `runTime` and `nEvents`

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+            'viewcone': 0 * u.deg,
+            'cscat': [10, 1500 * u.m, 0]
+        }
+
+
+    Example of configData for array simulations:
+
+    .. code-block:: python
+
+    configData = {
```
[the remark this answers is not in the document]
GernotMaier: No, line 59 following is for array_model == simtel.

simtools/simulator.py
GernotMaier: This is a layout as defined in data/layout/

simtools/simulator.py
```diff
@@ -0,0 +1,740 @@
...
+    printListOfLogFiles():
+        Print list of log files.
+    """
+
+    def __init__(
+        self,
+        label=None,
+        simulator=None,
+        filesLocation=None,
+        simtelSourcePath=None,
```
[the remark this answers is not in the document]
GernotMaier: OK!

simtools/simulator.py
GernotMaier: Fixed.

simtools/simulator.py
GernotMaier: OK

simtools/simulator.py
GernotMaier: Agree on the exceptions - I added it first here and then later in the io_handler.

And I didn't get the other point, which special function do you mean here?

simtools/simulator.py
GernotMaier: I didnt' do anything here...that is Raul's!

simtools/simulator.py
GernotMaier: Moving it somewhere else would make things much more complicated (as I would have to pass on the arrayModelConfig). To make things clearer, I've renamed the functions to `_load_configuration_and_simulation_model`

simtools/simulator.py
GernotMaier: OK!

tests/unit_tests/test_simulator.py
GernotMaier: Same comments as in previous issue.

simtools/simulator.py
GernotMaier: Agree

simtools/simulator.py
GernotMaier: Clarified docstring.

GernotMaier:
Thanks @orelgueta  - many good points. I think I've addressed all of them.

simtools/simulator.py
orelgueta: What I mean is that in `io_handler` we have functions called `getCorsikaOutputDirectory` and `getArraySimulatorOutputDirectory`. These essentially do what you do here, they build the output directory by calling `getOutputDirectory` and providing the label and file location. I am surprised you didn't use them (and maybe added one for sim_telarray) and instead decided to make `getOutputDirectory` public and added this code here. Unless there's a good reason for this, I really recommend changing it and using `io_handler` like we do everywhere else.

simtools/simulator.py
orelgueta: OK!

orelgueta (commented):
Almost everything was addressed indeed, thanks!

Only three points are pending,
1. The use of `io_handler` (see specific comment). I don't mind the change in strategy, but I think it should be well motivated.
2. The `list(set([]))` point, though this is minor and wouldn't block a merge.
3. I cannot tell if you changed the names of the test functions to be consistent, i.e., snake_case. GitHub really drives me crazy seeing all of the changes and my comments. If not, I also highly recommend doing it.

simtools/simulator.py
GernotMaier: Using getCorsikaOutputDirectory and getArraySimulatorOutputDirectory would mean to add ifs to distinguish between the simulation codes. I don't want to do it, it is messy and will get worse with more simulation types

orelgueta: I don't see why one `if` (or 2-3 in the future) is worse than changing the whole philosophy of `io_handler`. In addition, the `if` can be placed in `io_handler` rather than here. However, I will not insist on this point. If you prefer doing it this way, that's fine.

GernotMaier:
see issue #321. I think we should discuss this independently of this merge request.



> On 13. Sep 2022, at 16:14, orelgueta ***@***.***> wrote:
> 
> 
> @orelgueta commented on this pull request.
> 
> In simtools/simulator.py:
> 
> > +        Set file locations, input and output directories
> +
> +        Parameters
> +        ----------
> +        filesLocation: str or Path.
> +            Location of the output files. If not given, it will be set from \
> +            the config.yml file.
> +        simtelSourcePath: str or Path
> +            Location of source of the sim_telarray/CORSIKA package.
> +
> +        """
> +
> +        self._simtelSourcePath = Path(cfg.getConfigArg("simtelPath", simtelSourcePath))
> +        self._filesLocation = Path(cfg.getConfigArg("outputLocation", filesLocation))
> +        try:
> +            self._outputDirectory = io.getOutputDirectory(
> 
> I don't see why one if (or 2-3 in the future) is worse than changing the whole philosophy of io_handler. In addition, the if can be placed in io_handler rather than here. However, I will not insist on this point. If you prefer doing it this way, that's fine.
> 
> —
> You are receiving this because you were assigned.
> 



---
Dr Gernot Maier
CTA

Tel.: +49  33 7627 7598
Internet: www.desy.de/cta
Besucheradresse: 1X07

orelgueta:
What about the names of the test functions?

GernotMaier:
The inconsistency of methods names go through the whole code, and it is not connected to test functions. I always try to replicate the method name which is tested, e.g. `printListOfFile()` --> `test_printListOfFile() or `_set_simulator()` --> `test_set_simulator()`.

Not sure what to do now - in theory we should change it everywhere to be consistent. But I don't even know which way we should go...

orelgueta:
I thought we always have functions in camelCase (e.g., `printListOfFile`) and test names in snake_case (e.g., test_ print_list_of_file`). Are there cases in which it's not like that?

PR #320, opened 2022-09-09
Change to set just the DB rather than simtools

orelgueta:


GernotMaier (approved):
OK! That is a nice and small one...

PR #331, opened 2022-09-20
Fix camera plot and add requested option

### LINKED ISSUE #84: Testing validate_camera_fov.py
Works!

Looking at the plot, different items:

- is the coordinate system x_pix/y_pix in the bottom right rotated correctly? I would have expected that it aligns with the bin edges.
- minor option request: switch off plotting of pixel numbers, plot all, plot first N (could probably be handled with one command line parameter)

validate_camera_fov_North-LST-1_pixelLayout

orelgueta:
Fixes #84.

### DIFF
--- a/applications/validate_camera_fov.py
+++ b/applications/validate_camera_fov.py
@@ -70,6 +70,15 @@ def main():
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--printPixelsID",
+        help=(
+            "Up to which pixel ID to print (default: 50). "
+            "To suppress printing of pixel IDs, set to zero (--printPixelsID 0)."
+            "To print all pixels, set to 'All'."
+        ),
+        default=50,
+    )
 
     args = parser.parse_args()
     label = "validate_camera_fov"
@@ -105,7 +114,19 @@ def main():
     print("Avg. edge radius = {0:.3f} cm\n".format(rEdgeAvg))
 
     # Now plot the camera as well
-    fig = camera.plotPixelLayout(args.cameraInSkyCoor)
+    try:
+        pixelIDsToPrint = int(args.printPixelsID)
+        if pixelIDsToPrint == 0:
+            pixelIDsToPrint = -1  # so not print the zero pixel
+    except ValueError:
+        if args.printPixelsID.lower() == "all":
+            pixelIDsToPrint = camera.getNumberOfPixels()
+        else:
+            raise ValueError(
+                f"The value provided to --printPixelsID ({args.printPixelsID}) "
+                "should be an integer or All"
+            )
+    fig = camera.plotPixelLayout(args.cameraInSkyCoor, pixelIDsToPrint)
     plotFilePrefix = outputDir.joinpath(f"{label}_{telModel.name}_pixelLayout")
     for suffix in ["pdf", "png"]:
         fileName = f"{str(plotFilePrefix)}.{suffix}"
--- a/applications/validate_camera_fov.py
+++ b/applications/validate_camera_fov.py
@@ -61,6 +61,15 @@ def main():
     )
     parser.initialize_telescope_model_arguments()
     parser.initialize_default_arguments(add_workflow_config=False)
+    parser.add_argument(
+        "--cameraInSkyCoor",
+        help=(
+            "Plot the camera layout in sky coordinates "
+            "(akin to looking at it from behind for single mirror telesecopes)"
+        ),
+        action="store_true",
+        default=False,
+    )
 
     args = parser.parse_args()
     label = "validate_camera_fov"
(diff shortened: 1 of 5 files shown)


orelgueta:
CORSIKA compilation fails all of a sudden??

GernotMaier:
It fails with the line
```
checking whether the Fortran 77 compiler works... no
```
The answer should be yes (see older actions). Not sure why, is that a newer host system? Nothing changed since the last successful test 6 days ago.

I pressed 'rerun' at the top of the last successful job, let's wait a couple of minutes if this works: https://github.com/gammasim/gammasim-tools/actions/runs/3052873818/jobs/4999722211

applications/validate_camera_fov.py
```diff
@@ -96,12 +114,24 @@ def main():
...
     # Now plot the camera as well
-    fig = camera.plotPixelLayout()
-    plotFileName = label + "_" + telModel.name + "_pixelLayout"
-    plotFile = outputDir.joinpath(plotFileName)
-    for f in ["pdf", "png"]:
-        fig.savefig(str(plotFile) + "." + f, format=f, bbox_inches="tight")
-    print("\nPlotted camera in {}\n".format(plotFile))
+    try:
+        pixelIDsToPrint = int(args.printPixelsID)
+        if pixelIDsToPrint == 0:
```
GernotMaier, line 119: Why don't we require as an argument on the command line to run `--printPixelsID -1` instead of `--printPixelsID 0`? This would not require this line.

simtools/model/camera.py
```diff
@@ -177,21 +176,13 @@ def _rotatePixels(self, pixels):
...
-        is saved in the const dictionary CAMERA_ROTATE_ANGLE.
-        In the case of dual mirror telescopes, the axis is flipped in order to keep the same
-        axis definition as for single mirror telescopes.
-        One can check if the telescope is a two mirror one with isTwoMirrorTelescope.
         """
 
-        if isTwoMirrorTelescope(self._telescopeModelName):
-            pixels["y"] = [(-1) * yVal for yVal in pixels["y"]]
-
         rotateAngle = pixels["rotateAngle"]  # So not to change the original angle
```
GernotMaier, line 181: What does the comment 'so not to change...' mean? Is it that we don't want to change `pixels['rotationAngle']`? If yes, then the comment should say this.

simtools/model/camera.py
```diff
@@ -177,21 +176,13 @@ def _rotatePixels(self, pixels):
...
-        rotateAngle += np.deg2rad(CAMERA_ROTATE_ANGLE[self._cameraName])
+        # The original pixel list is given such that x -> North, y -> West,
+        # i.e., x points down and y points to the right (for single-mirror telescopes).
```
GernotMaier, line 183: I don't understand the connection of North -> down and West -> right. This is from looking at a map?

simtools/model/camera.py
```diff
@@ -809,10 +814,15 @@ def plotPixelLayout(self):
         plt.tick_params(axis="both", which="major", labelsize=15)
 
         self._plotAxesDef(plt, self._pixels["rotateAngle"])
+        description = "For an observer facing the camera"
+        if cameraInSkyCoor and not isTwoMirrorTelescope(self._telescopeModelName):
+            description = "For an observer behind the camera looking through"
+        if isTwoMirrorTelescope(self._telescopeModelName):
+            description = "For an observer looking from secondary to camera"
```
GernotMaier, line 821: I guess it is clear here that it is the secondary?

simtools/model/model_parameters.py
```diff
@@ -42,22 +42,6 @@
...
-    "FlashCam": 270,
-    "NectarCam": 248.214,
-    "SCT": 90,
-    "ASTRI": 90,
-    "1M": 270,
-    "GCT": 90,
-    "SST": 90,
-}
-
 CAMERA_RADIUS_CURV = {
```
GernotMaier, line 45: I guess this will also go at some point (independent of this PR)

GernotMaier (approved):
Very minor comment - so I approve and let you decide to merge now or address them.

applications/validate_camera_fov.py
orelgueta: I thought about it, but that's less intuitive to the user and I preferred to hide the ugliness of the implementation from the user. Also, in the case, if the user sets it to zero, it would print just the zero pixel which would be strange.

simtools/model/camera.py
orelgueta: In the next line I add 90 degrees to the rotation angle. I don't want to do `pixels["rotateAngle"] += np.deg2rad(90)` so not to change the value in `pixels["rotateAngle"]` (might be used later). That comment is so trivial that it might be confusing, so maybe I should remove it. I will think about it.

simtools/model/camera.py
orelgueta: The full explanation is (excerpt from an e-mail from Konrad):

```
x -> North, y -> West, z -> Up in the ground system.
At azimuth=0, zenith angle=0 all coordinate systems are aligned.
When you turn the telescope the "normal" way towards
the horizon, the x-axis points downwards, the z-axis
points in any case from (primary) dish towards camera
```

I wanted to avoid writing so much text in the code so I shortened it, but I guess it is necessary to understand. So I will add this entire explanation (unless you think that even with that it is not clear).

simtools/model/camera.py
orelgueta: I had `the` in the sentence but it became too long in the plot. I assume it is clear enough for the people looking at the these plots.

simtools/model/model_parameters.py
orelgueta: Yes! Kept this file open to fix this next (I don't think it's even used anywhere).

orelgueta:
I tend to ignore the Codacy error of increased complexity. A method for plotting the camera is bound to be complex. I can break it up into pieces just to make Codacy happy, but those pieces would not be useful anywhere else so I don't see a reason for it. 

All other comments were addressed/answered above, so I will merge (after a 2 hour grace period).

GernotMaier:
Agree, all good.

Issue #241, opened 2022-04-28
Check why the ray tracing does not repeat the simulation 20 times as expected


orelgueta:

orelgueta:
The issue is that this line

https://github.com/gammasim/gammasim-tools/blob/19d0133b0400c641ea7c5f6bf31bb52238677d85/simtools/simtel/simtel_runner.py#L68

is not overwritten by the setting in the inheriting class (ray-tracing). Related to that, the test setting is not propagated either. Should be an easy fix.

orelgueta:
The ray tracing actually does repeat the run 20 times as expected. The `simulate(force=False)` option was not propagated, but this will be fixed in #333. Perhaps the test option will be fixed in the same PR.

orelgueta:
Test is also propagated correctly so in the end this issue was not really an issue, except it made us fix the forcing of simulate which is important.

orelgueta:
Fixed in #333.

PR #333, opened 2022-09-22
Remove CAMERA_RADIUS_CURV

orelgueta:
The parameter itself is simply not necessary. To make a long story short, it's not the camera radius curvature but the mirror. It is used in testeff to calculate the effective mirror area, but instead of using that effective mirror area, we use the one derived from detailed ray tracing. Therefore, this parameter can simply be removed.

On the way there, a few other issues were "discovered" and fixed with the PR.

### DIFF
--- a/simtools/simtel/simtel_runner_ray_tracing.py
+++ b/simtools/simtel/simtel_runner_ray_tracing.py
@@ -97,7 +97,6 @@ class SimtelRunnerRayTracing(SimtelRunner):
 
         # File location
         self._baseDirectory = io.getOutputDirectory(self._filesLocation, self.label, "ray-tracing")
-        self._baseDirectory.mkdir(parents=True, exist_ok=True)
 
         self._singleMirrorMode = singleMirrorMode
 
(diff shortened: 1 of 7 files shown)


orelgueta:
In the end all sorts of other things were included in this PR. It also fixes #241.

simtools/simtel/simtel_runner_ray_tracing.py
```diff
@@ -237,7 +242,11 @@ def _checkRunResult(self, runNumber=None):
 
     def _isPhotonListFileOK(self):
         """Check if the photon list is valid,"""
+        nLines = 0
         with open(self._photonsFile, "r") as ff:
-            nLines = len(ff.readlines())
+            for _ in ff:
```
GernotMaier, line 247: Looks definitely more complicated - any particular reason why the old implementation didn't work?

GernotMaier (approved):
All good - please merge.

simtools/simtel/simtel_runner_ray_tracing.py
orelgueta: It was significantly slower, it read the entire file (~2 million lines) just to check if it has more than a 100 lines.

PR #336, opened 2022-09-28
Minor improvements for the production application

### LINKED ISSUE #283: Review application production.py
Related to additions in branch https://github.com/gammasim/gammasim-tools/compare/comp-resources

Notes:

- task `simulate` submits successfully CORSIKA and sim_telarray jobs
    -  Need to understand better why not all log files are in the expected place (possibly add log_out to sub string?)
    - some sim_telarray runs are not completing successfully, but unclear why.
- task `lists` is not implemented
- task `inspect` needs to be tested (fails at this point due to some missing runs, see issue above)
- task `resources` fails as submission log files do not exist.

GernotMaier:
This addresses the remaining points of issue #283: added functionality to printout the produced list of simulated corsika or sim_telarray files.

Closes #283

### DIFF
--- a/applications/production.py
+++ b/applications/production.py
@@ -29,7 +29,7 @@
         What task to execute. Options:
             simulate (perform simulations),
             lists (print list of output files)
-            inspect (plot sim_telarray histograms for quick inspection) [NOT IMPLEMENTED]
+            inspect (plot sim_telarray histograms for quick inspection)
             resources (print quicklook into used computational resources)
     array_only (activation mode)
         Simulates only array detector (no showers).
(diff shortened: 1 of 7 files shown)


VictorBarbosaMartins:
Hi Gernot. It looks good, though I was wondering if it would be more elegant to define a list with the choices for the "--task argument", i.e. "["simulate", "lists", "inspect", "resources"]" and then use a "for loop" through the "elif statements" instead of concatenating them in the code. Probably we would need another list with the texts to be printed too. what do you think?

GernotMaier:
Good point - I have simplified those lines a little bit - required to have a consistent naming of the functions call in the simulator model following the command line parameters (e.g., use simulate() instead of submit(). Please have another look.

orelgueta (approved):
All good, no comments from my side. I would suggest to change `filelist` to file-list or `file_list`, but that is very minor and can be ignored.

VictorBarbosaMartins (approved):
Good!

GernotMaier:
> All good, no comments from my side. I would suggest to change `filelist` to file-list or `file_list`, but that is very minor and can be ignored.

Do we actually have a logic when to use snake vs camel case? I am very confused by that right now.

orelgueta:
We have a logic everywhere except for input arguments from the user. There we never discussed it and I would vote for snake case.

Issue #258, opened 2022-06-10
The readout of the Rx program fails


orelgueta:
The output of the Rx program has some header lines which need to be skipped (are they new?).

GernotMaier:
Rx program require revisit, see also  also integration test in `tests/integration_tests/test_ray_tracing.py:test_rx` (at this point switched off)

PR #340, opened 2022-09-29
Fix usage of rx for ray tracing

GernotMaier:
Address #258 and allow to use the rx tool for the determination of the containment radius.

Change call to rx using subprocess with `Shell=True`.


GernotMaier:
Don't understand why Codacy is not running. It should also address the reported subProcess issue.

orelgueta (approved):
Looks good, once the test finishes we can merge.

PR #341, opened 2022-09-30
Telescope naming

### LINKED ISSUE #334: Check for naming of telescopes as L-/M/S- and replace by LST-, MST-, etc.
CTAO Naming of telescopes is `LSTS` or `MSTN`. Introduce here that telescopes are named with three letter codes: LST, MST, SST, SCT.

This includes also:

https://github.com/gammasim/gammasim-tools/blob/0ef78194dc002bec3c007b493f5ea8af042429cb/simtools/simulator.py#L330

Check what is done here and why it is needed - this is an incomplete list of telescope types.

GernotMaier:
Introduce telescope type names closer to the CTAO standard. Use now consistently LST, MST, SST, SCT instead of `[L-, M-, S-]`.

(note that for the CTAO convention, the site is added; e.g. MSTN-05)

The line in simtools/simulator.py is the one I am not so sure about (I didn't liked it before).

Closes #334


simtools/simulator.py
```diff
@@ -333,7 +333,7 @@ def _collectArrayModelParameters(self, configData):
             raise
 
         # Reading telescope keys
-        telKeys = [k for k in _restData.keys() if k[0:2] in ["L-", "M-", "S-"]]
+        telKeys = [k for k in _restData.keys() if k[1:4] in ["ST-", "SC-"]]
```
orelgueta, line 336: I don't understand this line. This would pick up all telescopes except SCTs. For SCT this should be `CT-`, no?

orelgueta (commented):
One comment on the line you are not sure about where I think there's a bug.

A more general question, shouldn't we take advantage and modify our naming in general to fit with the official CTAO naming? I.e., rather than `North-LST-1`, just `LSTN-1`? We will have an issue with the MSTs since they are referred to as `North-MST-NectarCam` and `South-MST-FlashCam`, but perhaps we can change it to fit with CTAO naming and remove the explicit names `FlashCam` and `NectarCam`. What do you think? Too early?

GernotMaier:
I agree that we should move towards the official names of MSTS and MSTN, etc. I would suggest to do this in a separate PR, as it requires probably quite a few changes (e.g., in names.py) - let's try and keep the PRs smaller. If ok, then I open an issue and start this at some point in the next week.

orelgueta:
OK. Though the largest change would be in the DB. We should discuss whether we want to change the DB completely or just the new versions.

GernotMaier:
(failure in integration test understood and fixed in PR 344 - merge anyway)

PR #343, opened 2022-09-30
Add internal application for adding units to the DB

### LINKED ISSUE #338: Update site center coordinates in DB (adding EPSG, adding units)
Site center parameters are stored in the db without units (e.g. altitude or lon/lat).

Also good to add EPSG, for easier reference:   North is EPSG=32628, for the South it is 32719

orelgueta:
The application added in this PR should not really be used so I did not write much documentation for it nor added it to the tests. We should probably come up with a solution for how to deal with such applications, maybe put them in a separate directory. Fixes #338.

### DIFF
--- a/applications/add_unit_to_parameter_in_db.py
+++ b/applications/add_unit_to_parameter_in_db.py
@@ -1,0 +1,64 @@
+#!/usr/bin/python3
+
+"""
+    Summary
+    -------
+    This application is used to add a unit to parameters in the DB.
+
+    This application should not be used by anyone but expert users and only in extreme cases.
+    Therefore, no additional documentation about this applications will be given.
+
+"""
+
+import logging
+
+import simtools.config as cfg
+import simtools.util.commandline_parser as argparser
+import simtools.util.general as gen
+from simtools import db_handler
+
+
+def main():
+
+    parser = argparser.CommandLineParser(description=("Add a unit field to a parameter in the DB."))
+    parser.initialize_default_arguments(add_workflow_config=False)
+
+    args = parser.parse_args()
+    cfg.setConfigFileName(args.configFile)
+
+    logger = logging.getLogger()
+    logger.setLevel(gen.getLogLevelFromUser(args.logLevel))
+
+    # parsToUpdate = ["altitude"]
+    parsToUpdate = ["ref_long", "ref_lat"]
+
+    # units = ["m"]
+    units = ["deg", "deg"]
+
+    db = db_handler.DatabaseHandler()
+
+    for site in ["North", "South"]:
+        for parNow, unitNow in zip(parsToUpdate, units):
+            allVersions = db.getAllVersions(
+                dbName=db.DB_CTA_SIMULATION_MODEL,
+                site=site,
+                parameter=parNow,
+                collectionName="sites",
+            )
+            for versionNow in allVersions:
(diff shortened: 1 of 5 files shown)


applications/add_unit_to_parameter_in_db.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python3
+
+"""
+    Summary
+    -------
+    This application is used to add a unit to parameters in the DB.
+
+    This application should not be used by anyone but expert users and only in extreme cases.
```
GernotMaier, line 8: Suggest to replace `extreme` by `unusual`.

applications/add_unit_to_parameter_in_db.py
```diff
@@ -0,0 +1,64 @@
...
+    Therefore, no additional documentation about this applications will be given.
```
GernotMaier, line 9: I don't think this is the right thing to do - safety by not giving documentation? Doesn't this require write access to the DB, which most users don't have?

applications/add_unit_to_parameter_in_db.py
```diff
@@ -0,0 +1,64 @@
...
+    cfg.setConfigFileName(args.configFile)
+
+    logger = logging.getLogger()
+    logger.setLevel(gen.getLogLevelFromUser(args.logLevel))
+
+    # parsToUpdate = ["altitude"]
+    parsToUpdate = ["ref_long", "ref_lat"]
+
+    # units = ["m"]
+    units = ["deg", "deg"]
```
GernotMaier, line 36: Oh, now I understand. The changes are hardwired? So you hardly ever expect to this be used?

simtools/db_handler.py
```diff
@@ -1262,3 +1283,60 @@ def insertFilesToDB(self, filesToAddToDB, dbName=DB_CTA_SIMULATION_MODEL):
...
+            _telModelNameValidated = names.validateTelescopeModelName(telescopeModelName)
+            _telNameDB = self._getTelescopeModelNameForDB(_siteValidated, _telModelNameValidated)
+            query["Telescope"] = _telNameDB
+        elif collectionName == "sites":
+            query["Site"] = _siteValidated
+        else:
+            raise ValueError("Can only get versions of the telescopes and sites collections.")
+
+        _allVersions = list()
+        for post in collection.find(query):
```
GernotMaier, line 1336: Ha! Suggest to replace the non-pythonic for loop by list comprehension...:-)

GernotMaier (approved):
Minor comment about the application and documentation.

applications/add_unit_to_parameter_in_db.py
orelgueta: Done

applications/add_unit_to_parameter_in_db.py
orelgueta: As discussed, moved this application to a sub directory called `db_development_tools` for specific tools like that which do not require documentation.

Technically the API user can make these changes as well. We also have a read-only user but we do not use it (maybe we should start).

simtools/db_handler.py
orelgueta: When you're right, you're right!

orelgueta:
Merging, failure in integration test is solved in #344

PR #347, opened 2022-10-05
Static legend handlers

orelgueta:
And a few small changes to the camera plotting, the only important one is a sign change for the rotation of the pixels which was not propagated before.


GernotMaier (approved):
(no comment)

Issue #327, opened 2022-09-14
Remove common parameters


orelgueta:
https://github.com/gammasim/gammasim-tools/blob/39dcc1dfc766c38508d0f3bc7a0f849ce4f44ee3/simtools/simtel/simtel_config_writer.py#L49

We shouldn't have hardcoded common parameters. We can actually take these parameters from the DB entry for each telescope or use the default values Konrad used for Prod6.

orelgueta:
Fixed in #350

Issue #349, opened 2022-10-05
Add a production test which compares to the official production


orelgueta:
We should add a test which generates events equivalent to the latest large production (Prod5 in this case). The comparison should probably be done using the sim_telarray histograms and using the same random seed in both productions. This would probably an integration test, unless we want to invent a new class of tests.

VictorBarbosaMartins:
Hi @orelgueta, this is a rather old issue, but it seems there is some overlap to the tool I was planning to develop (compare histograms). Would such a tool be sufficient, or what do you mean by "a test which generates events equivalent to the latest large production"?

orelgueta:
Actually this test would be simpler than the tool you are planning to develop. You would like to have a tool that compares two simulations on a statistical level to see if they are similar or different.

The idea here is to do the following:
 
- take a small file produced with the current production system (e.g., a DPPS test file produced with the Prod5 package);
- read the random seeds from that file (that will be done manually once)
- run the production tool (the new one) using the exact same random seeds
- compare the first few events and make sure they are exactly the same

So here the plan is to test that the configuration used is exactly equivalent and no mistakes are made in the simtools setup rather than statistically compare between two setups.

orelgueta:
It turns out that making an exact comparison of telescope trigger times or adc samples doesn't really work.

Even when using the exact same configuration and the same random seeds, some rounding difference sometimes causes the random number sequence to change and the output is no longer exactly the same. This happens when running on a different system (e.g., compiling with a different gcc, etc.).

This means this test is not really feasible. Even though I found a specific combination of random seeds that result in identical results, this test will randomly break in the long run. Therefore, I suggest we switch to a statistical test or a combination of the two. Perhaps it's best to discuss in person, but a few suggestions here are also welcome of course.

(BTW, for CORSIKA the values are always the same when using the same seeds and configuration.)

orelgueta:
I tried running on the same system, i.e., inside the container, and I still get slightly different results for certain random seeds. So it is not related to the system/gcc. Perhaps there is something in the configuration I am still missing. I will continue digging.

orelgueta:
It seems getting an exact comparison to work is not really possible and definitely not worth the effort. I am changing the strategy to having a "reasonable output" check and perhaps a statistical comparison in some nightly/weekly tests. I will keep this issue open for now, until these are implemented.

GernotMaier:
@orelgueta -  am going to close this issue. It describes long integration tests, which are discussed and implemented in issue #1867

PR #350, opened 2022-10-05
Remove common pars

orelgueta:
These parameters are already in the DB for all telescopes, so it seems safe to simply remove them. All of the tests passed locally as well. I would have liked to have the test described in #349 for this, but for now I think this PR is safe and we can recheck it when we add the test (and/or compare to Prod6).


orelgueta:
Fixes #327.

GernotMaier (approved):
(no comment)

PR #358, opened 2022-10-18
Create applications/retrieve file from db

### LINKED ISSUE #353: Create applications/retrieve_file_from_db.py
By updating the documentation of the applications (Issue #169 ), we realized that some input files were missing in the directories of gamma-simm tools. After discussions in #169, we decide that we can get the missing files from the DB as it is done in the integration tests.

VictorBarbosaMartins:
Closes #353

Please see the last commit message for more details.


applications/get_file_from_db.py
```diff
@@ -0,0 +1,113 @@
...
+
+    output_path = Path(args.outputFolder)
+    if output_path.exists() == False:
+        print("Folder{} does not exist" "It will be created now".format(args.outputFolder))
+        output_path.mkdir(parents=True, exist_ok=True)
+
+    if args.fileNames is not None:
+        if len(args.fileNames) > 0:
+            db.exportFilesDB(args.dbToGetFrom, args.outputFolder, args.fileNames)
+        elif len(args.fileNames) == 0:
```
orelgueta, line 101: You will never really get here unless it's an empty string.

applications/get_file_from_db.py
```diff
@@ -0,0 +1,113 @@
...
+            db.exportFileDB(args.dbToGetFrom, args.outputFolder, args.fileNames)
+        else:
+            print(
+                "Aborted, {} invalid. Valid formats are: str or list of str.".format(args.fileNames)
+            )
+
+    else:
```
orelgueta, line 108: I guess you will never get here either cause you cannot get a `None` due to the protection `argparse` provides you (required argument which has to be a string).

simtools/db_handler.py
```diff
@@ -198,6 +198,22 @@ def exportFileDB(self, dbName, dest, fileName):
         file = self._getFileMongoDB(dbName, fileName)
         self._writeFileFromMongoToDisk(dbName, dest, file)
 
+    def exportFilesDB(self, dbName, dest, fileNames):
```
orelgueta, line 201: I don't think we need to create this function. The two reasons I see are the following,
1. The user can easily just put this for loop in their application, it doesn't make a huge difference to the user to have it in the API instead.
2. Even if we do want to have this functionality in the API, it's much better to have a single function that deals both with a string or a list of strings then having two separate functions.

simtools/io_handler.py
```diff
@@ -96,3 +96,37 @@ def getInputDataFile(parentDir=None, fileName=None, test=False):
     else:
         filePrefix = Path(cfg.get("dataLocation")).joinpath(parentDir)
     return filePrefix.joinpath(fileName).absolute()
+
+
+def deleteOutputFile(fileName=None):
```
orelgueta, line 101: I think you added these functions solely for the use of the test. However, as I explain below, they are not needed for the test. I also don't really see why they would be needed by a user of gammasim-tools. So, I suggest not adding them.

tests/unit_tests/test_db_handler.py
```diff
@@ -316,23 +316,37 @@ def test_separating_get_and_write(db):
...
 
     logger.info("----Testing inserting files to the DB-----")
     logger.info(
         "Creating a temporary file in {}".format(io.getOutputDirectory(dirType="model", test=True))
     )
-    fileName = io.getOutputDirectory(dirType="model", test=True) / "test_file.dat"
-    with open(fileName, "w") as f:
-        f.write("# This is a test file")
+    outputDir = io.getOutputDirectory(dirType="model", test=True)
+    fileNames = ["test_file.dat", "test_file2.dat"]
```
orelgueta, line 326: I assume you added this additional file in order to test your new function which exports a list of files. As I mentioned above, I don't think we need this function in the API, so the test can go back to it's original form.

tests/unit_tests/test_db_handler.py
```diff
@@ -316,23 +316,37 @@ def test_separating_get_and_write(db):
...
+        assert file_id == db._getFileMongoDB("sandbox", file)._id
 
-    file_id = db.insertFileToDB(fileName, "sandbox")
-    assert file_id == db._getFileMongoDB("sandbox", "test_file.dat")._id
+    logger.debug("Getting files from DB to {}".format(outputDir))
+    db.exportFilesDB("sandbox", outputDir, fileNames)
+
+    logger.debug("Checking if files were written to {}".format(outputDir))
+    for file in fileNames:
+        assert io.getOutputFile(outputDir / file, dirType="model", test=True).exists()
```
orelgueta, line 338: This test was not in the original function, but I think it's a good one to have, so please keep this line (just not in a for loop).

tests/unit_tests/test_db_handler.py
```diff
@@ -316,23 +316,37 @@ def test_separating_get_and_write(db):
...
 
     logger.info("Dropping the temporary files in the sandbox")
     db.dbClient["sandbox"]["fs.chunks"].drop()
     db.dbClient["sandbox"]["fs.files"].drop()
 
+    logger.info("Dropping the temporary files in the output directory")
+    io.deleteOutputFiles([outputDir / fileNames[step] for step in range(len(fileNames))])
```
orelgueta, line 345: This is actually not necessary. The test system uses a pytest location for all output that is removed periodically. You do not need to delete the files explicitly. Therefore, you can remove these lines and the functions in `io_handler`

orelgueta (commented):
Please see comments for requested changes

applications/get_file_from_db.py
```diff
@@ -0,0 +1,113 @@
...
+    -------
+    This application gets a file from the DB.
+
+    The name and location of the file are required.
+    This application should complement the ones for getting parameters \
+    and adding entries and files to the DB.
+
+    Command line arguments
+    ----------------------
+    fileNames (str or list of str, required)
```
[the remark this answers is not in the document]
VictorBarbosaMartins: In this case, we have to change also all the parameters from all functions where it is still lower/upper case. See, for instance, production.py and add_file_to_db.py for other functions.  I am changing it here but for the rest we need another issue.

applications/get_file_from_db.py
```diff
@@ -0,0 +1,113 @@
...
+        Name of the file(s) to get including the full path. \
+        i.e., python applications/get_file_from_db.py -f file_1.dat.
+    outputFolder (str)
+        Name of the local output directory where to save the files. \
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Ok, agreed.

applications/get_file_from_db.py
```diff
@@ -0,0 +1,113 @@
...
+        If it does not exist, it will be created.
+    db (str)
```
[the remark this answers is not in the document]
VictorBarbosaMartins: well, at some point you might have different DBs (or not?). Furthermore, if the same file is in different DBs, retrieving the file wouldnt be straightforward. In that case, I would prefer to keep the DB option.

simtools/db_handler.py
VictorBarbosaMartins: Ok, I would go for your second point. In this case we should also merge insertFileToDB and insertFilesToDB in simtools/db_handler.py to match the style here.

tests/unit_tests/test_db_handler.py
VictorBarbosaMartins: Since I will merge, for simplicity, both functions into one and they will deal both with a single file and a list of files, I will keep the test (and rather adapt it to the new function). Do you agree in this case to keep the test?

tests/unit_tests/test_db_handler.py
VictorBarbosaMartins: good to know, thanks

applications/get_file_from_db.py
```diff
@@ -0,0 +1,113 @@
...
+    logger = logging.getLogger()
+    logger.setLevel(gen.getLogLevelFromUser(args.logLevel))
+
+    output_path = Path(args.outputFolder)
+    if output_path.exists() == False:
+        print("Folder{} does not exist" "It will be created now".format(args.outputFolder))
+        output_path.mkdir(parents=True, exist_ok=True)
+
+    if args.fileNames is not None:
+        if len(args.fileNames) > 0:
```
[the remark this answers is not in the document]
VictorBarbosaMartins: My mistake. It should be greater than 1.

applications/get_file_from_db.py
VictorBarbosaMartins: As I said, my mistake, it should be greater than 1 in the code. thanks for spotting that

applications/get_file_from_db.py
VictorBarbosaMartins: Ok, I will remove this else statement.

applications/get_file_from_db.py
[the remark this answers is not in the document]
orelgueta: Agree, we will change them slowly as we go along.

applications/get_file_from_db.py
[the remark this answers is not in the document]
orelgueta: We already have a few DBs (three to be exact), that's why I said I think you should simply go through all of them to search for the file. The case in which we have the same file in different DBs is not very likely. However, in that case you would simply provide the first one you find (and report which DB you got it from). 

You can keep the DB option if you want, but I highly recommend not making it a required parameter. The user (including myself), shouldn't really remember the names of the DBs or type them on the command line for such a simple use case.

applications/get_file_from_db.py
orelgueta: Actually, I think you should remove the original `if`. Since it's a required option, you will never have it as `None`.

simtools/db_handler.py
orelgueta: Notice that `insertFilesToDB` doesn't include the protection in the case it's not a list (perhaps it's worth changing it as well, not sure). Also it is divided into two functions which is unnecessary in the export case (might be worth changing the insert case as well). What I would do is change the export method do be something like the usual

```
if not isinstance(file, list):
    filesToExport = [file]
else:
    filesToExport = file
for fileNow in filesToExport:
    self._logger.debug(f"Getting {fileName} and writing it to {dest}")
    file = self._getFileMongoDB(dbName, fileName)
    self._writeFileFromMongoToDisk(dbName, dest, file)
```

tests/unit_tests/test_db_handler.py
orelgueta: Sure, that's fine

applications/get_file_from_db.py
orelgueta: OK, but what do you want to do if it's zero? I think that since you anyway will have just one function that can deal with a string or a list, you can simply remove this entire piece of code and pass the file names to that function, regardless if it's a list or not.

tests/unit_tests/test_db_handler.py
orelgueta: Actually, now you will have to do both. You will have to have a test where you pass a list to `exportFileDB` and one where you pass just a string.

simtools/db_handler.py
```diff
@@ -182,21 +180,26 @@ def getModelParameters(
...
             Name of the DB to search in.
         dest: str or Path
             Location where to write the file to.
-        fileName: str
+        file_name: str
             Name of the file to get.
+        Returns
+        -------
+        GridOut
+            :class:`~gridfs.grid_file.GridOut` or ``None`` if file not found
```
orelgueta, line 196: You are not actually returning GridOut, you are returning the file ID. The thing is, I don't think the file ID is so useful as a return value (nor the GridOut). If you wrote a file, that's it, no additional info is necessary in the return value. If the file was not found you are raising an error (I think we should add that info about the raising of the error to the docstring).

simtools/db_handler.py
```diff
@@ -182,21 +180,26 @@ def getModelParameters(
...
         """
 
         self._logger.debug(f"Getting {fileName} and writing it to {dest}")
-        file = self._getFileMongoDB(dbName, fileName)
-        self._writeFileFromMongoToDisk(dbName, dest, file)
+        filePathInstance = self._getFileMongoDB(dbName, fileName)
+        self._writeFileFromMongoToDisk(dbName, dest, filePathInstance)
+        return filePathInstance._id
```
orelgueta, line 202: See above, I don't think it's necessary.

simtools/db_handler.py
```diff
@@ -1267,63 +1271,44 @@ def _getTaggedVersion(dbName, version="Current"):
 
         return tags["Tags"][version]["Value"]
 
-    def insertFileToDB(self, file, dbName=DB_CTA_SIMULATION_MODEL, **kwargs):
+    def insertFileToDB(self, fileName, dbName=DB_CTA_SIMULATION_MODEL, **kwargs):
         """
-        Insert a file to the DB.
+        Insert file to the DB.
```
orelgueta, line 1276: The `a` should come back I guess cause it's just one file now.

simtools/db_handler.py
```diff
@@ -1267,63 +1271,44 @@ def _getTaggedVersion(dbName, version="Current"):
...
+            The name of the file to insert (full path).
         dbName: str
             the name of the DB
-        file: str or Path
-            The name of the file to insert (full path).
         **kwargs (optional): keyword arguments for file creation.
             The full list of arguments can be found in, \
             https://docs.mongodb.com/manual/core/gridfs/#the-files-collection
             mostly these are unnecessary though.
+            kwargs can be used only in case of passing a single file to the function.
```
orelgueta, line 1288: This is not necessary anymore I guess

simtools/db_handler.py
```diff
@@ -1267,63 +1271,44 @@ def _getTaggedVersion(dbName, version="Current"):
...
 
         Returns
         -------
         file_id: gridfs "_id"
-            If the file exists, returns the "_id" of that one, otherwise creates a new one.
+            If the file exist, returns its "_id", otherwise creates the file and returns its id.
```
orelgueta, line 1293: Both sentences are not great I think. I suggest to write

`If the file exists, return its "_id", otherwise insert the file and return its newly created DB "_id".`

tests/integration_tests/test_applications.py
```diff
@@ -189,6 +189,15 @@
         ]
     ],
     # Database
+    "get_file_from_db": [
+        ["--file_name", "mirror_CTA-S-LST_v2020-04-07.dat"],
+        ["--file_name", "PSFcurve_data_v2.txt"],
+        [
+            "--file_name",
+            "ray-tracing-North-LST-1-d10.0-za20.0_validate_optics\
```
orelgueta, line 197: You should break this string differently to be consistent with how we do it everywhere else. It should be

```
(
"ray-tracing-North-LST-1-d10.0-za20.0_validate_optics"
".ecsv"
)
```

(ignore the indentation I wrote above, just the fact that it's closed within parenthesis and without the `\`

tests/unit_tests/test_db_handler.py
```diff
@@ -325,6 +325,15 @@ def test_separating_get_and_write(db):
         assert io.getOutputFile(fileNow, dirType="model", test=True).exists()
 
 
+def test_export_file_db(db, db_cleanup_file_sandbox, caplog):
+    logger.info("----Testing exporting files from the DB-----")
+    outputDir = io.getOutputDirectory(dirType="model", test=True)
+    fileName = "mirror_CTA-S-LST_v2020-04-07.dat"
+    fileToInsert = outputDir / fileName
```
orelgueta, line 332: Not to insert, but rather `fileToExport`

orelgueta (commented):
Minor comments

applications/get_file_from_db.py
```diff
@@ -0,0 +1,96 @@
...
+        "sandbox",
+        "test-data",
+    ]
+    output_path = Path(args.output_directory)
+    if output_path.exists():
+        for db_name in available_dbs:
+            try:
+                db.exportFileDB(db_name, args.output_directory, args.file_name)
+            except FileNotFoundError:
+                continue
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Good point

simtools/db_handler.py
VictorBarbosaMartins: Well spotted, it is the ID that is returning. I do think it is good to return the id, at least for the case where we test the insert function. probably the user will not care for the id, in which case he/she can just ignore it and not even assign a variable for keeping it, no problem. I will add the warning to the docstring

simtools/db_handler.py
VictorBarbosaMartins: I propose we keep this return statement. it will help us to test the function and it does no harm to the code and the user

tests/integration_tests/test_applications.py
VictorBarbosaMartins: I break now for different lists of arguments, it should be more readable. Please have a look

applications/get_file_from_db.py
```diff
@@ -0,0 +1,104 @@
...
+                    "Got file {} from DB {} and saved into {}".format(
+                        args.file_name, dbName, args.output_directory
+                    )
+                )
+                break
+            except FileNotFoundError:
+                continue
+
+        if fileId is None:
+            raise FileNotFoundError
```
orelgueta, line 97: Please add an error message to the logger before you raise, saying that you couldn't find `file_name` in any of the available DBs.

applications/get_file_from_db.py
```diff
@@ -0,0 +1,104 @@
...
+
+    else:
+        print("Aborted, directory{} does not exist".format(args.output_directory))
```
orelgueta, line 100: Shouldn't there be a space before `{}`?

simtools/db_handler.py
```diff
@@ -182,21 +180,32 @@ def getModelParameters(
...
+        Get file from the DB and write to disk.
 
         Parameters
         ----------
         dbName: str
             Name of the DB to search in.
         dest: str or Path
             Location where to write the file to.
-        fileName: str
+        file_name: str
```
orelgueta, line 191: It says `file_name` here but the parameter passed is actually `fileName`

simtools/db_handler.py
```diff
@@ -182,21 +180,32 @@ def getModelParameters(
...
             Name of the file to get.
+
+        Returns
+        -------
+        GridOut._id
+            :class:`~gridfs.grid_file.GridOut._id`
```
orelgueta, line 197: I don't mind keeping the return ID, but maybe write this in a more friendly way? I.e., 
```
fileID: GridOut._id
    the database ID the file was assigned when it was inserted to the DB.
``` 
It's a bit clearer than GridOut info which someone has to look up.

simtools/db_handler.py
```diff
@@ -1267,63 +1282,45 @@ def _getTaggedVersion(dbName, version="Current"):
...
             The full list of arguments can be found in, \
             https://docs.mongodb.com/manual/core/gridfs/#the-files-collection
             mostly these are unnecessary though.
 
         Returns
         -------
-        file_id: gridfs "_id"
-            If the file exists, returns the "_id" of that one, otherwise creates a new one.
+        GridOut._id
+            :class:`~gridfs.grid_file.GridOut._id`
```
orelgueta, line 1303: As mentioned above, I don't think 
```
:class:`~gridfs.grid_file.GridOut._id`
```
is necessary, the text below it is enough.

orelgueta (commented):
Really minor comments left.

simtools/db_handler.py
VictorBarbosaMartins: I am using not fileName

simtools/db_handler.py
VictorBarbosaMartins: Hmm... I copied from the grid module exactly as they use there. shouldnt we be consistent?

simtools/db_handler.py
VictorBarbosaMartins: same comment as above, should we keep it for consistency?

orelgueta (approved):

PR #369, opened 2022-10-26
Fix gridfs return object

orelgueta:
Add also a test to make sure we get the warning we expect


VictorBarbosaMartins:
Good. wasnt really necessary but I will integrate into my branch and merge it

VictorBarbosaMartins (approved):

PR #373, opened 2022-10-28
Fix adding file to DB application

orelgueta:
Modify the add_file_to_db.py application to comply with the changes made in #358

### DIFF
--- a/applications/add_file_to_db.py
+++ b/applications/add_file_to_db.py
@@ -141,8 +141,9 @@ def main():
     print(*filesToInsert, sep="\n")
     print()
     if _userConfirm():
-        db.insertFilesToDB(filesToInsert, args.dbToInsertTo)
-        logger.info("File{} inserted to {} DB".format(plural, args.dbToInsertTo))
+        for fileToInsertNow in filesToInsert:
+            db.insertFileToDB(fileToInsertNow, args.dbToInsertTo)
+            logger.info(f"File {fileToInsertNow} inserted to {args.dbToInsertTo} DB")
     else:
         logger.info(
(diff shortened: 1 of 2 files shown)


VictorBarbosaMartins:
Interesting that no test detected that. I would suggest to add an integration test for this application. I know we dont want to write in the DB every time we test, but if we work only with the sandbox and clean it afterwards, as for the dbExportDb test, this shouldnt be a problem.

VictorBarbosaMartins:
There is another mention to this deprecated function here:
Please adjust it to use only insertFileToDB
https://github.com/gammasim/gammasim-tools/blob/742c77a5380e330d89ca6c77a7f074ea0c74ffb1/simtools/db_handler.py#L1241-L1244

orelgueta:
> There is another mention to this deprecated function here: Please adjust it to use only insertFileToDB
> 
> https://github.com/gammasim/gammasim-tools/blob/742c77a5380e330d89ca6c77a7f074ea0c74ffb1/simtools/db_handler.py#L1241-L1244

Oh, I remember discussing this one in your PR and I even remember you changing it. Guess it didn't propagate. Either way, I made that change now.

orelgueta:
> Interesting that no test detected that. I would suggest to add an integration test for this application. I know we dont want to write in the DB every time we test, but if we work only with the sandbox and clean it afterwards, as for the dbExportDb test, this shouldnt be a problem.

As much as I was against it before, I agree that this application should also be tested. It would have solved two issues in the two latest PRs. I will add a test to it but in a separate PR if you don't mind.

VictorBarbosaMartins (approved):
Good to go. Please remember to include the integration test in the new PR (you can assign me to do it as well :) )

Issue #368, opened 2022-10-25
Review file name generation / validation in names.py


GernotMaier:
names.py includes both validation of telescope elements and the generation of file names (e.g. corsikaRunLogFileName). The latter should be separated and moved possibly to the corresponding corsika/simtel directories.

orelgueta:
Solved in #379.

PR #379, opened 2022-11-10
Remove repetitive getting file names methods and move them to the appropriate place

### LINKED ISSUE #309: Review repetive method calls with many parameters
"Rich" (many parameters) and repetitive method calls makes the code sometimes hard to read (as it is e.g., hard to see if there is something different in the list of arguments or not). Wondering if there isn't a better solution, like working with dictionaries or so. 

Try and understand if there is another solution.

Example:
```
132     def getLogFile(self, run):
133         """Get full path of the simtel log file for a given run."""
134         fileName = names.simtelLogFileName(
135             run=run,
136             primary=self.config.primary,
137             arrayName=self.arrayModel.layoutName,
138             site=self.arrayModel.site,
139             zenith=self.config.zenithAngle,
140             azimuth=self.config.azimuthAngle,
141             label=self.label,
142         )
143         return self._simtelLogDir.joinpath(fileName)
144
145     def getSubLogFile(self, run, mode='out'):
146         ''' Get full path of the simtel log file for a given run. '''
147         fileName = names.simtelSubLogFileName(
148             run=run,
149             primary=self.config.primary,
150             arrayName=self.arrayModel.layoutName,
151             site=self.arrayModel.site,
152             zenith=self.config.zenithAngle,
153             azimuth=self.config.azimuthAngle,
154             label=self.label,
155             mode=mode
156         )
157         return self._simtelLogDir.joinpath(fileName)
158
159     def getHistogramFile(self, run):
160         """Get full path of the simtel histogram file for a given run."""
161         fileName = names.simtelHistogramFileName(
162             run=run,
163             primary=self.config.primary,
164             arrayName=self.arrayModel.layoutName,
165             site=self.arrayModel.site,
166             zenith=self.config.zenithAngle,
167             azimuth=self.config.azimuthAngle,
168             label=self.label,
169         )
170         return self._simtelDataDir.joinpath(fileName)
```

orelgueta:
Fixes #309 and #368.


simtools/simtel/simtel_runner_array.py
```diff
@@ -121,92 +121,102 @@ def _load_simtel_data_directories(self):
...
+                run: int
+                    Run number.
+                primary: str
+                    Primary particle (e.g gamma, proton etc).
+                zenith: float
+                    Zenith angle (deg).
+                azimuth: float
+                    Azimuth angle (deg).
+                site: str
+                    Paranal or LaPalma.
```
GernotMaier, line 163: Wondering if we need to say here that it has to be Paranal or La Palma. The code would work with any site, or?

simtools/simtel/simtel_runner_array.py
```diff
@@ -121,92 +121,102 @@ def _load_simtel_data_directories(self):
...
+        elif file_type == "histogram":
+            return self._simtel_log_dir.joinpath(f"{file_name}.hdata.zst")
+        elif file_type == "output":
+            return self._simtel_data_dir.joinpath(f"{file_name}.simtel.zst")
+        elif file_type == "sub_log":
+            suffix = ".log"
+            if "mode" in kwargs:
+                suffix = f".{kwargs['mode']}"
+            return self._simtel_log_dir.joinpath(f"log_sub_{file_name}{suffix}")
+        else:
```
GernotMaier, line 201: This else is not needed.

GernotMaier (approved):
All good, very minor comments. Good to see that it saved 200 lines, despite additional testing.

I still notice the many very similar methods in the corsika and simtel runners, but let's leave it like that.

simtools/corsika/corsika_config.py
```diff
@@ -458,23 +433,73 @@ def _get_text_multiple_lines(pars):
...
-            view_cone=self._user_parameters["VIEWCONE"],
-            label=self.label,
+    def get_file_name(self, file_type, run_number=None):
+        """
+        Get a CORSIKA config style file name for various file types.
+
+        Parameters
+        ----------
+        file_type: str
+            The type of file it is (determines the file suffix).
```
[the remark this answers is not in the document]
orelgueta: Done

simtools/simtel/simtel_runner_array.py
orelgueta: Yes, this specific function would work with any site so I could remove this. In general we of course force it to be Paranal or La Palma so it would fail elsewhere I assume. Either way, removed it from here.

simtools/simtel/simtel_runner_array.py
orelgueta: Why is it not needed? Don't we want to raise an error if an unknown option is given?

orelgueta:
I will wait with the merge for your reply about the unnecessary `else`.

I agree there are a few more functions which appear duplicated. There is the option of moving some of them to Simulator and making cases for them (essentially like I did in this PR), but then you risk making the code more complicated than it should be.

It is good to see a reduction of 200 lines though, experts say the best PRs are those which reduce code rather than add it.

simtools/simtel/simtel_runner_array.py
GernotMaier: Because all other if statements have a return value (oh no - see now that `file_type` == sub_log does not always result in a return. I think that it is fine. 

But what should happen new mode is not in kwargs?

orelgueta: All if statements have a return value, also the case of `file_type == "sub_log"`. The else is in case the `file_type` provided is not known. In that case I could either return a default file name (which I wouldn't know what it would be) or raise an error. I chose the latter.

> But what should happen new mode is not in kwargs?

if `mode` is not in kwargs then the suffix of the file is `.log` by default.

GernotMaier: OK - so you don't need the last else, as there is no decisions to be made. Anything, super small thing and some code linters will complain about it. Fine with me to keep it.

orelgueta: Finally understood what you mean, removed it. Will wait for tests to pass and then merge.

PR #380, opened 2022-11-11
Including set env db.sh

VictorBarbosaMartins:
Following the new scheme for configuration files, I included the set_env_db_template.sh in the gammasim-tools repository, updating the documentation.


docs/source/conf.py
```diff
@@ -24,7 +24,7 @@
 
 project = "gammasim-tools"
 copyright = "2022, gammasim-tools developers"
-author = "Raul R Prado and Orel Gueta and Gernot Maier"
+author = "Raul R Prado and Orel Gueta and Gernot Maier and Victor Barbosa Martins"
```
GernotMaier, line 27: Really wondering at how many places we have our names. Seems to me that there must be a better way to do that and have one single list of authors. But good that you found that!

docs/source/getting_started.rst
```diff
@@ -29,4 +29,12 @@ Each time you want to use the package, just activate the conda environment:
 
     $ conda activate gammasim-tools-dev
 
+If you need access to the DB, ask one of the developers, update the login \
+credentials in set_env_db_template.sh and source it to activate the \
```
GernotMaier, line 33: I thought you would want that users copy the template file to `set_env_db.sh` and not modify directly the template? Otherwise git will always complaint.

docs/source/getting_started.rst
```diff
@@ -29,4 +29,12 @@ Each time you want to use the package, just activate the conda environment:
...
+environmental variables:
+
+.. code-block:: console
+
+    $ source set_env_db_template.sh
+
 You are all set now =)
```
GernotMaier, line 40: I suggest to remove the smiley. Sometimes things don't work and I found smiley than always very annoying =)

set_env_db_template.sh
```diff
@@ -0,0 +1,10 @@
+export DB_API_PORT=27017 #Port on the MongoDB server
+export DB_SERVER='cta-simpipe-protodb.zeuthen.desy.de' # MongoDB server
+export DB_API_USER=YOUR USERNAME # username for MongoDB: ask the responsible person
```
GernotMaier, line 3: Suggest to have `YOUR_USERNAME`, as spaces are not allowed. Same for the password.

set_env_db_template.sh
```diff
@@ -0,0 +1,10 @@
+export DB_API_PORT=27017 #Port on the MongoDB server
```
GernotMaier, line 1: Add a shebang (that will make codacy also to be happy)

GernotMaier (approved):
Minor suggestions - please address them and merge then.

docs/source/getting_started.rst
VictorBarbosaMartins: thats a better idea. thanks

set_env_db_template.sh
VictorBarbosaMartins: Good, thanks for the advise, I didnt have to look into the codacy report

docs/source/conf.py
VictorBarbosaMartins: agreed. we need to make it more general and be able to change in a single place, I will have a look at it later

VictorBarbosaMartins:
I addressed the comments and now the tests are passing, so I am merging it. If something more is needed please undo it and let me know.

PR #388, opened 2022-11-15
Change logging levels for simtel_runner

### LINKED ISSUE #90: Logging levels
Running derive_mirror_rnda.py produces too many INFO printouts. Most of these are from the simtel_runner. I guess for many other applications we would want to print every call to sim_telarray, however maybe we could dynamically, based on the application running, change this printouts to be DEBUG level?

GernotMaier:
Addresses issue #90 and suppresses a lot of prints to the screen when running with default verbose options.

Closes #90

### DIFF
--- a/simtools/simtel/simtel_runner.py
+++ b/simtools/simtel/simtel_runner.py
@@ -163,7 +163,7 @@ class SimtelRunner:
             self._logger.info("Running (test) with command:{}".format(command))
             self._run_simtel_and_check_output(command)
         else:
-            self._logger.info("Running ({}x) with command:{}".format(self.RUNS_PER_SET, command))
+            self._logger.debug("Running ({}x) with command:{}".format(self.RUNS_PER_SET, command))
             self._run_simtel_and_check_output(command)
 
             for _ in range(self.RUNS_PER_SET - 1):


orelgueta:
This is not really what is described in the issue. I would prefer to have the command printed as INFO in general, just not for every mirror or parameter value. So I would like to change the printout it dynamically. Otherwise I would prefer to keep the printout in INFO and close the issue as "not critical" before changing to DEBUG.

GernotMaier:
OK, I misunderstood you - but why would the general user want to know all the sim_telarray calls?

Would you prefer to have printed out only the first N calls?

orelgueta:
Yes, just the first N calls is what I had in mind.

GernotMaier:
This is hard to do! SimTelRunner instantiated in the loop, so there is no record of previous printout when calling run().

I've also parsed through the python logger options and there is the possibility to add a filter/handler, but it seems to be overkill for this issue (see e.g. this entry on [stack overflow](https://stackoverflow.com/questions/6506146/logging-in-continouous-loop)).

orelgueta:
That's why I opened an issue and didn't just take 10 minutes to do it =)
If it's too complicated we can forget about it. We will anyway separate at some point printouts and entries in the log file. As printouts these messages are somewhat excessive. However, having them in a log file is not a big issue.

orelgueta:
After a short discussion, decided to change the original request in #90 and just change the sim_telarray printouts to be at  DEBUG level.

orelgueta (approved):

PR #389, opened 2022-11-16
Reduce footprint of conda image by >800MB 

GernotMaier:
The footprint of the conda packages is insane and slows down installation and tests.

Two quick changes, which should not have any impact on any results / running:

- change to conda-forge (numpy does not install the 800MB large mkl library)
- change to matploblib-base removes the >150 MB large QT installation, which we don't use)

Related to PR in the container section.


orelgueta (approved):
Nice!

Issue #223, opened 2022-03-25
Add functionality to retrieve gammasim-tools version


GernotMaier:
Software version is hardwired into simtools/version.py

This should be improved in future, see e.g.. the [ctapipe approach](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/version.py)

GernotMaier:
Would be nice to have this one solved - not relevant but simply important to have.

As we are now using setup tools, we should be able to use the solution indicated in  https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/version.py

orelgueta:
Coincidentally, the same was implemented just now in [this PR](https://github.com/cta-observatory/pycorsikaio/pull/17) in corsikaio. Maybe it can be used as an example.

GernotMaier:
I've looked into it and it is too complicated for me. Maybe we can ask Max to add it? Or does one of you get familiar with setuptools?

PR #393, opened 2022-11-17
Automatic extraction of version number and authors

GernotMaier:
Ignore the name of this PR - we should change the theme later to something pretty (e.g., using the [numpydoc extensions](https://numpydoc.readthedocs.io/en/latest/index.html)).

Changes here are to automatize the following:

- read version number from `simtools.version`
- read authors from CITATION.cff file (note that these are filled both as html authors and are available in the RST files using `|author|`

Removed also `docs/_config.yml`, this is a file from previous implementations of the documentation and not used anywhere.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -138,7 +138,7 @@ pygments_style = "sphinx"
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "bizstyle"
+html_theme = "sphinx_rtd_theme"
 
 html_title = f"{project} v{version} Manual"
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,7 +15,7 @@
 import os
 import sys
 
-from astropy.io.misc import yaml
+import yaml
 
 import simtools.version
 
@@ -31,7 +31,7 @@ def get_authors_from_citation_file(file_name):
     """
     try:
         with open("../../CITATION.cff") as file:
-            citation = yaml.load(file)
+            citation = yaml.safe_load(file)
     except FileNotFoundError:
         raise
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,22 +15,51 @@
 import os
 import sys
 
+from astropy.io.misc import yaml
+
+import simtools.version
+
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -15,22 +15,51 @@
...
 
 
+def get_authors_from_citation_file(file_name):
+    """
+    Read list of authors from CITATION.cff file
+
+    """
+    try:
+        with open("../../CITATION.cff") as file:
+            citation = yaml.load(file)
```
orelgueta, line 34: Don't we need to always use `safe_load` or something like that when reading yaml?

docs/source/conf.py
```diff
@@ -15,22 +15,51 @@
...
+        raise
+
+    author = ""
+    try:
+        for person in citation["authors"]:
+            author = author + person["given-names"] + " " + person["family-names"]
+            author += " (" + person["affiliation"] + "), "
+    except KeyError:
+        pass
+    return author[:-1]
```
orelgueta, line 45: Wouldn't this remove the last author?

docs/source/conf.py
```diff
@@ -15,22 +15,51 @@
...
+
+
 # -- Project information -----------------------------------------------------
 
 project = "gammasim-tools"
 copyright = "2022, gammasim-tools developers"
-author = "Raul R Prado and Orel Gueta and Gernot Maier and Victor Barbosa Martins"
+author = get_authors_from_citation_file("../CITATION.cff")
+rst_epilog = """
+.. |author| replace:: {author}
```
orelgueta, line 54: I have never seen this way of replacing. Is this python or some weird RST way of replacing?

orelgueta (approved):
Minor comments.
This doesn't solve the issue of getting the version automatically mentioned in #223, right?

docs/source/conf.py
GernotMaier: No, this removes the last `,` from the string build in this loop.

docs/source/conf.py
GernotMaier: This is weird sphinx stuff: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_epilog

GernotMaier:
And no, this doesn't solve #223 . It just reads the version number from `simtools.version`.

Issue #329, opened 2022-09-14
Integration of telescope_layout tools into gammasim-tools


GernotMaier:
Merge [telescope_layout](https://github.com/gammasim/telescope-layouts) repository into gammasim-tools. This avoids duplication of code. Archive the telescope_layouts repository after the successful integration into gammasim-tools.

Following applications will be included into gammasim-tools:

1. `print_layout.py` (rename to `print_array_elements.py`): prints coordinates of array elements (in most cases telescope positions) in the different used coordinate systems
2. `compare_layouts.py` (rename to `compare_array_elements.py`): compare the positions of two lists of array elements and print differences

Integrate updates in telescope_layout code:
- `telescope-layouts/layout_array.py` --> `simtools/layout/layout_array.py`
- `telescope-layouts/layout_telescope.py` --> `simtools/layout/telescope_position.py`
- review code in simtools/layout at the same time.

Additional functionality:
- read telescope positions from DB and print in positions in different coordinate systems.

Suggest not to include the plotting in this issue (see also Sites branch), but do this as a next step.

GernotMaier:
Some points to be discussed.

`simtools/layout/telescope_position.py`:
- large number of very similar functions (e.g., hasLocalCoordinates(), hasUtmCoordinates, etc). Generalize to `hasCoordinates(system='utm')`.
- units and names are defined in [data/parameters/telescope-position_parameters.yml](https://github.com/gammasim/gammasim-tools/blob/master/data/parameters/telescope-position_parameters.yml) - would simplify the code if we put this as a dict into a separate python module and import it (don't need any reading functions for yaml then, etc). Or is it needed at all? Seems that all units are hardcoded anyway using the `@u.quantity_input` decorators from astropy.units
- some of gammasim-tools classes have `fromKwargs` methods, others not. Same for `__repr__` functions. This is nice, but if we don't use them, do we want it?
- suggest to rename 'local coordinates' to 'CORSIKA coordinates'

`telescope_layout` is a bit complicated. Suggest to add a module called `array_element`, including the coordinates in a given coordinate system only. `telescope_layout` would than have instances for each coordinate system of `array_element`.

GernotMaier:
A couple discussion points, related to coordinate systems and telescope positions (see also draft PR #330)

### 1. Telescope altitude.
Decided to use the altitude (asl) as 'z' position for all coordinate system, including the corsika system. The logic is that the telescope position should be independent of the telescope type (i.e., sphere radius).

[...]

PR #396, opened 2022-11-17
Documentation of auxiliary files / data directory

GernotMaier:
Added one small page to describe the files in the data directory.

Note that I've moved [data/corsika/corsika_parameters.yml](https://github.com/gammasim/gammasim-tools/blob/master/data/corsika/corsika_parameters.yml) into /data/parameter/corsika_parameters.yml, as I had a hard time to write a sentence in the documentation justifying a separate directory (this is why some modules required changes).


orelgueta:
I think we decided to keep some files in this directory to allow for tests which do not require a DB connection. I might remember wrongly though.

VictorBarbosaMartins:
> I think we decided to keep some files in this directory to allow for tests which do not require a DB connection. I might remember wrongly though.

Hmm, I see. If we need those for tests I suggest using tests/resources instead, otherwise, I don't see the reason why we should keep two directories with data

GernotMaier:
Shouldn't start the habit of discussing code changes in pull requests...

In my understanding we do:
- add to database anything which is related to the MC model 
- add to the test/resources directory anything which is needed for tests.

These data files are different:
- the data/parameters files are required to test user input for e.g., correct units. 
- data data/layout files are required to run the trigger threshold application (there is a discussion in #329 on it as well). These are model version independent.

I think this is a consistent approach and ok for now.

In future, we might consider to write the layout files dynamically; it is so simple and therefore not necessary to really keep them as files.

VictorBarbosaMartins (approved):
Ok!

PR #398, opened 2022-11-17
Get parameter type from db and remove model_parameters.py

### LINKED ISSUE #10: Using yml files for parameters instead of modules
There are two modules (corsika_parameters.py and model_parameters.py) that should become yml files in data/parameter-files.

orelgueta:
Fixes #10 (after the decision made in the discussion in that issue).


GernotMaier (approved):
All good!

PR #399, opened 2022-11-17
Udpate doc

### LINKED ISSUE #345: Revision of documentation
The documentation requires a good read and some updates.

- check for missing application and module documentation
- check for consistent interface description of modules
- add add hints regarding the installation of CORSIKA/sim_telarray (issue #172)

VictorBarbosaMartins:
Closes #345 

Hello Orel, Gernot,

finally (as promised) I am sending the PR with the documentation update. It has became much larger than I first thought and it is not yet complete. It was a lot of work for me, because on the fly I am learning about documentation and about the code structure itself.

I updated (chronologically):
- documentation of applications
- rst files
- classes in simtools (but not inside the subdirectories):
   - class methods
   - function parameters
   - return and raise statements
   - class parameters
   - class atributes

We are still missing:
- classes inside subdirectories in simtools

I ignored throughout the code the private functions (starting with an '_'), which I think is fine.

In order to be complete we will have to address those classes inside the subdirectories of simtools as well (plus your comments and suggestions to come). I will be during the morning in the workshop again and will answer several of your points later. Please feel free to work on this branch as well, I am opening the PR for that.

Thank you.
Best regards,
Victor


GernotMaier:
Thanks for this - I will correct smallish typos directly without commenting.

docs/source/framework.rst
```diff
@@ -12,48 +12,53 @@ Framework
 Configuration
```
GernotMaier, line 12: This document needs a complete re-write given the new configurator. I will do this.

tests/integration_tests/test_applications.py
```diff
@@ -66,7 +63,7 @@
             " --test",
         ]
     ],
-    "derive_mirror_rnda::psf_notuning": [
+    "derive_mirror_rnda::psf_measurement": [
```
GernotMaier, line 66: I am trying to understand the difference between `derive_mirror_rnda::psf_measurement` and `derive_mirror_rnda::psf_notuning`. Isn't this exactly the same?

tests/integration_tests/test_applications.py
```diff
@@ -13,12 +13,20 @@
```
GernotMaier, line 13: What is the motivation to re-arrange the sequence of the tests? It makes it hard to compare! 

If there is no strong reason for this, could we go back to the original sequence? The possibility to track changes is important.

simtools/camera_efficiency.py
```diff
@@ -541,7 +588,8 @@ def plot_cherenkov_efficiency(self):
 
         Returns
         -------
-        plt
+        fig
+            The figure
```
GernotMaier, line 592: I see the difficult of writing a good sentence here, but maybe we can discuss more what actually is returned? A matplotlob figure object? A pdf? An ascii figure? A path to a figure?

simtools/camera_efficiency.py
```diff
@@ -574,7 +622,8 @@ def plot_nsb_efficiency(self):
...
```
GernotMaier, line 626: Same as previous comment.

simtools/camera_efficiency.py
```diff
@@ -411,6 +434,11 @@ def calc_tel_efficiency(self):
     def calc_camera_efficiency(self):
         """
         Calculate the camera nominal efficiency including gaps (as defined in B-TEL-1170).
+
+        Returns
+        -------
+        cam_efficiency: float
+            Camera efficiency
```
GernotMaier, line 441: Is this a single value at one wavelength or an array for all wavelengths / offsets? Maybe we can be clearer here.

docs/source/applications.rst
```diff
@@ -7,25 +7,28 @@ Applications
...
-These applications will be used as building blocks for the Simulation System Workflows.
+Applications are python scripts built on the :ref:`Library` that execute a simple, well defined
+task. These applications will be used as building blocks for the Simulation System Workflows.
 
 The application scripts can be found in gammasim-tools/applications.
 
-The applications expect a config file named config.yml in your gammasim-tools directory (unless specified differently
-using the function function config.setConfigFileName) . Please,
-follow the instructions in :ref:`Configuration` to setup your config file.
+The applications expect a config file named config.yml in your gammasim-tools directory (unless
```
GernotMaier, line 15: This is a comment to myself that this paragraph needs an update given the new Configuration class

applications/print_array_elements.py
```diff
@@ -29,12 +29,33 @@
 
     Example:
 
+    First get the array element list
```
GernotMaier: Here you didn't add `get ... from DB`, was that a conscious decision?

applications/derive_mirror_rnda.py
```diff
@@ -174,23 +176,26 @@ def _parse(label):
         "--mirror_list",
         help=(
             "Mirror list file to replace the default one. It should be used if"
-            " measured mirror focal lengths need to be accounted"
+            " measured mirror focal lengths need to be accounted. It contains the following"
```
GernotMaier, line 179: I think we should not describe the columns and data fields of the input files and keep the help messages short (if possible to a single line).

GernotMaier:
@VictorBarbosaMartins - I also suggest not to continue with this PR in the subdirectories for the modules but to do this in a separate PR. It will otherwise be hard to follow up!

applications/add_file_to_db.py
```diff
@@ -6,15 +6,15 @@
     This application adds a file to the DB.
 
     The name and location of the file are required.
-    This application should complement the ones for updating parameters \
-    and adding entries to the DB.
+    This application should complement the ones for updating parameters, \
+    adding entries to the DB and getting file from the DB.
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Good point. I think we can indicate that the DB is chosen by the --db argument, then it will work for all DBs

applications/compare_cumulative_psf.py
```diff
@@ -50,20 +50,32 @@
...
 
-    First, create an yml file named lst_pars.yml with the following content:
+    First, get the yml file named lst_pars_example.yml from the DB:
 
     .. code-block:: yaml
 
-        mirror_reflection_random_angle: '0.0075,0.15,0.035'
-        mirror_align_random_horizontal: '0.0040,28.,0.0,0.0'
-        mirror_align_random_vertical: '0.0040,28.,0.0,0.0'
+        python applications/get_file_from_db.py --file_name lst_pars_example.yml
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Im not familiar with the application but as far as I understood, we wanted to test changing the parameters as well. In any case, as part of the example in the doc string we dont need to cover this case I think, I will remove it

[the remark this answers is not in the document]
VictorBarbosaMartins: Im not familiar with the application but as far as I understood, we wanted to test changing the parameters as well. In any case, as part of the example in the doc string we dont need to cover this case I think, I will remove it

simtools/camera_efficiency.py
```diff
@@ -469,12 +508,20 @@ def calc_reflectivity(self):
...
     def calc_nsb_rate(self):
         """
         Calculate the NSB rate.
+
+        Returns
+        -------
+        nsb_rate
+            NSB rate
+        n1_sum
+            Sum of NSB counts
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I agree adding the units makes more sense. I don't know why this function needs to return the counts, since the integral flux is calculated inside it. I havent found any use in the code for the returned counts, I will remove it

simtools/camera_efficiency.py
```diff
@@ -469,12 +508,20 @@ def calc_reflectivity(self):
             [wl_now > 299 and wl_now < 551 for wl_now in self._results["wl"]]
         ]
         c2_sum = np.sum(c2_reduced_wl)
+        cher_spec_weighted_reflectivity = c2_sum / c1_sum / self._results["masts"][0]
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I only assigned it in order to mention it in the docstring, otherwise it would look ugly (see below), hence, I suggest to keep the defined variable, or any other idea?
"""
        """
        Calculate the Cherenkov spectrum weighted reflectivity in the range 300-550 nm.

        Returns
        -------
        c2_sum / c1_sum / self._results["masts"][0]
            Cherenkov spectrum weighted reflectivity (300-550 nm)
        """
"""

simtools/camera_efficiency.py
```diff
@@ -451,12 +484,18 @@ def calc_tot_efficiency(self, tel_effeciency):
         fill_factor = self._telescope_model.camera.get_camera_fill_factor()
 
         tel_effeciency_nsb = fill_factor * (n4_sum / (masts_factor * n1_sum))
+        tel_total_effeciency = tel_effeciency / np.sqrt(tel_effeciency_nsb)
```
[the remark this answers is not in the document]
VictorBarbosaMartins: same here, the assignment is only to facilitate the docstring, I wouldnt put any calculation in the docstring

docs/source/framework.rst
VictorBarbosaMartins: ok, thanks!

simtools/camera_efficiency.py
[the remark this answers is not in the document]
GernotMaier: Hmm, what does it add if you put a variable name into the docstring? What only setting:

```
Returns
--------
Float
   Cherenkov spectrum weighted reflectivity (300-550 nm)
```

The name of the parameter is not required, see https://numpydoc.readthedocs.io/en/latest/format.html#returns

tests/integration_tests/test_applications.py
VictorBarbosaMartins: Indeed, a `--no_tuning`  argument was missing in the test

tests/integration_tests/test_applications.py
VictorBarbosaMartins: I sorted it alphabetically, as it is displayed in my IDE. it makes more sense to me than chronologically. since we work on the tests throughout time, the chronology loses its meaning. There is the change to organize it according to the module, but the names of the applications do not always refer directly to a module, hence, it would be also difficult. My final suggestion is to keep it alphabetically.

simtools/camera_efficiency.py
VictorBarbosaMartins: That is right. I now looked at the return statement of the respective function and it is a pyplot.fig, so I suggest I write there "Instance of pyplot.figure", "pyplot.figure" or "instance of Figure" (then pyplot is omitted) which one would be better?

simtools/camera_efficiency.py
VictorBarbosaMartins: It is a single value (float) as indicated. It seems to me it is a wavelength-averaged mean. can you confirm that? it is hard to interpret when the variable is called "results"

applications/print_array_elements.py
VictorBarbosaMartins: No, this was probably a lapse of memory. I will add it compatible to the previous case

tests/integration_tests/test_applications.py
GernotMaier: You are right, the current sorting was kind of random. It just now almost impossible to compare the changes. I will have another look, but it makes reviewing much harder.

applications/derive_mirror_rnda.py
VictorBarbosaMartins: I agree. the problem here is that there is no doctring for the function and if we just remove this explanation we lose it for good. I will create a docstring for the function and include the detailed information there and keep it small in the help message

simtools/camera_efficiency.py
GernotMaier: Yes, this is correct.

applications/add_file_to_db.py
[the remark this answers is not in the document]
VictorBarbosaMartins: Ok, I decided instead of it to add the article 'a' before it. it indicates that there are more DBs. Please indicate me again if you don't like it

applications/print_array_elements.py
VictorBarbosaMartins: There is actually a difference between getting and adding a file to DB. In getting a file we don't need to indicate which db it is, we just search through all the available dbs, hence I will add here "from the DBs". On the other hand, in adding a file to DB  we have to indicate which DB, hence I used there "to a DB"

applications/derive_mirror_rnda.py
VictorBarbosaMartins: Ok, so I reduced the size and did not add a docstring (I actually think it would not be necessary)

simtools/camera_efficiency.py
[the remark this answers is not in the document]
VictorBarbosaMartins: Ok, then I will pass only the name, but the consistency will be lost since in all the other functions the variable is returned

applications/add_file_to_db.py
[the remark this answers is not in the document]
GernotMaier: Good

VictorBarbosaMartins:
Ok, I pushed again, please have a second look. I will let you close the conversations in case you are fine with my changes.

simtools/camera_efficiency.py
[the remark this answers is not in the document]
GernotMaier: But you agree that this is clearer and simpler? We should rather apply in future this to the other methods.

simtools/camera_efficiency.py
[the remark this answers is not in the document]
GernotMaier: I don't think 'code should serve docstrings', it should be the other way around...

simtools/camera_efficiency.py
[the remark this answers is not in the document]
VictorBarbosaMartins: Indeed, it is not necessary to know the name of the returned variable since one has to define anyways a new one to save the returned value

simtools/camera_efficiency.py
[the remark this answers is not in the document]
VictorBarbosaMartins: oh, I forgot to change here too

simtools/camera_efficiency.py
```diff
@@ -524,7 +522,7 @@ def calc_nsb_rate(self):
             * self._telescope_model.reference_data["nsb_reference_value"]["Value"]
             / nsb_integral
         )
-        return nsb_rate, n1_sum
+        return nsb_rate
```
GernotMaier, line 524: And you are sure that n1_sum is not sued when this function is called?

GernotMaier (approved):
Have a look at the one question, otherwise good!

simtools/camera_efficiency.py
GernotMaier: But removing it just because we don't know the units is maybe not correct. I think @orelgueta knows the answer here.

VictorBarbosaMartins: I removed it because the function was returning a variable that was never use in the code.

orelgueta: It's a remnant from the first version of the code which I had locally (there it was used). Either Raul or I copied it but removed its usage. It's safe to remove.

tests/integration_tests/test_applications.py
orelgueta: Wait, the order was not random!
I ordered it such that the first 5-10 applications were roughly the quickest ones. Originally we had the optics first which were the slowest tests first and I always had to wait 5-10 minutes just to pass the first ones. The tests were also bunched in Camera/Optics/DB/etc. That I actually liked a lot. Could we go back to the order before this change?

simtools/camera_efficiency.py
orelgueta: `Instance of pyplot.figure` is best

orelgueta (commented):
I couldn't find a way to remove Gernot's approval, so I comment here to please look first at my comment on the order of the applications before you merge.

tests/integration_tests/test_applications.py
VictorBarbosaMartins: Ok, I just returned to the previous order

VictorBarbosaMartins:
> I couldn't find a way to remove Gernot's approval, so I comment here to please look first at my comment on the order of the applications before you merge.

Maybe we should include that information in the developer guidelines regarding the order of the application tests

orelgueta:
Thanks! Feel free to merge then.

PR #397, opened 2022-11-17
Module organisation

GernotMaier:
This PR implements the module organisation as discuss in issue #395.

We will have to deal with the documentation in a separate PR (most of it should be done automatically)

I repeat here the re-organization outlined in the issue:

- new `configuration` directory including configuration.py, commandline_parser.py
- new 'data_model' directory including data_model.py, model_data_writer.py, validate_data.py, validate_schema.py, workflow_description.py
- `job_submission` - rename it to `job_execution`
- `util/model.py`  rename it and move it to model/model_utils.py

Closes #395

PR #403, opened 2022-11-18
Update Documents after module reorganisation

GernotMaier:
This is an update to the files in doc/resources to reflect the recent reorganisation of the modules (PR #397).
Fixes also the warnings / errors reported by sphinx.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -103,7 +103,7 @@ autodoc_mock_imports = [
 # Change the look of autodoc classes
 # napoleon_use_ivar = True
 
-numpydoc_show_class_members = True
+numpydoc_show_class_members = False
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
--- a/docs/source/library.rst
+++ b/docs/source/library.rst
@@ -13,9 +13,11 @@ gammasim-tools is composed of the following modules:
   :maxdepth: 1
   :glob:
 
-  simulators
+  camera_efficiency
+  configuration
+  data_model
+  layout_array
   mc_model
   ray_tracing
-  camera_efficiency
-  layout_array
+  simulators
   util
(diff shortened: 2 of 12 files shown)


docs/source/conf.py
```diff
@@ -103,7 +103,7 @@ def get_authors_from_citation_file(file_name):
 # Change the look of autodoc classes
 # napoleon_use_ivar = True
 
-numpydoc_show_class_members = True
+numpydoc_show_class_members = False
```
VictorBarbosaMartins, line 106: I am not sure this is having an effect, it seems that the class methods are still being listed in the documentation. Check for example the sphinx documentation of classdata_model.validate_data.DataValidator

docs/source/library.rst
```diff
@@ -13,9 +13,11 @@ gammasim-tools is composed of the following modules:
...
-  mc_model
-  ray_tracing
   camera_efficiency
+  configuration
+  data_model
   layout_array
+  mc_model
+  ray_tracing
+  simulators
   util
```
VictorBarbosaMartins, line 23: I am wondering about the modules in general. I would expect to have in this file all the listed modules, or at least all the repositories with modules. I propose we try to adjust this list to the code and dont come up with new names and categories that are not in the code. For instance, in the code we have `simulator.py` and here `simulators`. I also haven't seen the db_handler here. Should we first think about reordering the modules in the code and adjust the list here accordingly?

docs/source/conf.py
GernotMaier: This is the suggested configuration do get rid of hundreds of warnings like
```
simtools/simulator.py:docstring of simulator.Simulator:110: WARNING: autosummary: stub file not found 'simulator.Simulator.simulate'. Check your autosummary_generate setting
```
(and I can't find the website anymore suggesting this, somewhere in numpydocs)

PR #408, opened 2022-11-21
Improvements in metadata validation for submit_data_from_external

GernotMaier:
Maintenance and improvements of the metadata validation used in submit_data_from_external. There should be no change in results or functionality.

In general: this is very fine tuned and quite complicated; not very happy about these modules but I also don't see a better solutions.

Changes:
- renamed data_model.data_model to data_model.meta_data_model, as this module is about meta data only
- rewrite of meta data definitions to avoid duplications and have one single place for the definition of all fields
- this includes a combination of the metadata field info (type, required, default) and the possibility to get the dict with defaults filled (using `_metadata_dict_with_defaults`)
- added a module docstring trying to explain a bit the approach
- correct treatment of metadata entries with no required fields at all (e.g., the `CONTEXT:DOCUMENT`).
- generalised treatments of metadata entries which are list of dicts (e.g., of type `instrumentlist` or `documentlist`)
- fixed `general:change_dict_keys_case` for dicts which consist of list of strings

I've tried to add unittests for these cases to make it more explicit.


simtools/data_model/meta_data_model.py
```diff
@@ -0,0 +1,307 @@
+"""
+Definition of metadata model for input to and output of gammasim-tools.
+Follows CTA top-level data model definition.
+Metadata is broken into hierarchical categories in this module.
+
+Metadata definitions require the three following fields to be defined:
+
+* type: data type (str)
```
orelgueta, line 8: Is the `(str)` an example? I.e., can it be changed to be `(str, int, float, email, etc.)`? Or does it have to be a string? It's a bit confusing to me.

simtools/data_model/meta_data_model.py
```diff
@@ -0,0 +1,307 @@
...
+        "CREATION_TIME": {"type": "datetime", "required": True, "default": None},
+        "ID": {"type": "str", "required": False, "default": None},
+        "DATA": {
+            "CATEGORY": {"type": "str", "required": False, "default": None},
+            "LEVEL": {"type": "str", "required": False, "default": None},
+            "TYPE": {"type": "str", "required": False, "default": None},
+            "MODEL": {
+                "NAME": {"type": "str", "required": False, "default": None},
+                "VERSION": {"type": "str", "required": False, "default": None},
+                "URL": {"type": "str", "required": False, "default": None},
```
orelgueta, line 68: Could we add a `url` type like the email and instrumentlist ones?

simtools/data_model/meta_data_model.py
```diff
@@ -0,0 +1,307 @@
...
+
+    """
+    return {"DOCUMENT": {"type": "documentlist", "required": False, "default": None}}
+
+
+def _category_document():
+    """
+    Metadata field DOCUMENT.
+
+    Described a
```
orelgueta, line 108: Missing description

simtools/data_model/meta_data_model.py
```diff
@@ -0,0 +1,307 @@
...
+        "REFERENCE": _category_reference(),
+        "CONTACT": _category_contact(),
+        "PRODUCT": _category_product(),
+        "INSTRUMENT": _category_instrument(),
+        "PROCESS": _category_process(),
+        "CONTEXT": _category_context(),
+    }
+
+
+def metadata_input_reference_document_list(schema_list):
```
orelgueta, line 212: `schema_list` missing in docstring

simtools/data_model/meta_data_model.py
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+
+def _metadata_dict_with_defaults(meta_dict):
+    """
+    Prepare dictionary with default values filled and removal of all type/required/default dicts.
+
+    Parameters
+    ----------
+    dict
```
orelgueta, line 268: should be `meta_dict: dict`

simtools/data_model/validate_schema.py
```diff
@@ -163,8 +163,8 @@ def _validate_data_type(self, schema, key, data_field):
             self._validate_datetime(data_field, self._field_is_optional(schema))
         elif schema["type"] == "email":
             self._validate_email(data_field, key)
-        elif schema["type"] == "instrumentlist":
-            self._validate_instrument_list(data_field)
+        elif schema["type"].endswith("list"):
+            self._validate_list(schema["type"], data_field)
```
orelgueta, line 167: So you no longer validate that the instruments in the list make sense? Will try to come back to this comment later if I find the answer below.

tests/unit_tests/data_model/test_meta_data_model.py
```diff
@@ -0,0 +1,79 @@
+import copy
+
+import pytest
+
+from simtools.data_model import meta_data_model
+
+
+def test_top_level_reference_schema():
+    """(very hard to test this)"""
```
orelgueta, line 9: What about testing also that the output is what you expect? For example that you have at least one key you are expecting and in this case also a nested dict?

orelgueta (commented):
Looks good, just a few minor comments.
Is there also a way/need to add a test for providing a wrong or unknown type?

simtools/data_model/meta_data_model.py
GernotMaier: I removed as it is obviously confusing: with `str` it was meant the that data type is given as string (e.g. "float"). I don't think this is needed here, so removed it completely.

simtools/data_model/meta_data_model.py
GernotMaier: This is actually a rabbit hole and it seems really complicated to check for a valid url (I think it is an unsolved problem)?

See e.g., 
- https://stackoverflow.com/questions/7160737/how-to-validate-a-url-in-python-malformed-or-not
- https://stackoverflow.com/questions/827557/how-do-you-validate-a-url-with-a-regular-expression-in-python

It would be very nice to have, but...

simtools/data_model/meta_data_model.py
GernotMaier: Copy and paste error - the last line of the docstring is not needed.

simtools/data_model/meta_data_model.py
GernotMaier: yes, fixed it.

simtools/data_model/validate_schema.py
GernotMaier: `self._validate_list(schema["type"], data_field)` is replacing ` self._validate_instrument_list(data_field)` by calling the more generic function; `schema["type"],` can be of type instrumentlist or documentlist. So there should be no change to before when the list type is `instrumentlist`

tests/unit_tests/data_model/test_meta_data_model.py
GernotMaier: I've added an `assert` for a random key to be included. Actually, the test is not so bad and I removed the 'very hard to test comments'. We test three characteristics of the outcome of this module, that should be good!

GernotMaier:
@orelgueta - I think I've addressed your comments (as much as I could), let me know if there is anything remaining.

simtools/data_model/meta_data_model.py
orelgueta: OK, no problem! 
(I saw those stackoverflow entries before asking and figured maybe we can do something simple, but it's not really worth it, you are right.)

simtools/data_model/validate_schema.py
orelgueta: Yeah I realised that later and figured it's a shame that we don't validate that the input is a valid instrument. It would open somewhat of a can of worms because the list of valid instruments might change with time and then we would have to change this validation function. However, perhaps it's something to consider. I will approve this PR anyway and let you decide if you want to add it. If not, simply merge.

orelgueta (approved):
Thanks for addressing the comments. I am approving and will let you decide about the validation of instruments point.

simtools/data_model/validate_schema.py
GernotMaier: Wait - here we validate data types, not the actually values (this is consistent throughout the class). The name validation should happen somewhere, but not in the schema validation.

PR #407, opened 2022-11-21
Finishing up the documentation update

VictorBarbosaMartins:
Finishing up the documentation update started by #399.
- Update the docstrings from the modules subdirectories

PR #411, opened 2022-11-22
Move visualization stuff into a new directory called visualization

### LINKED ISSUE #410: Reordering of modules
In this issue I propose we discuss another reordering of the modules.
- As discussed before, we would like to create a directory (*visualization*) inside *gammasim-tools/simtools* and move both *visualize.py* and *legend_handlers.py* to it;
- I propose we unify categories: Use the subdirectories in gammasim-tools/simtools as categories and organize *tests/integration_tests/test_applications.py* according to it (and vice-versa!). For example, we can create a repository in *gammasim-tools/simtools* and gather the modules related to *optics* together, as it is already done in  *tests/integration_tests/test_applications.py* in the command line comments;
- I would like to leave no module without a category. By now, *camera_efficiency*, *db_handler*, *io_handler*, *psf_analysis*, *ray_tracing*, *simulator*, *version*, *visualize* are in *gammasim-tools/simtools* but dont belong to any subdirectory.

orelgueta:
Closes #410.


VictorBarbosaMartins (changes requested):
Hi, thanks again for it!
I've checked the occurrences of 'visualize' and it seems you got all of them. There is only this file, which I am not sure if it is important (simtools.egg-info/SOURCES.txt), where the old path is still there.

My only request is in the API. Please move the legend_handler section, which is now under util, to the new visualize section. Thank you!

orelgueta:
But simtools.egg-info shouldn't be pushed to gitlab, I don't see it here either. 

What do you mean by the API?

VictorBarbosaMartins:
> But simtools.egg-info shouldn't be pushed to gitlab, I don't see it here either.

Good point.

> What do you mean by the API?

I meant the documentation (https://github.com/gammasim/gammasim-tools/blob/move_visualize/docs/source/util.rst). It would be good to move legend_handlers from the util section to a new section called visualization. Also https://github.com/gammasim/gammasim-tools/blob/move_visualize/docs/source/library.rst needs an update. If you prefer, I can do it, but I would like to finish #407 first

orelgueta:
Added the requested changes to the documentation (sorry I missed it, not really used to making changes there as well). There's one conflict left to resolve and then we are probably ready to merge.

VictorBarbosaMartins (approved):
Good. thanks!

PR #412, opened 2022-11-23
Set matplotlib logger to warning level

### LINKED ISSUE #244: Large number of matplotlib warnings when running with DEBUG options
Running some applications (e.g. derrive_mirro_rnda) with -v DEBUG results in a large number of DEBUG output from matplotlib (matplotlib.font_manager and matplotlib.backend_pdf) which makes the output very hard to parse.

I can suppress these messages with
```
logging.getLogger('matplotlib.font_manager').disabled = True
logging.getLogger('matplotlib.backends.backend_pdf').disabled = True
```
and obtain a much more readable and therefore useful output.

I would like to add these two lines? Is there any issue with that?
Suppressing debug messages is not in all cases good, but do we expect that we need those matplotlib debug messages?

orelgueta:
This suppresses the infinite debug messages from matplotlib. It's far from the ideal solution, but the other option is to rewrite our entire logging functionality. Honestly, it's not such a bad solution either. Closes #244.

### DIFF
--- a/simtools/__init__.py
+++ b/simtools/__init__.py
@@ -1,3 +1,4 @@
 import logging
 
 logging.basicConfig(format="%(levelname)s::%(module)s(l%(lineno)s)::%(funcName)s::%(message)s")
+logging.getLogger("matplotlib").setLevel(logging.WARNING)


GernotMaier (approved):
Works, good solution!

PR #413, opened 2022-11-23
Add option to get site parameter, remove getting multiple versions

### LINKED ISSUE #384: applications/get_parameter.py does not work for getting parameter from a site
When writing the documentation for applications/get_parameter.py, I realized it does not work for getting parameter from a site, for instance, the altitude. It would be nice to have it implemented. Adding such a test to the integration tests would also be necessary.

orelgueta:
Fixes #384

### DIFF
--- a/applications/get_parameter.py
+++ b/applications/get_parameter.py
@@ -22,6 +22,10 @@
     log_level (str, optional)
         Log level to print (default=INFO).
 
+    Raises
+    ------
+    KeyError in case the parameter requested does not exist in the model parameters.
+
     Example
     -------
     Get the mirror_list parameter from the DB.
--- a/applications/get_parameter.py
+++ b/applications/get_parameter.py
@@ -3,11 +3,10 @@
 """
     Summary
     -------
-    Get a parameter entry from DB for a specific telescope.
+    Get a parameter entry from DB for a specific telescope or a site.
     The application receives a parameter name, a site, a telescope (if applicable) and \
     optionally a version. It then prints out the parameter entry.
-    If no version is provided, the entries of the last 5 versions are printed.
-    The name of the file is required.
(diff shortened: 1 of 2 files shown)


VictorBarbosaMartins (changes requested):
Hi @orelgueta, thanks for adding this functionality, it looks good. Please, complement with the following:
- add to the docstring also the raise statement
- add a integration test for this functionality, ex. `python applications/get_parameter.py --parameter altitude --site North --model_version prod5`

orelgueta:
Thanks for the review. I added those two points. I should mention though that we never had `Raises` in the applications. I added it here cause it's always good to have, but we might want to add it to all other applications as well.

VictorBarbosaMartins (approved):
Great, thanks. lets merge once the tests are successful

VictorBarbosaMartins:
> Thanks for the review. I added those two points. I should mention though that we never had `Raises` in the applications. I added it here cause it's always good to have, but we might want to add it to all other applications as well.

Indeed, I just realized that now. I will do the rest in the documentation PR

GernotMaier:
> Thanks for the review. I added those two points. I should mention though that we never had `Raises` in the applications. I added it here cause it's always good to have, but we might want to add it to all other applications as well.

I this something we want to have? My idea was always:
- modules and classes raise exceptions, as they are called by applications or other classes which can / should handle the exception (depend on the use case)
- errors in applications should trigger a logger.error or logger.critical and then an application quit. This is a level with direct communication with the users, do we expect them to understand KeyErrors or similar? I am really not sure.

orelgueta:
I agree, but we also print out a meaningful error to the user. So essentially we are just using python's built in `raise` function instead of some `os.exit` command to exit. It seems reasonable to me. Whether it should be in the documentation I am not sure, but the code I am confident about.

Issue #360, opened 2022-10-20
Development of the atmosphere tool


VictorBarbosaMartins:
I created this issue to guide us through the development of the atmosphere tool. 
### **Motivation**
We want to understand the impact of a change in the atmospheric conditions into our ability to reliably reconstruct the events.
### **Goal of the tool**
The tool will be able to compare the results of Corsika simulations generated with different atmospheric conditions. The main parameters to be investigated are the photon density distribution on the ground and its wavelength distribution. 

#### **One atmosphere**
- [ ] Create an array of telescopes that covers up to 1 km on the ground (using class LayoutArray)
- [ ] Generate a corsika config file with appropriate parameters to save the list of photons with wavelength (corsika_config.py and corsika_runner.py). Options for including different atmosphere files should be given.
- [ ] Use the array and config file created to run a production (production.py)
- [ ] Create a Corsika output handler (using pyeventio)
- [ ] In the Corsika output handler, estimate the following variables (with plots):
  - [ ] photon density as function of distance to array center;
  - [ ] wavelength distribution;
  - [ ] 2D distribution (distance to the center vs wavelength)
  - [ ] 2D distribution (X vs Y on the ground)
  - [ ] integrated photon counts;
#### **Compare two atmospheres**
- [ ] Plot the distributions from the two atmospheres together (for visual comparison)
- [ ] Use a statistical test to evaluate how two distributions differ from each other (should be generic and able to be used for other kinds of distribution functions)

GernotMaier:
Sounds good, two comments:

- applications/make_regular_arrays.py is the application used to make the regular arrays
- have a look at https://github.com/cta-observatory/pycorsikaio for reading corsika binary output

VictorBarbosaMartins:
> Sounds good, two comments:
> 
> * applications/make_regular_arrays.py is the application used to make the regular arrays
> * have a look at https://github.com/cta-observatory/pycorsikaio for reading corsika binary output
Thanks. 
- Yes, I am basing on applications/make_regular_arrays.py to generate the "large" array.
- I've been using pyeventio (https://github.com/cta-observatory/pyeventio) and getting all the information I need from the telescope file. Do we have a preference for which Corsika output file to use?

GernotMaier:
Ah - you are reading the sim_telarray file, not the corsika file. All good, although for looking at the atmosphere it would be enough to test the corsika file.

VictorBarbosaMartins:
[...]

Issue #339, opened 2022-09-29
Layout plotting (sites branch)


GernotMaier:
Integrate plotting of telescope layouts.

- [ ] sites branch integration
- [ ] check and possible use plotting from ctapipe or ctaplot

VictorBarbosaMartins:
Hi Gernot. Could I integrate this into the atmosphere tool issue and work on it? It is slightly related to the production of a large array of telescopes that I'm working on. I've been looking into the code and I haven't found a plotting application for the array yet, only the printing of array position. Is that right or have I missed something?

GernotMaier:
Probably won't want to mix it with the atmosphere tool, but this is still something to do. @orelgueta , can you comment on the plotting?

orelgueta:
There are simple plotting tools in the sites branch already. Just need to decide if we want them and if we do, where to add them (in visualize or in layout class).

Another option is to use this opportunity and import the array plotting from ctapipe (though it doesn't really fit our needs exactly since we want the telescope numbers and stuff)

PR #429, opened 2022-11-29
Layout array

### LINKED ISSUE #418: Integrate old 'sites' branch to master
As part of #360, we need to develop the array layout (plotting), which was already started in the 'sites' branch (https://github.com/gammasim/gammasim-tools/tree/sites). In order not to duplicate functionalities I am adapting the old code to the new state of the master branch, implementing the documentation and the respective tests. This is part of #360

VictorBarbosaMartins:
Closes #418 and #339

I adapted and integrated part of the old branch 'sites' into the current version of the code. The following changed in comparison to the old branch:
- sites.py was a module at simtools. I renamed it to layout_array_builder and put inside simtools/layout since it is closely related to simtools/layout_array. I intend to expand this same module to include functions which create an array, according to a specific shape, for instance, the arrays we will need for the atmospheric tool (#360). This will still come, but I am already justifying its name;
- I made the tests and the documentation of the functions that were missing in the original branch;
- changed the mpatches of MST and the sizes, because the mpatches.Rectangle for the MST wasn't making sense to me. The sizes of the circles now reproduce in scale the differences in sizes of the telescope types. Furthermore, the MST mpatches.circle is filled to distinguish it more clearly from the LST (see image attached). Naturally in the image, the legend is malfunctioning, but this would be subject to a change in the legend_handlers.py. Let me know if you prefer it implemented here.
test

- The next step is to write the application for plotting the array, which I can do once we agree on these changes here. Thank you.

### DIFF
--- a/applications/print_array_elements.py
+++ b/applications/print_array_elements.py
@@ -132,7 +132,9 @@ def main():
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
     layout = layout_array.LayoutArray(site="North", mongo_db_config=db_config)
-    layout.read_telescope_list_file(telescope_list_file=args_dict["array_element_list"])
+    layout.initialize_layout_array_from_telescope_file(
+        telescope_list_file=args_dict["array_element_list"]
+    )
     layout.convert_coordinates()
     if args_dict["export"] is not None:
         layout.export_telescope_list(
--- a/applications/print_array_elements.py
+++ b/applications/print_array_elements.py
@@ -120,18 +120,18 @@ def _parse(label=None, description=None):
         default=False,
         action="store_true",
     )
-    return config.initialize()
+    return config.initialize(db_config=True)
 
 
 def main():
 
     label = Path(__file__).stem
-    args_dict, _ = _parse(label, description=("Print a list of array element positions"))
+    args_dict, db_config = _parse(label, description=("Print a list of array element positions"))
 
     _logger = logging.getLogger()
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
-    layout = layout_array.LayoutArray(site="North")
+    layout = layout_array.LayoutArray(site="North", mongo_db_config=db_config)
     layout.read_telescope_list_file(telescope_list_file=args_dict["array_element_list"])
     layout.convert_coordinates()
     if args_dict["export"] is not None:
--- a/applications/print_array_elements.py
+++ b/applications/print_array_elements.py
@@ -131,10 +131,7 @@ def main():
     _logger = logging.getLogger()
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
-    layout = layout_array.LayoutArray(site="North", mongo_db_config=db_config)
-    layout.initialize_layout_array_from_telescope_file(
-        telescope_list_file=args_dict["array_element_list"]
-    )
+    layout = layout_array.LayoutArray(telescope_list_file=args_dict["array_element_list"])
     layout.convert_coordinates()
     if args_dict["export"] is not None:
         layout.export_telescope_list(
--- a/applications/print_array_elements.py
+++ b/applications/print_array_elements.py
@@ -126,7 +126,7 @@ def _parse(label=None, description=None):
 def main():
 
     label = Path(__file__).stem
-    args_dict = _parse(label, description=("Print a list of array element positions"))
+    args_dict, _ = _parse(label, description=("Print a list of array element positions"))
 
     _logger = logging.getLogger()
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
--- a/applications/print_array_elements.py
+++ b/applications/print_array_elements.py
@@ -120,13 +120,13 @@ def _parse(label=None, description=None):
         default=False,
         action="store_true",
     )
-    return config.initialize(db_config=True)
+    return config.initialize()
 
 
 def main():
 
     label = Path(__file__).stem
-    args_dict, db_config = _parse(label, description=("Print a list of array element positions"))
+    args_dict = _parse(label, description=("Print a list of array element positions"))
 
     _logger = logging.getLogger()
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -386,7 +386,7 @@ class LayoutArray:
 
     def _load_telescope_names(self, row):
         """
-        Read and set telescope names
+        Read and set telescope names.
 
         Parameters
         ----------
@@ -460,9 +460,58 @@ class LayoutArray:
                     raise
         return value * unit
 
-    def _if_no_key_in_dict_pass(self, dictionary, key):
+    def _try_set_coordinate(self, row, tel, table, crs_name, key1, key2):
+        """Function auxiliary to self._load_telescope_list. It sets the coordinates.
+
+        Parameters
+        ----------
+        row: dict
+            A row of the astropy.table.Table with array element coordinates.
+        tel: TelescopePosition
+            Instance of TelescopePosition.
+        table: astropy.table.Table or astropy.table.QTable
+            data table with array element coordinates.
+        crs_name: str
+            Name of coordinate system.
+        key1: str
+            Name of x-coordinate.
+        key2: str
+            Name of y-coordinate.
+        """
         try:
-            return dictionary[key]
+            tel.set_coordinates(
+                crs_name,
+                self._assign_unit_to_quantity(row[key1], table[key1].unit),
+                self._assign_unit_to_quantity(row[key2], table[key2].unit),
+            )
+        except KeyError:
+            pass
+
+    def _try_set_altitude(self, tel, table, row):
+        """
+        Function auxiliary to self._load_telescope_list. It sets the altitude of the
+        TelescopePosition instance.
+
+        Parameters
+        ----------
+        row: dict
+            A row of the astropy.table.Table with array element coordinates.
+        tel: TelescopePosition
+            Instance of TelescopePosition.
+        table: astropy.table.Table or astropy.table.QTable
+            data table with array element coordinates.
+        """
+        try:
+            tel.set_altitude(
+                self._altitude_from_corsika_z(
+                    pos_z=self._assign_unit_to_quantity(row["pos_z"], table["pos_z"].unit),
+                    tel_name=tel.name,
+                )
+            )
+        except KeyError:
+            pass
+        try:
+            tel.set_altitude(self._assign_unit_to_quantity(row["alt"], table["alt"].unit))
         except KeyError:
             pass
 
@@ -476,47 +525,13 @@ class LayoutArray:
             data table with array element coordinates
 
         """
-
         for row in table:
             tel = self._load_telescope_names(row)
+            self._try_set_coordinate(row, tel, table, "corsika", "pos_x", "pos_y")
+            self._try_set_coordinate(row, tel, table, "utm", "utm_east", "utm_north")
+            self._try_set_coordinate(row, tel, table, "mercator", "mercator", "lon")
+            self._try_set_altitude(row, tel, table)
 
-            try:
-                tel.set_coordinates(
-                    "corsika",
-                    self._assign_unit_to_quantity(row["pos_x"], table["pos_x"].unit),
-                    self._assign_unit_to_quantity(row["pos_y"], table["pos_y"].unit),
-                )
-            except KeyError:
-                pass
-            try:
-                tel.set_coordinates(
-                    "utm",
-                    self._assign_unit_to_quantity(row["utm_east"], table["utm_east"].unit),
-                    self._assign_unit_to_quantity(row["utm_north"], table["utm_north"].unit),
-                )
-            except KeyError:
-                pass
-            try:
-                tel.set_coordinates(
-                    "mercator",
-                    self._assign_unit_to_quantity(row["lat"], table["lat"].unit),
-                    self._assign_unit_to_quantity(row["lon"], table["lon"].unit),
-                )
-            except KeyError:
-                pass
-            try:
-                tel.set_altitude(
-                    self._altitude_from_corsika_z(
-                        pos_z=self._assign_unit_to_quantity(row["pos_z"], table["pos_z"].unit),
-                        tel_name=tel.name,
-                    )
-                )
-            except KeyError:
-                pass
-            try:
-                tel.set_altitude(self._assign_unit_to_quantity(row["alt"], table["alt"].unit))
-            except KeyError:
-                pass
             self._telescope_list.append(tel)
 
     @staticmethod
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -487,7 +487,7 @@ class LayoutArray:
         except KeyError:
             pass
 
-    def _try_set_altitude(self, tel, table, row):
+    def _try_set_altitude(self, row, tel, table):
         """
         Function auxiliary to self._load_telescope_list. It sets the altitude of the
         TelescopePosition instance.
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -10,7 +10,7 @@ from simtools import db_handler, io_handler
 from simtools.layout.telescope_position import TelescopePosition
 from simtools.util import names
 from simtools.util.general import collect_data_from_yaml_or_dict
-from simtools.util.names import all_telescope_class_names
+from simtools.util.names import all_telescope_class_names, lst
 
 __all__ = ["InvalidTelescopeListFile", "LayoutArray"]
 
@@ -178,7 +178,7 @@ class LayoutArray:
         for simtools_par in corsika_pars:
             corsika_par = names.translate_simtools_to_corsika(simtools_par)
             corsika_dict[simtools_par] = {}
-            for tel_type in names.all_telescope_class_names:
+            for tel_type in all_telescope_class_names:
                 corsika_dict[simtools_par][tel_type] = corsika_parameters_dict[corsika_par][
                     tel_type
                 ]["value"]
@@ -871,22 +871,25 @@ class LayoutArray:
 
         Parameters
         ----------
-        file_name: str
-            Name of the ecsv file with telescope layout.
+        telescope_table: astropy.QTable
+            Astropy QTable with telescope information.
 
         Returns
         -------
-        dict
-            Dictionary with the telescope layout information.
+        astropy.QTable
+            Astropy QTable with telescope information updated with the radius.
         """
 
         telescope_table["radius"] = [
-            float(telescope_table.meta["corsika_sphere_radius"][tel_name_now[:3]].split()[0])
+            float(
+                telescope_table.meta["corsika_sphere_radius"][
+                    names.get_telescope_type(tel_name_now)
+                ].split()[0]
+            )
             for tel_name_now in telescope_table["telescope_name"]
         ]
+
         telescope_table["radius"].unit = u.Unit(
-            telescope_table.meta["corsika_sphere_radius"][all_telescope_class_names["LST"]].split()[
-                1
-            ]
+            telescope_table.meta["corsika_sphere_radius"][lst].split()[1]
         )
         return telescope_table
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -43,8 +43,8 @@ class LayoutArray:
 
     def __init__(
         self,
-        mongo_db_config,
-        site,
+        mongo_db_config=None,
+        site=None,
         label=None,
         name=None,
         layout_center_data=None,
@@ -117,6 +117,34 @@ class LayoutArray:
 
     def __getitem__(self, i):
         return self._telescope_list[i]
+
+    @property
+    def site(self):
+        """
+        Get the site
+        """
+        return self.site
+
+    @site.setter
+    def site(self, site):
+        """
+        Set the site
+        """
+        self.site = site
+
+    @property
+    def mongo_db_config(self):
+        """
+        Get the MongoDB config
+        """
+        return self.mongo_db_config
+
+    @mongo_db_config.setter
+    def mongo_db_config(self, mongo_db_config):
+        """
+        Set the MongoDB config
+        """
+        self.site = mongo_db_config
 
     def _initialize_corsika_telescope(self, corsika_dict=None):
         """
@@ -194,6 +222,13 @@ class LayoutArray:
                         "the unit."
                     )
                     corsika_dict[simtools_par][tel_type] = corsika_dict[simtools_par][tel_type]
+
+        if self.mongo_db_config is None:
+            self._logger.error("DB connection info was not provided, cannot set telescope altitude")
+            raise ValueError
+        if self.site is None:
+            self._logger.error("Site was not provided, cannot set telescope altitude")
+            raise ValueError
 
         db = db_handler.DatabaseHandler(mongo_db_config=self.mongo_db_config)
         self._logger.debug("Reading site parameters from DB")
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -144,7 +144,7 @@ class LayoutArray:
         """
         Set the MongoDB config
         """
-        self.site = mongo_db_config
+        self.mongo_db_config = mongo_db_config
 
     def _initialize_corsika_telescope(self, corsika_dict=None):
         """
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -430,6 +430,36 @@ class LayoutArray:
 
         return tel
 
+    def _assign_unit_to_quantity(self, value, unit):
+        """
+        Assign unit to quantity.
+
+        Parameters
+        ----------
+        value:
+            value to get a unit. It can be a float, int, or a Quantity (convertible to 'unit').
+        unit: astropy.units.Unit
+            Unit to apply to 'quantity'.
+
+        Returns
+        -------
+        astropy.units.Quantity
+            Quantity of value 'quantity' and unit 'unit'.
+        """
+        if isinstance(value, u.Quantity):
+            self._logger.debug(f"Value {value} is already of type astropy.units.Quantity.")
+            if isinstance(value.unit, type(unit)):
+                self._logger.debug(f"Quantity {value} has already unit {unit}. Returning {value}")
+                return value
+            else:
+                try:
+                    value = value.to(unit)
+                    return value
+                except u.UnitConversionError:
+                    self._logger.error(f"Cannot convert {value.unit} to {unit}.")
+                    raise
+        return value * unit
+
     def _load_telescope_list(self, table):
         """
         Load list of telescope from an astropy table (support both QTable and Table)
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -198,17 +198,9 @@ class LayoutArray:
         db = db_handler.DatabaseHandler(mongo_db_config=self.mongo_db_config)
         self._logger.debug("Reading site parameters from DB")
         _site_pars = db.get_site_parameters(self.site, "Current", only_applicable=True)
-        corsika_dict["corsika_obs_level"] = _site_pars["altitude"]["Value"]
-
-        try:
-            corsika_dict["corsika_obs_level"] = corsika_dict["corsika_obs_level"] * u.Unit(
-                _site_pars["altitude"]["units"]
-            )
-        except KeyError:
-            corsika_dict["corsika_obs_level"] = corsika_dict["corsika_obs_level"]
-            logger.warning(
-                "Key not valid. Dictionary does not have a key 'unit'. Continuing without the unit."
-            )
+        corsika_dict["corsika_obs_level"] = _site_pars["altitude"]["Value"] * u.Unit(
+            _site_pars["altitude"]["units"]
+        )
 
         return corsika_dict
 
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -865,30 +865,6 @@ class LayoutArray:
         return pyproj.CRS("EPSG:4326")
 
     @staticmethod
-    def get_telescope_type(telescope_name):
-        """
-        Guess telescope type from name, e.g. "LST", "MST", "SST", "SCT".
-
-        Parameters
-        ----------
-        telescope_name: str
-            Telescope name
-
-        Returns
-        -------
-        str
-            Telescope type.
-        """
-
-        _class, _ = names.split_telescope_model_name(telescope_name)
-        try:
-            if _class[0:3] in (names.all_telescope_class_names):
-                return _class[0:3]
-
-        except IndexError:
-            pass
-
-    @staticmethod
     def include_radius_into_telescope_table(telescope_table):
         """
         Include the radius of the telescopes types into the astropy.table.QTable telescopes_table
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -4,14 +4,13 @@ from pathlib import Path
 import astropy.units as u
 import numpy as np
 import pyproj
-from astropy.coordinates.errors import UnitsError
 from astropy.table import QTable
 
 from simtools import db_handler, io_handler
 from simtools.layout.telescope_position import TelescopePosition
 from simtools.util import names
 from simtools.util.general import collect_data_from_yaml_or_dict
-from simtools.util.names import all_telescope_class_names, lst
+from simtools.util.names import all_telescope_class_names
 
 __all__ = ["InvalidTelescopeListFile", "LayoutArray"]
 
@@ -878,37 +877,10 @@ class LayoutArray:
         -------
         astropy.QTable
             Astropy QTable with telescope information updated with the radius.
-
-        Raises
-        ------
-        UnitsError:
-            If the units of radii given in the table metadata are not the same.
         """
 
         telescope_table["radius"] = [
-            float(
-                telescope_table.meta["corsika_sphere_radius"][
-                    names.get_telescope_type(tel_name_now)
-                ].split()[0]
-            )
+            telescope_table.meta["corsika_sphere_radius"][names.get_telescope_type(tel_name_now)]
             for tel_name_now in telescope_table["telescope_name"]
         ]
-
-        for tel_type in names.all_telescope_class_names:
-            if tel_type in telescope_table.meta["corsika_sphere_radius"]:
-                if (
-                    telescope_table.meta["corsika_sphere_radius"][tel_type].split()[1]
-                    == telescope_table.meta["corsika_sphere_radius"][lst].split()[1]
-                ):
-                    continue
-                else:
-                    msg = "The units of the telescope radii in the QTable are not the same. \
-                    Aborting."
-                    self._logger.error(msg)
-                    raise UnitsError(msg)
-
-        telescope_table["radius"].unit = u.Unit(
-            telescope_table.meta["corsika_sphere_radius"][lst].split()[1]
-        )
-
         return telescope_table
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -507,8 +507,8 @@ class LayoutArray:
 
         Returns
         -------
-        dict
-            Dictionary with the telescope layout information.
+        astropy.QTable
+            Astropy table with the telescope layout information.
 
         Raises
         ------
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -189,8 +189,11 @@ class LayoutArray:
                         tel_type
                     ] * u.Unit(unit)
                 except KeyError:
-                    logger.error("Key not valid. Dictionary does not have a key 'unit'.")
-                    pass
+                    logger.warning(
+                        "Key not valid. Dictionary does not have a key 'unit'. Continuing without "
+                        "the unit."
+                    )
+                    corsika_dict[simtools_par][tel_type] = corsika_dict[simtools_par][tel_type]
 
         db = db_handler.DatabaseHandler(mongo_db_config=self.mongo_db_config)
         self._logger.debug("Reading site parameters from DB")
@@ -202,7 +205,10 @@ class LayoutArray:
                 _site_pars["altitude"]["units"]
             )
         except KeyError:
-            pass
+            corsika_dict["corsika_obs_level"] = corsika_dict["corsika_obs_level"]
+            logger.warning(
+                "Key not valid. Dictionary does not have a key 'unit'. Continuing without the unit."
+            )
 
         return corsika_dict
 
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -24,10 +24,10 @@ class LayoutArray:
 
     Parameters
     ----------
-    site: str
-        Site name or location (e.g., North/South or LaPalma/Paranal)
     mongo_db_config: dict
         MongoDB configuration.
+    site: str
+        Site name or location (e.g., North/South or LaPalma/Paranal)
     label: str
         Instance label.
     name: str
@@ -42,8 +42,8 @@ class LayoutArray:
 
     def __init__(
         self,
-        site,
         mongo_db_config,
+        site,
         label=None,
         name=None,
         layout_center_data=None,
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -79,8 +79,8 @@ class LayoutArray:
 
         Parameters
         ----------
-        layout_center_data: dict
-            Dict describing array center coordinates.
+        mongo_db_config: dict
+            MongoDB configuration.
         layout_array_name: str
             e.g. South-4LST, North-Prod5 ...
         label: str
@@ -97,7 +97,12 @@ class LayoutArray:
         array_name = names.validate_layout_array_name(split_name[1])
         valid_layout_array_name = site_name + "-" + array_name
 
-        layout = cls(site_name, name=valid_layout_array_name, label=label)
+        layout = cls(
+            site=site_name,
+            mongo_db_config=mongo_db_config,
+            name=valid_layout_array_name,
+            label=label,
+        )
 
         telescope_list_file = layout.io_handler.get_input_data_file(
             "layout", f"telescope_positions-{valid_layout_array_name}.ecsv"
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -179,17 +179,29 @@ class LayoutArray:
             corsika_par = names.translate_simtools_to_corsika(simtools_par)
             corsika_dict[simtools_par] = {}
             for tel_type in names.all_telescope_class_names:
-                unit = corsika_parameters_dict[corsika_par][tel_type]["unit"]
                 corsika_dict[simtools_par][tel_type] = corsika_parameters_dict[corsika_par][
                     tel_type
-                ]["value"] * u.Unit(unit)
+                ]["value"]
+
+                try:
+                    unit = corsika_parameters_dict[corsika_par][tel_type]["unit"]
+                    corsika_dict[simtools_par][tel_type] = corsika_dict[simtools_par][
+                        tel_type
+                    ] * u.Unit(unit)
+                except KeyError:
+                    pass
 
         db = db_handler.DatabaseHandler(mongo_db_config=self.mongo_db_config)
         self._logger.debug("Reading site parameters from DB")
         _site_pars = db.get_site_parameters(self.site, "Current", only_applicable=True)
-        corsika_dict["corsika_obs_level"] = _site_pars["altitude"]["Value"] * u.Unit(
-            _site_pars["altitude"]["units"]
-        )
+        corsika_dict["corsika_obs_level"] = _site_pars["altitude"]["Value"]
+
+        try:
+            corsika_dict["corsika_obs_level"] = corsika_dict["corsika_obs_level"] * u.Unit(
+                _site_pars["altitude"]["units"]
+            )
+        except KeyError:
+            pass
 
         return corsika_dict
 
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import astropy.units as u
 import numpy as np
 import pyproj
+from astropy.coordinates.errors import UnitsError
 from astropy.table import QTable
 
 from simtools import db_handler, io_handler
@@ -864,8 +865,7 @@ class LayoutArray:
         """
         return pyproj.CRS("EPSG:4326")
 
-    @staticmethod
-    def include_radius_into_telescope_table(telescope_table):
+    def include_radius_into_telescope_table(self, telescope_table):
         """
         Include the radius of the telescopes types into the astropy.table.QTable telescopes_table
 
@@ -878,6 +878,11 @@ class LayoutArray:
         -------
         astropy.QTable
             Astropy QTable with telescope information updated with the radius.
+
+        Raises
+        ------
+        UnitsError:
+            If the units of radii given in the table metadata are not the same.
         """
 
         telescope_table["radius"] = [
@@ -889,7 +894,21 @@ class LayoutArray:
             for tel_name_now in telescope_table["telescope_name"]
         ]
 
+        for tel_type in names.all_telescope_class_names:
+            if tel_type in telescope_table.meta["corsika_sphere_radius"]:
+                if (
+                    telescope_table.meta["corsika_sphere_radius"][tel_type].split()[1]
+                    == telescope_table.meta["corsika_sphere_radius"][lst].split()[1]
+                ):
+                    continue
+                else:
+                    msg = "The units of the telescope radii in the QTable are not the same. \
+                    Aborting."
+                    self._logger.error(msg)
+                    raise UnitsError(msg)
+
         telescope_table["radius"].unit = u.Unit(
             telescope_table.meta["corsika_sphere_radius"][lst].split()[1]
         )
+
         return telescope_table
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -61,7 +61,7 @@ class LayoutArray:
         self.mongo_db_config = mongo_db_config
         self.label = label
         self.name = name
-        self.site = names.validate_site_name(site)
+        self.site = None if site is None else names.validate_site_name(site)
         self.io_handler = io_handler.IOHandler()
 
         self._telescope_list = []
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -56,10 +56,7 @@ class LayoutArray:
 
         self.label = label
         self.name = name
-        if site not in ["North", "South"]:
-            raise ValueError("Site must be 'North' or 'South'!")
-        else:
-            self.site = site
+        self.site = names.validate_site_name(site)
         self.io_handler = io_handler.IOHandler()
 
         self._telescope_list = []
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -58,10 +58,8 @@ class LayoutArray:
         self._logger = logging.getLogger(__name__)
         self._logger.debug("Init LayoutArray")
 
-        self.mongo_db_config = mongo_db_config
         self.label = label
         self.name = name
-        self.site = names.validate_site_name(site)
         self.io_handler = io_handler.IOHandler()
 
         self._telescope_list = []
@@ -130,7 +128,7 @@ class LayoutArray:
         """
         Set the site
         """
-        self.site = site
+        self.site = names.validate_site_name(site)
 
     @property
     def mongo_db_config(self):
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -865,7 +865,8 @@ class LayoutArray:
         """
         return pyproj.CRS("EPSG:4326")
 
-    def include_radius_into_telescope_table(self, telescope_table):
+    @staticmethod
+    def include_radius_into_telescope_table(telescope_table):
         """
         Include the radius of the telescopes types into the astropy.table.QTable telescopes_table
 
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -160,7 +160,7 @@ class LayoutArray:
         FileNotFoundError:
             If file_name does not exist.
         """
-
+        logger = logging.getLogger(__name__)
         if file_name is None:
             corsika_parameters_dict = collect_data_from_yaml_or_dict(
                 self.io_handler.get_input_data_file("parameters", "corsika_parameters.yml"), None
@@ -189,6 +189,7 @@ class LayoutArray:
                         tel_type
                     ] * u.Unit(unit)
                 except KeyError:
+                    logger.error("Key not valid. Dictionary does not have a key 'unit'.")
                     pass
 
         db = db_handler.DatabaseHandler(mongo_db_config=self.mongo_db_config)
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -58,8 +58,10 @@ class LayoutArray:
         self._logger = logging.getLogger(__name__)
         self._logger.debug("Init LayoutArray")
 
+        self.mongo_db_config = mongo_db_config
         self.label = label
         self.name = name
+        self.site = names.validate_site_name(site)
         self.io_handler = io_handler.IOHandler()
 
         self._telescope_list = []
@@ -115,34 +117,6 @@ class LayoutArray:
 
     def __getitem__(self, i):
         return self._telescope_list[i]
-
-    @property
-    def site(self):
-        """
-        Get the site
-        """
-        return self._site
-
-    @site.setter
-    def site(self, site):
-        """
-        Set the site
-        """
-        self._site = names.validate_site_name(site)
-
-    @property
-    def mongo_db_config(self):
-        """
-        Get the MongoDB config
-        """
-        return self._mongo_db_config
-
-    @mongo_db_config.setter
-    def mongo_db_config(self, mongo_db_config):
-        """
-        Set the MongoDB config
-        """
-        self._mongo_db_config = mongo_db_config
 
     def _initialize_corsika_telescope(self, corsika_dict=None):
         """
@@ -221,16 +195,16 @@ class LayoutArray:
                     )
                     corsika_dict[simtools_par][tel_type] = corsika_dict[simtools_par][tel_type]
 
-        if self._mongo_db_config is None:
+        if self.mongo_db_config is None:
             self._logger.error("DB connection info was not provided, cannot set telescope altitude")
             raise ValueError
-        if self._site is None:
+        if self.site is None:
             self._logger.error("Site was not provided, cannot set telescope altitude")
             raise ValueError
 
-        db = db_handler.DatabaseHandler(mongo_db_config=self._mongo_db_config)
+        db = db_handler.DatabaseHandler(mongo_db_config=self.mongo_db_config)
         self._logger.debug("Reading site parameters from DB")
-        _site_pars = db.get_site_parameters(self._site, "Current", only_applicable=True)
+        _site_pars = db.get_site_parameters(self.site, "Current", only_applicable=True)
         corsika_dict["corsika_obs_level"] = _site_pars["altitude"]["Value"] * u.Unit(
             _site_pars["altitude"]["units"]
         )
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -120,7 +120,7 @@ class LayoutArray:
 
     def _initialize_corsika_telescope(self, corsika_dict=None):
         """
-        Initialize Dictionary for CORSIKA telescope parameters. Allow input from different sources\
+        Initialize Dictionary for CORSIKA telescope parameters. Allow input from different sources
         (dictionary, yaml, ecsv header), which require checks to handle units correctly.
 
         Parameters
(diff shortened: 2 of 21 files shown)


VictorBarbosaMartins:
Codacy still complains about this.
```
Minor icon MINOR
Code Style
Method could be a function
def telescope_layout_file_to_dict(self, file_name):
```
I am wondering if we should, instead of using these plot functions in a class, to have a simtools/util/layout.py ? What do you think @orelgueta ?

orelgueta:
You can simply remove the `self` argument and add a `@staticmethod` decorator. We have quite a few of those if you want an example. (I started a review but will finish it tomorrow.)

VictorBarbosaMartins:
> You can simply remove the `self` argument and add a `@staticmethod` decorator. We have quite a few of those if you want an example. (I started a review but will finish it tomorrow.)

thanks. that is indeed a good idea

tests/unit_tests/layout/test_layout_array_builder.py
```diff
@@ -0,0 +1,56 @@
...
+        assert telescopes_dict[MST10_index][keys[key_step]].value[0] == values_from_file[key_step]
+
+
+def test_get_telescope_patch(corsika_telescope_data_dict, layout_builder_instance):
+    for tel_type in np.array(list(corsika_telescope_data_dict["corsika_sphere_radius"].keys())):
+        radius = corsika_telescope_data_dict["corsika_sphere_radius"][tel_type]
+        patch = layout_builder_instance._get_telescope_patch(
+            tel_type, 0 * u.m, 0 * u.m, radius.value * u.m
+        )
+        assert mpatches.Circle == type(patch)
```
orelgueta, line 37: This test would fail if you add a different shape. Maybe also test the radius?

orelgueta (commented):
All good, really minor comments.

simtools/visualization/legend_handlers.py
```diff
@@ -230,12 +230,14 @@ class MSTHandler(object):
...
-            width,
-            height,
+        center = (
+            handlebox.xdescent + 0.25 * handlebox.width,
+            handlebox.ydescent + 0.5 * handlebox.height,
+        )
+        radius = handlebox.height
+        patch = mpatches.Circle(
+            xy=center,
+            radius=radius * (11.5 / 23),
```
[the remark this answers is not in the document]
VictorBarbosaMartins: good idea. will just wait for your comment above to adjust the sizes. 
by the way, I see that in legend_handlers the parameters "legend, orig_handle, fontsize" are never used by the methods. Do we need those?

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,211 @@
...
+        plt.xlabel(x_title, fontsize=18, labelpad=0)
+        plt.ylabel(y_title, fontsize=18, labelpad=0)
+        plt.tick_params(axis="both", which="major", labelsize=15)
+
+        plt.legend(
+            legend_objects,
+            legend_labels,
+            handler_map=legend_handler_map,
+            prop={"size": 11},
+            loc="upper left",
```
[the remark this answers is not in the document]
VictorBarbosaMartins: good. done.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,214 @@
...
+        for tel_name_now in telescopes["telescope_name"]:
+            for tel_type in tel_counters:
+                if tel_type in tel_name_now:
+                    tel_counters[tel_type] += 1
+
+        if tel_counters["LST"] > 0:
+            legend_objects.append(leg_h.LSTObject())
+            legend_labels.append("LST ({})".format(tel_counters["LST"]))
+        if tel_counters["MST"] > 0:
+            legend_objects.append(leg_h.MSTObject())
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Indeed. I just added it (also the handler)

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Returns
+        -------
+        dict
+            Dictionary with the telescope layout information.
+        """
+
+        telescopes = QTable.read(file_name)
+        telescopes["radius"] = [
+            float(telescopes.meta["corsika_sphere_radius"][tel_name_now[:3]].split()[0])
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I agree, just moved it there

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,214 @@
...
+            Y positions of the telescopes in meters.
+
+        Returns
+        -------
+        2-tuple of numpy.Array of float
+            X and Y positions of the rotated telescopes positions.
+
+        """
+        if not isinstance(x, type(y)):
+            raise RuntimeError("x and y are not the same type! " "Cannot perform transformation.")
```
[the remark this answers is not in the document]
VictorBarbosaMartins: good question. I adjusted the message.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,214 @@
...
+        if not isinstance(x, (list, np.ndarray)):
+            x = [x]
+            y = [y]
+
+        if len(x) != len(y):
+            raise RuntimeError(
+                "Cannot perform coordinate transformation " "when x and y have different lengths."
+            )
+        x_trans, y_trans = list(), list()
```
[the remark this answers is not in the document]
VictorBarbosaMartins: done

tests/unit_tests/layout/test_layout_array.py
```diff
@@ -26,95 +25,78 @@ def layout_center_data_dict():
...
-        dest=io_handler.get_output_directory(dir_type="model", test=True),
-        file_name=test_file_name,
-    )
-
-    cfg_file = gen.find_file(
-        test_file_name,
-        io_handler.get_output_directory(dir_type="model", test=True),
-    )
-    return cfg_file
+def layout_array_instance(io_handler):
```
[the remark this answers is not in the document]
VictorBarbosaMartins: yes, I was using it before but now it is not necessary anymore, I deleted it from conftest.py

tests/unit_tests/layout/test_layout_array.py
```diff
@@ -158,11 +140,10 @@ def test_build_layout(
     layout.export_telescope_list(crs_name="corsika")
 
     # Building a second layout from the file exported by the first one
-    layout_2 = LayoutArray("test_layout_2")
-    layout_2.read_telescope_list_file(layout.telescope_list_file)
+    layout_array_instance.read_telescope_list_file(layout.telescope_list_file)
```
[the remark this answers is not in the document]
VictorBarbosaMartins: yes, it is the good thing about the fixtures :) we shall only not forget to pass the fixture as argument to the function

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,214 @@
...
+            (x.value, y.value), radius=radius.value, fill=fill_flag, color=colors[valid_name]
+        )
+        return patch
+
+    @staticmethod
+    def _rotate(rotation_angle, x, y):
+        """
+        Rotate x and y by the rotation angle given in rotation_angle.
+        The function returns the rotated x and y values.
+        Notice that rotation_angle must be given in degrees!
```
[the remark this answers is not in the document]
VictorBarbosaMartins: yes. I am including the decorators. I am only missing adjust the tests to it

simtools/visualization/legend_handlers.py
```diff
@@ -258,7 +260,7 @@ def legend_artist(legend, orig_handle, fontsize, handlebox):
         radius = handlebox.height
         patch = mpatches.Circle(
             xy=center,
-            radius=radius * (2.8 / 12),
+            radius=radius * (9.7 / 23),
```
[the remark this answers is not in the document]
orelgueta: They are a bit larger than the telescope dish radius and it's 12.5 m for LST, 9.15 for MST, 7.15 m for SCT and 3.0 m for SST.

simtools/visualization/legend_handlers.py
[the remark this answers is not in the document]
orelgueta: Yes, you need those, it's the general structure of legend handlers, nothing we could do about it.

kosack:
Array plotting is something that you would really benefit from using ctapipe's functionality so we do it in one common way.  We have both coordinate systems  like `ctapipe.coordinates.EastingNorthingFrame` and  `ctapipe.visualization.ArrayDisplay` that could be used here.  I worry that we will and up with slightly different conventions, which should be avoided. 

The only feature missing so far is to map the correct telescope names instead of simtel numbers, but we have an issue opened for that.

Using the `ctapipe.instrument` model even automatically can create an ArrayDisplay:
```python
source.subarray.peek()
```
image

orelgueta:
Hi Karl,

I agree using ctapipe plotting and coordinate transformation capabilities would be great. We are just a bit hesitant at the moment with adding ctapipe as a dependency. I am sure it will happen at some point (or perhaps some "core" ctapipe package), but it would probably happen when we need to use the analysis capabilities first like for muon analysis. 

BTW, if we wanted to use the ctapipe array plotting for example, it would mean we would have to also use the subarray functionality. That in itself would involve a rethinking of our current telescope model in gammasim-tools. Not necessarily a bad idea, but it's not a simple "import and plot" decision.

GernotMaier:
I agree here with Orel - duplication is definitely bad, but it would be a large effort to adapt gammasim-tools to the ctapipe style.

Maybe easier to start introduce a DPPS core package with minimal interface for those standard tasks.

tests/unit_tests/layout/test_layout_array_builder.py
VictorBarbosaMartins: Good. I implemented a bit more on these tests.

VictorBarbosaMartins:
Hi all, thanks for the discussions. Indeed we should try to avoid duplications of large amounts of code (maybe small internal functions would be ok?), though I am not sure at the moment how is the best way to avoid it.

@orelgueta, I implemented all the changes we discussed here and in the mattermost, please have a look again once the tests pass. Thank you.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+from matplotlib.collections import PatchCollection
+
+from simtools import io_handler
+from simtools.util import names
+from simtools.visualization import legend_handlers as leg_h
+
+
+class LayoutArrayBuilder:
+    """
+    Manage the array layout. It provides methods to create and plot a LayoutArray.
```
GernotMaier, line 15: These are two very different tasks in one class (create a layout and plot it). Would it make sense to separate that?

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+            prop={"size": 11},
+            loc="best",
+        )
+
+        plt.tight_layout()
+
+        return fig
+
+    @staticmethod
+    def telescope_layout_file_to_dict(file_name):
```
GernotMaier, line 123: Is that similar to LayoutArray::read_telescope_list_file?

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Returns
+        -------
+        patch
+            Instance of mpatches.Circle.
+        """
+
+        valid_name = names.get_telescope_type(name)
+        fill_flag = False
+        if valid_name == "MST":
```
GernotMaier, line 175: Hardcoded telescopes names. Wonder if there is another solution; does the introduction of a new telescope type in CTA means that we have to update this code? 

There is also MST and SCT here, why no SST or LST?

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
     name += f"_{label}" if label is not None else ""
     name += ".log"
     return name
+
+
+def get_telescope_type(telescope_name):
+    """
+    Guess telescope type from name. Types are "LST", "MST", "SST", "SCT".
```
GernotMaier, line 762: Maybe see e.g., "LST", "MST", ...

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+        Telescope name
+    Returns
+    -------
+    str
+        Telescope type.
+    """
+
+    _class, _ = split_telescope_model_name(telescope_name)
+    try:
+        if _class[0:3] in ("LST", "MST", "SST", "SCT"):
```
GernotMaier, line 775: There is a names:all_telescope_class_names dict, should we use this here instead of listing all telescope types? If we hardcode the telescope types in gammasim-tools (which I don't like at all), we should have it at one place only.

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+
+    except IndexError:
+        pass
+
+    return ""
+
+
+def get_corsika_telescope_data_dict():
+    io_handler = simtools.io_handler.IOHandler()
+    io_handler.set_paths(output_path="./simtools-output/", data_path="./data/")
```
GernotMaier, line 786: These are configuration parameters, paths shouldn't be hard coded.

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+        io_handler.get_input_data_file("parameters", "corsika_parameters.yml"), None
+    )
+    keys = ["corsika_sphere_center", "corsika_sphere_radius"]
+    final_dict = {}
+    for key in keys:
+        final_dict[key] = {}
+        for tel_type in file_data[key].keys():
+            final_dict[key][tel_type] = file_data[key][tel_type]["value"] * u.meter
+    final_dict["corsika_obs_level"] = file_data["SITE_PARAMETERS"]["North"]["OBSLEV"][0] * u.cm
+    final_dict["corsika_obs_level"] = final_dict["corsika_obs_level"].to(u.m) * u.meter
```
GernotMaier, line 797: Not sure about this - isn't it already in cm?

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+    file_data = collect_data_from_yaml_or_dict(
+        io_handler.get_input_data_file("parameters", "corsika_parameters.yml"), None
+    )
+    keys = ["corsika_sphere_center", "corsika_sphere_radius"]
+    final_dict = {}
+    for key in keys:
+        final_dict[key] = {}
+        for tel_type in file_data[key].keys():
+            final_dict[key][tel_type] = file_data[key][tel_type]["value"] * u.meter
+    final_dict["corsika_obs_level"] = file_data["SITE_PARAMETERS"]["North"]["OBSLEV"][0] * u.cm
```
GernotMaier, line 796: Why `NORTH` here?

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+        self.telescope_types = ["LST", "MST", "SCT", "SST"]
+        telescope_object_list = ["LSTObject", "MSTObject", "SCTObject", "SSTObject"]
+        telescope_colors = ["darkorange", "dodgerblue", "lightsteelblue", "black"]
+        self.colors_dict = {}
+        self.telescope_object_dict = {}
+        for step, telescope_type in enumerate(self.telescope_types):
+            self.colors_dict[telescope_type] = telescope_colors[step]
+            self.telescope_object_dict[telescope_type] = telescope_object_list[step]
+
+    @u.quantity_input(rotatio_angle=u.deg)
```
orelgueta, line 30: There's a typo in the name of the parameter

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+    def plot_array(self, telescopes, rotate_angle=0 * u.deg):
```
orelgueta, line 31: Do you really need to multiply by the unit here despite the decorator? (asking, not sure)

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        Plot the array of telescopes.
+
+        Parameters
+        ----------
+        telescopes: dict
+            Dictionary with the telescope position and names.
+        rotate_angle:
+            Angle to rotate the plot. For rotate_angle = 0 the X-axis points towards the east, and\
```
orelgueta, line 40: `x-axis` and `y-axis`, i.e., not in capital letters

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+class LayoutArrayBuilder:
+    """
+    Manage the array layout. It provides methods to create and plot a LayoutArray.
+    """
+
+    def __init__(self):
+        """Initialize LayoutArrayBuilder."""
+        self.io_handler = io_handler.IOHandler()
+        self.telescope_types = ["LST", "MST", "SCT", "SST"]
+        telescope_object_list = ["LSTObject", "MSTObject", "SCTObject", "SSTObject"]
```
orelgueta, line 22: I don't think this and the colours should be defined in the init since it ties the plotting into this class while we actually wanted to keep the plotting as independent as possible so it might move to visualisation.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+        Returns
+        -------
+        plt.figure
+            Instance of plt.figure with the array of telescopes plotted.
+
+        """
+        fig, ax = plt.subplots(1)
+        legend_objects = list()
+        legend_labels = list()
+        tel_counters = {one_telescope: 0 for one_telescope in self.telescope_types}
```
orelgueta, line 52: Was this here before? I don't really understand why it's necessary. (I mean in general the counters.)

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+        if rotate_angle != 0:
+            telescopes["pos_x"], telescopes["pos_y"] = self._rotate(
+                rotate_angle, telescopes["pos_x"], telescopes["pos_y"]
+            )
+
+        size_factor = max(np.max(telescopes["pos_x"]), np.max(telescopes["pos_y"])) / (300.0 * u.m)
+        fontsize = 12
+        for tel_name_now in telescopes["telescope_name"]:
```
orelgueta, line 60: I suggest we talk about this part tomorrow cause I am a bit confused about it.

simtools/layout/layout_array_builder.py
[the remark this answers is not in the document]
orelgueta: But it's still used here

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+    def _get_telescope_patch(self, name, x, y, radius):
+        """
+        Collect the patch of one telescope to be plotted by self.plot_array.
+
+        Parameters
+        ----------
+        name: str
+            Name of the telescope (type).
+        x: astropy.units.m
+            X position of the telescope in meters.
```
orelgueta, line 161: `x` (i.e., not capital). same for `y`.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+
+        if len(x) != len(y):
+            raise RuntimeError(
+                "Cannot perform coordinate transformation when x and y have different lengths."
+            )
+
+        try:
+            x = x.to(u.m)
+            y = y.to(u.m)
+            if isinstance(x[0].unit, type(u.m)) and isinstance(y[0].unit, type(u.m)):
```
orelgueta, line 231: You don't need to check this, if the conversion fails then you won't get here.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+                x = x.value
```
orelgueta, line 232: why do you do this? If you are given quantities, you should return quantities rather than removing the units.

simtools/layout/layout_array_builder.py
```diff
@@ -0,0 +1,244 @@
...
+                y = y.value
+        except AttributeError:
+            pass
+
+        x_trans, y_trans = [], []
+        rotate_angle = np.deg2rad(rotation_angle.value)
```
orelgueta, line 238: I recommend using `to('rad')` instead. You don't have to enforce degrees as well in this case.

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+        if _class[0:3] in ("LST", "MST", "SST", "SCT"):
+            return _class[0:3]
+
+    except IndexError:
+        pass
+
+    return ""
+
+
+def get_corsika_telescope_data_dict():
```
orelgueta, line 784: I am sure we read this dict somewhere else, probably in the corsika config module. Perhaps call the function there or write a new function there and call it? Either way, I don't think this should be here in `names.py`

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+    io_handler.set_paths(output_path="./simtools-output/", data_path="./data/")
+    file_data = collect_data_from_yaml_or_dict(
+        io_handler.get_input_data_file("parameters", "corsika_parameters.yml"), None
+    )
+    keys = ["corsika_sphere_center", "corsika_sphere_radius"]
+    final_dict = {}
+    for key in keys:
+        final_dict[key] = {}
+        for tel_type in file_data[key].keys():
+            final_dict[key][tel_type] = file_data[key][tel_type]["value"] * u.meter
```
orelgueta, line 795: I thought the unit is given in the file, why hardcode it here?

simtools/util/names.py
```diff
@@ -748,3 +755,44 @@ def camera_efficiency_log_file_name(site, telescope_model_name, zenith_angle, la
...
+    return ""
+
+
+def get_corsika_telescope_data_dict():
+    io_handler = simtools.io_handler.IOHandler()
+    io_handler.set_paths(output_path="./simtools-output/", data_path="./data/")
+    file_data = collect_data_from_yaml_or_dict(
+        io_handler.get_input_data_file("parameters", "corsika_parameters.yml"), None
+    )
+    keys = ["corsika_sphere_center", "corsika_sphere_radius"]
```
orelgueta, line 790: You should return the full dictionary, not just these keys.

orelgueta (commented):
Thanks for taking into account my comments! I made a few new comments on code I think is new.

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: Yes, indeed I moved now plot-related stuff to visualization/visualize.py

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: Similar yes but not the same. One initializes the class and this one  is to return a dict

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: I use now names.all_telescope_types for accessing the names but individual types have to be specifically written. we intentionally chose this plot marker for the MST, as we decided also for a specific marker for the SCT. We dont expect telescope types to change very often but definitely in case CTA eventually decides for another type of telescope, we would have to include it in the code as well.

The other telescopes will get the fill_flag false as defined before and the Circle as a patch as given by the else statement in the sequence.

simtools/util/names.py
VictorBarbosaMartins: Good. I just adjusted to this dict

simtools/util/names.py
VictorBarbosaMartins: I am having a hard time adjusting to the configuration parameters. should I initialize here the configurator to load the default paths?

simtools/util/names.py
VictorBarbosaMartins: it is in cm but the astropy.units is not read automatically. but I saw an error here and corrected it: I did not need to multiply again "* u.meter" since it is already in meter (given the '.to(u.m) '). thanks for pointing this out

simtools/util/names.py
VictorBarbosaMartins: this function converts the information we need from the corsika_parameters.yml file. before this, we had a manually defined dictionary. In this manually defined dictionary, the observation level was 2158 m, that is, the observation level of the north. That is the reason. but if you prefer, I can pass a flag and get the site through the flag?

simtools/layout/layout_array.py
```diff
@@ -16,6 +16,8 @@
 class InvalidTelescopeListFile(Exception):
     """Exception for invalid telescope list file."""
 
+    pass
```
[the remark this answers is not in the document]
VictorBarbosaMartins: hmm interesting. I dont recall adding those statements, I think it came with merging the master into the branch. In any case, I removed it now

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: yes, otherwise you pass an invalid argument as a default parameter

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: I moved now all things related to plotting to the simtools/visualization/visualize

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: I changed this function, please take a look again. now it tests also if the units are the same

simtools/layout/layout_array_builder.py
VictorBarbosaMartins: Indeed, now we return the same unit as given

simtools/util/names.py
VictorBarbosaMartins: It is indeed used in other places. My argument to keep it here is that the function takes parameters from the yml file, i.e. it is related to the names of telescopes and to parameters that are not going to change as their sizes and observation level. If you prefer I can move it to util.general ?

simtools/util/names.py
VictorBarbosaMartins: I tried to be more consistent now and got it from the file

simtools/util/names.py
VictorBarbosaMartins: I disagree. I implemented this function to replace the previous manually defined dictionary and all info we need is passed here. in case at some point we need more info from this file we can implement it in this function or directly read the file again.

VictorBarbosaMartins:
Hi @GernotMaier , @orelgueta . I worked out most of the suggestions. I argue in some cases against the suggestions and in some cases I could not find a quick solution (in the case of configuration, I am not quite used to this module yet). Please have a look at the opened conversations. Thank you!

VictorBarbosaMartins:
Codacy complains about this `eval`
https://github.com/gammasim/gammasim-tools/blob/8e5fd1396057fb047b03275cb161fcf7bac6edbd/simtools/util/names.py#L797

Including another dependency (import ast) as it suggests doesnt seem to be a wise solution either. what do you think?

orelgueta:
> Codacy complains about this `eval`
> 
> https://github.com/gammasim/gammasim-tools/blob/8e5fd1396057fb047b03275cb161fcf7bac6edbd/simtools/util/names.py#L797
> 
> Including another dependency (import ast) as it suggests doesnt seem to be a wise solution either. what do you think?

I think you can do `* u.Unit(unit)` instead.

simtools/util/names.py
orelgueta: Perhaps I misunderstand and a quick chat would explain it, but the way I see it, we should read this file once for each "run" of gammasim-tools. Then you have the entire dictionary "in memory" and each module that needs it retrieves it and extracts the info it needs from it. This is exactly what we do with the model parameters and it works great. 

That's what I suggest to do here, read this file in `corsika_config` (or somewhere there) and call that function here. It's better to keep it in a class like `corsika_config` rather than in `util.general` cause then you can make sure you read the file only once per instance of `corsika_config` (which is essentially only once per "run" of gammasim-tools) rather than every time you call the namespace `util.general`.

simtools/util/names.py
orelgueta: My answer above should answer this as well I think

orelgueta:
Thanks for the changes @VictorBarbosaMartins! 
I answered some points but will do a more thorough review on Monday most likely.

simtools/layout/layout_array_builder.py
GernotMaier: But both are reading the telescope list from a file of the same type? This looks very similar to me.

simtools/layout/layout_array_builder.py
GernotMaier: But we haven't even agreed (in CTA) how telescopes are named. If we hardwire names now in many places, it will be very hard to change later on. I propose to have telescope names in one single place only.

simtools/util/names.py
GernotMaier: But why should a method called `get_corsika_telescope_data_dict()` return a dict with the observation level for `North` only? What if you want to do that for Paranal?

@orelgueta - do we have the `corsika_sphere_radius` in the database? This is all what is needed here.

I also think that `names.py` is not the correct place to put this method. This is about names....

orelgueta: I made those comments as well and explained my suggestions for them (not much different from yours), see below.

It is not in the DB yet. Generally we don't have CORSIKA in the DB yet as I wanted Victor to finish his tool first and then we can see what exactly is needed. So for now we use this yaml file which has everything.

kosack:
> I agree here with Orel - duplication is definitely bad, but it would be a large effort to adapt gammasim-tools to the ctapipe style.
> Maybe easier to start introduce a DPPS core package with minimal interface for those standard tasks.

ctapipe *is* the core package.  I don't really understand the hesitancy to use it as one.  It is a very small package with minimal required dependencies (and most are the same).   You can't really separate the visualization from the coordinates from the instrument model, as they are all connected. 

But this can be done incrementally, by for example simply using ctapipe's visualization tools to start, and then later other parts like the config system.  I would not suggest migrating everything.   

The reason this is important is from a lesson learned"from HESS (and probably others).  Due to a slight difference in conventions between the simulation software and the data processing software, there were numerous bugs in the software that persisted for years without being caught, event though making a single plot using common plotting utilities would have caught them instantly.   We should avoid that.

simtools/layout/layout_array.py
```diff
@@ -424,6 +489,7 @@ def read_telescope_list_file(self, telescope_list_file):
         self._initialize_corsika_telescope(table.meta)
         self._initialize_coordinate_systems(table.meta)
         self._load_telescope_list(table)
+        return table
```
orelgueta, line 489: It seems you added this return of the table just for the tests. One of the tests (`test_include_radius_into_telescope_table`) doesn't seem necessary because the function `include_radius_into_telescope_table` is not actually used anywhere (and seems weird to me). The other test `test_plot_array` I think should be done by actually initialising the LayoutArray class instead and maybe adding a function to return the entire table after initialisation. 

I am not 100% sure about the last point since the telescope table in LayoutArray include all coordinates in it, but it's still probably good to do. We can discuss this point if you wish.

simtools/layout/layout_array.py
```diff
@@ -763,7 +829,7 @@ def _get_crs_wgs84():
     @staticmethod
     def get_telescope_type(telescope_name):
```
orelgueta, line 827: This function moved to names, so I suggest to remove it from here.

simtools/layout/layout_array.py
```diff
@@ -778,10 +844,33 @@ def get_telescope_type(telescope_name):
...
-            if _class[0:3] in ("LST", "MST", "SST", "SCT"):
+            if _class[0:3] in (names.all_telescope_class_names):
                 return _class[0:3]
 
         except IndexError:
             pass
 
-        return ""
+    @staticmethod
+    def include_radius_into_telescope_table(telescope_table):
```
orelgueta, line 851: This function is not used anywhere but in tests and also seems strange to me. I suggest to remove it, but we can discuss it.

tests/unit_tests/layout/test_layout_array.py
```diff
@@ -232,3 +229,59 @@ def test_altitude_from_corsika_z(layout_center_data_dict, corsika_telescope_data
...
+    corsika_dict = layout_array_instance._from_corsika_file_to_dict()
+    for key, value in corsika_dict.items():
+        if key == "corsika_obs_level":
+            assert value == get_manual_corsika_dict_north["corsika_obs_level"]
+        else:
+            for tel_type, subvalue in value.items():
+                assert subvalue == get_manual_corsika_dict_north[key][tel_type]
+
+
+def test_initialize_corsika_telescope_from_dict(
```
orelgueta, line 282: I modified this code a bit (and the one in the function above). What's interesting is that when I tried running pytest just of this test it failed cause io_handler wasn't initialised. You didn't see it because when you run all tests one after the other, one of them initialises also io_handler. Therefore, I added `io_handler` in the conftest (made a comment there as well).

orelgueta (commented):
Finished going over all the changes and added a few more comments. All in all it looks very nice. Just a few more changes are necessary and we can merge.

simtools/layout/layout_array.py
```diff
@@ -117,15 +124,48 @@ def _initialize_corsika_telescope(self, corsika_dict=None):
...
+            corsika_dict[simtools_par] = {}
+            for tel_type in names.all_telescope_class_names:
+                unit = corsika_parameters_dict[corsika_par][tel_type]["unit"]
+                corsika_dict[simtools_par][tel_type] = corsika_parameters_dict[corsika_par][
+                    tel_type
+                ]["value"] * u.Unit(unit)
+
+        corsika_obs_level = names.translate_simtools_to_corsika("corsika_obs_level")
+        site_par_label = names.translate_simtools_to_corsika("SITE_PARAMETERS")
+        corsika_dict["corsika_obs_level"] = (
```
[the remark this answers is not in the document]
orelgueta: I added the functionality to read from the DB and removed these few lines of code.

applications/print_array_elements.py
```diff
@@ -131,7 +131,7 @@ def main():
     _logger = logging.getLogger()
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
-    layout = layout_array.LayoutArray()
+    layout = layout_array.LayoutArray(site="North")
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I will write an issue for this one

tests/unit_tests/visualization/test_visualize.py
```diff
@@ -91,3 +95,52 @@ def test_add_unit():
...
+    with pytest.raises(TypeError):
+        gen.rotate(angle_deg, x, y[0])
+    with pytest.raises(RuntimeError):
+        gen.rotate(angle_deg, x[:-1], y)
+    with pytest.raises(UnitsError):
+        gen.rotate(angle_deg, x_new_array.to(u.cm), y_new_array)
+
+
+def test_plot_array(telescope_test_file, layout_array_instance):
+    telescope_table = layout_array_instance.read_telescope_list_file(telescope_test_file)
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I will change the docstring to (table)

simtools/layout/layout_array.py
VictorBarbosaMartins: if I remember correctly you agreed to return the table? 

this function is not used anywhere because we still dont have an application for plotting the array. it seems to me the function is necessary if you want to get a dictionary from a telescope file and then pass to the plot_array function. The ideia is to later create an application to do something similar to what we are doing in the test_plot_array. 

the LayoutArray instance is in any case used in the test_plot_array through 'layout_array_north_instance'.

I propose we keep it like this and if we dont like the application later on we can change both

simtools/layout/layout_array.py
VictorBarbosaMartins: same as above, I believe we will need this to the plotting application

simtools/util/names.py
```diff
@@ -27,6 +28,81 @@
     "validate_telescope_name_db",
 ]
 
+lst = "LST"
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I think it would be better to read the code like it is now, but I returned it to the way it was to avoid duplication

simtools/layout/layout_array.py
orelgueta: Yes, I agreed to keeping the part of returning the table. I think it's a good idea to have this option.

simtools/layout/layout_array.py
orelgueta: Technically for plotting you don't need to add a column to the table, you can simply read the radius of each telescope from the metadata in the table (or the DB). If you really want to make the plotting function not dependent on the metadata or the DB and want to keep this function, I won't object. However, in that case, please don't use `tel_name_now[:3]` to get the telescope class but the actual function you added. Also, I really don't understand why you need to use `.split()[0]` and `.split()[1]` to get the radius from the metadata. I thought the metadata is a dictionary.

VictorBarbosaMartins: Indeed I prefer to keep it. I will change the tel_name_now as you suggest

the split is because the telescope name comes as MST-10, then we want only the first part. the second split is to get only the string for the unit to be passed to u.Unit. I dont think astropy would understand if we give directly the string such as `u.Quantity("10 m")` or would it?

tests/unit_tests/layout/test_layout_array.py
```diff
@@ -232,3 +229,59 @@ def test_altitude_from_corsika_z(layout_center_data_dict, corsika_telescope_data
     )
     with pytest.raises(TypeError):
         layout._altitude_from_corsika_z(5.0, None, "LST-01")
+
+
+def test_include_radius_into_telescope_table(layout_array_instance, telescope_test_file):
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Since we decided to keep this function to maintain the plot function independent of the metadata and DB, I am closing here, but feel free to reopen it if you want to further discuss it

simtools/layout/layout_array.py
orelgueta: Wait, but if I understand the code correctly, in this function you refer to the metadata dictionary, not the list of telescopes. So in this line below
```
telescope_table.meta["corsika_sphere_radius"][tel_name_now[:3]].split()[0]
``` 
should actually be 
```
telescope_table.meta["corsika_sphere_radius"][get_telescope_class(tel_name_now)]["value"]
```
and for the unit it should be the same but with unit
```
telescope_table.meta["corsika_sphere_radius"][get_telescope_class(tel_name_now)]["unit"]
```
(right now you are always using the LST unit which should be fine, but not the best practice.

tests/unit_tests/layout/test_layout_array.py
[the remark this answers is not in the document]
orelgueta: OK in that case allow me to review it then. 

I think you can simply use a list below, no need for a numpy array
```
values_from_file = np.array([20.190000534057617, -352.4599914550781, 62.29999923706055, 9.6])
```

I think it would be nicer to run on the actual values of the two lists with zip rather than with enumerate in the line below
```
for key_step, _ in enumerate(keys):
```

Why do you need the `[0]` in
```
telescope_table_with_radius[mst_10_index][keys[key_step]].value[0]
```

Lastly, you shouldn't assert floats with `==` but use `pytest.approx`

simtools/layout/layout_array.py
VictorBarbosaMartins: This does not work. I think you are mixing up with the corsika_parameters.yml where we do have "value" and "unit".
Here, if I print the following for instance:
```
        print(type(telescope_table.meta["corsika_sphere_radius"]))
        print((telescope_table.meta["corsika_sphere_radius"]))
```
I get the following:
```
<class 'dict'>
{'LST': '12.5 m', 'MST': '9.6 m', 'SST': '3 m'}
```
Therefore, there is no "value" and "unit" separated in the metadata. We have to get them manually.

```

Comment (VictorBarbosaMartins) [reply simtools/layout/layout_array.py:868]: Regarding the unit, I agree that it is not the best way. Nevertheless, we cannot pass all the units since we need the unit of the column "radius" in the table, regardless of the type of telescope. In this case, I changed the code to check if the units given in the metadata are the same for the telescope types and extracted one of them to set the table column unit. Please have a look, if you like the way it is I will implement the test for this new raise statement.

File: simtools/util/names.py (hunk 41 above)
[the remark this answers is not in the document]
Comment (VictorBarbosaMartins) [reply simtools/util/names.py:31]: actually changing the dictionary to value: key would disrupt the logic of the variables defined in names. Please have a look at the other all_* variables defined there. using single lst, mst, sct, sst parameters makes the code more readable despite the apparent duplication in names. therefore, I prefer to keep the logic and the readability, but if you prefer you can change it in names. in case you do please make sure you propagate the change to the validation functions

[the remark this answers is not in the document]
Comment (orelgueta) [reply simtools/util/names.py:31]: You can keep it as you had it originally, no problem. We'll see if Gernot complains =)

File: simtools/layout/layout_array.py (hunk 36 above)
Comment (orelgueta) [reply simtools/layout/layout_array.py]: OK, this issue made us realise that the layout files were not all the same. The metadata in one of the files was different to the others. This is now fixed and I think my code suggestion should therefore work now.

File: simtools/visualization/visualize.py [hunk 43]
```diff
@@ -531,3 +542,141 @@ def plot_hist_2D(data, **kwargs):
...
+    x_title = "East [m]"
+    y_title = "North [m]"
+    plt.axis("square")
+    plt.grid(True)
+    plt.gca().set_axisbelow(True)
+    plt.xlabel(x_title, fontsize=18, labelpad=0)
+    plt.ylabel(y_title, fontsize=18, labelpad=0)
+    plt.tick_params(axis="both", which="major", labelsize=15)
+
+    legend_handler_map = {
```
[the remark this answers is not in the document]
VictorBarbosaMartins: I just remember the reason. it is because here we are initializing the handlers. by defining this in the legend_handlers, every time we import leg_h the handlers will be initialized and this leads to an error. I dont see a problem in defining it here where it is actually used

simtools/layout/layout_array.py
```diff
@@ -117,15 +133,84 @@ def _initialize_corsika_telescope(self, corsika_dict=None):
...
+        else:
+            if not isinstance(file_name, Path):
+                file_name = Path(file_name)
+            if file_name.exists():
+                corsika_parameters_dict = collect_data_from_yaml_or_dict(file_name, None)
+            else:
+                raise FileNotFoundError
+
+        corsika_dict = {}
+        corsika_pars = ["corsika_sphere_radius", "corsika_sphere_center"]
```
GernotMaier, line 177: side remark: this should go away in future, as both parameters are not really related to layouts (and partly dependent on the structure)

(but now ok!)

simtools/layout/layout_array.py
```diff
@@ -362,40 +445,62 @@ def _load_telescope_list(self, table):
...
+                    tel.set_coordinates("corsika", row["pos_x"], row["pos_y"])
             except KeyError:
                 pass
             try:
-                tel.set_coordinates(
-                    "utm",
-                    row["utm_east"] * table["utm_east"].unit,
-                    row["utm_north"] * table["utm_north"].unit,
-                )
+                if not isinstance(row["utm_east"], u.Quantity) and not isinstance(
```
GernotMaier, line 461: Here and all the following if statements: this is a lot of repetition and maybe we can simplify the code by:
```
tel.set_coordinates("utm", 
      assign_unit_to_quantity( row["utm_east"], table["utm_east"].unit),
      assign_unit_to_quantity( row["utm_north"], table["utm_north"].unit ))
```
and a small helper function `assign_unit_to_quantity` which does the job of all those if statements.


```

File: simtools/util/general.py [hunk 46]
```diff
@@ -647,3 +650,69 @@ def change_dict_keys_case(data_dict, lower_case=True):
         raise
 
     return _return_dict
+
+
+@u.quantity_input(rotation_angle=u.deg)
+def rotate(rotation_angle, x, y):
+    """
+    Rotate x and y by the rotation angle given in rotation_angle, in degrees.
+    The function returns the rotated x and y values in the same unit given.
```
GernotMaier, line 659: Could you indicate the rotation direction? Clockwise, counterclockwise?

simtools/util/general.py
```diff
@@ -647,3 +650,69 @@ def change_dict_keys_case(data_dict, lower_case=True):
...
+
+    Parameters
+    ----------
+    rotation_angle: astropy.units.deg
+        Angle to rotate the array in degrees.
```
GernotMaier, line 664: Would it be complicated to generalize and accept also rad?

simtools/util/general.py
```diff
@@ -647,3 +650,69 @@ def change_dict_keys_case(data_dict, lower_case=True):
...
+    x: numpy.array or list
+        x positions of the telescopes, usually in meters.
```
GernotMaier, line 666: The function seems to be generic (it is in `general.py`), but here it is about telescope positions? Why e.g., can't we replace  `model/camera.py::_rotate_pixels()` and have only one rotation function?

simtools/visualization/legend_handlers.py
```diff
@@ -15,55 +20,72 @@
...
+            "parameters", "corsika_parameters.yml"
+        )
+        corsika_info = CorsikaConfig.load_corsika_parameters_file(corsika_parameters_file)
+
+        self.radius_dict = {}
+        self.colors_dict = {
+            lst: "darkorange",
+            mst: "dodgerblue",
+            sct: "black",
+            sst: "lightsteelblue",
```
GernotMaier, line 52: I indicated before that I don't like having telescope types hardwired in too many places. Leave it for now, but it is not nice.

simtools/visualization/legend_handlers.py
```diff
@@ -200,67 +222,86 @@ def legend_artist(legend, orig_handle, fontsize, handlebox):
...
+            radius=radius * self.radius_dict[sst] / self.radius_dict[lst],
+            facecolor=self.colors_dict[sst],
+            edgecolor=self.colors_dict[sst],
+            transform=handlebox.get_transform(),
+        )
+        handlebox.add_artist(patch)
+        return patch
+
+
+class SCTHandler(object):
```
GernotMaier, line 291: I know that plotting is always messy, but having to duplicated the NNNHandler for telescope type is really very weird. I guess you've discussed this with Orel and there is no better solution?

simtools/visualization/visualize.py
```diff
@@ -531,3 +542,137 @@ def plot_hist_2D(data, **kwargs):
...
+def get_telescope_patch(name, x, y, radius):
+    """
+    Collect the patch of one telescope to be plotted by plot_array.
+
+    Parameters
+    ----------
+    name: str
+        Name of the telescope (type).
+    x: astropy.units.m
+        x position of the telescope in meters.
```
GernotMaier, line 557: Generally: is there a strong reason to ask for meters, and not for units of lengths? If you anyway convert to value, you can also convert to meters first.

simtools/visualization/visualize.py
```diff
@@ -531,3 +542,137 @@ def plot_hist_2D(data, **kwargs):
...
+        Instance of plt.figure with the array of telescopes plotted.
+
+    """
+
+    fig, ax = plt.subplots(1)
+    legend_objects = []
+    legend_labels = []
+    tel_counters = {one_telescope: 0 for one_telescope in names.all_telescope_class_names}
+    if rotate_angle != 0:
+        telescopes["pos_x"], telescopes["pos_y"] = gen.rotate(
```
GernotMaier, line 616: Does this modify the original positions in the telescope variables? Meaning calling `plot_array` twice with a non-zero rotation angles will result in different plots?

tests/conftest.py
```diff
@@ -186,3 +188,29 @@ def telescope_model_sst(db, db_config, io_handler):
         label="test-telescope-model-sst",
     )
     return telescope_model_SST
+
+
+@pytest.fixture
+def telescope_test_file(db, args_dict, io_handler):
+    test_file_name = "telescope_positions-North-TestLayout.ecsv"
```
GernotMaier, line 195: Want to note that all tests seems to use `North` layouts, which is the one without SSTs. Not sure if this might cause an issue, just think about it.

applications/print_array_elements.py
```diff
@@ -120,18 +120,18 @@ def _parse(label=None, description=None):
...
 
     label = Path(__file__).stem
-    args_dict, _ = _parse(label, description=("Print a list of array element positions"))
+    args_dict, db_config = _parse(label, description=("Print a list of array element positions"))
 
     _logger = logging.getLogger()
     _logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
-    layout = layout_array.LayoutArray()
+    layout = layout_array.LayoutArray(site="North", mongo_db_config=db_config)
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Good question. This is not the best solution. It would be better to get the site directly from the file (given by `args_dict["array_element_list"]`) that is anyways used in the line below this one. To do that we would have to read the file before initializing the class, therefore, I changed the code now and divided the `read_telescope_list_file` into two files: `read_telescope_list_file` (now static method, i.e. one can use without initializing the class) and `initialize_layout_array_from_telescope_file` which does exactly what `read_telescope_list_file` was doing before. With this we are able to implement a function that reads the metadata from the telescope file and give us the site, then we initialize the class. Nevertheless, in discussions with Orel we realized that we want to drastically reduce the metadata in the telescope file and instead of passing all the coordinates info, we simply pass the site and then the rest of the information is read from the DB. Since we will probably implement that later, I suggest we keep this "North" here and open an issue so we dont forget it.

simtools/layout/layout_array.py
VictorBarbosaMartins: That is right. We want to read all this from the database, which will have separated site and model parameters.

simtools/layout/layout_array.py
VictorBarbosaMartins: That is a good idea. I now created not only the `_assign_unit_to_quantity` but also the `_try_set_coordinate` to avoid the excessive number of `try and excepts` and the `_try_set_altitude` for consistency. The three new functions have also unit tests now. Please have a look and let me know what you think. 

There are other places with a lot of if and try and except statements, such as  many of the _initialize fuinctions but I wouldnt further extend this PR but rather open an issue for that?

simtools/util/general.py
VictorBarbosaMartins: not so much, I just implemented it, with the test.

simtools/util/general.py
VictorBarbosaMartins: Yes, that is the idea. Orel is going to check if that is possible. I will open an issue for that so we dont forget it.

simtools/visualization/legend_handlers.py
VictorBarbosaMartins: the telescopes are not hardwired here, these are parameters which define the strings we use for the telescopes. the definition is centralized in `names.py`. everywhere else we import and use these variables. if at some point we want to change, we do it in a single place. If at some point we want to add a different kind of telescope we would indeed need to add it to the definitions to `names.py` and add the variables throughout the code (DB and so on..). I dont see a way to cover all kinds of telescopes other than this. Do you have an idea how to improve this?

By the way, I centralized the colors now as well. Now they are defined only in TelescopeHandler for all telescopes.

simtools/visualization/legend_handlers.py
VictorBarbosaMartins: Yes, this is something Orel did. I think we do have to have it like this, since the legends are different for the different type of telescopes.

simtools/visualization/visualize.py
VictorBarbosaMartins: hmm yes, because we want to plot it in meters. This function receives the positions from  "pos_x" and "pos_y" keywords, which we define in meters. I dont see a reason to expand the function here. Do you think it is necessary?

simtools/visualization/visualize.py
VictorBarbosaMartins: I tested it now and indeed this happens. do you think that it is a problem? 
what about introducing a flag to ask if we want to permanently rotate or just for the plot?

tests/conftest.py
VictorBarbosaMartins: Good point. This would interfere with the entire layout array. I will open an issue on that to check how can we make sure to test both sites.

VictorBarbosaMartins:
> Thanks for this - a few minor comments and considerations.
> 
> `tests/unit_tests/visualization/test_legend_handlers.py` is empty - this is what you want? Hard do test?

Ah I just saw this one comment. It is in this PR empty but I started implementing it in #434 and Orel said it is better to finish this PR first.

applications/print_array_elements.py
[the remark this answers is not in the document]
GernotMaier: Do I understand it correctly that `site` is needed because now also `mongo_db_config` is added to the constructed, although it is only needed in `from_layout_array_name`? My suggestion would be to remove site and mongo_db_config from the `_init_` function. This means you wouldn't need the db config in print_array_elements.

simtools/util/general.py
GernotMaier: This is #438

simtools/layout/layout_array.py
GernotMaier: Great, thanks (and I believe it was me introducing the repetitive code)

simtools/visualization/visualize.py
GernotMaier: All what you need would be to use instead of `x` the converted value `x.to(u.m)`, so that is not a large expansion.

simtools/visualization/visualize.py
GernotMaier: Yes, we will have forget soon that a plotting function is modifying the values. So better to copy it into new variables.

GernotMaier (changes requested):
One thing left: please have a look at the init function of `LayoutArray`.

applications/print_array_elements.py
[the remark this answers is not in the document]
orelgueta: `mongo_db_config` will anyway be needed here cause we will always want to read the site info from the DB then from the file. So at least that we should keep.

simtools/util/general.py
orelgueta: This should be possible, the only difference between the rotations is the direction. For the pixels the rotation is clockwise and here anticlockwise. The only question is whether we should add an option in this rotation function for the direction or should I change some code in the pixel rotations to accept also anticlockwise rotation (I think I prefer the former).
(will add the same comment to the issue.)

simtools/visualization/legend_handlers.py
orelgueta: Yes, there's no choice, that's how matplotlib handles legend handlers.

simtools/visualization/visualize.py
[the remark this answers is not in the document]
orelgueta: Oh right, then it's perfectly fine. What about the question whether you can write it explicitly instead of building the map from the lists?
(I admit I did not recheck if that question is still relevant considering your answer, so if it isn't, please just say that it isn't relevant anymore.)

applications/print_array_elements.py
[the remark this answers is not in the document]
GernotMaier: Maybe I misunderstand the changes - but in the current implementation `applications/print_array_elements.py` does not need the database, `LayoutArray:_init()_` does not need the DB - is only needed when calling `LayoutArray::from_layout_array_name`. So why adding a dependency, in the init function?

I think it is important to keep the code always as simple as possible - this is a tool used a lot when discussing telescope positions. Keeping the option of having all information in the metadata header of the ecsv files about sites is an important feature and should be kept.

simtools/visualization/visualize.py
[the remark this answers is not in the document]
VictorBarbosaMartins: We could write it explicitly but that would mean we would be hard-wiring the telescope handlers. with the current implementation we get the list the list with all the handlers defined in legend_handlers. we would have to make sure to keep that one list updated but we dont need to worry about the use of the handlers elsewhere.

applications/print_array_elements.py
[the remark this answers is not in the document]
VictorBarbosaMartins: it seems to me we need to decide first if we want to get the site parameters from the DB or always from the header in the telescope files. Please see my question here:

#https://github.com/gammasim/gammasim-tools/issues/435#issuecomment-1354725181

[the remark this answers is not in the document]
VictorBarbosaMartins: I was trying to remove the dependency on the site parameter, I found only one place where it is important now:
https://github.com/gammasim/gammasim-tools/blob/2e118f51f26bd53830035fcd46013cefbe6f94e0/simtools/layout/layout_array.py#L142

This function is now being used in the initialization of the class and it is important to collect the corsika parameters to build the array. especially the observational level is of importance here and it depends on which site we are talking about. To summarize, we need the site as a parameter to load the corsika parameters into the class instance.

I cant find a good solution now to get rid of the site parameter and at the same time keep the way we initialize the class here. Does any of you have an idea?

GernotMaier:
We might have discussed it offline, but I repeat my question about the change of the API for the init function of LayoutArray. What do I miss here to understand why it is needed? Neither site or mongo_db are needed in the init and also not for when using the class as a simple converter of a ecsv file. If both of you insist, then we can do it - but I would like to understand...

orelgueta:
I agree that site is not always needed and we can make it optional (in case it is given in the file). The MongoDB info is a bit more necessary in case you want to limit the info in the file to just the site. However, we can make that optional as well in case all of the location information is given in the file (as is done now). Supporting both options would definitely make the class more flexible. The only question is whether this flexibility is necessary. I will anyway change the site requirement since it's an easy change. With the MongoDB info I will wait for your opinion since generally I don't mind the dependence on the DB.

GernotMaier (approved):
I approve it now - we've discussed everything and the issue on site / mongodb should be clear. I leave it to both of you to decide what to do.

orelgueta:
Thanks @GernotMaier!
@VictorBarbosaMartins, please wait for me to do what I mentioned in the previous comment with the site and DB config before merging. It should be fairly simple and since it's part of this same PR, I think it should be done here.

orelgueta:
I removed the requirement to provide the site and the MongoDB connection details in the case the info exists in the layout file. However the change discussed in #436 is necessary for the completeness of this PR I think. Nevertheless, @VictorBarbosaMartins, please check my changes and if you are happy with them I think we can merge. The rest can be done in separate PRs.

tests/unit_tests/layout/test_layout_array.py
```diff
@@ -205,7 +205,7 @@ def test_build_layout(layout_array_north_four_LST_instance, tmp_test_directory,
     layout.export_telescope_list(crs_name="corsika")
 
     # Building a second layout from the file exported by the first one
-    layout_2 = LayoutArray(layout.telescope_list_file, name="test_layout_2")
+    layout_2 = LayoutArray(site=layout.site, mongo_db_config=db_config, name="test_layout_2")
```
VictorBarbosaMartins, line 208: I dont think this is the correct way. Here we are generating the second layout instance based on an attribute from the first layout instance, which is not the idea of the test. The idea is generating a second class instance based on a telescope file. But I agree that in order to do that we have to be able to get the site from the telescope file (as discussed in #436). I propose we return this to how it was and comment it out until we have this new feature ready (and make sure we dont forget to uncomment later).

simtools/layout/layout_array.py
```diff
@@ -195,6 +223,13 @@ def _from_corsika_file_to_dict(self, file_name=None):
                     )
                     corsika_dict[simtools_par][tel_type] = corsika_dict[simtools_par][tel_type]
 
+        if self.mongo_db_config is None:
+            self._logger.error("DB connection info was not provided, cannot set telescope altitude")
+            raise ValueError
+        if self.site is None:
+            self._logger.error("Site was not provided, cannot set telescope altitude")
+            raise ValueError
```
VictorBarbosaMartins, line 203: Just to make sure we are on the same line of thoughts. Once we implement the feature to get the site from `telescope_list_file` in case the file is passed as argument (as discussed in #436), we will also accept `self.site` and `self.mongo_db_config` as None and will not raise these errors. is that right?

simtools/layout/layout_array.py
```diff
@@ -195,6 +223,13 @@ def _from_corsika_file_to_dict(self, file_name=None):
                     )
                     corsika_dict[simtools_par][tel_type] = corsika_dict[simtools_par][tel_type]
 
+        if self.mongo_db_config is None:
+            self._logger.error("DB connection info was not provided, cannot set telescope altitude")
```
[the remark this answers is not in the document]
orelgueta: Fixed

tests/unit_tests/layout/test_layout_array.py
orelgueta: This was a mistake. I changed it for a second to test something and forgot to change it back (my docker was broken so I had to run the tests using GitHub). I changed it back now, specifying the attribute name. 

To clarify, at the moment we don't need to have the site in the init if we provide a file in the init. We currently assume that the file contains the CORSIKA location and altitude. As part of #436 we should add a check that this is really true and if not read it from the DB.

simtools/layout/layout_array.py
orelgueta: As I mentioned above, we already accept them as `None` if we provide a `telescope_list_file`. What will change later is that we could read the site from the file sometimes. I suggest we have a quick chat before we start implementing #436 to make sure it's all clear.

tests/unit_tests/layout/test_layout_array.py
VictorBarbosaMartins: Ok, that sounds good. I think I understand it now. we only demand a site parameter in `_from_corsika_file_to_dict`, which is specifically used in case the telescope file is not passed. that is fine by now. we will have to make sure to set the self.site parameter from the telescope file in #436, but I agree we can chat a bit before implementing anything there.

VictorBarbosaMartins (commented):
Ok, I think we can finally merge this, after more than 1 month (I cannot approve since I am also author here). Thanks for finishing these last changes @orelgueta!

Issue #417, opened 2022-11-28
Unused parameter in corsika_runner


orelgueta:
https://github.com/gammasim/gammasim-tools/blob/8cdb37a97cf9c4fcf37ed0380927c28e4735cb3a/simtools/corsika/corsika_runner.py#L84

This parameter is set in the simulator, but is not actually used anywhere in the CORSIKA runner. It's seems weird to simply remove it. Any ideas @GernotMaier?

GernotMaier:
`CorsikaConfig` has an option to load CORSIKA parameters from a yaml file, but this is not set in corsika_runner. 

This needs a bit more investigation to understand better if we want that option or not (if not, than we should remove the code consistently also in corsikaconfig. 

I assigned the issue to myself and will look into it.

GernotMaier:
`corsika_parameters_file` is used (implemented in one of the merge request listed above). I therefore close this issue.

PR #430, opened 2022-11-29
Fix pylint issues

### LINKED ISSUE #415: Remove unused defaults_init option in layout_array?
The option `defaults_init` in

https://github.com/gammasim/gammasim-tools/blob/8cdb37a97cf9c4fcf37ed0380927c28e4735cb3a/simtools/layout/layout_array.py#L193

is not used. @GernotMaier, can I remove it or should I implement a usage for it?

orelgueta:
Fixing most issues reported by pylint. A few were ignored, in particular these ones with the flag `--disable=`

- `W1203` - Use lazy % formatting in logging functions
- `C0103` - Variable name "..." doesn't conform to snake_case naming style
- `E0401` - unable to import '...'
- `R0902` - Too many instance attributes
- `R0913` - Too many arguments
- `R0914` - Too many local variables

I also used the flag `--logging-format-style=new` when running pylint in an attempt to avoid the issue about lazy logging, but it doesn't always work. I don't think we should switch to lazy logging (and have most of the internet backing me up).

I added some pylint messages in the code as well so that it ignores specific lines or pieces of code which we know are OK.

Remaining issues that we should actually resolve are missing module docstrings, TODOs, FIXMEs and some specific issues like the one reported in #417. 

pylint also complains about duplication of code. Not sure if we can or should resolve these. I suggest we take a closer look in the future.

In general I think we can start using pylint as a prehook if we decide to officially ignore the messages above and fix or decide about the few remaining others. We should probably ignore the module docstring as well in this case until they are implemented.

Lastly, I also changed everywhere to use f-string, please use that from now on exclusively.

This PR closes #415.


GernotMaier (approved):
Very nice! 
```
Your code has been rated at 9.02/10 (previous run: 7.69/10, +1.33)
```
So a really significant improvement.
(and I didn't ignore your sugesstions on `-disable` flags.

GernotMaier:
Agree with your suggestions on moved the pre-commit topic to #431

(and not sure if I should merge, as it is not clear if you ask for 2 or at least 1 out of 2 reviewers)

orelgueta:
Thanks for the review and for opening the issue. I merged the PR. I asked for 2 reviewers as a way to ask both of you and planned to accept the quickest person. Sometimes I ask a specific person because it relates to code they wrote.

Issue #428, opened 2022-11-29
LayoutArray


maxnoe:
The LayoutArray (why the strange order name?) has `add_telescope` which uses `xx`, `yy` and `corsika_z`. 

However, other parts of the documentation note that the goal is to be disentangled as much as possible from the simulation software.

So why not specify the real z coordinate and then compute the value needed for CORSIKA in the code path dealing specifically with CORSIKA?

maxnoe:
Also: xx and yy do not provide any further information over just using x and y, it just looks weird.

GernotMaier:
Forgot about this one, sorry @maxnoe 

The logic is that `telescope_position` handles all relevant coordinate systems (UTM, mercator, CORSIKA) and this is the only place in the code where we do transformation of spatial coordinates. Don't think we really introduce a dependence, I would call it an option to use CORSIKA coordinates (as `add_telescope(self, telescope_name, crs_name, xx, yy, altitude=None, tel_corsika_z=None)` allows to use either altitude or corsika_z)

For xx, yy, zz: probably the original thought was to distinguish it from x, y, z (is it can be long, lat, altitude). I'll change it to x, y, z if this is clearer.

GernotMaier:
I remember now why `xx` etc is used: we are using the pyproj package for transformation, and this is the convention for variable naming there (e.g., see https://pyproj4.github.io/pyproj/stable/api/transformer.html#pyproj.transformer.Transformer.transform)

So I think this is fine and well motivated.

GernotMaier:
Closing this issue, as I think the naming question has been addressed and using the convention from pyproj is well motivated.

maxnoe:
> The logic is that telescope_position handles all relevant coordinate systems

Then I'd expect an `ArrayLayout` to be a collection of `TelescopePosition` s

GernotMaier:
It is kind of: `LayoutArray._telescope_list` contains all `TelescopePositions`s .

(and I agree on the naming, don't remember why we called it LayoutArray)

PR #441, opened 2022-12-19
Application for plotting array

### LINKED ISSUE #339: Layout plotting (sites branch)

VictorBarbosaMartins:
Closes #339

Implemented the application for plotting the array. For more examples please see the integration tests.
This PR is built upon #428. Please approve and merge that one first.


GernotMaier:
Did you also update the documentation? Better to check the docu how to use the application then the integration tests...

VictorBarbosaMartins:
> Did you also update the documentation? Better to check the docu how to use the application then the integration tests...

I just updated the documentation as well, thanks for the reminder. 

I implemented the application from scratch, there was no documentation. The tests were failing due to a feature in the code, they should pass now.

Before reviewing here, we need to approve and merge #428 . There we haven't agreed yet on passing 'site' as a parameter to LayoutArray. Orel and myself are for passing it. I suggest we approve there and if later we decide not to pass 'site' anymore we adapt the code but it is important now for the further development to have it.

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+        "--figure_name",
+        help="Name of the output figure to be saved into as a pdf.",
+        type=str,
+        required=False,
+        default=None,
+    )
+    config.parser.add_argument(
+        "--rotate_angle",
+        help="Angle to rotate the array before plotting (in degrees).",
+        type=str,
```
orelgueta, line 84: Why is this a string and not a float?

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+    input_group.add_argument(
+        "--layout_array_name",
+        help="Name of the layout array.",
+        nargs="+",
+        type=str,
+        required=False,
+        default=None,
+    )
+
+    return config.initialize(db_config=True)
```
orelgueta, line 114: Why is the db_config set to True here? I do not see the DB actually used below.

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+    description = "Plots layout array."
+    usage = "python applications/plot_layout_array.py --layout_array_name test_layout"
+    args_dict, _ = _parse(label, description, usage)
+    io_handler_instance = io_handler.IOHandler()
+
+    if args_dict["rotate_angle"] is None:
+        rotate_angles = [0 * u.deg]
+    else:
+        rotate_angles = []
+        for _, one_angle in enumerate(args_dict["rotate_angle"]):
```
orelgueta, line 129: Why do you use `enumerate` without actually using the counter? Seems unnecessary.

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+        logger.info("Plotting array from telescope list file(s).")
+        telescope_file = args_dict["telescope_list"]
+
+    elif args_dict["layout_array_name"] is not None:
+        logger.info("Plotting array from layout array name(s).")
+        telescope_file = [
+            io_handler_instance.get_input_data_file(
+                "layout", f"telescope_positions-{one_array}.ecsv"
+            )
+            for _, one_array in enumerate(args_dict["layout_array_name"])
```
orelgueta, line 145: Same question as above about `enumerate`

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+        Shows the telescope labels in the plot.
+    verbosity (str, optional)
+        Log level to print (default=INFO).
+
+    Example
+    -------
+    .. code-block:: console
+
+        python applications/plot_layout_array.py --figure_name northern_array_alpha \
+        --layout_array_name test_layout
```
orelgueta, line 36: I modified this to be a working example

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+            telescopes_dict = LayoutArray.include_radius_into_telescope_table(telescope_table)
+            fig_out = plot_array(
+                telescopes_dict, rotate_angle=one_angle, show_tel_label=args_dict["show_tel_label"]
+            )
+            output_dir = io_handler_instance.get_output_directory(
+                label, dir_type="application-plots"
+            )
+            plot_file = output_dir.joinpath(plot_file_name)
+
+            for f in ["pdf", "png"]:
```
orelgueta, line 175: I modified this to be `ext`, but it's not important.

orelgueta (commented):
A few minor comments and made some very minor changes. 

However, the real issue is when I looked at the plots. They are rotated with respect to reality. The south is rotated by 90 degrees, the north I think as well, but I am not sure. Also, the SSTs are way too small and are essentially invisible with this colour. 

Also, when plotting just 4 LSTs with `--layout_array_name North-4LST`, the plot is cut and therefore shows only two LSTs.

applications/plot_layout_array.py
```diff
@@ -0,0 +1,182 @@
...
+
+    A rotation angle in degrees can be passed in case the array should be rotated before plotting.
+    A sequence of arguments for the rotation angle is also permitted, in which case all of them
+    are plotted and saved separately.
+
+
+    Command line arguments
+    ----------------------
+    figure_name (str, optional)
+        File name for the pdf output (without extension).
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Indeed it would be better. I implemented my idea, please take a look it thats fine for you.

applications/plot_layout_array.py
VictorBarbosaMartins: I had some troubles with initializing it as None (and type float) so I defined it to be a str. would it be better to define it as float and initialize as 0.0?

applications/plot_layout_array.py
VictorBarbosaMartins: I have copied it from the other application but If we dont need it I am removing the dependence now

applications/plot_layout_array.py
VictorBarbosaMartins: I thought codacy would complain but you are right, since I am using only the value and not the counter I dont need enumerate. I am removing it now

applications/plot_layout_array.py
VictorBarbosaMartins: I dont understand. What would be a working example? And have you pushed any changes to the PR ? I don't see your commit

applications/plot_layout_array.py
VictorBarbosaMartins: again, I dont see your commit. please remember to checkout the branch after my changes in case you want to change directly something in the code

VictorBarbosaMartins:
@orelgueta thanks for the comments. Some points are still open, and I am still going to check if the arrays are rotate and why. I am still going to adjust the size of the sst markers as you suggested.

VictorBarbosaMartins:
> However, the real issue is when I looked at the plots. They are rotated with respect to reality. The south is rotated by 90 degrees, the north I think as well, but I am not sure. Also, the SSTs are way too small and are essentially invisible with this colour.
I have checked and the arrays are not rotated. You can confirm it by checking the plotted image with the values in the ecsv file. Here I am assuming that east is the positive X axis and north is the positive Y axis. If this is not the case, please let me know.

VictorBarbosaMartins:
> Also, when plotting just 4 LSTs with `--layout_array_name North-4LST`, the plot is cut and therefore shows only two LSTs.
The plot is not cut but if you look at `data/layout/telescope_positions-4LST.ecsv` you will see that two telescopes are placed at the same position twice. I don't actually know why this was done (I havent created those files) or whether it makes sense, but the problem here is not with the plotting. Please let me know if the positions of telescopes have to be changed in the file.

VictorBarbosaMartins:
> Also, the SSTs are way too small and are essentially invisible with this colour.

I returned to the definition of a scale parameter to increase the size of the markers in case the total number of telescopes is larger than 30 and changed the color of the SST marker. Furthermore, I increased the dpi of the saved image, which is necessary in case we want to zoom in.

Printing the label of the telescopes is still not ideal, but it will hardly be, since it is hard to plot 100 markers with a label in a reasonable size without overlaying each other (Try `python applications/plot_layout_array.py --telescope_list data/layout/telescope_positions-South-TestLayout.ecsv --figure_name test.png --show_tel_label` and look at `simtools-output/plot_array_layout/applications-plots/test.png` to see what I mean)

applications/plot_layout_array.py
orelgueta: Sorry, I committed but forgot to push...
See now the changes.

applications/plot_layout_array.py
orelgueta: Again, forgot to push. This resulted in a conflict so I reimplemented the small change after your modifications. Your modifications are a bit more complex than I expected though, so I will review that now.

applications/plot_layout_array.py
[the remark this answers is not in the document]
orelgueta: Yes, the current implementation is fine, resolving this thread.

orelgueta:
> > However, the real issue is when I looked at the plots. They are rotated with respect to reality. The south is rotated by 90 degrees, the north I think as well, but I am not sure. Also, the SSTs are way too small and are essentially invisible with this colour.
> 
> I have checked and the arrays are not rotated. You can confirm it by checking the plotted image with the values in the ecsv file. Here I am assuming that east is the positive X axis and north is the positive Y axis. If this is not the case, please let me know.

The coordinate system of the CORSIKA inputs is what's rotated I think. We'll confirm and continue discussing this on Mattermost.

orelgueta:
> > Also, when plotting just 4 LSTs with `--layout_array_name North-4LST`, the plot is cut and therefore shows only two LSTs.
> 
> The plot is not cut but if you look at `data/layout/telescope_positions-4LST.ecsv` you will see that two telescopes are placed at the same position twice. I don't actually know why this was done (I havent created those files) or whether it makes sense, but the problem here is not with the plotting. Please let me know if the positions of telescopes have to be changed in the file.

Yes, I noticed it now when reviewing the other PR. That is clearly a typo in the telescope positions which should be fixed in those files.

orelgueta:
> > Also, the SSTs are way too small and are essentially invisible with this colour.
> 
> I returned to the definition of a scale parameter to increase the size of the markers in case the total number of telescopes is larger than 30 and changed the color of the SST marker. Furthermore, I increased the dpi of the saved image, which is necessary in case we want to zoom in.
> 
> Printing the label of the telescopes is still not ideal, but it will hardly be, since it is hard to plot 100 markers with a label in a reasonable size without overlaying each other (Try `python applications/plot_layout_array.py --telescope_list data/layout/telescope_positions-South-TestLayout.ecsv --figure_name test.png --show_tel_label` and look at `simtools-output/plot_array_layout/applications-plots/test.png` to see what I mean)

The plot looks good now. The SSTs are much more visible and even when printing the labels everything is fairly clear. It might be even clearer after rotating the plot.

orelgueta (approved):
Approving after the various changes implemented and discussion on Mattermost

PR #447, opened 2023-01-06
Ignore the py23 deprecated warning

orelgueta:
Matplotlib depends on fontools which comes with this annoying deprecated warning. Apparently everyone (including matplotlib), simply adds an ignore hook to pytest to ignore this warning instead of fixing it. Since we anyway cannot do anything about it, I suggest to add the same ignore hook.


GernotMaier (approved):
Nice.

PR #459, opened 2023-05-31
Allow to print heights for non-telescope-type array elements

GernotMaier:
Some telescope lists involve non-telescope array elements (e.g., positions of illuminators). The conversion from UTM to corsika coordinates and altitude did not work as a `corsika_sphere_center` entry was required.

- return `0. * u.m` as `corsika_sphere_center`  for non-telescope elements
- the test telescope list [tests/resources/telescope_positions-North-utm.ecsv](tests/resources/telescope_positions-North-utm.ecsv) missed the `corsika_sphere_center`  entries
-  added an integration test for this use case

### DIFF
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -376,7 +376,9 @@ class LayoutArray:
 
         """
 
-        if names.get_telescope_type(tel_name) is not None:
+        if (names.get_telescope_type(tel_name) is not None
+                and len(names.get_telescope_type(tel_name)) > 0):
+
             return self._corsika_telescope["corsika_sphere_center"][
                 names.get_telescope_type(tel_name)
             ]
--- a/tests/integration_tests/test_applications.py
+++ b/tests/integration_tests/test_applications.py
@@ -368,6 +368,15 @@ APP_LIST = {
             "corsika",
         ],
     ],
+    "print_array_elements::print_compact_cors_telescopeheights": [
+        [
+            "--array_element_list",
+            "tests/resources/telescope_positions-North-utm.ecsv",
+            "--export",
+            "corsika",
+            "--use_corsika_telescope_height",
+        ],
+    ],
 }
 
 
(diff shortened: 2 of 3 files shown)


VictorBarbosaMartins (approved):

tests/integration_tests/test_applications.py
```diff
@@ -368,6 +368,15 @@
             "corsika",
         ],
     ],
+    "print_array_elements::print_compact_cors_telescopeheights": [
```
orelgueta, line 371: Is "cors" here short for corsika? I suggest to change to corsika in that case.

simtools/layout/layout_array.py
```diff
@@ -376,7 +376,9 @@ def _get_corsika_sphere_center(self, tel_name):
 
         """
 
-        if names.get_telescope_type(tel_name) is not None:
+        if (names.get_telescope_type(tel_name) is not None
```
orelgueta, line 379: When would `get_telescope_type` return `None`? Looking at the code, I am not sure it ever would.

orelgueta (commented):
Very minor comments

orelgueta:
> It looks good. I wanted to cross check the values of the CORSIKA parameters that you wrote onto the file with the telescope files from the DB but I haven't found the files in the test-data database. Do you know what happened to them @orelgueta ?

I think we removed these files from the DB and kept them in `tests/resources` and `data/layout/` instead. However, I cannot find that conversation in any of the histories. Do you remember where we made that decision?

tests/integration_tests/test_applications.py
GernotMaier: Fixed.

simtools/layout/layout_array.py
GernotMaier: True, fixed!

GernotMaier:
> > It looks good. I wanted to cross check the values of the CORSIKA parameters that you wrote onto the file with the telescope files from the DB but I haven't found the files in the test-data database. Do you know what happened to them @orelgueta ?
> 
> I think we removed these files from the DB and kept them in `tests/resources` and `data/layout/` instead. However, I cannot find that conversation in any of the histories. Do you remember where we made that decision?

Don't remember - but note that we will modify this application in future to allow reading the corsika_sphere etc from the DB. For now, we use this application for the array layout definition and information is needed in the file.

orelgueta (approved):

PR #461, opened 2023-06-12
First working example of production script for rel 0

orelgueta:
There are still many things to do (some are marked with TODOs), but I think it's a first place to stop and ask for a review.


applications/simulate_prod.py
```diff
@@ -0,0 +1,346 @@
...
+
+def _proccess_simulation_config_file(config_file, primary_config, logger):
+    """
+    Read simulation configuration file with details on shower
+    and array simulations
+
+    Attributes
+    ----------
+    config_file: str
+        Name of simulation configuration file
```
GernotMaier, line 212: I have a colleague who would remark on the missing `.` at the end of the sentence (and in other places).

applications/simulate_prod.py
```diff
@@ -0,0 +1,346 @@
...
+        Name of the primary selected from the configuration file.
+
+    Returns
+    -------
+    str
+        label of simulation configuration
+    dict
+        configuration of shower simulations
+    dict
+        configuration of array simulations
```
GernotMaier, line 223: Do we agree that `array simulations` are a good expression for the telescope simulations? I had to go forth and back to look up what `config_array` is. What would be wrong with `telescope simulations`? Not good because we have several telescopes?

simtools/configuration/commandline_parser.py
```diff
@@ -197,20 +197,20 @@ def initialize_telescope_model_arguments(self, add_model_version=True, add_teles
 
         Parameters
         ----------
-        add_model_version: str
-            Model version.
-        add_telescope: str
-            Telescope model name (e.g. LST-1, SST-D, ...).
+        add_model_version: bool
+            Set if to allow a simulation model argument.
```
GernotMaier, line 201: Shouldn't it be 'Set to allow'...?

simtools/configuration/commandline_parser.py
```diff
@@ -197,20 +197,20 @@ def initialize_telescope_model_arguments(self, add_model_version=True, add_teles
...
+        add_telescope: bool
+            Set if to allow a telescope name argument.
```
GernotMaier, line 203: Shouldn't it be 'Set to allow'...?

applications/simulate_prod.py
```diff
@@ -0,0 +1,346 @@
+#!/usr/bin/python3
+
+"""
+    Summary
+    -------
+    This application is used to run simulations for productions (typically on the grid).
+    It allows to run a Paranal (CTA-South) or La Palma (CTA-North) array layout simulation
```
GernotMaier, line 7: Further down, it is `CTAO-Site`, maybe we should consistently use `CTAO` (what a useless comment..)

simtools/configuration/commandline_parser.py
```diff
@@ -197,20 +197,20 @@ def initialize_telescope_model_arguments(self, add_model_version=True, add_teles
...
         """
 
         _job_group = self.add_argument_group("telescope model")
         _job_group.add_argument(
-            "--site", help="CTAO site (e.g. North, South)", type=self.site, required=False
+            "--site", help="CTAO site (e.g., North, South)", type=self.site, required=False
```
GernotMaier, line 208: in `simulate_prod` we are using paranal and la palma for the sites, and not North/South. Not sure if this is an issue, but we should be consistent.

simtools/corsika_simtel/corsika_simtel_runner.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def __init__(self, common_args=None, corsika_args=None, simtel_args=None):
+
+        CorsikaRunner.__init__(self, use_multipipe=True, **(common_args | corsika_args))
+        SimtelRunnerArray.__init__(self, **(common_args | simtel_args))
+
+    def get_run_script(self, **kwargs):
+        """
+        Get the full path of the run script file for a given run number.
+
```
GernotMaier, line 53: Do I understand it correctly that this function is more than a getter, as it calls the export functions and writes files? Suggest to use a different names, as getters (at least in my understanding) should simply get something. Maybe `prepare_run_script`?

simtools/simulator.py
```diff
@@ -208,6 +213,12 @@ def _load_corsika_config_and_model(self, config_data):
             "corsika_parameters_file", None
         )
 
+        # Remove sim_telarray parameters from the CORSIKA config dictionary
```
GernotMaier, line 216: is this necessary to run the code or is it simply that you don't want to have sim_telarray configuration parameter in the corsika config.

simtools/util/general.py
```diff
@@ -72,7 +72,7 @@ def file_has_text(file, text):
         return True
 
 
-def validate_config_data(config_data, parameters):
+def validate_config_data(config_data, parameters, ignore_unidentified=False):
```
GernotMaier, line 75: ignore_unidentified missing from docstring

simtools/simulator.py
```diff
@@ -224,7 +235,9 @@ def _load_sim_tel_config_and_model(self, config_data):
             "parameters", "array-simulator_parameters.yml"
         )
         _parameters = gen.collect_data_from_yaml_or_dict(_parameter_file, None)
-        self.config = gen.validate_config_data(_rest_config, _parameters)
+        # The solution below with ignore_unidentified might be temporary
+        # TODO - Think if there's a better solution
+        self.config = gen.validate_config_data(_rest_config, _parameters, ignore_unidentified=True)
```
GernotMaier, line 240: Don't see anything bad about this solution.

simtools/simulator.py
```diff
@@ -313,42 +328,55 @@ def _set_simulation_runner(self):
...
+        }
+        corsika_args = {
+            "mongo_db_config": self._mongo_db_config,
+            "site": self.site,
+            "layout_name": self.layout_name,
+            "corsika_parameters_file": self._corsika_parameters_file,
+            "corsika_config_data": self._corsika_config_data,
+        }
+        # TODO: This is not very elegant, find a nicer solution?
+        if self.simulator in ["simtel", "corsika_simtel"]:
```
GernotMaier, line 343: Explain to me what the issue is here.

simtools/corsika_simtel/__init__.py
GernotMaier, line 1: Is this file needed?

GernotMaier (commented):
Very nice and I am actually surprised that everything fit kind of into the current design of the code. Don't have really comments on the code, but more general comments. Wait for your demonstration before approving it.

applications/simulate_prod.py
```diff
@@ -0,0 +1,346 @@
...
+        Simulation configuration file
+        (contains the default setup which can be overwritten by the command line options)
+    prod_tag (str, required)
+        The production tag (ID) to use (e.g., Prod5)
+    site (str, required)
+        Paranal or LaPalma (case insensitive)
+    primary (str, required)
+        Name of the primary particle to simulate. The available options are
+        gamma, gamma_diffuse, electron, proton, muon, helium, nitrogen, silicon, and iron.
+    from_direction (str, required)
```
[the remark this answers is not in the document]
orelgueta: Done.

applications/simulate_prod.py
```diff
@@ -0,0 +1,346 @@
...
+        Should be one of North, South, East, West (case insensitive)
+    zenith_angle (float, required)
+        Zenith angle in degrees
+    nshow (int, optional)
+        Number of showers to simulate
```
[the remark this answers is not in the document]
orelgueta: Done and added an explanation that this number is prior to reuse factors.

applications/simulate_prod.py
```diff
@@ -0,0 +1,346 @@
...
+
+    .. code-block:: console
+
+        INFO::layout_array(l569)::read_telescope_list_file::Reading array elements from ...
+        WARNING::corsika_runner(l127)::_load_corsika_config_data::data_directory not given
+        in corsika_config - default output directory will be set.
+        INFO::layout_array(l569)::read_telescope_list_file::Reading array elements from ...
+        INFO::corsika_config(l493)::_set_output_file_and_directory::Creating directory
+        INFO::simulator(l405)::simulate::Submission command: local
+        INFO::simulator(l410)::simulate::Starting submission for 1 runs
```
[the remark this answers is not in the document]
orelgueta: Added an `if` to the code to avoid this in the future as well.

applications/simulate_prod.py
orelgueta: There are some crazy people out there... 
Fixed.

applications/simulate_prod.py
orelgueta: This is more widespread than this application. We have an entire module which is called `SimtelRunnerArray`. Generally gammasim-tools makes this distinction between telescope and array simulation and think it makes sense to be honest. 
If you feel strongly I can be persuaded to change it everywhere though.

simtools/configuration/commandline_parser.py
orelgueta: Yes, fixed.

simtools/configuration/commandline_parser.py
orelgueta: Yes, fixed.

applications/simulate_prod.py
orelgueta: Yes, fixed.

simtools/configuration/commandline_parser.py
orelgueta: I actually switched to use the default configurator option instead of redefining it in the application. I was a bit hesitant originally cause I was worried about having both `--site north --from_azimuth_direction north` since it might be a bit confusing. However, I guess for the expected users of this application it should be very clear.
I was also not too happy about defining the `--telescope` option unnecessarily which the default configurator does, but it's not a real issue.

simtools/corsika_simtel/corsika_simtel_runner.py
orelgueta: Yes, `prepare_run_script` is better so I changed it everywhere to that.

simtools/simulator.py
orelgueta: Yes it's actually necessary cause otherwise the corsika config complains about unknown parameters provided. I would like to find a better solution in the future but as a first draft it's not too bad.

simtools/util/general.py
orelgueta: Fixed.

simtools/simulator.py
orelgueta: Yeah I guess it's not too bad. Do you think I can remove the TODO then?

simtools/simulator.py
orelgueta: I am not happy about the fact there are two very similar `if`s one after the other:

```
 # TODO: This is not very elegant, find a nicer solution?
        if self.simulator in ["simtel", "corsika_simtel"]:
            simtel_args = {
                "array_model": self.array_model,
                "config_data": {
                    "simtel_data_directory": self.config.data_directory,
                    "primary": self.config.primary,
                    "zenith_angle": self.config.zenith_angle * u.deg,
                    "azimuth_angle": self.config.azimuth_angle * u.deg,
                },
            }
        if self.simulator == "corsika":
            self._set_corsika_runner(common_args | corsika_args)
        if self.simulator == "simtel":
            self._set_simtel_runner(common_args | simtel_args)
        if self.simulator == "corsika_simtel":
            self._set_corsika_simtel_runner(common_args, corsika_args, simtel_args)
```

This is just because the code crashes otherwise in case we set `corsika` as the simulator and `self.config` is `None`. I would have liked to avoid this first if and somehow call `config.data_directory` without crashing. Perhaps if I change it to a dictionary or call it like a dictionary it would work. I would like to try anyway.

simtools/corsika_simtel/__init__.py
orelgueta: If I understand correctly my python, it is no longer 100% necessary for python 3.3 and up but it is recommended to add since it makes this package into a regular package instead of a namespace package (which is what we want in this case). I admit I did not look into it too deeply before creating it though since it is a sort of habit which we always did. 

I actually tried now without this file and the code still works. Should we then remove these files if they are empty from the entire package?

applications/simulate_prod.py
GernotMaier: No, let's leave it.

simtools/simulator.py
GernotMaier: OK!

simtools/simulator.py
GernotMaier: Yes, remove the TODO.

simtools/corsika_simtel/__init__.py
GernotMaier: I think we should remove them. We are way beyond 3.3. and no need to keep unnecessary files.

simtools/simulator.py
GernotMaier: OK, understand. I don't have a better solution to suggest.

GernotMaier (approved):
All good from my side, thanks!

PR #470, opened 2023-06-21
Docstr coverage

GernotMaier:
I do not know why I have started this branch many months ago and I am not sure if we need / want it: this is a precommit hook to check docstring coverage, see https://github.com/HunterMcGushion/docstr_coverage

My feeling is that it is nice, but not absolutely necessary.  I'll let you two decide if we want to integrate it or if we should delete the branch.


orelgueta:
I don't mind either way, if Victor likes it, let's include it.

GernotMaier:
I've updated the version and set the coverage to 70. Let's first try it out and increase it slowly.

VictorBarbosaMartins (approved):
Good to go, thanks!

PR #474, opened 2023-06-22
Merge production script for Rel 0 to master

orelgueta:
This PR is essentially identical to #461, only this time it's to the master. Doesn't really require a review but if @VictorBarbosaMartins wants to take advantage, feel free.


applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    model_version (str, required)
+        The telescope model version to use (e.g., Prod5).
+    site (str, required)
+        North or South (case insensitive).
+    primary (str, required)
+        Name of the primary particle to simulate. The available options are
+        gamma, gamma_diffuse, electron, proton, muon, helium, nitrogen, silicon, and iron.
+    from_azimuth_direction (str, required)
+        Should be one of North, South, East, West (case insensitive).
+    zenith_angle (float, required)
```
VictorBarbosaMartins, line 35: Why dont we have azimuth_angle as well?

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+        Zenith angle in degrees.
+    nshow (int, optional)
+        Number of showers to simulate.
+        The Number of simulated events depends on the number of times a shower is re-used in the
+        telescope simulation. The number provided here is before any reuse factors.
+    start_run (int, required)
```
VictorBarbosaMartins, line 41: start_run_number?

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+        Start run number such that the actual run number will be 'start_run' + 'run'.
+        This is useful in case a new transform is submitted for the same production.
+        It allows the transformation system to keep using sequential run numbers without repetition.
+    run (int, required)
```
VictorBarbosaMartins, line 45: run_number ?

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    )
+    return config.initialize(db_config=True, telescope_model=True)
+
+
+def _proccess_simulation_config_file(config_file, primary_config, logger):
+    """
+    Read the simulation configuration file with all the details
+    on shower and array simulations.
+
+    Attributes
```
VictorBarbosaMartins, line 195: these are not attributes, but parameters, no? We do not list attributes.

The _logger is missing

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+
+    Returns
+    -------
+    str
+        Label of simulation configuration.
+    dict
+        Configuration of shower simulations.
+    dict
+        Configuration of array simulations.
+
```
VictorBarbosaMartins, line 210: please add the Raises statement here in the docstring as well

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+            config_arrays[key] = value
+
+    config_arrays["data_directory"] = config_showers["data_directory"]
+    config_arrays["site"] = config_showers["site"]
+    config_arrays["layout_name"] = config_showers["layout_name"]
+
+    return label, config_showers, config_arrays
+
+
+def _translate_from_azimuth_direction_to_phi(logger, from_azimuth_direction):
```
VictorBarbosaMartins, line 262: since you have to translate it anyways, I would also support the option of passing directly the azimuth angle. I know that we dont need that for e.g. prod6 but it is strange to me not to allow it

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    if from_azimuth_direction == "north":
+        return 0 * u.deg
+    if from_azimuth_direction == "south":
+        return 180 * u.deg
+    if from_azimuth_direction == "east":
+        return 90 * u.deg
+    if from_azimuth_direction == "west":
+        return 270 * u.deg
+
+    logger.error(f"The direction {from_azimuth_direction} to simulate from was not recognised")
```
VictorBarbosaMartins, line 287: recognised x recognized? Do we have a preference for BE or AE?

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    config_arrays["layout_name"] = config_showers["layout_name"]
+
+    return label, config_showers, config_arrays
+
+
+def _translate_from_azimuth_direction_to_phi(logger, from_azimuth_direction):
+    """
+    Translate the direction particles are coming from to an azimuth angle.
+
+    Attributes
```
VictorBarbosaMartins, line 266: Parameters

simtools/configuration/commandline_parser.py
```diff
@@ -197,20 +197,20 @@ def initialize_telescope_model_arguments(self, add_model_version=True, add_teles
...
+        add_model_version: bool
+            Set to allow a simulation model argument.
+        add_telescope: bool
+            Set to allow a telescope name argument.
         """
 
         _job_group = self.add_argument_group("telescope model")
         _job_group.add_argument(
-            "--site", help="CTAO site (e.g. North, South)", type=self.site, required=False
+            "--site", help="CTAO site (e.g., North, South)", type=self.site, required=False
```
VictorBarbosaMartins, line 208: North, South are examples or the only options?

simtools/configuration/commandline_parser.py
```diff
@@ -288,3 +288,32 @@ def efficiency_interval(value):
...
+        Argument parser type to check that the zenith angle provided is in the interval [0, 180].
+        We allow here zenith angles larger than 90 degrees in the improbable case
+        such simulations are requested. It is not guaranteed that the actual simulation software
+        supports such angles!
+
+        Parameters
+        ----------
+        angle: float
+            zenith angle to verify
+
```
VictorBarbosaMartins, line 304: add Returns here

simtools/configuration/commandline_parser.py
```diff
@@ -288,3 +288,32 @@ def efficiency_interval(value):
             raise argparse.ArgumentTypeError(f"{value} outside of allowed [0,1] interval")
 
         return fvalue
+
+    @staticmethod
+    def zenith_angle(angle):
```
VictorBarbosaMartins, line 293: the name does not seem to be for a method since it does not indicate an action. I suggest e.g. "validate_zenith_angle" ?

simtools/corsika/corsika_config.py
```diff
@@ -498,7 +512,7 @@ def _write_seeds(self, file):
         for s in corsika_seeds:
             file.write(f"SEED {s} 0 0\n")
 
-    def get_input_file(self):
+    def get_input_file(self, use_multipipe=False):
         """
         Get the full path of the CORSIKA input file.
```
VictorBarbosaMartins, line 518: Please add the Parameters section to the docstring

simtools/corsika/corsika_runner.py
```diff
@@ -164,12 +172,14 @@ def _load_corsika_data_directories(self):
...
-    def get_run_script(self, **kwargs):
+    def prepare_run_script(self, use_pfp=True, **kwargs):
         """
         Get the full path of the run script file for a given run number.
 
         Parameters
         ----------
+        use_pfp: bool
+            Whether to use the preprocessor in preparing the CORSIKA input file
         run_number: int
```
VictorBarbosaMartins, line 183: I dont see these 2 following parameters in the declaration of the function. are they part of kwargs? maybe one should indicate that in case affirmative

simtools/corsika_simtel/corsika_simtel_runner.py
```diff
@@ -0,0 +1,202 @@
+import stat
+from pathlib import Path
+
+from simtools.corsika.corsika_runner import CorsikaRunner
+from simtools.simtel.simtel_runner_array import SimtelRunnerArray
+
+__all__ = ["CorsikaSimtelRunner"]
```
VictorBarbosaMartins, line 7: I would add here to the list all classes present in the file including the error classes. Otherwise you are not able to import them in the tests later I think. Please add unit tests for these error as well

simtools/corsika_simtel/corsika_simtel_runner.py
```diff
@@ -0,0 +1,202 @@
...
+class CorsikaSimtelRunner(CorsikaRunner, SimtelRunnerArray):
+    """
+    CorsikaSimtelRunner is responsible for running CORSIKA and piping it to sim_telarray
+    using the multipipe functionality. CORSIKA is set up using corsika_autoinputs program
+    provided by the sim_telarray package. It creates the multipipe script and sim_telarray command
+    corresponding to the requested configuration.
+
+    It uses CorsikaConfig to manage the CORSIKA configuration and SimtelRunnerArray
+    for the sim_telarray configuration. User parameters must be given by the
+    common_args, corsika_args and simtel_args arguments.
```
VictorBarbosaMartins, line 23: I dont understand why so many line breaks, since it did not reach the limit and it also did not start a new sentence. Are these really necessary?

simtools/simulator.py
```diff
@@ -208,6 +213,12 @@ def _load_corsika_config_and_model(self, config_data):
             "corsika_parameters_file", None
         )
 
+        # Remove sim_telarray parameters from the CORSIKA config dictionary
+        # TODO - Replace this with a more elegant solution!
```
VictorBarbosaMartins, line 217: I suggest creating issues for the TODOs, otherwise they get lost in the code (and pylint will probably complain)

simtools/simulator.py
```diff
@@ -313,42 +326,55 @@ def _set_simulation_runner(self):
...
+            }
+        if self.simulator == "corsika":
+            self._set_corsika_runner(common_args | corsika_args)
         if self.simulator == "simtel":
-            self._set_simtel_runner()
-        elif self.simulator == "corsika":
-            self._set_corsika_runner()
+            self._set_simtel_runner(common_args | simtel_args)
+        if self.simulator == "corsika_simtel":
+            self._set_corsika_simtel_runner(common_args, corsika_args, simtel_args)
```
VictorBarbosaMartins, line 357: Dont we need a final 'else' here to check for possible invalid `self.simulator`?

simtools/simulator.py
```diff
@@ -313,42 +326,55 @@ def _set_simulation_runner(self):
...
-            config_data={
-                "simtel_data_directory": self.config.data_directory,
-                "primary": self.config.primary,
-                "zenith_angle": self.config.zenith_angle * u.deg,
-                "azimuth_angle": self.config.azimuth_angle * u.deg,
-            },
-        )
+        self._simulation_runner = SimtelRunnerArray(**simulator_args)
+
+    def _set_corsika_simtel_runner(self, common_args, corsika_args, simtel_args):
```
VictorBarbosaMartins, line 372: I know it is a bit annoying, but it would be nice if we document also the internal functions for ourselves (the developers).

VictorBarbosaMartins (changes requested):
I quickly went through the code but did not test it myself nor looked into the tests. I give some suggestions in the comments, mostly organisational. 
Thanks for implementing this!

applications/simulate_prod.py
orelgueta: This is because the direction at the moment cannot be an angle (will explain further below).

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    site (str, required)
+        North or South (case insensitive).
+    primary (str, required)
+        Name of the primary particle to simulate. The available options are
+        gamma, gamma_diffuse, electron, proton, muon, helium, nitrogen, silicon, and iron.
+    from_azimuth_direction (str, required)
+        Should be one of North, South, East, West (case insensitive).
+    zenith_angle (float, required)
+        Zenith angle in degrees.
+    nshow (int, optional)
```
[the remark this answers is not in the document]
orelgueta: The following statement is going to be a repeating theme throughout my replies: I tried to keep the same idea/names/options as the current production script in order to make it easier to interface with workload. If we see that we have time for Rel 0 to make this script better/different from the current one, I will. This is just the first draft of it.

applications/simulate_prod.py
orelgueta: Same statement as above.

applications/simulate_prod.py
orelgueta: Same statement as above.

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    .. code-block:: console
+
+        python applications/simulate_prod.py \
+        --production_config tests/resources/prod_multi_config_test.yml --model_version Prod5 \
+        --site north --primary gamma --from_azimuth_direction north --zenith_angle 20 \
+         --start_run 0 --run 1
+
+    By default the configuration is saved in simtools-output/test-production
+    and the output in corsika-data and simtel-data. The location of the latter directories
+    can be set to a different location via the option --data_directory).
```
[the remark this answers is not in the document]
orelgueta: Unnecessary parentheses, yes.
It has to be different directories for CORSIKA and sim_telarray cause that's how the simulation package (not simtools) does it. The user sets the output directory with `--data_directory` and then two directories are created within. It should be noted that this script at the moment is not meant for a user but for production so the output will anyway be moved around to fit with grid requirements. Later we will make it more friendly to occasional users (if there are any).

applications/simulate_prod.py
orelgueta: You are right, fixed.

applications/simulate_prod.py
orelgueta: Done.

applications/simulate_prod.py
orelgueta: It will be done, but not now (see statement above).

applications/simulate_prod.py
```diff
@@ -0,0 +1,332 @@
...
+    Attributes
+    ----------
+    from_azimuth_direction: str (north, south, east, west)
+        The direction particles are coming from.
+
+    Returns
+    -------
+    float (Astropy.Quantity)
+        The phi angle for CORSIKA configuration.
+
```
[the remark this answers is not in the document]
orelgueta: Done

applications/simulate_prod.py
orelgueta: Generally we write code in AE but comments in BE. However, for `z` vs. `s` it is harder cause one can use BE while still using `z`. I usually use `s`, but not always and there's no way I will go through the code to make this consistent =)

applications/simulate_prod.py
orelgueta: Done (this issue is also in `production.py` then cause this code was taken from there. I will try to fix it there as well.)

simtools/configuration/commandline_parser.py
orelgueta: It can actually be paranal, south, lapalma or north. I don't think we should list all of these options, the error message would be clear if the user makes a mistake.

simtools/configuration/commandline_parser.py
orelgueta: Fixed (also the other cases in this file).

simtools/configuration/commandline_parser.py
orelgueta: That's cause it's not really a method it's a `type`. See where and how it's used and it might become clearer.

simtools/corsika/corsika_config.py
orelgueta: Done

simtools/corsika/corsika_runner.py
orelgueta: Yes, they are in kwargs. As far as I remember we don't have to indicate that arguments are in kwargs instead of being named arguments. However, if you feel strongly or find that we have to, please let me know and I change it.

simtools/corsika_simtel/corsika_simtel_runner.py
orelgueta: I actually realised I don't need that error class, it stayed from the original class I copied from, so I removed it which solves that problem.

Generally we don't have to list stuff in `__all__`, it is there just in case a user imports using `*` which is really discouraged. For the tests it's not an issue.

simtools/corsika_simtel/corsika_simtel_runner.py
orelgueta: Yes it reached the limit so I tried breaking it where it is somewhat easier to read. In `read_the_docs` these are anyway meaningless no?

simtools/corsika_simtel/corsika_simtel_runner.py
```diff
@@ -0,0 +1,202 @@
...
+        Parameters
+        ----------
+        multipipe_file: str or Path
+            The name of the multipipe file which contains all of the multipipe commands.
+        """
+
+        multipipe_executable = Path(self.corsika_config._config_file_path.parent).joinpath(
+            "run_cta_multipipe"
+        )
+        with open(multipipe_executable, "w") as file:
```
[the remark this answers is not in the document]
orelgueta: We create that path before calling this function (in `corsika_config`), so I think we can skip the test here.
(And generally I am more relaxed about testing things which anyway python would complain about if they are wrong and give a fairly clear trace).

simtools/simulator.py
orelgueta: They will definitely not get lost in the code cause it's something I am working on actively. Generally I agree with you, I wouldn't have merged this to master with TODOs, but we decided to do it anyway just this time. Once I reach a state in which I can say the Rel 0 script is done I will remove them if they are still there and open issues.

simtools/simulator.py
orelgueta: No, cause it is checked already in the setter:
https://github.com/gammasim/gammasim-tools/blob/4a1b76426f5fb852e48d453b089c42bb955f265d/simtools/simulator.py#L159

However, this comment made me realise that this is not a proper setter so I will change that later.

simtools/simulator.py
orelgueta: This would be a lot of work to do consistently everywhere. Maybe in the next "maintenance period"?

tests/resources/prod_multi_config_test.yml
```diff
@@ -0,0 +1,49 @@
...
+  array:
+    model_version: Prod5
+    zenith: 20 deg
+    phi: 0 deg
+    default:
+      LST: 'D234'
+      MST: NectarCam-D
+    LST-01: '1'
+
+# These are the default settings for each primary for 20 deg zenith angle, La Palma.
```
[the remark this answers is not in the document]
orelgueta: We haven't decided yet if we should put these CORSIKA parameters in the DB or keep a yaml file and how to organise it. This is the question we discussed a couple of weeks ago in the chat if you remember. The decision we made was to put the majority in the code actually as look up tables. Until I implement it I won't know if it's the best solution though.

orelgueta:
Hi @VictorBarbosaMartins, I fixed most things and replied to others. I also fixed the failing unit tests.

applications/simulate_prod.py
VictorBarbosaMartins: Sorry, I did not find the explanation below. is it the explanation in which you mention that this is planned for afterwards?

applications/simulate_prod.py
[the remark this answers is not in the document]
VictorBarbosaMartins: Ok, if you can keep track of the TODOs without opening issues, good.
However, let me point out that a single dictionary would do the job of translating our parameter names to the ones needed in workload, and in that way we would keep parameter names in simtools consistent. but if you prefer to do this later, ok.

applications/simulate_prod.py
VictorBarbosaMartins: Uh, that is odd, why comments in BE and code in AE? I have asked the question before to propose a discussion but I did not want to propose you to make the entire code consistent. Let's then simply ignore this for the moment

simtools/corsika/corsika_runner.py
VictorBarbosaMartins: what I meant is something similar to the this (visualize.py):
```
def plot_hist_2D(data, **kwargs):
    """
    Produce a two dimensional histogram plot. Any option that can be changed after plotting (e.g.,\
    axes limits, log scale, etc.) should be done using the returned plt instance.

    Parameters
    ----------
    data: numpy structured array
          The columns of the structured array are used as the x-axis and y-axis titles.
    **kwargs:
        * title: set a plot title.
        * Any additional kwargs for plt.hist2d
```
I would propose to do the same here but we don't need to go through the code and fix everywhere

simtools/corsika_simtel/corsika_simtel_runner.py
VictorBarbosaMartins: ok that we dont need.

Your second statement is probably true but there exception for cases in which the function is in the __init__.py file (I am having this problem now in the other PR). nevermind, all good then.

VictorBarbosaMartins (approved):
I left some more comments, but I will no longer hold the PR back. Despite many TODOs opens, I believe you will come back to them soonish. Thank you again for implementing it

applications/simulate_prod.py
orelgueta: Yes, I meant the statement about keeping things similar to the current production (no angles). Sorry, should have made it clearer here.

applications/simulate_prod.py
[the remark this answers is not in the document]
orelgueta: I would actually change the parameters used in workload later (would be better). I just want to do the changes gradually.

applications/simulate_prod.py
orelgueta: It's actually standard that code is in AE (see for example `color` in matplotlib). Not sure why, maybe it's cause usually AE is shorter or cause they contribute more code to the world. Either way, comments/docstrings are then a free choice and as European scientists we usually use BE everywhere.

simtools/corsika_simtel/corsika_simtel_runner.py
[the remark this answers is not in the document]
orelgueta: The comment in the other PR wasn't from me =)

tests/resources/prod_multi_config_test.yml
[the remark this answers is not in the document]
orelgueta: But there is a TODO here, two lines below this one. (And anyway it can't be forgotten, if you try simulating anything which is not La Palma 20 deg you will get an error.

orelgueta:
Thanks for the extra checks. I fixed a few more things and if the tests pass I will merge.

PR #478, opened 2023-06-27
Fix isort pre-commit issue

orelgueta:
The name of the branch is because I opened it to make small changes to installation instructions. However, this PR actually solves a very cryptic error that happens when trying to commit using python 3.9 and the old `isort` we were using (cost half an hour).

The build of the docs still fails, but I am not sure if it should block this PR (I guess it fails in any PR, the changes here should not affect that).


docs/source/conf.py
```diff
@@ -235,7 +235,7 @@ def get_authors_from_citation_file(file_name):
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
+intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
```
GernotMaier, line 238: This really makes a difference?

GernotMaier (approved):
All good - thanks for spending time on this....

docs/source/conf.py
orelgueta: This wasn't done by me but by `black` in the pre-commit (see commit message in [ad1769c](https://github.com/gammasim/simtools/pull/478/commits/ad1769cbb3048164e7e44b0e352b6a49cf6c4057)). This is why I said I think you didn't use the pre-commit hook.

PR #481, opened 2023-06-28
Set tests to run on main

orelgueta:
With the switch to main instead of master, we need to tell the workflows to run on main instead of master.


VictorBarbosaMartins (approved):
Good! thanks for fixing that

PR #484, opened 2023-07-06
Pack output for grid registering

orelgueta:
Also, gzipped log files, added zero padding to file names (see discussion with DPPS about it) and modified tests accordingly.


simtools/applications/simulate_prod.py
```diff
@@ -335,6 +344,34 @@ def main():
...
+        f"{shower_configs['zenith']} at the {args_dict['site']} site, "
+        f"using the {array_configs['model_version']} telescope model."
+    )
+
+    if args_dict["pack_for_grid_register"]:
+        logger.info("Packing the output files for registering on the grid")
+        output_files = simulator.get_list_of_output_files()
+        log_files = simulator.get_list_of_log_files()
+        histogram_files = simulator.get_list_of_histogram_files()
+        tar_file_name = Path(log_files[0]).name.replace("log.gz", "log_hist.tar.gz")
```
GernotMaier, line 359: I guess I do not need to know the logic of this renaming.

simtools/corsika/corsika_runner.py
```diff
@@ -329,12 +331,12 @@ def get_file_name(self, file_type, **kwargs):
             f"_{kwargs['label']}" if "label" in kwargs and kwargs["label"] is not None else ""
         )
         file_name = (
-            f"corsika_run{kwargs['run']}_{kwargs['primary']}_"
+            f"corsika_run{kwargs['run']:06}_{kwargs['primary']}_"
```
GernotMaier, line 334: Do I understand it correctly that this is due to the 6-digit run numbers in corsika and you want to make sure that this is enforced?

GernotMaier (approved):
All good, two questions (no comments) and two messages from codecov about lines not covered by tests.

simtools/applications/simulate_prod.py
orelgueta: This is because we tar together the log and hist files and upload to the grid that one tarball as the log of the job instead of uploading them separately.

simtools/corsika/corsika_runner.py
GernotMaier: OK - but it defines a max range of run numbers, or?

And people should use `sort -n` to solve above issue.

orelgueta: It doesn't limit the run numbers, they can be larger than 999,999, it's just that after that we won't guarantee the consecutive order.

> And people should use sort -n to solve above issue.

Sure, we could ask people to use that and implement similar "tools" in the various places (when dumping file lists from the grid, listing files in ctapipe, etc.). However, it's probably easier to simply pad the numbers =)

Issue #463, opened 2023-06-19
Set output directory for data products from applications.


GernotMaier:
Output path for data products from applications requires knowledge of the full logic of how these directories are created. 

I suggest that when the user gives as a configuration option `--output_path`, all files are written into this directory and no subdirectories are created.

This is the issue:
e.g., when running `print_array_elements.py`, the converted coordinate file is written to currently `./tmp/simtools-output/d-2023-06-18/layout/telescope_positions-corsika.ecsv`, when providing `--output_path ./tmp/` . Including dates in a non-standard format (`d-2023`) does make it even more difficult.

Solution:
Suggest at least that when `output_path` is given, that the files are written directly into this directory. We could also discuss if one could give explicitly the name of the output file(s).

orelgueta:
As mentioned when we talked, I think it's clear that if a user provides an `output_path`, that should be the exact output path without any additions/omissions.

PR #487, opened 2023-07-11
explicit setting of output path

GernotMaier:
This PR addresses issue #463: add a command line flag to explicitly set the `output_path` without adding simtools tool name and date. The name of the command line flag is `--plain_output_path`.

### DIFF
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -106,9 +106,9 @@ class CommandLineParser(argparse.ArgumentParser):
             required=False,
         )
         _job_group.add_argument(
-            "--no_simtools_output_path_naming",
-            help="do not add date and tool name to output path",
-            action="store_false",
+            "--plain_output_path",
+            help="use plain output path (without adding the tool name and dates)",
+            action="store_true",
             required=False,
         )
         _job_group.add_argument(
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -106,7 +106,7 @@ class CommandLineParser(argparse.ArgumentParser):
             required=False,
         )
         _job_group.add_argument(
-            "--plain_output_path",
+            "--_use_plain_output_path",
             help="use plain output path (without adding the tool name and dates)",
             action="store_true",
             required=False,
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -106,6 +106,12 @@ class CommandLineParser(argparse.ArgumentParser):
             required=False,
         )
         _job_group.add_argument(
+            "--no_simtools_output_path_naming",
+            help="do not add date and tool name to output path",
+            action="store_false",
+            required=False,
+        )
+        _job_group.add_argument(
             "--model_path",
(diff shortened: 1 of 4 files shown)


simtools/configuration/commandline_parser.py
```diff
@@ -105,6 +105,12 @@ def initialize_path_arguments(self):
             default="./",
             required=False,
         )
+        _job_group.add_argument(
+            "--plain_output_path",
```
VictorBarbosaMartins, line 109: I was actually expecting that we would replace the other `output_path` by the `plain_output_path` instead of adding a flag. But I am also fine with this. however, I would make clearer in the parameter name that this is a flag, e.g., `output_is_plain`

simtools/io_handler.py
```diff
@@ -32,10 +32,15 @@ def __init__(self):
...
+        self.plain_output_path = False
         self.data_path = None
         self.model_path = None
 
-    def set_paths(self, output_path=None, data_path=None, model_path=None):
+    def set_paths(self,
+                  output_path=None,
+                  data_path=None,
+                  model_path=None,
+                  plain_output_path=False):
```
VictorBarbosaMartins, line 43: Commenting here (out of place): Please add to the Parameters doc string the new parameter

tests/unit_tests/test_io_handler.py
```diff
@@ -25,6 +25,20 @@ def test_get_output_directory(args_dict, io_handler):
         label="test-io-handler", dir_type="model", test=True
     ) == Path(f"{args_dict['output_path']}/output/test-output/test-io-handler/model")
 
+    io_handler.plain_output_path = True
+
+    assert io_handler.get_output_directory(label="test-io-handler") == Path(
```
VictorBarbosaMartins, line 30: we probably also want to get a warning (or info) when we pass both `label` and `plain_output_path` to inform the user that the latter is preferred over the first. and then I would also test the warning with caplog.

VictorBarbosaMartins (changes requested):
Hi Gernot. thanks for implementing this. I made only some minor comments.

simtools/configuration/commandline_parser.py
GernotMaier: `output_path`  without `plain_output_path` is still setting a path, including the tools and dates. I see the issue with the parameter name and renamed it to `use_plain_output_path`.

simtools/io_handler.py
GernotMaier: Done.

tests/unit_tests/test_io_handler.py
GernotMaier: Not sure about this - we generally don't check for every function the consistency of all arguments. In this case, the `use_plain_output_path` is clearly overwriting everything else and I would prefer not to add a info / warning.

VictorBarbosaMartins (approved):
All good. Lets wait for the tests and merge. Thanks!

tests/unit_tests/test_io_handler.py
VictorBarbosaMartins: yes, in that case let's keep it like it is

PR #496, opened 2023-07-19
Add missing boost-histogram to pyproject.toml

GernotMaier:


VictorBarbosaMartins (approved):
Hope that solves the problem with the package versions :) 
sorry for forgetting it

Issue #494, opened 2023-07-18
pytest tests/unit_tests/test_simulator.py::test_submitting fails (locally)


GernotMaier:
See attached log files with all the errors.

This is for running locally in a conda environment without (!) a simtel_array installation.

If the missing simtel is the reason: should unit tests notice this and be skipped or better fail like in this case?

[test_error.log](https://github.com/gammasim/simtools/files/12080277/test_error.log)

VictorBarbosaMartins:
Hi.
I think failing is good, since it is testing the simulator itself (which depends on simtel_array). I would though prefer to treat the error and try to indicate in the error message that the problem is the missing simtel_array installation

GernotMaier:
A lot of our applications do not need simtel_array, and I think we should allow to run pytests without the simulation software (as we want to test simtools).

VictorBarbosaMartins:
agreed. but we test applications through the integration tests. the unit tests should be dedicated tests. in any case, I think you are right. If this specific unit test needs the simtel_array installation to test a minimal function, we can skip the test if the installation is not there. However, it is important to have a dedicated test that fails if the installation is not there, which would probably make sense to have in the tests for the container, as we spoke yesterday.

GernotMaier:
This is a different discussion. What we need is that tests are running without sim_telarray / corsika installation.

orelgueta:
Quick answer (partly so that I don't forget as well), the issue is indeed that sim_telarray isn't available and the current plan is to simply skip this test (and perhaps others) if it doesn't. Fairly easy to implement, just need to make sure that we want to skip the test and not perhaps modify it to test stuff around running sim_telarray.

orelgueta:
Wow, that took a turn...

Forget the earlier explanation, this error has nothing to do with sim_telarray being present or not. I will explain here the full chain of events (even if it's not interesting, I wasted half of the morning on this so I need to write about it).

- The error itself is the result of a bug in the `test_submitting` function where we did not set the simulator to a test run without actually trying to run sim_telarray (this is a unit test, not an integration test). This bug led to the script calling sim_telarray to actually be called.
- We never saw this error on ubuntu because the script which is called when "submitting" has a few lines in it and the last line which reports the exit code is a simple `echo`. Therefore, we always got a successful run exit code (another bug). This will be fixed in a PR now, where each call is checked if it succeeded, not just the last one.
[...]

PR #497, opened 2023-07-19
Improve CI by adding pip installation

GernotMaier:
Improve CI by adding running of unit tests on:
- pip and conda installation
- ubuntu-latest and macos-latest

Do we have a strong reason to add other python version (3.10, 3.11)?

Inspired by https://github.com/cta-observatory/ctapipe/blob/main/.github/workflows/ci.yml

### DIFF
--- a/.github/workflows/CI-integrationtests.yml
+++ b/.github/workflows/CI-integrationtests.yml
@@ -15,17 +15,26 @@ on:
     types: [opened, synchronize]
   push:
     branches:
-    - main
+      - main
 
 jobs:
 
   integrationtests:
-    runs-on: ubuntu-latest
-
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        #        python-version: [3.8, 3.9, 3.10.4]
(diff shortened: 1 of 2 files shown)


GernotMaier:
@orel - could you have a look at https://github.com/gammasim/simtools/actions/runs/5599215251/jobs/10239886414

Is this test actually calling corsika/simtel_array or just trying to access a file in test resources? I think this is the same issue as in #494.

orelgueta:
> @orel - could you have a look at https://github.com/gammasim/simtools/actions/runs/5599215251/jobs/10239886414
> 
> Is this test actually calling corsika/simtel_array or just trying to access a file in test resources? I think this is the same issue as in #494.

This should be solved in #499.

I don't have an objection to add python 3.10 and 3.11. I think we removed them just to shorten the tests (especially with the integration tests taking 15 minutes). Maybe we can try re-introducing them just for the unit tests?

orelgueta (approved):
Looks good!

PR #499, opened 2023-07-20
Fix exit codes, submitting test and shebang

### LINKED ISSUE #494: pytest tests/unit_tests/test_simulator.py::test_submitting fails (locally)

orelgueta:
Very simple PR in the end which fixes #494. 

One thing that worries me is that in this branch both the integration tests and the unit tests ran much quicker (3 times quicker). They are still successful so maybe these are good news, but we should be alert for any hidden issues that might lead to this.

### DIFF
--- a/simtools/corsika/corsika_runner.py
+++ b/simtools/corsika/corsika_runner.py
@@ -218,7 +218,7 @@ class CorsikaRunner:
 
         with open(script_file_path, "w") as file:
             # shebang
-            file.write("#!/usr/bin/bash\n")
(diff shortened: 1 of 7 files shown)


GernotMaier (approved):
All good. Tests in #497 are also working fine with these changes.

PR #503, opened 2023-07-25
Bug fix in determination of reference coordinate system

GernotMaier:
The coordinate transformation code for the array layouts tries to find the `reference` coordinate system, meaning the `from system` for all coordinate transformation.

This did not work correctly and in the test cases always the crs_local (corsika system) was chosen, even if the array centre coordinates where all `np.nan`.

The logic to determine if a system has coordinates corrected and simplified with this PR.

### DIFF
--- a/simtools/layout/telescope_position.py
+++ b/simtools/layout/telescope_position.py
@@ -493,23 +493,19 @@ class TelescopePosition:
         if _crs_from is None:
             return
 
-        try:
-            for _crs_to_name, _crs_to in self.crs.items():
-                if _crs_to_name == _crs_from_name:
-                    continue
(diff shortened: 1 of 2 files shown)


orelgueta (commented):
Changes look good, but the unit test fails at the moment. 
Also, notice that with this change you are still going in order CORSIKA->Mercator->UTM if they all exist (probably that's fine).

GernotMaier:
Note the removal of the `try...except (InvalidCoordSystem, TypeError) as e:` this is not necessary anymore, due to the logic correction (we could end up here only because of the logical error fixed with this PR).

orelgueta (approved):
Looks good now

PR #505, opened 2023-07-27
Azimuth angle in degrees

orelgueta:
Allow the azimuth angle to be given in string or in degrees.
A couple of other fixes were included in this PR because the error messages I got while debugging were wrong.

### DIFF
--- a/simtools/applications/simulate_prod.py
+++ b/simtools/applications/simulate_prod.py
@@ -86,7 +86,6 @@ import tarfile
 from copy import copy
 from pathlib import Path
 
-import astropy.units as u
 from astropy.io.misc import yaml
 
 import simtools.util.general as gen
@@ -140,25 +139,22 @@ def _parse(description=None):
     )
     config.parser.add_argument(
         "--from_azimuth_direction",
-        help="Direction from which the primary reaches the atmosphere",
-        type=str.lower,
+        help=(
+            "Direction from which the primary reaches the atmosphere."
+            "It can be in degrees between 0 and 360 or one of (north, south, east, west)."
+        ),
+        type=CommandLineParser.azimuth_angle,
         required=True,
-        choices=[
-            "north",
-            "south",
-            "east",
-            "west",
-        ],
     )
     config.parser.add_argument(
         "--zenith_angle",
-        help="Zenith angle in degrees",
+        help="Zenith angle in degrees (between 0 and 180).",
         type=CommandLineParser.zenith_angle,
         required=True,
     )
     config.parser.add_argument(
         "--nshow",
-        help="Number of showers to simulate",
+        help="Number of showers to simulate.",
         type=int,
         required=False,
     )
@@ -273,38 +269,6 @@ def _proccess_simulation_config_file(config_file, primary_config, logger):
     return label, config_showers, config_arrays
 
 
-def _translate_from_azimuth_direction_to_phi(logger, from_azimuth_direction):
-    """
-    Translate the direction particles are coming from to an azimuth angle.
-
-    Parameters
-    ----------
-    from_azimuth_direction: str (north, south, east, west)
-        The direction particles are coming from.
-
-    Returns
-    -------
-    float (Astropy.Quantity)
-        The phi angle for CORSIKA configuration.
-
-    Raises
-    ------
-    ValueError
-    """
-
-    if from_azimuth_direction == "north":
-        return 0 * u.deg
-    if from_azimuth_direction == "south":
-        return 180 * u.deg
-    if from_azimuth_direction == "east":
-        return 90 * u.deg
-    if from_azimuth_direction == "west":
-        return 270 * u.deg
-
-    logger.error(f"The direction {from_azimuth_direction} to simulate from was not recognised")
-    raise ValueError
-
-
 def main():
 
     args_dict, db_config = _parse(description=("Run simulations for productions"))
@@ -319,10 +283,9 @@ def main():
     # Overwrite default and optional settings
     shower_configs["run_list"] = args_dict["run"] + args_dict["start_run"]
     array_configs["site"] = shower_configs["site"] = args_dict["site"]
-    array_configs["zenith"] = shower_configs["zenith"] = args_dict["zenith_angle"] * u.deg
-    array_configs["phi"] = shower_configs["phi"] = _translate_from_azimuth_direction_to_phi(
-        logger, args_dict["from_azimuth_direction"]
-    )
+    array_configs["zenith"] = shower_configs["zenith"] = args_dict["zenith_angle"]
+    array_configs["phi"] = shower_configs["phi"] = args_dict["from_azimuth_direction"]
+
     if args_dict["nshow"] is not None:
         shower_configs["nshow"] = args_dict["nshow"]
     if args_dict["label"] is not None:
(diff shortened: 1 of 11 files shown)


simtools/applications/simulate_prod.py
```diff
@@ -140,25 +139,22 @@ def _parse(description=None):
     )
     config.parser.add_argument(
         "--from_azimuth_direction",
-        help="Direction from which the primary reaches the atmosphere",
-        type=str.lower,
+        help=(
+            "Direction from which the primary reaches the atmosphere."
```
GernotMaier, line 143: `reaches the atmosphere` is really confusing me. Would it be different if the Earth would have no atmosphere? (we ignore now other facts like breathing difficulties). I guess this is the direction the particle is traveling / shower axis is pointing towards. This would be clearer.

simtools/applications/simulate_prod.py
```diff
@@ -140,25 +139,22 @@ def _parse(description=None):
...
+            "It can be in degrees between 0 and 360 or one of (north, south, east, west)."
```
GernotMaier, line 144: Is this the right place to mention that North=0 deg (or 180deg) and that it is clockwise / anticlockwise?

simtools/applications/simulate_prod.py
```diff
@@ -140,25 +139,22 @@ def _parse(description=None):
...
-            "north",
-            "south",
-            "east",
-            "west",
-        ],
     )
     config.parser.add_argument(
         "--zenith_angle",
-        help="Zenith angle in degrees",
+        help="Zenith angle in degrees (between 0 and 180).",
```
GernotMaier, line 151: Do we allow showers coming from below? 180 deg?

simtools/configuration/commandline_parser.py
```diff
@@ -335,11 +338,78 @@ def zenith_angle(angle):
...
+
+            return fangle * u.deg
+        except ValueError:
+            logger.debug(
+                "The azimuth angle provided is not a valid numeric value. "
+                "Will check if it is (north, south, east, west) instead"
+            )
+        if isinstance(angle, str):
+            from_azimuth_direction = angle.lower()
+            if from_azimuth_direction == "north":
```
GernotMaier, line 399: Suggest to add a `lower()`, as it should be ok to say "North", or "NORTH".  Some for EWS

simtools/configuration/commandline_parser.py
```diff
@@ -335,11 +338,78 @@ def zenith_angle(angle):
...
+                return 90 * u.deg
+            if from_azimuth_direction == "west":
+                return 270 * u.deg
+            raise argparse.ArgumentTypeError(
+                "The azimuth angle can only be a number or one of "
+                f"(north, south, east, west), not {angle}"
+            )
+        logger.error(
+            f"The azimuth value provided, {angle}, is not a valid number "
+            "nor one of (north, south, east, west), quitting"
```
GernotMaier, line 413: Do we need the `, quitting`? We don't really quit, but raise a `TypeError` (which we might catch when calling this function).

simtools/corsika/corsika_runner.py
```diff
@@ -220,6 +220,10 @@ def prepare_run_script(self, use_pfp=True, **kwargs):
             # shebang
             file.write("#!/usr/bin/env bash\n")
 
+            # Make sure to exit on failed comands and report their error code
```
GernotMaier, line 223: typo, should be `commands`

tests/unit_tests/configuration/test_commandline_parser.py
```diff
@@ -61,3 +63,36 @@ def test_zenith_angle():
         match=r"The provided zenith angle, 190.0, is outside of the allowed \[0, 180\] interval",
     ):
         parser.CommandLineParser.zenith_angle(190)
+    with pytest.raises(ValueError):
+        parser.CommandLineParser.zenith_angle("North")
```
GernotMaier, line 67: See my other comment, I really like to give `North`.

tests/unit_tests/utils/test_general.py
```diff
@@ -300,3 +301,22 @@ def test_save_dict_to_file(tmp_test_directory, caplog):
     with pytest.raises(IOError):
         gen.save_dict_to_file(example_dict, path)
         assert "Failed to write to" in caplog.text
+
+
+def test_get_file_age(tmp_test_directory):
+    # Create a temporary file and wait for 2 seconds before accessing it
```
GernotMaier, line 307: I think we wait for 1 s: `sleep(1)`.

GernotMaier (approved):
Very minor comment - I'll approve now assuming that you fix the typos etc. No need to look into it again. Thanks!

GernotMaier:
(and I think we can ignore the failed unit test - this is the DB issue with multiple connections)

simtools/applications/simulate_prod.py
orelgueta: You made me realise that it's actually wrong, it's the pointing direction, not the `from_azimuth_direction`. We do the conversion to CORSIKA `PHI` later. I suggest to instead change the option to a simple `azimuth_angle` and change the docstring to be `Telescope pointing direction in azimuth`, OK?

simtools/applications/simulate_prod.py
orelgueta: Yes, I will add that as well.

simtools/applications/simulate_prod.py
orelgueta: CORSIKA can do it and MAGIC sometimes does it, so I don't see a reason to limit it here.

simtools/configuration/commandline_parser.py
orelgueta: But there is a `lower()` there just before the comparison (to avoid repeating it 4 times). Am I missing something?

simtools/configuration/commandline_parser.py
orelgueta: Removed.

simtools/corsika/corsika_runner.py
orelgueta: Fixed.

tests/unit_tests/configuration/test_commandline_parser.py
orelgueta: Well here it's irrelevant because for the zenith angle any string should fail

tests/unit_tests/utils/test_general.py
orelgueta: True, forgot to change the docstring.

orelgueta:
> (and I think we can ignore the failed unit test - this is the DB issue with multiple connections)

I agree we can ignore this. 
I just need your OK about the azimuth angle (mostly to make sure I am not wrong about it) and then I will push the fixes and merge. Thanks for the review!

simtools/applications/simulate_prod.py
GernotMaier: OK!

simtools/applications/simulate_prod.py
GernotMaier: OK!

simtools/configuration/commandline_parser.py
GernotMaier: Did not see that, perfect!

tests/unit_tests/configuration/test_commandline_parser.py
GernotMaier: I think I have to agree with that....

orelgueta:
OK, thanks! I will merge once the tests pass.

orelgueta:
Merging despite the failing tests since they are related to the DB and the issue was thought to be solved. It is apparently not, but I will look into it again elsewhere.

PR #513, opened 2023-07-31
Fix test insert file db

### LINKED ISSUE #512: Test of Insert files to DB issue
I need some input on how to deal with the current DB issue since it requires taking a decision. First I will describe the test:
- One of the tests, `test_insert_files_db`, tests the function which inserts a file to the DB. 
- The test inserts a test file to the sandbox DB and checks that it is there.
- pytest then drops the entire "collection" of files in the sandbox DB so that we do not accumulate "junk".

The issue is the following:
- When we test various versions of python in the CI simultaneously, each pytest instance inserts a different test file into the sandbox DB.
- When pytest finishes it drops the **entire** collection of files in the sandbox DB. That means it drops also the files of the other tests for the other python versions currently running.
- Then the second part of the test fails because it can't find the file it inserted.
- We cannot use a random DB for this test because the API user we are using for accessing our MongoDB instance does not have permissions to drop an entire DB (rightly so). 
- Assuming we do not want to collect junk just for this test, we need to find a different solution.

Possible solutions:
- Easiest - don't test inserting files to the DB. Not ideal because the `insert_file_to_db` method is quite central in simtools, but it's also fairly simple so it's not a big deal.
- Second easiest - make an exception if we get a `None` when we try to read the file we inserted. This should solve the issue when we run simultaneous tests in the CI, but will keep the original functionality of the test when running locally as developing (I vote for this solution for now).
- We could try using [mongomock](https://github.com/mongomock/mongomock) to mock the MongoDB instance we have just for this instance. This might be the best solution in the long run, but requires a bit of time to figure it out and implement (and there's no guarantee this would actually work since gridfs is not a basic MongoDB functionality). If we vote for this I suggest we implement it only for this test and not as a general way to test our `db_handler` because it's always better to test the real DB when possible in my opinion.
- Remove the matrix in the CI and test only one python version for now (or forever).

Thoughts @GernotMaier, @VictorBarbosaMartins?

orelgueta:
Another attempt which is a fairly simple solution. See more details in #512. This also fixes #512.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -158,15 +158,6 @@ def db_no_config_file():
     return db
 
 
-@pytest.fixture()
-def db_cleanup_file_sandbox(db_no_config_file):
-    yield
-    # Cleanup
-    logger.info("Dropping the temporary files in the sandbox")
-    db_no_config_file.db_client["sandbox"]["fs.chunks"].drop()
-    db_no_config_file.db_client["sandbox"]["fs.files"].drop()
-
-
 @pytest.fixture
 def telescope_model_lst(db, db_config, io_handler):
     telescope_model_LST = TelescopeModel(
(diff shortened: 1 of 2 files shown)


orelgueta:
We need to keep an eye on the DB in the next few weeks to make sure we are not filling it with random DBs and that the usual sandbox DB isn't filled with stuff, but from local tests it looks fine.

GernotMaier (approved):
All good!

PR #514, opened 2023-08-02
version update pre-commit

GernotMaier:
This PR updates the version of the tools used in pre-commit. Especially flake8 made quite a large version jump since our last update.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,19 +6,19 @@ repos:
       - id: isort
         args: ["--profile", "black", "--filter-files"]
   - repo: https://github.com/psf/black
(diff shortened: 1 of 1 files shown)


orelgueta (approved):
How do we make sure we do not forget to update this every once in a while? Is there a way to always use the latest stable version or update it automatically?

GernotMaier:
I searched a bit around, and the only solution I found is to write a GitHub action and let it run once a week. There is a `pre-commit autoupdate` command, which updates the `.pre-commit-config.yaml`. One would do this and automatically push? 

(but it seems a bit over the top, maybe we try and remember)

orelgueta:
OK, let's just try to remember.

PR #515, opened 2023-08-03
Application to validate schema files

GernotMaier:
Add a small application using [json-schema](https://json-schema.org/) to validate schema files used to define input and parameter definitions to simtools applications.

- requires additional package jsonschema
- application without module; not sure about how to to unit tests in this case
- reference schema is in the [workflow](https://github.com/gammasim/workflows/blob/prototype-DeriveArrayElementCoordinates/schemas/jsonschema.yml) (still in a branch) repository; this file is needed as a schema in general for simtools

### DIFF
--- a/simtools/applications/validate_schema_files.py
+++ b/simtools/applications/validate_schema_files.py
@@ -6,7 +6,7 @@
 
     Command line arguments
     ----------------------
-    file-name (str)
+    file_name (str)
       input file to be validated
     schema (str)
       schema file (jsonschema format) used for validation
@@ -17,7 +17,7 @@
     .. code-block:: console
 
         simtools-validate-schema-file \
-         --file-name tests/resources/MST_mirror_2f_measurements.schema.yml \
+         --file_name tests/resources/MST_mirror_2f_measurements.schema.yml \
          --schema jsonschema.yml
 
 """
--- a/simtools/applications/validate_schema_files.py
+++ b/simtools/applications/validate_schema_files.py
@@ -51,8 +51,8 @@ def _parse(label, description):
     """
 
     config = configurator.Configurator(label=label, description=description)
-    config.parser.add_argument("-f", "--file_name", help="file to be validated", required=True)
-    config.parser.add_argument("-s", "--schema", help="json schema file", required=True)
+    config.parser.add_argument("--file_name", help="file to be validated", required=True)
+    config.parser.add_argument("--schema", help="json schema file", required=True)
     return config.initialize()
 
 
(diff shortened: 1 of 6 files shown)


GernotMaier:
Added also an integration test to run this application. Uses a file called `jsonschema.yml` which is downloaded from the data database test-db.

simtools/applications/validate_schema_files.py
```diff
@@ -0,0 +1,134 @@
...
+        with open(input_file, "r", encoding="utf-8") as file:
+            data = yaml.safe_load(file)
+    except FileNotFoundError:
+        logging.error(f"Input file {input_file} not found")
+        raise
+
+    try:
+        jsonschema.validate(data, schema=schema)
+    except jsonschema.exceptions.ValidationError:
+        logging.error(f"Schema validation failed for {input_file} using {schema}")
```
orelgueta, line 118: What would it print in `schema`? Would it be the entire defined schema as given in the schema file? That might make the log a bit long and confusing. Perhaps it's better to give the name of the schema file in that case?

simtools/applications/validate_schema_files.py
```diff
@@ -0,0 +1,134 @@
...
+        raise
+
+    print(f"Schema validation successful for {input_file}")
```
orelgueta, line 121: You shouldn't use `print` here but rather the logger. In fact, I just realised that you use the `logging` directly instead of getting a logger. That is not how we've done it elsewhere and I think the format of the logging messages would be different if you do it like that and the level set by the user will not be respected (but I am not sure).

orelgueta (commented):
A few minor comments.

orelgueta:
Oh and I also made one small change that you should make sure you agree with.

simtools/applications/validate_schema_files.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns
+    -------
+    config (Configurator)
+        application configuration
+
+    """
+
+    config = configurator.Configurator(label=label, description=description)
+    config.parser.add_argument("-f", "--file_name", help="file to be validated", required=True)
```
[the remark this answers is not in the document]
GernotMaier: Agree, forgot about it. Remove the short options.

simtools/applications/validate_schema_files.py
GernotMaier: Yes, it would print the long schema. This has been proven to be very useful, as it allows a quick comparison to find the issue.

simtools/applications/validate_schema_files.py
GernotMaier: Changed this to logger.info(...)

GernotMaier:
(and I agree with all your changed)

orelgueta (approved):

PR #519, opened 2023-08-03
Use command line tools for integration tests on pip installation

GernotMaier:
Use `simtools-...` to run integration tests for pip installation.

Integration test checks if conda is installed. If not, it uses to `simtools-...` to call integration tests (this should work on the GithHub action environment; but will not work locally in case you have conda installed. In this case, integration tests will call `python simtools/applications/...`).

### DIFF
--- a/tests/integration_tests/test_applications.py
+++ b/tests/integration_tests/test_applications.py
@@ -3,6 +3,7 @@
 
 import logging
 import os
+import subprocess
 from io import StringIO
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_applications.py
```diff
@@ -401,16 +401,34 @@ def prepare_one_file(file_name):
...
+    def get_application_name(app):
+        """
+        Get the application name from the app string.
+        Return name of command line tool in case pip installation
+        is used.
+        """
+
+        # check if conda is installed on this machine
+
+        if shutil.which("conda") is not None:
```
orelgueta, line 424: Is this the best way to check if we are in the "dev" case? Probably it is, just wondering if you considered other options.

orelgueta (approved):

tests/integration_tests/test_applications.py
GernotMaier: As discussed in your office: this is a simple solution for running integration tests on the github runners (with a clean environment). I assume that if we make the move away for the dev environment from conda, it will be straight forward to change this one line?

PR #520, opened 2023-08-03
Automatic version setting

GernotMaier:
Addresses #223 

Automatic version setting using the same approach as in astropy (which is also used in ctapipe).

Note that for the development, this includes the full git hash:
```
simtools-plot-simtel-histograms --version
plot_simtel_histograms 0.4.1.dev4027+g18da645b
```

I did not understand why it says `0.4.1`. The last release is `0.4.0`, it seems it adds a micro version, but I can't see where this happens.

### DIFF
--- a/simtools/version.py
+++ b/simtools/version.py
@@ -1,5 +1,5 @@
-# this is adaped from https://github.com/cta-observatory/ctapipe/blob/main/ctapipe/version.py
-# this is adapted from https://github.com/astropy/astropy/blob/master/astropy/version.py
+# this is adapted from https://github.com/cta-observatory/ctapipe/blob/main/ctapipe/version.py
+# which is adapted from https://github.com/astropy/astropy/blob/master/astropy/version.py
 # see https://github.com/astropy/astropy/pull/10774 for a discussion on why this needed.
 
 try:
--- a/simtools/version.py
+++ b/simtools/version.py
@@ -1,5 +1,17 @@
-# preliminary simtools versioning; should be replaced by
-# cleaner way of automatically retrieve the version. e.g.,
-# https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/version.py
+# this is adaped from https://github.com/cta-observatory/ctapipe/blob/main/ctapipe/version.py
+# this is adapted from https://github.com/astropy/astropy/blob/master/astropy/version.py
+# see https://github.com/astropy/astropy/pull/10774 for a discussion on why this needed.
 
-__version__ = "0.4.0-dev"
+try:
+    try:
+        from ._dev_version import version
+    except ImportError:
+        from ._version import version
+except Exception:
+    import warnings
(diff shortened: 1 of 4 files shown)


orelgueta (approved):
Tiny question and also make sure that you agree with my small textual changes in the last commit.

simtools/version.py
```diff
@@ -1,5 +1,17 @@
...
+# see https://github.com/astropy/astropy/pull/10774 for a discussion on why this needed.
 
-__version__ = "0.4.0-dev"
+try:
+    try:
+        from ._dev_version import version
+    except ImportError:
+        from ._version import version
+except Exception:
+    import warnings
```
orelgueta, line 11: Do we want to switch to using our "normal" logger or is there a specific reason to use `warnings` here?

orelgueta:
hmm, shame. Should we also add a test for this then?

GernotMaier:
Our production environment is actually not a normal pip installation (which would download the package from pypi).

I've followed now the instruction from [this website](https://www.freecodecamp.org/news/how-to-create-and-upload-your-first-python-package-to-pypi/) to upload simtools to test-pypi. I had to tweak the pyproject.toml file (will be in a separate PR) to upload it, see https://test.pypi.org/project/gammasimtools/0.4.0/#description

`pip install -i https://test.pypi.org/simple/ gammasimtools==0.4.0` unfortunately does not work due to

```
Discarding https://test-files.pythonhosted.org/packages/13/04/279627b71a95358b954fcd05ca64ced7ec20aca8bcde852e20ef30529531/astropy-1.0.1.post1.dev12058.tar.gz (from https://test.pypi.org/simple/astropy/): Requested astropy from https://test-files.pythonhosted.org/packages/13/04/279627b71a95358b954fcd05ca64ced7ec20aca8bcde852e20ef30529531/astropy-1.0.1.post1.dev12058.tar.gz (from gammasimtools==0.4.0) has inconsistent version: expected '1.0.1.post1.dev12058', but metadata has '1.0.1.post1.dev0'
  Downloading https://test-files.pythonhosted.org/packages/ff/d2/d5f707a1219894f55f706a6cecf784a92c70ce7ca8b04148f310fa3194ab/astropy-1.0.1.dev12058.tar.gz (7.2 MB)
```

For this PR and the question here (is the version set correctly), I've ignored above, download the tar package from pypip ([files section](https://test.pypi.org/project/gammasimtools/0.4.0/#files) and installed this using `pip install ..tar.gz`. 

I can confirm that the version number is set correctly when running the different `simtools-...`

orelgueta:
OK, good to know it works as intended then.

simtools/version.py
GernotMaier: No good answer here. This is used in a similar way by astropy and ctapipe. I guess both solutions work and I vote to keep using what everyone else does. OK?

Issue #431, opened 2022-11-30
Add pylint as prehook (define pylint.rc)


GernotMaier:
Add a pre-commit hook for pylint.

Suggestions (from PR #430) on settings:
- `W1203` - Use lazy % formatting in logging functions
- `C0103` - Variable name "..." doesn't conform to snake_case naming style
- `E0401` - unable to import '...'
- `R0902` - Too many instance attributes
- `R0913` - Too many arguments
- `R0914` - Too many local variables

(these should be defined in a pylintrc file)

GernotMaier:
Why do you want to ignore `C0130`? is this mostly because of common short variables? If it is that, maybe we can add a list to pylintrc using `good-names` (https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#good-names)

orelgueta:
Yes, we could add a list of good-names. It would be quite a few and some of them would be quite specific, but perhaps it's best to change some variable names in the code and then add a list of `good-names` for the others. I suggest we don't ignore `C0130` in the first instance and I will try to fix it later.

PR #521, opened 2023-08-04
Pylint updates

GernotMaier:
(a bit lower priority - something on the side)

This is a follow up of PR #430 and of issue #431:

- add a [.pylintrc](.pylintrc) file
- correct some of the issues reported by pylint.

The pylintrc files includes:
- exceptions as discussed in #430 and #431
- list of good names (short variable names; should discuss if can shorted this
- case settings for module/variable/method naming
- no docstring required for main() function

There are several discussion points, see below.

TODO:

- [ ] add prehook (pylint is slow - check if this is reasonable or if we should add it to the CI)
- [x] mv .pylintrc to pyproject.toml


GernotMaier:
I am attaching all pylint messages I get with
```
find . -name "[a-z]*.py" -exec pylint {} \; > tt.txt
```

Discussion points:
-  C0114: Missing module docstring (missing-module-docstring): most of our modules don't have doc strings, but the methods / classes are all well documented. Ignore?
- tests code is not as 'clean' and this is probably ok. Do we want  to exclude this from pylinting?
- R1734: Consider using [] instead of list() (use-list-literal) - simply fix this? (many responses; can be fixed using sed)
- R1735: Consider using '{}' instead of a call to 'dict'. (use-dict-literal) - simply fix this? (many responses; can be fixed using sed)
- W1514: Using open without explicitly specifying an encoding (unspecified-encoding): should we us `UTF-8` in general for encoding when opening ascii files?

The rating is all over the place (from 0. to 10). We probably want to get in the >8 area.

GernotMaier:
[pylint.txt](https://github.com/gammasim/simtools/files/12258322/pylint.txt)

simtools/layout/telescope_position.py
```diff
@@ -293,7 +293,7 @@ def _convert(self, crs_from, crs_to, xx, yy):
             raise
         if xx is None or yy is None:
             return np.nan, np.nan
-        _to_x, _to_y = transformer.transform(xx=xx, yy=yy)
+        _to_x, _to_y = transformer.transform(xx, yy)
```
orelgueta, line 296: This one I don't actually agree with.

orelgueta:
> I am attaching all pylint messages I get with
> 
> ```
> find . -name "[a-z]*.py" -exec pylint {} \; > tt.txt
> ```
I didn't go over all of the messages there obviously. Let's first agree on a few points below. However, it also looks like you didn't use all of the `disable` options mentioned in #430. The might have been intentional though.

 
> * C0114: Missing module docstring (missing-module-docstring): most of our modules don't have doc strings, but the methods / classes are all well documented. Ignore?

Yes, I think so.


> * tests code is not as 'clean' and this is probably ok. Do we want  to exclude this from pylinting?

For now I would vote to ignore, but suggest to open an issue and fix also the tests at some point cause it's not too much work and there's no reason not to have it as "nice".

> * R1734: Consider using [] instead of list() (use-list-literal) - simply fix this? (many responses; can be fixed using sed)
> * R1735: Consider using '{}' instead of a call to 'dict'. (use-dict-literal) - simply fix this? (many responses; can be fixed using sed)

Yes, let's fix these (I have been doing it slowly wherever I see it in the meantime)

> * W1514: Using open without explicitly specifying an encoding (unspecified-encoding): should we us `UTF-8` in general for encoding when opening ascii files?

Yes, that would probably work. 

> The rating is all over the place (from 0. to 10). We probably want to get in the >8 area.

True.

GernotMaier:
Several updates:

- solved the easy issues (e.g., used sed to replace dict, list, etc)
- moved .pylintrc into pylint.toml (used `pylint --generate-toml-config`, which read the existing .pylintrc and generated the toml outout)
- disabled all options discussed in #430 

Attached the most recent messages.

[pylint.txt](https://github.com/gammasim/simtools/files/12273412/pylint.txt)

simtools/layout/telescope_position.py
GernotMaier: You are right, my mistake. Misunderstood the pylint messages.

GernotMaier:
I suggest additionally to add the following option: `disable=unpacking-non-sequence`.

The reason is:
```
simtools/layout/telescope_position.py:296:8: E0633: Attempting to unpack a non-sequence (unpacking-non-sequence)
```
which appears to be related to a pylint issue not being able to handle @overload (see [here](https://stackoverflow.com/questions/74323399/attempting-to-unpack-a-non-sequencepylintunpacking-non-sequence-with-pyproj-tr) and [here](https://github.com/pylint-dev/astroid/issues/1015))

simtools/layout/telescope_position.py
```diff
@@ -293,7 +293,10 @@ def _convert(self, crs_from, crs_to, xx, yy):
             raise
         if xx is None or yy is None:
             return np.nan, np.nan
-        _to_x, _to_y = transformer.transform(xx=xx, yy=yy)
+        (
```
orelgueta, line 296: Is this really what black decided to do with this line? I recommend changing it to 
```
 _to_x, _to_y, = transformer.transform(
     xx=xx,
     yy=yy
)
```

which I assume is what black would do in normal circumstances.

GernotMaier: Really, for such a short line? I have black as pre-commit hook installed and it didn't change anything in this line. I changed it now to a single line.

orelgueta: At some point there was a change here, otherwise it wouldn't have popped up in the review. The original change was very weird so I suggested something more traditional. However, if it fits in one line, that's obviously better.

GernotMaier:
We should discuss how we continue with this PR - do we first want to solve issues or do we agree on the pylint settings, merge this PR and continuously improve the code?

GernotMaier:
Also want to note that I've added to the pylint settings:

```
# see https://github.com/pylint-dev/pylint/issues/2289
generated-members = ["gist_heat_r"]
```

Somehow pylint cannot handle generated items.

GernotMaier:
I would suggest to review and merge this PR and slowly work on the remaining items. Not a high priority!

orelgueta (approved):
All the dict/list changes to {}/[] I am sure I made in the previous pylint round but somehow it never propagated to the code. Never mind, as long as it is done now.

GernotMaier:
Thanks for looking it it. I will merge when all tests are completed.

Note that we didn't add the pre-hook - let's do this at a later stage after resolving some of the issues.

Issue #502, opened 2023-07-21
Change from util to utils


orelgueta:
There's a mismatch between the directory name in the tests and the actual code

**Util**
[https://github.com/gammasim/simtools/tree/fix_logger_general/simtools/util](https://github.com/gammasim/simtools/tree/fix_logger_general/simtools/util)

**Utils**
[https://github.com/gammasim/simtools/tree/fix_logger_general/tests/unit_tests/utils](https://github.com/gammasim/simtools/tree/fix_logger_general/tests/unit_tests/utils)

From a grammatical point of view and other examples, `utils` would be better. Therefore, the suggestion is to change everywhere to `utils` (would require changing imports mostly).

PR #529, opened 2023-08-11
Change from util to utils

GernotMaier:
This PR fixes issue #502 and changes the `util` directory to `utils`.


GernotMaier:
One of you is enough to review - whoever wants to do it. I've used sed, and it is very boring.

Note that there had been a couple of format changes (removed lines), introduced by the pre-commit hooks. Not sure why they have not been applied before (but we have recently updated the pre-commit hooks, maybe therefore)

Issue #516, opened 2023-08-03
Repository organisation


GernotMaier:
This issue is to discuss the organisation of the gammasim repositories and includes a suggestion to merge different repositories.

As a reminder, the simulation pipeline consists of five main components:
- simulation software (like CORSIKA, sim_telarray); not considered here
- simtools - the python library with applications to setup productions, derive parameters, etc
- workflows - binding several simtools together (e.g. for model parameters the assert - derive - validate - review workflow; [early example](https://github.com/gammasim/workflows/blob/prototype-DeriveArrayElementCoordinates/workflows/DeriveArrayElementCoordinates.cwl))
- schemas - describing input data for the simulation pipelines  (e.g., [this one](https://github.com/gammasim/workflows/blob/prototype-DeriveArrayElementCoordinates/schemas/array_coordinates.schema.yml))
- docker files for production and development (see [here](https://github.com/gammasim/containers))

Having these items in three different repositories appeared kind of logical when we started. It was partly motivated to experiment with technologies new to us (e.g., docker, or now the CWL stuff).

It however generates a couple of issues:

- containers generation is not triggered by a simtools pull request or releases. They are triggered by hand: this is not nice; especially when the most recent simtools is required in containers for development.
- unit and integration tests need files across the repository (e.g., to test the simtool `validate_schema_files.py`, we need the [json schema](https://github.com/gammasim/workflows/blob/prototype-DeriveArrayElementCoordinates/schemas/jsonschema.yml); additional some files in [tests/resources](tests/resources) are duplication of files from other repositories.

My suggest therefore is to merge docker and schemas into simtools. We wait with the CWL workflows until we are convinced that this is the solution to go ahead and we make the step beyond playing around with cwl.

Suggestion:
- add a `docker` directory to simtools with the dev/prod/simtelarray/tools directory as it is now in the [containers repository](https://github.com/gammasim/containers)
- add a `schemas` directory to simtools and move the [schemas](https://github.com/gammasim/workflows/tree/prototype-DeriveArrayElementCoordinates/schemas) from the workflow repository. Note that this will be one file per model parameters (roughly 230 files) and maybe another 20-50 for the description of the input data

In case we go ahead with workflows, these would move into simtools/workflows. These will be 20-40 different workflow definitions.

This is for everyone to comment on. @orelgueta @VictorBarbosaMartins

GernotMaier:
[...]

PR #532, opened 2023-08-11
Move docker files from container repository to simtools repository

GernotMaier:
This PR is a follow up on the discussion in issue #516 and moves Dockerfiles from the [gammasim/containers](https://github.com/gammasim/containers) repository in a [./docker](./docker) directory.

Changes with respect to the implementation in containers:
- prod and dev containers are using as base images now the simtelarray image (so no need to recompile CORSIKA/sim_telarray all the time); this significantly reduces as well the building of these images.
- ...

TODO:
- [ ] add CI files

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -2,15 +2,64 @@
 
 Docker files for [simtools](https://github.com/gammasim/simtools) development and applications.
 
-[Docker](https://www.docker.com/community-edition#/download) or any compatible software are required to run these images.
+[Docker](https://www.docker.com/community-edition#/download) or any compatible software are required to run or build these images.
 
 Types of dockerfiles and containers available:
 
 - simtools users: [Dockerfile-prod](./Dockerfile-prod) provides a container with all software installed (CORSIKA, sim\_telarray, simtools python environment, simtools). Pull latest release with: `docker pull ghcr.io/gammasim/simtools-prod:latest`
-- simtools development: [Dockerfile-dev](./Dockerfile-dev) provides a container with CORSIKA, sim\_telarray, and simtools conda environment installed. Pull latest release with: `docker pull ghcr.io/gammasim/simtools-dev:latest`
-- sim\_telarray: [Dockerfile-simtelarray](./Dockerfile-simtelarray) provides a container with the CORSIKA and sim\_telarray installed. Pull latest release with: `docker pull ghcr.io/gammasim/simtools-simtelarray:latest`
+- simtools developers: [Dockerfile-dev](./Dockerfile-dev) provides a container with CORSIKA, sim\_telarray, and simtools conda environment installed. Pull latest release with: `docker pull ghcr.io/gammasim/simtools-dev:latest`
+- sim\_telarray: [Dockerfile-simtelarray](./Dockerfile-simtelarray) provides a container with the CORSIKA and sim\_telarray installed. This provides the base image for the previously listed containers. Pull latest release with: `docker pull ghcr.io/gammasim/simtools-simtelarray:latest`
 
-The CORSIKA / sim\_telarray packages can be downloaded from MPIK (authentication required).
+## Container for simtools users
+
+## Container for simtools developers
+
+Provide a container for testing and development of simtools. This container is not optimised for size, but for completeness of development tools.
+
+Container includes installation of:
+
+- corsika and sim\_telarray
+- packages required by simtools (from pyproject.toml)
+
+The container does not include the simtools code, which should be cloned in the ./external directory (see below).
+
+There are two options on how to use this container:
+
+1. Download from [simtools container repository](https://github.com/gammasim/containers/pkgs/container/simtools-dev) **TODO**
+2. Build a new container from the available Dockerfile (requires access to sim\_telarray package)
+
+### Run a container using the prepared Docker image available from repository
+
+Packages are available from the [simtools container repository](https://github.com/gammasim/containers/pkgs/container/simtools-dev)**TODO**
+
+**Preparation:**
+
+Create a new directory for your development and clone simtools into a subdirectory:
+
+```bash
+mkdir -p simtools-dev && simtools-dev
+git clone git@github.com:gammasim/simtools.git
+```
+
+To download and run a prepared container in bash:
+
+```bash
+docker run --rm -it -v "$(pwd)/:/workdir/external" ghcr.io/gammasim/simtools-dev:latest bash -c "$(cat ./simtools/docker/entrypoint.sh) && bash"
+```
+
+This additionally executes the `entrypoint.sh` script (e.g., for pip install or set the database environment).
+
+Remember you need to `docker login` to the GitHub package repository with a personal token in order to download an image (follow [these instructions](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)).
+
+### Build a new developers container locally
+
+To build a new container locally run:
+
+```bash
+docker build -f Dockerfile-dev -t simtools-dev .
+```
+
+Use the docker container in the same way as above, replacing `ghcr.io/gammasim/simtools-dev:latest` by `simtools-dev`.
 
 ## Container for simulation software CORSIKA / sim\_telarray
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,55 @@
+# Docker files for simtools
+
+Docker files for [simtools](https://github.com/gammasim/simtools) development and applications.
+
+[Docker](https://www.docker.com/community-edition#/download) or any compatible software are required to run these images.
+
+Types of dockerfiles and containers available:
+
+- simtools users: [Dockerfile-prod](./Dockerfile-prod) provides a container with all software installed (CORSIKA, sim\_telarray, simtools python environment, simtools). Pull latest release with: `docker pull ghcr.io/gammasim/simtools-prod:latest`
+- simtools development: [Dockerfile-dev](./Dockerfile-dev) provides a container with CORSIKA, sim\_telarray, and simtools conda environment installed. Pull latest release with: `docker pull ghcr.io/gammasim/simtools-dev:latest`
+- sim\_telarray: [Dockerfile-simtelarray](./Dockerfile-simtelarray) provides a container with the CORSIKA and sim\_telarray installed. Pull latest release with: `docker pull ghcr.io/gammasim/simtools-simtelarray:latest`
+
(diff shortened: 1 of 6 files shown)


docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+```bash
+mkdir -p simtools-dev && simtools-dev
+git clone git@github.com:gammasim/simtools.git
+```
+
+To download and run a prepared container in bash:
+
+```bash
+docker run --rm -it -v "$(pwd)/:/workdir/external" ghcr.io/gammasim/simtools-dev:latest bash -c "$(cat ./simtools/docker/entrypoint.sh) && bash"
+```
```
VictorBarbosaMartins, line 86: General comment:

when running the docker we still get messages regarding gammasim-tools and not simtools. I just want to make sure that this is the way we want?

```
Successfully built gammasimtools
Installing collected packages: gammasimtools
Successfully installed gammasimtools-0.4.1.dev4154+g1f06507a
```

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+This additionally executes the `entrypoint.sh` script (e.g., for pip install or to set the database environment).
```
VictorBarbosaMartins, line 88: I would like to add the information here that at this point we already need a working `set_DB_environ.sh` inside the simtools directory.

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Build a new developers container locally
+
+To build a new container locally run:
+
+```bash
+docker build -f Dockerfile-dev -t simtools-dev .
+```
+
+Use the docker container in the same way as above, replacing `ghcr.io/gammasim/simtools-dev:latest` by `simtools-dev`.
```
VictorBarbosaMartins, line 100: not a big issue but it is better to quote the entire command instead. As I commented above, please include the path to the correct repository before this command, i.e., since we are at this point inside the `simtools/docker` repository:

```bash
cd ../..
docker run --rm -it -v "$(pwd)/:/workdir/external" simtools-dev bash -c "$(cat ./simtools/docker/entrypoint.sh) && bash"

File: docker/README.md [hunk 4]
```diff
@@ -0,0 +1,137 @@
...
+This additionally executes the `entrypoint.sh` script (e.g., for pip install or to set the database environment).
+
+Remember you need to `docker login` to the GitHub package repository with a personal token in order to download an image (follow [these instructions](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)).
+
+### Build a new developers container locally
+
+To build a new container locally run:
+
+```bash
+docker build -f Dockerfile-dev -t simtools-dev .
```
VictorBarbosaMartins, line 97: This commands assumes that you are running it from the `simtools/docker` repository but the command below assumes you are running from the `simtools-dev` repository. Please include before this command `cd simtools/docker`.

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+```bash
+docker run --rm -it -v "$(pwd)/external:/workdir/external" ghcr.io/gammasim/simtools-simtelarray:latest bash
+```
+
+## Build a new container locally
+
+To build a new container locally run:
+
+```bash
+docker build -f Dockerfile-simtelarray  -t sim_telarray .
```
VictorBarbosaMartins, line 121: Please include also here explicitly the path before this command: `cd simtools/docker`

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+To download and run a prepared container in bash:
+
+```bash
+docker run --rm -it -v "$(pwd)/external:/workdir/external" ghcr.io/gammasim/simtools-simtelarray:latest bash
+```
+
+## Build a new container locally
+
+To build a new container locally run:
```
VictorBarbosaMartins, line 118: Please include already here a note (and not after the build command) that at this point we strictly need `corsika7.7_simtelarray.tar.gz` inside the `simtools/docker` directory for the following commands to work.

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+Download the tar package from the MPIK website (password applies) with
+
+```bash
+./download_simulationsoftware.sh
+```
+
+Run the newly build container:
+
+```bash
+docker run --rm -it -v "$(pwd)/external:/workdir/external" simtelarray bash
```
VictorBarbosaMartins, line 134: This only worked for me when using the docker from remote, because you called the locally mounted docker image `sim_telarray` and not `simtelarray`. Please replace `sim_telarray` by `simtelarray` in l121 so that the name matches.

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+- corsika and sim\_telarray
+- packages required by simtools (from pyproject.toml)
+- simtools (main branch)
+
+### Run a simtools-prod container
+
+To run the container in bash
+
+```bash
+docker run ghcr.io/gammasim/simtools-prod:latest bash
```
VictorBarbosaMartins, line 26: I am testing this with `docker run ghcr.io/gammasim/simtools-prod:pr-532 bash`. I was expecting, this command would bring me to inside the container, but it didn't. Am I missing something?

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+In the container, simtools applications are installed and can be called directly (e.g., `simtools-print-array-elements -h`).
+
```
VictorBarbosaMartins, line 30: It would be interesting to provide a list of all commands (applications) available, or point to the file where we can find the possible commands?

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+In case file exchange with the local file system is required, use the docker syntax to mount a directory. Example:
```
VictorBarbosaMartins, line 31: I don't understand what you mean with this sentence. Could you please rephrase it? What do you mean by "exchange with the local file system"?

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+Output files can be found `./simtools-output/`.
+
+### Building a simtools-prod container
+
+To build a new container locally run:
+
+```bash
+docker build -f Dockerfile-prod  -t simtools-prod .
```
VictorBarbosaMartins, line 55: Please include here the path to the appropriate directory before this command:
`cd simtools/docker`

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+Building will take a while and the image is large (~1.4 GB). For using images build on your own, replace in all examples `ghcr.io/gammasim/simtools-prod:latest` by the local image name `simtools-prod`.
```
VictorBarbosaMartins, line 58: Same comment as for the dev container. Please add the full command here instead, and add `cd ../..` before that such that we are in the correct directory

VictorBarbosaMartins (changes requested):
Hi @GernotMaier.

I tested every container (prod and dev && remove and local) and provided some minor comments below. Most of the comments are suggestions for improving the clarity of the step by step (which is important for non-experts as myself). Some comments are just due to my ignorance. Furthermore, as @orelgueta  already looked through the code, I have not reviewed it now line by line. Most importantly, the containers seem to work fine for me.

By the way, it is probably better if you start reading the comments to the dev container part, where I started.

Thanks for this working implementation!

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+### Run a container using the prepared Docker image available from repository
+
+Packages are available from the [simtools container repository](https://github.com/orgs/gammasim/packages?repo_name=simtools).
+
+**Preparation:**
+
+Create a new directory for your development and clone simtools into a subdirectory:
+
+```bash
+mkdir -p simtools-dev && simtools-dev
```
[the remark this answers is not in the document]
GernotMaier: Yes, fixed.

docker/README.md
```diff
@@ -0,0 +1,137 @@
...
+git clone git@github.com:gammasim/simtools.git
+```
+
+To download and run a prepared container in bash:
+
+```bash
+docker run --rm -it -v "$(pwd)/:/workdir/external" ghcr.io/gammasim/simtools-dev:latest bash -c "$(cat ./simtools/docker/entrypoint.sh) && bash"
```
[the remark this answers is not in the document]
GernotMaier: It is not `/:/`, the directory "(pwd)/" is into the container directory `/workdir/externeal`

docker/README.md
GernotMaier: But this would be a repetition? I think the instructions are clear and repeating means that if this changes in future, we have one more line to correct (or miss it). I would prefer to leave it as it is.

docker/README.md
GernotMaier: OK - my assumption was that when you read this readme, then you are in this directory. Anyway, added it.

docker/README.md
GernotMaier: Added

docker/README.md
GernotMaier: I've corrected this.

docker/README.md
GernotMaier: This exists: https://gammasim.github.io/simtools/applications.html
(and we have to put some work into the documentation at some point)

docker/README.md
GernotMaier: Exactly what is written: if you want to exchange files (move a file from local file system to the container file system or the other way around). The docker is a closer box otherwise!

docker/README.md
GernotMaier: I changed the text to avoid to many confusing `cd`s (the cd command don't work if you start reading this readme not from the top).

GernotMaier:
Thanks for the careful review. I've updated the readme, let me know if this covers your comments.

docker/README.md
VictorBarbosaMartins: Ok, changing in only one place in the future is a good argument.

docker/README.md
VictorBarbosaMartins: yes, it is clear. my comment was more on the position in the text where the note appears. in my opinion it should come before running the command, otherwise one can run the command and only later realize that the file was needed.

docker/README.md
VictorBarbosaMartins: Thank you, I agree with the changes.

VictorBarbosaMartins (approved):
Thank you for the corrections! let's merge this (unless @orelgueta has something else to say?). Would you then also update the containers repository and archive it as we discussed yesterday?

GernotMaier:
Thank you for the review!

(and Orel confirmed that this is fine)

Issue #530, opened 2023-08-11
Sphinx extension "sphinx.ext.ifconfig" not used.


GernotMaier:
Any objections to remove "sphinx.ext.ifconfig" from [docs/source/conf.py](docs/source/conf.py)?

We are not using it (and at least my local installation does not have it included.)

GernotMaier:
This package allows you to put 'if' statements into the documentation files, e.g.

```
.. ifconfig:: platform == 'linux'

   This content is for Linux platform only.

.. ifconfig:: platform == 'windows'

   This content is for Windows platform only.
```
(chatgtp example)

We are not using it.

PR #533, opened 2023-08-11
Remove "sphinx.ext.ifconfig" from docs/source/conf.py

GernotMaier:
remove "sphinx.ext.ifconfig" from [docs/source/conf.py](https://github.com/gammasim/simtools/issues/docs/source/conf.py), see #530


orelgueta (approved):
I would have preferred deleting the line rather than commenting it out, but it's not important. Approving.

GernotMaier:
As you wish - I've removed the line.

PR #535, opened 2023-08-11
Check for github actions versions

GernotMaier:
Let's see if we can keep actions update to date with https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/.dependabot.yml
+++ b/.github/.dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
```
orelgueta, line 9: I assume it should say daily (or the command change to weekly)

orelgueta (changes requested):

.github/dependabot.yml
GernotMaier: It should be weekly, just corrected it (I thought for a quick test I try it with daily, but this seems to work only in the default branch)

PR #538, opened 2023-08-15
Build on merge / push to main

GernotMaier:
New simtools container (simtools-dev and simtools-prod) should be build when merging a pull request to main.

Packages are then available from here: https://github.com/orgs/gammasim/packages?repo_name=simtools


VictorBarbosaMartins (approved):
all good!

PR #555, opened 2023-08-18
simplify and fix building of corsika / simtel image

GernotMaier:
This PR fixes the corsika / simtelarray container generation (also simplifies it, as it is now a separate step).

Although strictly not necessary, this workflow is executed for each merge to main (but not for each pull request).  Main reason is to have it appear in the `package` on the right column of the simtools main repository page (arghh). 

Note to myself: when moving container files from other repositories, make sure that the package right are set correctly and that this one has write permissions.

Tests are successful as new packages appear here: https://github.com/gammasim/containers/pkgs/container/simtools-simtelarray and the action here: https://github.com/gammasim/simtools/actions/runs/5903392923/job/16061510040

### DIFF
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -3,6 +3,15 @@ name: docker-corsika-simtelarray
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+  release:
+    types: [published]
+  schedule:
+    - cron: '0 0 * * 0'  # Every Sunday at 00:00 UTC
 
 env:
   REGISTRY: ghcr.io
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
 
-  build-simtools-container:
+  corsika-simtelarray-container:
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -10,23 +10,7 @@ env:
 
 jobs:
 
-  corsikasimtelpackage:
-    runs-on: ubuntu-latest
-    steps:
-      - name: download corsikasimtelpackage
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docker-corsika-simtelarray-image.yml
```diff
@@ -36,15 +29,14 @@ jobs:
...
-          name: corsika7.7_simtelarray.tar.gz
-          path: corsika7.7_simtelarray.tar.gz
-
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: download corsikasimtelpackage
+        run: |
+          wget --no-verbose https://syncandshare.desy.de/index.php/s/${{ secrets.CLOUD_SIMTEL }}/download
+          mv download corsika7.7_simtelarray.tar.gz
```
VictorBarbosaMartins, line 38: not optimal to hardcode the version in the code. Couldn't we get the version already in the name of the downloaded file (question regardless of the discussion in #566)?

VictorBarbosaMartins (approved):
Hi @GernotMaier. I reviewed here and left two comments. I tested building and running the docker container and it worked locally for me as well, therefore I approve. (For the record, the review took me 22 min)

.github/workflows/build-docker-corsika-simtelarray-image.yml
```diff
@@ -3,30 +3,23 @@ name: docker-corsika-simtelarray
...
-      - name: Publish Artifact
-        uses: actions/upload-artifact@v3
-        with:
-          name: corsika7.7_simtelarray.tar.gz
-          path: corsika7.7_simtelarray.tar.gz
-          retention-days: 1
-
-  build-corsikasimtel-container:
-    needs: corsikasimtelpackage
+  corsika-simtelarray-container:
```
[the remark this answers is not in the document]
GernotMaier: Not sure, `corsika-simtelarray-container` is building and pushing the container and `corsikasimtelpackage` is compiling locally corsika and simtel_array. So these are different steps and probably good to have different names.

.github/workflows/build-docker-corsika-simtelarray-image.yml
GernotMaier: Yes and no. I agree that hardcoded version numbers might create issues. But in this case it is a strong reminder on what we are doing included the problems discussed in #566. So I would prefer to leave it there until we have found a more flexible solution in obtaining corsika and sim_telarray.

orelgueta: I would like to add as well that the version `7.7` mentioned here is not expected to change essentially "ever" (until we switch to CORSIKA 8 and then this would be the least of our problems).

Issue #553, opened 2023-08-18
layout_array prints too many debug messages


orelgueta:
Message are like the ones following:
```
DEBUG    simtools.layout.layout_array:layout_array.py:909 UTM system: EPSG:32628
DEBUG    simtools.layout.layout_array:layout_array.py:457 Value -70.98999786376953 m is already of type astropy.units.Quantity.
DEBUG    simtools.layout.layout_array:layout_array.py:459 Quantity -70.98999786376953 m has already unit m. Returning -70.98999786376953 m
DEBUG    simtools.layout.layout_array:layout_array.py:457 Value -52.08000183105469 m is already of type astropy.units.Quantity.
DEBUG    simtools.layout.layout_array:layout_array.py:459 Quantity -52.08000183105469 m has already unit m. Returning -52.08000183105469 m
DEBUG    simtools.layout.layout_array:layout_array.py:493 pos_x and pos_y are given. Setting coordinates.
DEBUG    simtools.layout.layout_array:layout_array.py:496 utm_east and utm_north are not given. Coordinates not set.
DEBUG    simtools.layout.layout_array:layout_array.py:496 mercator and lon are not given. Coordinates not set.
DEBUG    simtools.layout.layout_array:layout_array.py:457 Value 43.0 m is already of type astropy.units.Quantity.
```

It becomes hard to debug with so many messages. Can we reduce these?

PR #556, opened 2023-08-21
Reduce debug warnings from layout_array

GernotMaier:
This issue addresses #553 and reduces the amount of scrolling for @orelgueta .


orelgueta (approved):
Removing these messages is the only solution? They are really not necessary?

GernotMaier:
I really don't think they are necessary. I looked through them and they are all more 'development print statements', and not really debug statements.

PR #557, opened 2023-08-21
Important: remove temporary link to json schema

GernotMaier:
Remote temporary link to json schema in workflows repository.

Note that integration tests in all branches are failing right now. This issue resolves this.

### DIFF
--- a/tests/integration_tests/test_applications.py
+++ b/tests/integration_tests/test_applications.py
@@ -655,7 +655,9 @@ def test_applications(application, io_handler, monkeypatch, db):
 
     prepare_one_file("PSFcurve_data_v2.txt")
     prepare_one_file("MLTdata-preproduction.ecsv")
-    download_file("https://github.com/gammasim/workflows/blob/main/schemas/jsonschema.yml")
+    download_file(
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_applications.py
```diff
@@ -655,10 +655,7 @@ def download_file(url):
 
     prepare_one_file("PSFcurve_data_v2.txt")
     prepare_one_file("MLTdata-preproduction.ecsv")
-    # TODO - temporary path until workflows/documentation PR is merged
-    download_file(
-        "https://raw.githubusercontent.com/gammasim/workflows/documentation/schemas/jsonschema.yml"
-    )
+    download_file("https://github.com/gammasim/workflows/blob/main/schemas/jsonschema.yml")
```
orelgueta, line 658: Usually the links start with `raw`, it is not important here?

orelgueta (approved):
Approving but asked a question about a possible "issue".

tests/integration_tests/test_applications.py
GernotMaier: You are right - asking the internet and it seems the majority (but not all) pages recommend using raw. Changed it. Will merge after tests are successful.

PR #560, opened 2023-08-22
Reduce number of tests for pull requests.

GernotMaier:
Reduce the number of tests for draft pull requests:

- don't run the docker image building for drafts.
- reduce matrix dimension for unit tests (see discussions #561 )

This PR also includes a correction on the job names for the build-docker-corsika-simtelarray-image.yml (this name was duplicated in build-docker-images.yml)


.github/workflows/CI-unittests.yml
```diff
@@ -26,25 +26,25 @@ jobs:
             python-version: "3.9"
```
VictorBarbosaMartins, line 26: I did not understand why we keep ubuntu-latest with mamba and macos-latest for version 3.9 if you decided not to support 3.9. I would expect either we support it completely or we don't support at all. What is the idea behind it?

.github/workflows/CI-unittests.yml
```diff
@@ -26,25 +26,25 @@ jobs:
...
             install-method: mamba
 
-          - os: ubuntu-latest
-            python-version: "3.9"
-            install-method: pip
+#          - os: ubuntu-latest
```
VictorBarbosaMartins, line 29: We usually don't comment out code (bad practice). I suggest we remove the comments

VictorBarbosaMartins (commented):
The title of the PR says "reduce number of tests for draft pull requests" but it actually reduced the number of tests for all kinds of pull requests and we now only test PRs which are ready for review. I just want to make sure that this is the effect you expect, since the title of the PR was not updated after discussions in #561 and for this PR only 10 tests were ran (and not 16 as we used to have before). Apart from that and the two comments, all good.

.github/workflows/CI-unittests.yml
GernotMaier: I took your comment in the discussions into account and decided to keep 3.9 as long as DPPS supports 3.9. I hope this is ok - we lower the number of tests by a x2, so this is progress.

.github/workflows/CI-unittests.yml
GernotMaier: Yes I know. My suggestion is to keep it, as this allows us to switch it on much easier again.

GernotMaier:
Thanks @VictorBarbosaMartins  for the review. I've changed the title of the pull request to make the intention clearer (yes, we want to change it for all).

Let me know if there is anything left.

VictorBarbosaMartins (approved):
All good, thanks!

PR #563, opened 2023-08-23
Consistent usage of MLTData test file

GernotMaier:
Unit and integration tests were using different version of the mirror test file.

Replace the one in `tests/resources` and remove the DB call in integration tests.


tests/integration_tests/test_applications.py
```diff
@@ -509,8 +509,8 @@
     "add_file_to_db": [
         [
             "--file_name",
-            "TESTMODELDIR/MLTdata-preproduction.usermeta.yml",
```
orelgueta, line 512: Are you sure about this one?

orelgueta (commented):
Did you already remove the file from the DB?

GernotMaier:
> Did you already remove the file from the DB?

How do I do this?

The answer is no.

orelgueta:
Good, please let me know when this PR is merged and I will remove it (it's a bit tricky, maybe I should write an application for this).

orelgueta (approved):
OK, I am approving then and you can merge when tests pass.

VictorBarbosaMartins (approved):
Not sure you were waiting for me to review here (as Orel already approved) but I have only a general comment: 
We have three different ways to get files for the tests:
- resources directory
- get from DB
- TESTMODELDIR

This is a bit confusing, even for us developing new code. Where should we put new files that will accompany new developed code? Shouldn't we have a standard way of doing that?

(For the record, the review took me 9 minutes)

GernotMaier:
> Not sure you were waiting for me to review here (as Orel already approved) but I have only a general comment: We have three different ways to get files for the tests:
> 
>     * resources directory
> 
>     * get from DB
> 
>     * TESTMODELDIR
> 
> 
> This is a bit confusing, even for us developing new code. Where should we put new files that will accompany new developed code? Shouldn't we have a standard way of doing that?
> 
> (For the record, the review took me 9 minutes)

I think it is only two - resources directory and getting it from the DB.

There are a couple of thoughts behind it

- minimize the number of files in tests/resources (as they will have to be synced with files in the DB at some points; which might lead it issues)
- unit tests avoid calls to the DB to be as quick as possible (it is unavoidable for some tests)
- integration tests are allowed to call the DB (however, if a file is used in the unit tests and already in tests/resources, then we use that)

Does this make sense?

VictorBarbosaMartins:
> > Not sure you were waiting for me to review here (as Orel already approved) but I have only a general comment: We have three different ways to get files for the tests:
> > ```
> > * resources directory
> > 
> > * get from DB
> > 
> > * TESTMODELDIR
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is a bit confusing, even for us developing new code. Where should we put new files that will accompany new developed code? Shouldn't we have a standard way of doing that?
> > (For the record, the review took me 9 minutes)
> 
> I think it is only two - resources directory and getting it from the DB.
> 
> There are a couple of thoughts behind it
> 
> * minimize the number of files in tests/resources (as they will have to be synced with files in the DB at some points; which might lead it issues)
> * unit tests avoid calls to the DB to be as quick as possible (it is unavoidable for some tests)
> * integration tests are allowed to call the DB (however, if a file is used in the unit tests and already in tests/resources, then we use that)
> 
> Does this make sense?

It makes sense, thank you. Should we write it down somewhere for guidance?

PR #572, opened 2023-08-28
Corsika histograms with hd5 format

### LINKED ISSUE #547: Write corsika histograms output in hd5 files instead of ecsv
The last point is what was just raised in the meeting. 
We need to consider writing out hd5 files instead of ecsv files. It will probably not require any real changes to the code, just a quick change.

_Originally posted by @orelgueta in https://github.com/gammasim/simtools/pull/492#pullrequestreview-1582319165_

_UPDATE:_ 27.08.2023
**Status now:** 
- At the moment we save corsika output histograms [(generate_corsika_histograms.py)](https://github.com/gammasim/simtools/blob/main/simtools/applications/generate_corsika_histograms.py) into ecsv files. Multidimensional arrays are flattened (x and y coordinates for each element in the 2D histograms are passed at each row);
- One ecsv file is saved per histogram.

**Goal:** Change the format of the corsika output histograms to hd5 format. The hd5 format allows us to save multidimensional histograms, while the ecsv format allows only 2D arrays.

**Method:** 
- Since we already use astropy QTables, the change of format is expect to be rather simple. 
- Instead of having one hd5 file per histogram, I will allow the application to write the output of all the histograms into a single hd5 file, which is a cleaner solution and would not pollute the output directory with a multitude of files.

Please @GernotMaier @orelgueta, let me know if you are happy with this solution or if you have different ideas. Thank you.

VictorBarbosaMartins:
Closes #547.

Here I changed the code to instead of saving the histograms into ecsv files, it saves now into a hdf5 file. The following changes were made:

- Corrected the pylint, black etc which had not been applied before to some files;
- Changed ecsv to hdf5 files everywhere;
- Instead of saving one ecsv file per histogram type (e.g. wavelength, counts, etc) we now save all the histograms into a single hdf5 file, inside different "groups";
- Adapted the unit tests accordingly;
- Allowed the user to pass the desired hdf5 file name as a parameter to the application through `--hdf5_file_name`. If the user dosnt give any name but want the hdf5 file to be generated the name is taken from the input IACT file, e.g., if the input file is `tel_output_10GeV-2-gamma-20deg-CTAO-South.corsikaio`, the hdf5 output file is set to `tel_output_10GeV-2-gamma-20deg-CTAO-South.hdf5`.
- Created the integration test to test this last feature;
- Changed course to adapt for the use of `ctapipe.io` astropy helper functions according to https://github.com/gammasim/simtools/issues/547#issuecomment-1697117785. This includes ctapipe for the first time as a dependency in simtools, which is only justified because we intend to use it more often in the development of upcoming applications.

**Good starting point for review:** I think that a good starting point is looking at the application first and see what it does, e.g. by trying itself locally (and if you have other means to open the hdf5 file, it is worth having a look at its content.


GernotMaier:
@VictorBarbosaMartins - can you give me an example on how to run it on the command line?

VictorBarbosaMartins:
> @VictorBarbosaMartins - can you give me an example on how to run it on the command line?

Hi @GernotMaier. Sure, you can try the following:
```
simtools-generate-corsika-histograms --IACT_file ./tests/resources/tel_output_10GeV-2-gamma-20deg-CTAO-South.corsikaio --png --hdf5
```

and then look inside `/workdir/external/simtools/simtools-output/generate_corsika_histograms/application-plots/` for the output files, which will be a sequence of pngs files and a hdf5 table.

There is no application yet to read the hdf5 file but I implemented a function in the module. My suggestion to the following to check the content of the file:
- go to the `test_export_and_read_histograms` in `tests/unit_tests/corsika/test_corsika_histograms.py`,
- add a print statement `print(list_of_tables)` and an `assert False` before the end of the test;
- run this specific test (`pytest tests/unit_tests/corsika/test_corsika_histograms::test_export_and_read_histograms`) and you will see the content of the file printed out in the terminal. obs.: you can also print the meta data in the for loop and check its content.

I hope that helps, please let me know if you need more guidance.

simtools/corsika/corsika_histograms.py
```diff
@@ -1329,41 +1419,50 @@ def fill_ecsv_table(self, hist, x_edges, y_edges, x_label, y_label):
             The x edges of the histograms.
         y_edges: numpy.array
             The y edges of the histograms.
+            Use None if dealing with 1D histogram.
```
GernotMaier, line 1422: Suggest to shorten this "Use None for 1D histogram."
(unclear otherwise what dealing is)

simtools/corsika/corsika_histograms.py
```diff
@@ -1329,41 +1419,50 @@ def fill_ecsv_table(self, hist, x_edges, y_edges, x_label, y_label):
...
         x_label: str
             X edges label.
         y_label: str
             Y edges label.
+            Use None if dealing with 1D histogram.
```
GernotMaier, line 1427: Suggest to shorten this "Use None for 1D histogram."
(unclear otherwise what dealing is)

tests/unit_tests/corsika/test_corsika_histograms.py
```diff
@@ -783,52 +784,28 @@ def test_dict_1D_distributions(corsika_histograms_instance_set_histograms):
...
-        "hist_2D_photon_density_distr_all_tels.ecsv",
-    ]:
-        assert io_handler.get_output_directory(dir_type="test").joinpath(file_name).exists()
+    # Read hdf5 file
+    list_of_tables = corsika_histograms_instance_set_histograms.read_hdf5(output_file)
+    assert len(list_of_tables) == 12
+    for table in list_of_tables:
+        assert isinstance(table, Table)
+    # Check piece of metadata
+    assert list_of_tables[-1].meta["corsika_version"] == 7.741
```
GernotMaier, line 808: Could you use here the value from the metadata before writing (earlier in the function) instead of putting the corsika version to the last digit into the file?

tests/unit_tests/utils/test_names.py
```diff
@@ -75,3 +75,10 @@ def test_translate_simtools_to_corsika():
     simtools_pars = ["corsika_obs_level", "corsika_sphere_radius", "corsika_sphere_center"]
     for step, simtools_par in enumerate(simtools_pars):
         assert names.translate_simtools_to_corsika(simtools_par) == corsika_pars[step]
+
+
+def test_sanitize_name():
+    assert names.sanitize_name("y_edges unit") == "y_edges_unit"
+    assert names.sanitize_name("Y_EDGES UNIT") == "y_edges_unit"
+    assert names.sanitize_name("123name") == "_123name"
```
GernotMaier, line 83: Is this standard / usual behaviour to add a "_" for strings starting with a digit? Just asking, as I haven't seen it before.

simtools/utils/names.py
```diff
@@ -823,3 +824,35 @@ def translate_simtools_to_corsika(simtools_par):
...
+
+    # Convert to lowercase
+    sanitized = name.lower()
+
+    # Replace spaces with underscores
+    sanitized = sanitized.replace(" ", "_")
+
+    # Remove characters that are not alphanumerics or underscores
+    sanitized = re.sub(r"\W|^(?=\d)", "_", sanitized)
+
```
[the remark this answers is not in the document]
VictorBarbosaMartins: At the moment I think it is not necessary. I added this because I was getting lots of warnings from the Tables module that I wouldn't be able to use dot ('.') to access attributes if the attribute names are not sanitized. That means, if the attribute name is not sanitized, we get a warning. And the name of the attributes are fixed (should not be changed).

simtools/corsika/corsika_histograms.py
```diff
@@ -37,14 +41,16 @@ class CorsikaHistograms:
         Instance label.
     output_path: str
         Path where to save the output of the class methods.
+    hdf5_file_name: str
+        Name of the hdf5 file where to save the histograms.
```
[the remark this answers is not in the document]
VictorBarbosaMartins: It indeed reads strangely. I adopted now your suggestion, thanks.

simtools/corsika/corsika_histograms.py
```diff
@@ -58,10 +64,18 @@ def __init__(self, input_file, label=None, output_path=None):
...
         self.io_handler = io_handler.IOHandler()
         _default_output_path = self.io_handler.get_output_directory(self.label, "corsika")
+
         if output_path is None:
             self.output_path = _default_output_path
         else:
             self.output_path = Path(output_path)
+
+        if hdf5_file_name is None:
+            default_hdf5_file_name = re.split(r"\.", self.input_file.name)[0] + ".hdf5"
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Good point. I adopted your solution.

simtools/corsika/corsika_histograms.py
VictorBarbosaMartins: Done.

simtools/corsika/corsika_histograms.py
VictorBarbosaMartins: Done

tests/unit_tests/corsika/test_corsika_histograms.py
VictorBarbosaMartins: yes, I changed it now. now we dont have to change this test everytime the corsika version changes.

tests/unit_tests/utils/test_names.py
VictorBarbosaMartins: Good question. To use all the digits passed and in the same order, we have to add some character before the name. This functionality is not strictly necessary, because we don't expect names starting with numbers but it makes the function more complete (maybe more useful in the future).

simtools/utils/names.py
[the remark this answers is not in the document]
GernotMaier: But are you sure that you really catching all cases of input strings to sanitize them? Adding a "if not sanitized.isidentifier() raise WhateverError" would make this function more robust.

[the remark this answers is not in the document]
VictorBarbosaMartins: I added now as you suggested (I didnt know about the "isidentifier" method). however, I do not know how to include a test, since I do expect that every string would be covered by the function.

VictorBarbosaMartins:
Hi @GernotMaier, I had to change the flags and if statements again but I think tests are fine now. Please have a look again if tests pass. Thanks.

simtools/utils/names.py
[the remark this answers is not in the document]
GernotMaier: OK!

GernotMaier (approved):
All good now, thank you!

PR #575, opened 2023-08-28
Remove dependency of ray-tracing on files

GernotMaier:
Ray-tracing classes RayTracing and SimtelRunnerRayTracing are reading their default configuration values from files in ./data/parameters. File names are hard wired in the class inits. Parameter files for both classes are almost identical.

Changed this:
- default parameters are defined in SimtelRunnerRayTracing
- removed files in data/parameters

This will allow to run those tools without having the github repo cloned.

There should be no changes in results, printouts, or configuration of these tools.


GernotMaier:
Thanks for the quick feedback!

PR #576, opened 2023-08-28
Add pytest-xdist

GernotMaier:
Allows to run tests in parallel on all available core in your local environment. This significantly accelerates testing. This is for developers only.

Use with `pytest -n auto tests/unit_tests`.

More info at https://pytest-xdist.readthedocs.io/en/latest/

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,6 +11,7 @@ dependencies:
   - pre-commit
   - pytest
   - pytest-cov
+  - pytest-xdist
   - scipy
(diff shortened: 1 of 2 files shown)


orelgueta (approved):
Haven't tried it but the changes themselves look good so I approve.

GernotMaier:
Thanks for merging - just want to add that I've added also a remark in the documentation (documentation-update-developer-section branch)

PR #577, opened 2023-08-29
Fix issues with installed packages in the containers

orelgueta:
Remove unnecessary packages from the dev/prod docker files since they are already included in the sim_telarray docker files. Add bc and zstd to the sim_telarray image. Remove unnecessary krb packages from the sim_telarray docker file.


GernotMaier (approved):
All good, thanks!.

PR #579, opened 2023-08-29
Add tests for general.py

orelgueta:


tests/unit_tests/utils/test_general.py
```diff
@@ -337,3 +413,258 @@ def test_get_file_age(tmp_test_directory):
...
+    )
+
+
+def test_file_has_text(tmp_test_directory) -> None:
+    file = tmp_test_directory / "test_file_has_text.txt"
+    text = "test"
+    with open(file, "w") as f:
+        f.write(text)
+    assert gen.file_has_text(file, text)
+    assert not gen.file_has_text(file, "test2")
```
GernotMaier, line 449: What is the expected return for an empty file? I guess False? Worth testing?

tests/unit_tests/utils/test_general.py
```diff
@@ -337,3 +413,258 @@ def test_get_file_age(tmp_test_directory):
...
+def test_log_level_from_user() -> None:
+    """
+    Test get_log_level_from_user() function.
+    """
+    assert gen.get_log_level_from_user("info") == logging.INFO
+    assert gen.get_log_level_from_user("debug") == logging.DEBUG
+    assert gen.get_log_level_from_user("warn") == logging.WARNING
+    assert gen.get_log_level_from_user("warning") == logging.WARNING
+    assert gen.get_log_level_from_user("error") == logging.ERROR
+    assert gen.get_log_level_from_user("critical") == logging.CRITICAL
```
GernotMaier, line 583: We use nowhere in the code critical. Is this necessary?

GernotMaier (changes requested):
Very minor comments. (quite some functions of general.py I did not know about)

tests/unit_tests/utils/test_general.py
orelgueta: The function would have failed cause you cannot mmap an empty file. Generally I am of the opinion that we should allow python to raise its own error in this case. However, in this specific case I added a `try...except` to deal with that case and the case where the file does not exist (and the appropriate tests).

tests/unit_tests/utils/test_general.py
orelgueta: Yes the test is necessary cause it's an available option for logging and we might use it at some point in the future.

orelgueta:
Thanks for the review! I added the requested test/use case.

tests/unit_tests/utils/test_general.py
GernotMaier: ok!

GernotMaier (approved):

Issue #420, opened 2022-11-29
Use `dotenv` for loading environment variables?


maxnoe:
There is a very commonly used convention of having environment-variables in a `.env` file and there is a python package to automatically load those:

https://pypi.org/project/python-dotenv/

this could replace the custom `set_DB_environ.sh` script.

PR #582, opened 2023-08-30
Read environment variables using the dotenv package

GernotMaier:
Add functionality to read environment variables using the dotenv package (as suggested by Max in #420).

- adds additional dependency on [python-dotenv](https://pypi.org/project/python-dotenv/)
- replaces the current way of setting environment variables using `set_DB_environ.sh`
- variables are now set in a `.env` file, see `.env_template` (this replaces `set_DB_environ_template.sh`)
- minimal change to load the variables in the configurator module
- default file is `.env`, although other files can be set using the `--env_file` command line parameter 
- no change in functionality when using the 'old way' of setting environment variables. This is still used in the github CI, as secrets are given as environment variables (required two-lines addition in conf.py)
- updated the units tests to catch the different ways of setting environment variables

Note the changes in starting the docker-dev container. This is a minimal longer command line but I think still acceptable.

The `.env` file is compatible with the file format expected by docker to load environment variables (with `--env-file`, see updated instructions for usage of simtools-prod). This is important, as we didn't have a consistent way before to set environment variables in the simtools-prod container.

(note that this pull request also includes the changes of the sorting of the dependencies #581; so maybe better first to #581)


GernotMaier:
@orelgueta , @VictorBarbosaMartins - whoever wants to pick up this PR, please self-assign. I would say the PR is of size "S" (assuming that we start the scale at "XS").

tests/unit_tests/configuration/test_configurator.py
```diff
@@ -187,3 +181,49 @@ def test_initialize_output(configurator, args_dict):
...
+    configurator._fill_from_environmental_variables()
+    if "TEST_ENV_VARIABLE" in os.environ:
+        del os.environ["TEST_ENV_VARIABLE"]
+    assert "test_env_variable" not in configurator.config
+
+    # this is a valid configuration parameter, but already configured
+    # _fill_from_environmental_variables() should not change it
+    os.environ["LOG_LEVEL"] = "DEBUG"
+    configurator._fill_from_environmental_variables()
+    assert configurator.config["log_level"] == _config_save["log_level"]
```
orelgueta, line 203: Should this also compare to the actual log_level (e.g., `== INFO`) so that the test does not pass accidentally if the default configuration is "DEBUG"?

orelgueta (commented):
Looks good. Tiny question/comment about a test. It's very minor but I don't approve just in case.

tests/unit_tests/configuration/test_configurator.py
GernotMaier: The test depends on that the default value for log_level as defined in [here](https://github.com/gammasim/simtools/blob/e55809c6c90f3ed8b041d30ff9fd7542e84dab48/simtools/configuration/commandline_parser.py#L159). 

Do you see this an issue? Will we ever change the default level to "debug"? 

Note that we nowhere change the actually value of logging, so what I would have to add is a check if `config["log_level"] = "DEBUG"` , in this case use "INFO". Should I do this?

orelgueta: This test, as it is, depends on a value set somewhere else. If, for whatever reason, you change the default value to "debug" (extremely unlikely, I know, but it doesn't matter), then this test would not cover this case anymore. Since the tests have to stand on their own and always test what they are meant to test, I suggest to change it to 
```
assert configurator.config["log_level"] == _config_save["log_level"] == "INFO"
```
(not that I am not sure if `"INFO"` is the exact representation, I did not check.)
This will make this test robust against changes somewhere else and would fail if you do change somewhere else and alert you to change also this test.

orelgueta: Perfect, thanks!

orelgueta (approved):

GernotMaier:
Thanks a lot for the quick review!

PR #586, opened 2023-08-31
Env variables names

### LINKED ISSUE #583: Naming of Environmental Variables
It's convention to have these prefixed with the project, e.g. `SIMTOOLS_DB_PORT`, `SIMTOOLS_DB_PW`

_Originally posted by @maxnoe in https://github.com/gammasim/simtools/pull/582#discussion_r1310482483_

orelgueta:
Fixes #583. 
By the way, one of the fixes, in `simtel_path_no_mock` is not really related to this PR and I am surprised that the tests passed before this fix. It might point to an issue somewhere but I couldn't find it. Maybe @GernotMaier would know.

### DIFF
--- a/.env_template
+++ b/.env_template
@@ -1,12 +1,12 @@
 # Template file for environment variables settings
 # Copy this file to '.env' and add the required
 # credentials for simulation model database access
-DB_API_PORT=27017 #Port on the MongoDB server
-DB_SERVER='cta-simpipe-protodb.zeuthen.desy.de' # MongoDB server
-DB_API_USER=YOUR_USERNAME # username for MongoDB: ask the responsible person
-DB_API_PW=YOUR_PASSWORD # Password for MongoDB: ask the responsible person
-DB_API_AUTHENTICATION_DATABASE='admin'
-SIMTEL_PATH='/workdir/sim_telarray'
+SIMTOOLS_DB_API_PORT=27017 #Port on the MongoDB server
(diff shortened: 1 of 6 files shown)


GernotMaier (approved):

GernotMaier:
On `simtel_path_no_mock` - it seems that in at least some of the test this is not necessary and `simtel_path` is enough. This can be understood, e.g. ./unit_tests/corsika/test_corsika_runner.py is not running simtools, so the empty string passed for simtel_path is rightly accepted.

GernotMaier:
@orelgueta  - forgot this: could you check also the github action workflows? We are also setting environmental variables there.

orelgueta:
> On `simtel_path_no_mock` - it seems that in at least some of the test this is not necessary and `simtel_path` is enough. This can be understood, e.g. ./unit_tests/corsika/test_corsika_runner.py is not running simtools, so the empty string passed for simtel_path is rightly accepted.

Yes but in `test_ray_tracing.py` we use the no_mock version so I am quite surprised that it worked thus far after changing to use dotenv. My current guess is that it worked on GitHub because it sets the env variables in a different way, but I didn't check this.

orelgueta:
> @orelgueta - forgot this: could you check also the github action workflows? We are also setting environmental variables there.

Right, I forgot about that. That hopefully will explain why the integration tests failed. Fixed now, hopefully tests will pass, in which case I will merge.

GernotMaier:
All good now; I think this is good to be merged.

orelgueta:
Thanks for the review and the hint about the workflows!

PR #587, opened 2023-09-01
Remove unneeded calls to simtel_path_no_mock in unit tests

GernotMaier:
Related to #586 

The function `simtel_path_no_mock` (defined in conf.py) is not needed for some (all?) of the unit tests.


GernotMaier:
Note: only tests/integration_tests/test_ray_tracing.py requires simtel_no_mock - but see #588 - this should be changed.

orelgueta:
Let's discuss first in #588 before you continue with this PR.

GernotMaier:
I think we should review and merge this one. Should use the mock function (unless simtel_array is really needed).

PR #595, opened 2023-09-06
Add simplest integration test for simulate_prod

orelgueta:
Add integration test for simulate_prod and fix that the output directory is not the model directory.

### DIFF
--- a/tests/integration_tests/test_applications.py
+++ b/tests/integration_tests/test_applications.py
@@ -610,6 +610,7 @@ APP_LIST = {
             "10",
             "--data_directory",
             "TESTOUTPUTDIR/",
+            "--pack_for_grid_register",
         ]
     ],
     # print_array
--- a/tests/integration_tests/test_applications.py
+++ b/tests/integration_tests/test_applications.py
@@ -36,7 +36,7 @@ APP_LIST = {
             "20",
             "--test",
             "--output_path",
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_applications.py
```diff
@@ -559,9 +559,57 @@
...
+            "0",
+            "--run",
+            "1",
+            "--nshow",
+            "10",
+            "--data_directory",
+            "TESTOUTPUTDIR/",
+        ]
+    ],
+    "simulate_prod::gamma_20_deg_pack_for_grid": [
```
GernotMaier, line 591: I don't see what is different in simulate_prod::gamma_20_deg_pack_for_grid compared to "simulate_prod::gamma_20_deg. It is only the `--run` option - does this need to be tested?

GernotMaier (commented):
One important question regarding the two tests.

tests/integration_tests/test_applications.py
orelgueta: You are right, while changing options back and forth I forgot to put back the `"--pack_for_grid_register",` option for the second test. I added it now.

GernotMaier: Good!

GernotMaier (approved):

PR #598, opened 2023-09-06
Add event to push on merged pull request

### LINKED ISSUE #597: sim_telarray container not pushed to github packages
The sim_telarray container is not pushed to the packages and doesn't appear in the list of packages. 

I am pretty sure that when we use `ghcr.io/gammasim/simtools-simtelarray:latest`, we are actually using the old one from the archived container repo. This causes an issue that zstd is not available in that old container, which means productions fail at the moment.

I couldn't find any difference between the settings of two containers which do get published and the sim_telarray one except two places:
```
pull_request:
    branches: ["main"]
    types: [review_requested, ready_for_review]
```
exists only for the dev/prod containers and the same is true for
```
github.event.pull_request.merged == true
```

I assume it is the latter which is responsible for the issue, since It seems the containers are built, they are just not pushed (we get the following warning at the end of the build process - `WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load`).

The easiest way to test this is by manually triggering the workflow (which I just did). If that works, I will open a short PR to add the event above to the workflow. If not, I will ask @GernotMaier to look into it.

orelgueta:
See #597 for details.
Fixes #597.

### DIFF
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -68,7 +68,7 @@ jobs:
           context: .
           platforms: linux/amd64,linux/arm64/v8
           push:
-            ${{ github.event_name == 'pull_request' || github.event_name == 'release' || github.event_name == 'workflow_dispatch' }}
+            ${{ github.event_name == 'pull_request' || github.event_name == 'release' || github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true}}
           file: ./docker/Dockerfile-${{ matrix.type }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}-${{ matrix.type }}
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   pull_request:
     paths:
-      - 'docker/Dockerfile-simtelarray'
+      - './docker/Dockerfile-simtelarray'
   push:
     branches:
       - main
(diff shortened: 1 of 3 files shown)


GernotMaier (commented):
Just to understand: almost all merges do not affect the corsika/simtelarray image. Do you want to make sure that if we change something in the corresponding Dockerfile that a new image is build? We can restrict this workflow to run only when changes to this files are applied.

orelgueta:
As we said, we will discuss it together but I think it is already clear that my solution is not really what we want. However, to clarify, the issue is the push part, not the building part. 

So, I think your solution of building when a specific file is changed is good. Then we need to add that event to the push part and I suggest to also do that for the "scheduled" event so that we get one once a week in case we modify the sim_telarray tarball.

orelgueta:
I think I made all the necessary changes, including changing the name of the published container. I am not sure how to test it though without merging this PR first...

GernotMaier (commented):
Requires also changes to 
- https://github.com/gammasim/simtools/blob/cf135add71d43a7816812bb2f60997b91f2b13af/docker/Dockerfile-dev#L1
- https://github.com/gammasim/simtools/blob/cf135add71d43a7816812bb2f60997b91f2b13af/docker/Dockerfile-prod#L1

as both simtools-prod and simtools-dev are based on this container. The change in name should be reflected there. 

I leave it to you to merge this PR and do these updates in a separate PR, or do it here (it is two words to be changed)

orelgueta:
You are right, I forgot about changing those names, thanks for spotting that. I did that now so please review when you can.

GernotMaier:
Looks good - fine to merge after successful tests.

orelgueta:
Can you please approve then? Merging is currently blocked (will wait for tests to pass anyway).

GernotMaier (approved):
(sorry, thought I did approve)

PR #599, opened 2023-09-07
Refactor prod configuration

orelgueta:
This PR should be looked at after #596 and the ones previous.
The main goal of the PR is to move code from the application into the simulator class and simplify it. It also gets rid of a hack we had before. Tests were changed accordingly and simplified as well.

### DIFF
--- a/simtools/corsika/corsika_config.py
+++ b/simtools/corsika/corsika_config.py
@@ -90,6 +90,7 @@ class CorsikaConfig:
         self.label = label
         self.site = names.validate_site_name(site)
         self.primary = None
+        self.eslope = None
         self._config_file_path = None
         self._output_generic_file_name = None
         self._simtel_source_path = simtel_source_path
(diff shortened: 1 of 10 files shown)


simtools/corsika/corsika_config.py
```diff
@@ -246,6 +247,9 @@ def _validate_and_convert_argument(self, par_name, par_info, value_args_in):
             value_args = [0 * par_info["unit"][0], value_args[0]]
         elif par_name == "PRMPAR":
             value_args = self._convert_primary_input_and_store_primary_name(value_args)
+        elif par_name == "ESLOPE":
+            # To make it easier to pass to sim_telarray
```
GernotMaier, line 251: I am not sure if this comment is helpful. It didn't tell me anything. Is this more a motivational comment why self.eslope was introduced? If yes, then it should be removed.

GernotMaier (approved):
Thanks. Approved with one minor comment.

simtools/corsika/corsika_config.py
orelgueta: Yes it is exactly that, a motivational comment cause otherwise it seems sort of out of place. I agree it should be removed though so I will remove it.

PR #602, opened 2023-09-11
Push also on scheduled build

orelgueta:
This PR allows to also test the pull_request.merged rule.

### DIFF
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -71,7 +71,7 @@ jobs:
           context: .
           platforms: linux/amd64,linux/arm64/v8
           push:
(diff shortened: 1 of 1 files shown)


GernotMaier (approved):

orelgueta:
The `pull_request.merged` rule doesn't work, the image isn't pushed - https://github.com/gammasim/simtools/actions/runs/6147199985/job/16678126586

Should I open dummy PRs to continue trying/testing this? Do you have a better way to test it? Perhaps locally? (it's really slow on github)

orelgueta:
Local test confirmed that the "schedule" should work once this PR is merged. The `event.pull_request.merged` trigger does not work for pushing, so I changed it to the usual `github.event_name == 'pull_request'`. With the addition of the exact path provided now for this action, it might not lead to unnecessary runs. If it does, I suggest we remove it. 

@GernotMaier, could you please review again?

GernotMaier (approved):
OK - was also confused in the based by the differences of github.event_name and github.event.

Issue #605, opened 2023-09-18
Strip comments from environment variables


GernotMaier:
We use the `.env` files to set environmental variables (both in python and in docker).

For docker, the comments are not stripped correctly, e.g., the file [.env_template](.env_template) results in

```
SIMTOOLS_DB_API_PORT="27017 #Port on the MongoDB server"
```

when setting variables with `docker run --rm -it --env-file .env_template....`.

This means one line of string gymnastics to remove the comment.

This occurs e.g., when using the simtools-prod container to get a file from the DB.

PR #611, opened 2023-09-19
Strip comments from env

GernotMaier:
Strip comments, spaces, and '' from env variables.
This is relevant when using docker with the `--env-file`
option.

I've tested that this works by building a docker container locally and running ` simtools-get-file-from-db --file_name mirror_CTA-N-LST1_v2019-03-31.dat`

Review after pull request #603.

Addresses issue #605

### DIFF
--- a/simtools/configuration/configurator.py
+++ b/simtools/configuration/configurator.py
@@ -277,6 +277,10 @@ class Configurator:
                 env_variable_to_read = f"SIMTOOLS_{key.upper()}"
                 if value is None:
                     _env_dict[key] = os.environ.get(env_variable_to_read)
+                if key in _env_dict and _env_dict[key] is not None:
+                    _env_dict[key] = (
+                        _env_dict[key].split("#")[0].strip().replace(" ", "").replace("'", "")
+                    )
         except AttributeError:
             pass
 
--- a/simtools/configuration/configurator.py
+++ b/simtools/configuration/configurator.py
@@ -276,11 +276,12 @@ class Configurator:
                 # environmental variables for simtools should always start with SIMTOOLS_
                 env_variable_to_read = f"SIMTOOLS_{key.upper()}"
                 if value is None:
-                    _env_dict[key] = os.environ.get(env_variable_to_read)
-                if key in _env_dict and _env_dict[key] is not None:
-                    _env_dict[key] = (
-                        _env_dict[key].split("#")[0].strip().replace(" ", "").replace("'", "")
-                    )
+                    env_value = os.environ.get(env_variable_to_read)
+                    if env_value is not None:
+                        env_value = (
+                            env_value.split("#")[0].strip().replace('"', "").replace("'", "")
+                        )
+                    _env_dict[key] = env_value
         except AttributeError:
             pass
 
--- a/tests/unit_tests/configuration/test_configurator.py
+++ b/tests/unit_tests/configuration/test_configurator.py
@@ -185,6 +185,7 @@ def test_initialize_output(configurator):
 
 def test_fill_from_environmental_variables(configurator):
     configurator.parser.initialize_output_arguments()
+    configurator.parser.initialize_db_config_arguments()
     configurator._fill_from_command_line(arg_list=[])
 
     _config_save = copy(configurator.config)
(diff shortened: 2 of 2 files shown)


simtools/configuration/configurator.py
```diff
@@ -277,6 +277,10 @@ def _fill_from_environmental_variables(self):
                 env_variable_to_read = f"SIMTOOLS_{key.upper()}"
                 if value is None:
                     _env_dict[key] = os.environ.get(env_variable_to_read)
+                if key in _env_dict and _env_dict[key] is not None:
```
orelgueta, line 280: Why a separate for loop? We don't read env vars anywhere else so I think it can be in the same loop.

simtools/configuration/configurator.py
```diff
@@ -277,6 +277,10 @@ def _fill_from_environmental_variables(self):
...
+                    _env_dict[key] = (
+                        _env_dict[key].split("#")[0].strip().replace(" ", "").replace("'", "")
```
orelgueta, line 282: When would `replace(" ","")` be relevant (in an appropriate way)?
Do you think you need also `.replace('"', '')`?

tests/unit_tests/configuration/test_configurator.py
```diff
@@ -211,6 +212,19 @@ def test_fill_from_environmental_variables(configurator):
...
 
+    # using .dotenv files with docker: comments are not removed by docker
+    os.environ["SIMTOOLS_DB_API_PORT"] = "27017 #Port on the MongoDB server"
+    os.environ["SIMTOOLS_DB_SERVER"] = "'abc@def.de' # MongoDB server"
+    configurator.config["db_api_port"] = None
+    configurator.config["db_server"] = None
+    configurator._fill_from_environmental_variables()
+    assert configurator.config["db_api_port"] == 27017
+    assert configurator.config["db_server"] == "abc@def.de"
+    if "SIMTOOLS_DB_API_PORT" in os.environ:
```
orelgueta, line 223: Why do you need to delete these from the env? Wouldn't they be cleared anyway once the test finishes?

orelgueta (commented):
Some minor comments

simtools/configuration/configurator.py
GernotMaier: The reason is that `os.environ.get(env_variable_to_read)` can return None, so we need to check again before applying the strip command.

simtools/configuration/configurator.py
GernotMaier: The environmental variable could be "SIMTOOLS_DB_API_PORT=27017_____#Port on the MongoDB server" (where '_" stands for spaces.

I haven't seen `"` anywhere, so I don't think we need them. Or should we add them to be on the safe side?

tests/unit_tests/configuration/test_configurator.py
GernotMaier: You are right - I just following the examples before, where also for the last line the env variables are deleted. I would suggest to leave it at it is, as it makes it easier to add additional lines (without having to think about the previous ones).

simtools/configuration/configurator.py
orelgueta: But `strip` should take care of the extra spaces before `#`:
```
>>> a = "27017     # Port on the MongoDB server"
>>> a.split("#")[0].strip()
'27017'
```
Yes, I haven't seen any `"` anywhere either, but my suggestion was in order to be on the safe side.

tests/unit_tests/configuration/test_configurator.py
orelgueta: No problem

simtools/configuration/configurator.py
orelgueta: Sure, but you can do 
```
for key, value in self.config.items():
    # environmental variables for simtools should always start with SIMTOOLS_
    env_variable_to_read = f"SIMTOOLS_{key.upper()}"
    if value is None:
        env_value = os.environ.get(env_variable_to_read)
        if env_value is not None:
            env_value = env_value.split("#")[0].strip().replace('"', "").replace("'", "")
        _env_dict[key] = env_value
```
This would save the extra for loop and keep a `None` value if that is what is provided. 
(I included in the example above also the fixes in the next comment.)

GernotMaier: OK,, I've changed it accordingly (although I don't really see the differences - same amount of if statements...)

simtools/configuration/configurator.py
GernotMaier: Agree

orelgueta (approved):

PR #612, opened 2023-09-19
Remove wrong try..except in configurator::_fill_from_environmental_variables

GernotMaier:
Unclear why this has been added. At this point, it skips all remaining variables in case of an ATTRIBUTEERROR, which is clearly a bug.

To be reviewed after #611


GernotMaier:
Thanks - I was waiting for the approval of #611 . As this is about the same lines of code, I wanted to be sure that there are no conflicts.

Thanks for the approval, I wait for the tests and merge then.

PR #616, opened 2023-09-19
Remove CORSIKA source code from docker image

GernotMaier:
Remove the CORSIKA source code from the docker image to follow the non-distribution statement from CORSIKA authors (discussed with them).

I've tested this approach with a local build, but clearly we need some tests for containers.

### DIFF
--- a/docker/Dockerfile-simtelarray
+++ b/docker/Dockerfile-simtelarray
@@ -16,12 +16,16 @@ RUN apt-get update && apt-get install -y \
     apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # corsika and sim_telarray
+# (corsika source code is removed to follow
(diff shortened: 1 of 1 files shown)


orelgueta (approved):
I assume you ran the integration tests after this change and they work? (I think they should.) 

I think this should also work if we replace the interaction model because it is supposed to be copied/moved to the run directory prior or after compilation. Since we will need separate sim_telarray containers for different interaction models anyway, it should be fine.

GernotMaier:
Yes, I have been running the integration tests successfully.

There are data files from other hadronic interaction models in this directory, so I assume it will work (but we would have to compile it differently)

Note a cleanup script as part of simtel_array which removes everything not needed to run (e.g., there are pdfs from documentation, a lot of example files, ...). But this is another issue.

PR #618, opened 2023-09-22
Address issues with pyflakes and use flake8 instead

GernotMaier:
This addresses issues with pyflakes for files generated by setuptools (e.g. ./simtools/_version.py), see e.g., the failed tests [pyflakes runs here](https://github.com/gammasim/simtools/actions/runs/6273528145/job/17037966050#step:9:21).

pyflakes unfortunately doesn't allow a lot of configuration. Decided to replace it by flake8 (which is a wrapper for several tools, among them pyflakes) and configure the workflow such that the relevant issues with _version.py are ignored.

This pull requests also includes several issues noted by flake8 (but not pyflakes).

Note that `conda env update -f environment.yml` (with flak8 replacing pyflakes) did not result in a working environment for me and I had to remove it and start again.

### DIFF
--- a/.github/workflows/CI-unittests.yml
+++ b/.github/workflows/CI-unittests.yml
@@ -1,6 +1,6 @@
 ---
 name: CI-unittests
-# Execute unit tests - pyflakes and pytest
+# Execute unit tests - flake8 and pytest
 #
 env:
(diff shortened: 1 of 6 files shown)


orelgueta (approved):
Looks good.

As a side note, I think you might need to add `--prune` to the update command in order to get it to work.

PR #619, opened 2023-09-25
Ensure float64 for coordinate transformations, improved coordinate transformations

GernotMaier:
1. Coordinate transformations require float64 to ensure sub-cm precision. 

Abort with `InvalidCoordinateDataType` if values are given in float32.

Further changes:

- changes of all test files to float32
- update of metadata format of files in data/layout to be consistent with format using in main gitlab repo (only for ...TestLayout files; keep others for now to not make this specific PR too large).

Note that no changes are made regarding telescope naming (e.g., MSTN vs MST).

No need to review telescope positions. Changes are in the metadata section.

See also issue #620.

2. Transformation from UTM to CORSIKA coordinates requires to apply a scale factor depending on the height of the observatory.

Implemented a solution developed and cross checked with KB taking into account the differences between geodetic and geocentric altitude for the WGS84 ellipsoid. The solution should be correct on the sub-mm scale.

The module layout/layout_array is rather long. Therefore decide to move all functions related to geospatial coordinates into a new module called layout/geo_coordinates.py.


orelgueta (commented):
A couple of comments, mostly minor

data/layout/telescope_positions-North-TestLayout.ecsv
```diff
@@ -2,66 +2,41 @@
...
+# - {name: pos_y, unit: m, datatype: float64, description: telescope position towards West relative to the array center.}
+# - {name: pos_z, unit: m, datatype: float64, description: telescope position upwards relative to the CORSIKA observation level.}
+# - {name: geo_code, datatype: string, description: Geographical Position Code}
+# meta: !!omap
+# - data_type: position_file
+# - description: "CTAO Northern Array Element Coordinates (prod6)"
+# - center_northing: "3185067.275 m"
+# - center_easting: "217608.975 m"
+# - corsika_obs_level: "2156 m"
+# - corsika_sphere_radius: {LST: 12.5 m, MST: 9.15 m, SST: 3 m}
```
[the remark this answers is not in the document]
GernotMaier: Removed the SST value (I've copied and pasted the values simply).

I still struggle to understand if the files in this directory are more than just some examples. They shouldn't be used for any production. I've reduced the MST sphere radius to be consistent with the values we have in the layout gitlabs.

data/layout/telescope_positions-South-TestLayout.ecsv
```diff
@@ -2,61 +2,23 @@
...
+# - {name: pos_x, unit: m, datatype: float64, description: telescope position towards North relative to the array center.}
+# - {name: pos_y, unit: m, datatype: float64, description: telescope position towards West relative to the array center.}
+# - {name: pos_z, unit: m, datatype: float64, description: telescope position upwards relative to the CORSIKA observation level.}
+# meta: !!omap
+# - data_type: position_file
+# - description: "CTAO Southern Array Element Coordinates (prod6)"
+# - center_northing: "7269466 m"
+# - center_easting: "366822 m"
+# - corsika_obs_level: "2147 m"
+# - corsika_sphere_radius: {LST: 12.5 m, MST: 9.15 m, SST: 3 m}
```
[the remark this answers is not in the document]
GernotMaier: Yes, but this is consistent, as there are no SCTs in this file.

simtools/layout/layout_array.py
```diff
@@ -563,6 +567,9 @@ def read_telescope_list_file(telescope_list_file):
             _logger.error(f"Error reading list of array elements from {telescope_list_file}")
             raise
         _logger.info(f"Reading array elements from {telescope_list_file}")
+        if np.any([col.dtype == np.float32 for col in table.columns.values()]):
```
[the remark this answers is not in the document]
GernotMaier: This was a conscious decisions, as right now we test all columns not be float32 (e.g., also the telescope names, geocode, counters). So we allow for string and int. I do think this is enough in this case.

Alternatively: would have to have a list of all allowed axes and test them. This would be a lot more code, not sure if this is needed.

simtools/layout/layout_array.py
```diff
@@ -930,9 +937,16 @@ def include_radius_into_telescope_table(telescope_table):
...
+            u.Quantity(
+                telescope_table.meta["corsika_sphere_radius"][
+                    names.get_telescope_type(tel_name_now)
+                ]
+            )
+            .to("m")
+            .value
             for tel_name_now in telescope_table["telescope_name"]
         ]
+        telescope_table["radius"] = telescope_table["radius"].quantity * u.m
```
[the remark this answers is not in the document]
GernotMaier: This is the move the unit from each entry of this table column into the column header. So instead of `"radius:" 1m, 5m, 20m` we have now `"radius | m" 1, 5, 20`, consistent with all other table columns.

GernotMaier:
@orelgueta  - please have a look again, there is one open point about reading in integer.

data/layout/telescope_positions-North-TestLayout.ecsv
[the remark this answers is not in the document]
orelgueta: We don't really run production using simtools right now, so it's all theoretical at the moment, but I don't see why these files (or equivalent ones) shouldn't be used for production. We can put them in the DB if it makes it more "official", but I do think we should use your ecsv layout files for production in the future.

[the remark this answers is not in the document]
orelgueta: Either way this is resolved.

data/layout/telescope_positions-South-TestLayout.ecsv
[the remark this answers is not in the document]
orelgueta: OK!

simtools/layout/layout_array.py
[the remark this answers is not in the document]
orelgueta: OK, this solution should be fine then (perhaps will change in the future with a full schema, we'll see if it's necessary).

simtools/layout/layout_array.py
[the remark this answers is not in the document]
orelgueta: Oh, OK, didn't know it makes a difference but it's no problem of course.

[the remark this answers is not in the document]
GernotMaier: Don't know if it makes a differences, but it is consistent with all other columns in the table (where the columns have values, and not quantities)

GernotMaier:
@orelgueta - this is now ready for review.

I've added to the testing routines the scale factors we compared with Konrad. So this should work correctly!

simtools/layout/geo_coordinates.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        WGS84 coordinate system definition.
+
+        Returns
+        -------
+        pyproj.CRS
+            WGS84 coordinate system.
+
+        """
+        return pyproj.CRS("EPSG:4326")
```
orelgueta, line 55: Not my area of expertise so I can only ask, this hardcoded EPSG is OK? Applicable for both sites?

simtools/layout/geo_coordinates.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns
+        -------
+        pyproj.CRS
+            local coordinate system.
+
+        """
+        try:
+            _center_lat, _center_lon, _ = reference_point.get_coordinates("mercator")
+            _scale_factor_k_0 = self._coordinate_scale_factor(reference_point)
+            if not np.isnan(_center_lat.value) and not np.isnan(_center_lon.value):
```
orelgueta, line 75: Should you add an `else` here to protect against the case one of them is `nan`? I think that without it you would simply return `None`, which might be OK, or perhaps you would want to raise an error like the missing reference point one.

simtools/layout/geo_coordinates.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns
+        -------
+        k_0: float
+            Scale factor for local coordinate system.
+
+        """
+
+        try:
+            _center_lat, _, _centre_altitude = reference_point.get_coordinates("mercator")
+        except TypeError:
```
orelgueta, line 113: Now I am a bit confused. In the previous function you tried and excepted an `AttributeError`. Now the exception is a `TypeError`. Is the `try` and `except` in the previous function specifically aimed to catch the `AttributeError` raised in `_coordinate_scale_factor` (below)? If so, why is the try..except around the entire block instead of just specifically this function?

simtools/layout/geo_coordinates.py
```diff
@@ -0,0 +1,163 @@
...
+            Semi-minor axis of ellipsoid in meters.
+
+        Returns
+        -------
+        float
+            Ellipsoid radius at given latitude.
+
+        """
+
+        _lat_rad = np.deg2rad(latitude)
```
orelgueta, line 154: I suggest to force the latitude be in deg units at the argument level (with a decorator) and then do the conversion using astropy units, it would be safer. However, it's far from critical so I won't insist on it.

tests/unit_tests/layout/test_geo_coordinates.py
```diff
@@ -0,0 +1,109 @@
...
+import warnings
+
+import pyproj
+import pytest
+
+from simtools.layout.geo_coordinates import GeoCoordinates
+from simtools.layout.telescope_position import TelescopePosition
+
+
+def LaPalma_reference_point():
```
orelgueta, line 12: Why not use fixtures for these like we usually do?

tests/unit_tests/layout/test_geo_coordinates.py
```diff
@@ -0,0 +1,109 @@
...
+    _coord = GeoCoordinates()
+
+    ref_equator = TelescopePosition(name="Equator")
+    ref_equator.set_coordinates("mercator", 0.0, 0.0, 0.0)
+    scale_factor_equator = _coord._coordinate_scale_factor(ref_equator)
+    assert scale_factor_equator == pytest.approx(1.0, rel=1.0e-4)
+
+    ref_lapalma = LaPalma_reference_point()
+    scale_factor_lapalma = _coord._coordinate_scale_factor(ref_lapalma)
+    assert scale_factor_lapalma == pytest.approx(1.000341, rel=1.0e-4)
```
orelgueta, line 83: The relative accuracy of `1.0e-4` isn't enough here. For example, if you provide the Paranal value and compare it to the LaPalma value with that accuracy, the test will still pass. I suggest you change it to `1.0e-6` to at least fit with the last digit you provide in the test (might be able to go even higher to `1.0e-7`, but not sure if it's necessary).

tests/unit_tests/layout/test_geo_coordinates.py
```diff
@@ -0,0 +1,109 @@
...
+    _coord = GeoCoordinates()
+
+    _semi_major_axis = 6378137.0
+    _semi_minor_axis = 6356752.314245179
+
+    radius_equator = _coord._geocentric_radius(0.0, _semi_major_axis, _semi_minor_axis)
+    assert radius_equator == pytest.approx(_semi_major_axis, rel=1e-3)
+
+    radius_pole = _coord._geocentric_radius(90.0, _semi_major_axis, _semi_minor_axis)
+    assert radius_pole == pytest.approx(_semi_minor_axis, rel=1e-3)
```
orelgueta, line 100: Considering the point above about the relative accuracy, could you please recheck the value of `1.0e-3` here? It's harder for me to check since in this case I need to checkout the branch.

orelgueta (commented):
A few minor comments, one actual change required in the tests.

simtools/layout/geo_coordinates.py
GernotMaier: Yes, EPSG:4326 is the lat/long coordinate system and the same everywhere on the globe.

simtools/layout/geo_coordinates.py
GernotMaier: Good point. I've replace the if-statement by a try..expect pyproj.exceptions.CRSError: statement.

simtools/layout/geo_coordinates.py
GernotMaier: Thanks for the comment, this was important:

- raise now an AttributeError when reference point is None
- remove the second try..except, as we are pulling the WGS84 Geodetics from pyproj (depends on no input parameters and therefore should always work)

simtools/layout/geo_coordinates.py
GernotMaier: As mentioned in the chat, this is apparently not necessary, as numpy.deg2rad returns the input value if it is already given in u.rad.

tests/unit_tests/layout/test_geo_coordinates.py
GernotMaier: Conscious decision. Fixtures are making in this case simple code much more complicated.

tests/unit_tests/layout/test_geo_coordinates.py
GernotMaier: Very good point. Changed this now and put as well the test values in full accuracy into the code.

tests/unit_tests/layout/test_geo_coordinates.py
GernotMaier: Yes and no. For the tests listed, the accuracy was sufficient. But it is logical to increase the precision (as we have done everywhere). This requires also to give the lat of the side in sufficiently high precision.

GernotMaier:
I think I've addressed everything. Let me know if the pytest.fixture decorators are important (not sure why we use them inside a single testing module).

simtools/layout/geo_coordinates.py
GernotMaier: I've reverted the changes back to simply return 'None' in case of `nan` inputs. This is the expected behavior of the function and there is not need for an exception.

tests/unit_tests/layout/test_geo_coordinates.py
```diff
@@ -0,0 +1,118 @@
...
+    }
+    # ignore warnings from pyproj that te crs.to_dict()
+    # method is not sufficient to fully describe a
+    # coordinate system - this is not relevant for the
+    # simple comparison of the crs parameters
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        lapalma_dict = lapalma_crs.to_dict()
+    assert lapalma_dict["proj"] == _crs_test_dict["proj"]
+    assert lapalma_dict["k"] == pytest.approx(_crs_test_dict["k"], rel=1.0e-4)
```
orelgueta, line 69: I think we missed this low accuracy requirement as well. Should probably change to `1e-6` as well.

orelgueta (commented):
Last place where I think we missed the accuracy thing. If tests pass and you agree, I would take a look into converting to fixtures myself (since I tend to think it makes sense to do so).

orelgueta (approved):
All good now, thanks!

GernotMaier:
Thanks you @orelgueta

Issue #623, opened 2023-09-29
Default constructor layout = LayoutArray() not working.


GernotMaier:
Trying to write unit tests for `LayoutArray` using a simple constructor:

```
layout = LayoutArray()
```

This gives a TypError in this [line in IOHandler](https://github.com/gammasim/simtools/blob/f9ce4ad946f3bf5451c2cf5d82dce6aac54025b4/simtools/io_handler.py#L144), as `data_path` is None.

Requires 
- review behavior of the iohandler when no paths are set.
- review and update tests for LayoutArray

VictorBarbosaMartins:
I was wondering, should we actually allow an empty definition `layout = LayoutArray()` ? The empty definition would serve no purpose I think

PR #624, opened 2023-09-29
Ensure latitude calculation required for local coordinate calculation

GernotMaier:
Telescope position files might have the array centre defined in UTM coordinates only (this is actually the default).

This PR ensures that the latitude is calculated before the definition of the CORSIKA / local coordinate system. This allows to calculate the height and latitude dependent scale factor correctly.

Changed also the default for the centre altitude to `np.nan` (from 0 m). Ensures that this value is set and we do not use by accident the sea level value.

Tests are not entirely complete and require to solve issue #623 . Both layout_array.py and telescope_position.py have a test coverage of 80-85% only, this should be improved in a future pull request.

### DIFF
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -306,15 +306,9 @@ class LayoutArray:
         self._array_center.set_coordinates("corsika", 0.0 * u.m, 0.0 * u.m, 0.0 * u.m)
         self._set_array_center_mercator(center_dict)
         self._set_array_center_utm(center_dict)
-        try:
-            self._array_center.set_altitude(u.Quantity(center_dict.get("center_alt", np.nan * u.m)))
-        except TypeError:
-            pass
-        try:
-            _name = center_dict.get("array_name")
-            self.name = _name if _name is not None else self.name
-        except KeyError:
-            pass
+        self._array_center.set_altitude(u.Quantity(center_dict.get("center_alt", np.nan * u.m)))
+        _name = center_dict.get("array_name")
+        self.name = _name if _name is not None else self.name
 
         self._array_center.convert_all(
             crs_local=self.geo_coordinates.crs_local(self._array_center),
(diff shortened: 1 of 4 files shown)


simtools/layout/layout_array.py
```diff
@@ -297,11 +299,35 @@ def _initialize_coordinate_systems(self, center_dict=None):
...
+        center_dict = {} if center_dict is None else center_dict
+
         self._array_center = TelescopePosition()
         self._array_center.name = "array_center"
         self._array_center.set_coordinates("corsika", 0.0 * u.m, 0.0 * u.m, 0.0 * u.m)
+        self._set_array_center_mercator(center_dict)
+        self._set_array_center_utm(center_dict)
+        try:
+            self._array_center.set_altitude(u.Quantity(center_dict.get("center_alt", np.nan * u.m)))
+        except TypeError:
```
orelgueta, line 311: Are you sure there's a case where you get TypeError? Everything here should be defined so I don't see how it would happen (and you didn't write a test for it).

simtools/layout/layout_array.py
```diff
@@ -297,11 +299,35 @@ def _initialize_coordinate_systems(self, center_dict=None):
...
+            pass
+        try:
+            _name = center_dict.get("array_name")
+            self.name = _name if _name is not None else self.name
+        except KeyError:
```
orelgueta, line 316: Don't think you can get a KeyError when using `get`. It will just return `None` without raising an exception so the `try` is not necessary (no test for it either).

orelgueta (commented):
I think there are two unnecessary try/excepts, otherwise looks good.

simtools/layout/layout_array.py
GernotMaier: Yes, correct!

orelgueta (approved):

PR #625, opened 2023-10-02
Introduce more telescope classes and improve handling of corsika telescope height

GernotMaier:
(note a telescope class in simtools is what everyone else thinks is a telescope type)

This pull request:

- adds **simply** another telescope class (e.g., required by prod6, where MAGIC is included)
- improved handling of names of telescope class (allow for telescope classes >3 character length)
- ~~raises an error if `simtools/applications/print_array_elements.py` is called with the option `--use_corsika_telescope_height` and no CORSIKA sphere radius for the given telescope class is found~~
- adds and adjust unit tests to take these changes into account

### DIFF
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -389,7 +389,7 @@ class LayoutArray:
         Returns
         -------
         astropy.Quantity
-            Telescope sphere center value
+            Telescope sphere center value (0.0*u.m if sphere center is not defined).
 
         Raises
         ------
--- a/simtools/layout/layout_array.py
+++ b/simtools/layout/layout_array.py
@@ -10,7 +10,6 @@ from simtools.layout.geo_coordinates import GeoCoordinates
 from simtools.layout.telescope_position import TelescopePosition
 from simtools.utils import names
 from simtools.utils.general import collect_data_from_yaml_or_dict
-from simtools.utils.names import all_telescope_class_names
 
 __all__ = ["InvalidTelescopeListFile", "LayoutArray"]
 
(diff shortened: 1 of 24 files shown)


GernotMaier:
This is important and comes up due to the failure of the integration tests: we want the tool to be able to convert also non-telescope array elements (e.g., the position of the weather station). These elements do not have a sphere_center and sphere_radius defined.

Therefore: layout_array:get_corsika_sphere_center() is now issuing a warning if there is no sphere_center defined and return `0.0 u.m` in this case (and not an Error).

GernotMaier:
@orelgueta - note that some of the changes are form the legend handler updates and you can simply click them away.

simtools/layout/layout_array.py
```diff
@@ -396,14 +389,24 @@ def _get_corsika_sphere_center(self, tel_name):
         Returns
         -------
         astropy.Quantity
-            Telescope sphere center value (0.0*u.m if sphere center is not defined).
+            Telescope sphere center value
```
orelgueta, line 392: Why did you remove the `(0.0*u.m if sphere center is not defined)` if it's still true?

tests/unit_tests/layout/test_layout_array.py
```diff
@@ -39,7 +39,7 @@ def south_layout_center_data_dict():
 
 @pytest.fixture
 def layout_array_north_four_LST_instance(
-    north_layout_center_data_dict, manual_corsika_dict_north, io_handler, db_config
+    north_layout_center_data_dict, manual_corsika_dict_north, db_config
```
orelgueta, line 42: I think we sometimes pass the `io_handler` fixture because it sets the temporary output directories. Did you check that after removing this argument, no files/output is written to the local directory?

orelgueta (commented):
Two very minor comments.

tests/unit_tests/layout/test_layout_array.py
GernotMaier: Just checked this again and no files are written to the local directories for these tests.

simtools/layout/layout_array.py
GernotMaier: I found it too detailed, but looking at it again I can see as well reasons to keep it (e.g., this might be important when reading the sphinx-generated documentation). So I add it again.

orelgueta (approved):

PR #629, opened 2023-10-06
Add legend handlers and colours for HESS/MAGIC/VERTIAS

### LINKED ISSUE #626: Telescope classes and legend handler
The legend_handles have trouble accepting additional telescope types and I need help in understanding how this can be improved.

Especially for PR 625, where we fail in 

```
   radius_dict = [12.5, 9.15, 7.15, 3]
    for step, tel_type in enumerate(names.all_telescope_class_names):
        assert tel_handler.radius_dict[tel_type] == pytest.approx(radius_dict[step], 1.0e-3)
        assert tel_handler.colors_dict[tel_type] == colors[step]
```
I think this even has the order of the definition of telescopes class in names.py encoded into the test?

orelgueta:
Tests were also modified accordingly.
Fixes #626.

### DIFF
--- a/simtools/visualization/legend_handlers.py
+++ b/simtools/visualization/legend_handlers.py
@@ -5,7 +5,7 @@ import numpy as np
 from simtools.corsika.corsika_config import CorsikaConfig
 from simtools.io_handler import IOHandler
 from simtools.utils import names
-from simtools.utils.names import lst, mst, sct, sst
+from simtools.utils.names import hess, lst, magic, mst, sct, sst, veritas
 
 __all__ = [
     "EdgePixelObject",
@@ -50,6 +50,9 @@ class TelescopeHandler(object):
             mst: "dodgerblue",
             sct: "black",
             sst: "darkgreen",
+            hess: "grey",
+            magic: "grey",
+            veritas: "grey",
         }
         for key, value in corsika_info["corsika_sphere_radius"].items():
             self.radius_dict[key] = value["value"]
@@ -81,6 +84,18 @@ class SCTObject(object):
 
 class SSTObject(object):
     """SST Object."""
+
+
+class HESSObject(object):
+    """HESS Object."""
+
+
+class MAGICObject(object):
+    """MAGIC Object."""
+
+
+class VERITASObject(object):
+    """VERITAS Object."""
 
 
 class MeanRadiusOuterEdgeObject(object):
@@ -288,7 +303,7 @@ class SSTHandler(TelescopeHandler):
         return patch
 
 
-class SCTHandler(object):
+class SCTHandler(TelescopeHandler):
     """
     Legend handler class to plot a representation of an SCT in an array layout.
     """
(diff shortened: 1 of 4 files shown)


simtools/visualization/legend_handlers.py
```diff
@@ -5,7 +5,7 @@
 from simtools.corsika.corsika_config import CorsikaConfig
 from simtools.io_handler import IOHandler
 from simtools.utils import names
-from simtools.utils.names import lst, mst, sct, sst
+from simtools.utils.names import hess, lst, magic, mst, sct, sst, veritas
```
GernotMaier, line 8: Really thing that this is not good. Could we open an issue to rethink the handling of telescope types in names.py? Best to decouple this issue from this pull request.

simtools/visualization/legend_handlers.py
```diff
@@ -50,6 +50,9 @@ def __init__(self):
             mst: "dodgerblue",
             sct: "black",
             sst: "darkgreen",
+            hess: "grey",
```
GernotMaier, line 53: Here and in the following: there is no way to have an "OtherTelescopeObject" and have them all in grey?

simtools/visualization/legend_handlers.py
```diff
@@ -308,6 +323,78 @@ def legend_artist(self, legend, orig_handle, fontsize, handlebox):
         return patch
 
 
+class HESSHandler(TelescopeHandler):
```
GernotMaier, line 326: Again, all of the following functions are exactly the same? There must be a better way (sorry to be annoying).

GernotMaier (approved):
I honestly don't like the amount of duplication, but I also think that legends are not of most important, so we have to balance the time invested. Merge it if there no clear and fast way to avoid duplication.

simtools/visualization/legend_handlers.py
orelgueta: We could probably modify this `import` to import the dictionary that contains the telescope names, but then we would need additional changes in many other places, which I don't think would make the code more "legible".

Specifically this implementation in `names.py` of the explicit names of the telescopes was done by @VictorBarbosaMartins. At the time he convinced me that it's the best solution, so I trust it is. Perhaps @VictorBarbosaMartins can explain better.

simtools/visualization/legend_handlers.py
orelgueta: I understand the wish to avoid duplication, but I think that in general it will not make a large difference. 

We have to explicitly implement MAGIC because we will actually plot MAGIC. So, we could avoid some duplication by defining "OtherTelescope" for VERITAS/HESS, but that will reduce only those two to one type and even that reduction is limited. We anyway will have to list explicitly VERITAS/HESS as keys, so in the end we will save only one function (`class HESSHandler(TelescopeHandler):`). 
Once again, I do not really think it's worth it.

simtools/visualization/legend_handlers.py
orelgueta: See reply above

simtools/visualization/legend_handlers.py
GernotMaier: OK, I think this is fine for this PR and we should merge.

Do you want to open an issue that we revisit this and optimise it? I think this cannot be the best approach, I don't understand we can't have a variable/list "telescope_types" and simply loop? 

As said, let's decouple this from this PR so that we can move forward.

orelgueta:
OK, opened a new issue to revisit this point. Merging this PR.

simtools/visualization/legend_handlers.py
VictorBarbosaMartins: > We could probably modify this `import` to import the dictionary that contains the telescope names, but then we would need additional changes in many other places, which I don't think would make the code more "legible".
> 
> Specifically this implementation in `names.py` of the explicit names of the telescopes was done by @VictorBarbosaMartins. At the time he convinced me that it's the best solution, so I trust it is. Perhaps @VictorBarbosaMartins can explain better.

This was the solution we found to avoid writing the telescope names as strings throughout the code, after @GernotMaier's commented on the previous implementation. This way, the telescope names are defined once, imported and used as an object instead of a string, which is more elegant and centralized. A loop through all the telescope name types will not always serve our purposes, because in several situations we do need to name the telescope type explicitly, and in that case we thought it is better to name it through the object `mst` instead of the non-centralized string `"mst"`. Please feel free to rethink it and make other suggestions.

PR #636, opened 2023-10-16
This is a tiny PR correcting some very minor typos

GernotMaier:
(really not important)

For the command line parameters: I don't think we can say it is a boolean if it is simply set by a parameter using `action="store_true"` (e.g., by `--pdf`). For a boolean, I would expect to use `--pdf=True`.
(really very minor)


VictorBarbosaMartins (approved):
That is true. Probably reminiscence from previous implementation. Thanks for correcting it.

PR #639, opened 2023-10-16
Switch to alma container

### LINKED ISSUE #621: Switch to using almalinux
We agreed that in general it would be good if our containers would be based on almalinux 9 which is the OS expected to replace SL7 on the grid. In addition, I suspect that if we manage to do the switch, we will reduce the size of the container significantly. Conservatively, I would estimate a reduction of 50%.

The process of compiling sim_telarray in this container is not smooth, so I suggest to use this issue to keep track of our progress until we get it to work. (Particularly today I had a call in which I could play around and make some progress.)

I started from the base almalinux image, i.e., `almalinux`. I then had to install the following packages to get the compilation started 
```
yum install csh perl which gcc-fortran
```
I suspect the majority of these can be removed after compilation is done to reduce the size of the container further.

At this stage there's an error in compilation regarding multiple definitions in `out/EventIO.lo` which I will try to figure out at a later stage, but I think we are making progress.

orelgueta:
In the end, after installing all of the stuff for simtools, the container is unfortunately 1.7 GB (the simtools ubuntu based container is 1.9 GB). So we only save 200 MB with this change, which is not as much as I had hoped since the minimal almalinux container is about 1 GB smaller than the ubuntu one we've been using. 
Either way, it's a necessary change probably, so that we are in line with the production environment.

My local tests all worked, hopefully no one runs into any other changes.

Fixes #621.

### DIFF
--- a/docker/Dockerfile-dev
+++ b/docker/Dockerfile-dev
@@ -2,16 +2,17 @@ FROM ghcr.io/gammasim/simtools-corsika_sim_telarray
 WORKDIR /workdir
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y \
+RUN microdnf update -y && microdnf install -y \
     git \
     python3-pip \
-    python3-venv \
+    python3-devel \
     wget && \
-    apt-get clean && rm -rf /var/lib/apt/lists/*
+    microdnf clean all
 
 RUN wget --quiet https://raw.githubusercontent.com/gammasim/simtools/main/pyproject.toml && \
     python3 -m venv env && \
     source env/bin/activate && \
+    pip install -U pip && \
     pip install toml-to-requirements && \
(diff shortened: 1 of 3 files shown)


GernotMaier (approved):
All good. I've checked the container building locally and it works.

Also looked at the container size and it seems hard to reduce - the 880 MB of python is mostly numpy and llvmlite (half of it). This makes almost 400 MB.

orelgueta:
Thanks for looking into it. Yes, reducing the final container will be hard, but the reduction of the sim_telarray container will be useful in the future when we switch to use that externally.

I am running the workflow for building the sim_telarray container now. If that is successful, I will rerun the tests here and hopefully all will pass and we can merge.

orelgueta:
All passed, merging. Don't be surprised by the different look and feel of the new container when you switch to it.

PR #641, opened 2023-10-17
Simulator corsika parameter file

### LINKED ISSUE #640: Allow simulator.py to get as input the corsika_parameters.yml file
The current implementations of [simulator.py](https://github.com/gammasim/simtools/blob/main/simtools/simulator.py) uses the corsika parameter file hard coded from [corsika_parameters.yml]([corsika_parameters.yml](https://github.com/gammasim/simtools/blob/main/data/parameters/corsika_parameters.yml)). However, to be able to pass a configuration with a different atmospheric absorption file, we need to include the number of the atmosphere or the atmospheric absorption file in the "IACT_TUNING_PARAMETERS" key inside `corsika_parameters.yml`.

Therefore, I propose allowing the user to pass a different file with a different corsika parameter configuration through the `self._corsika_parameters_file` attribute of the Simulator() class in simulator.py. With that we will be able to simulate with different atmospheres.

Do you agree with the solution @GernotMaier , @orelgueta ?

VictorBarbosaMartins:
Closes #640.

Implemented the option to pass the corsika parameter file to the old production application. This allows us to run simulations for two different atmosphere (more complete description in the related issue).

### DIFF
--- a/simtools/applications/production.py
+++ b/simtools/applications/production.py
@@ -108,6 +108,13 @@ def _parse(description=None):
         required=True,
     )
     config.parser.add_argument(
+        "--corsika_parameters_file",
+        help="Yaml file with corsika parameters.",
+        type=str,
+        required=False,
+        default=None,
+    )
+    config.parser.add_argument(
         "--task",
         help=(
             "What task to execute. Options: "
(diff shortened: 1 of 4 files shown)


VictorBarbosaMartins:
Ok, closing this PR, as it is not really necessary. We can pass the `corsika_parameters_file` already inside the main configuration file as mentioned here https://github.com/gammasim/simtools/issues/640#issuecomment-1766533670

PR #650, opened 2023-10-20
change to if `self.combined_hists is None:`

### LINKED ISSUE #649: Bug with file list to plot-simtel-histograms application
Running ```simtools-plot-simtel-histograms --file_lists file_list.txt --figure_name histograms_simtel``` leads to the following error:
```
  File "/workdir/env/bin/simtools-plot-simtel-histograms", line 8, in <module>
    sys.exit(main())
  File "/workdir/external/simtools/simtools/applications/plot_simtel_histograms.py", line 77, in main
    number_of_hists = [sh.number_of_histograms for sh in simtel_histograms]
  File "/workdir/external/simtools/simtools/applications/plot_simtel_histograms.py", line 77, in <listcomp>
    number_of_hists = [sh.number_of_histograms for sh in simtel_histograms]
  File "/workdir/external/simtools/simtools/simtel/simtel_histograms.py", line 58, in number_of_histograms
    return len(self.combined_hists)
TypeError: object of type 'NoneType' has no len()
```

The problem is that https://github.com/gammasim/simtools/blob/d83f7eab25ce52da559e14e561039c09b845d1cf/simtools/simtel/simtel_histograms.py#L73 is never activated because the class instance always has "combined_hists" as attribute, once it is defined in the __init__ of the class.

Solution: Change https://github.com/gammasim/simtools/blob/d83f7eab25ce52da559e14e561039c09b845d1cf/simtools/simtel/simtel_histograms.py#L72 to check whether it is None instead.

VictorBarbosaMartins:
Closes #649

This simple solution solves the problem described in #649. Probably at some point we decided to initialize the attribute in the class init and forgot about the rest. I tested and it is working. Following PRs will continue the work with the application.

### DIFF
--- a/simtools/simtel/simtel_histograms.py
+++ b/simtools/simtel/simtel_histograms.py
@@ -51,7 +51,7 @@ class SimtelHistograms:
     @property
     def number_of_histograms(self):
         """Returns number of histograms."""
-        if not hasattr(self, "combined_hists"):
+        if self.combined_hists is None:
             self._combine_histogram_files()
         return len(self.combined_hists)
 
@@ -69,7 +69,7 @@ class SimtelHistograms:
         str
             Histogram title.
         """
-        if not hasattr(self, "combined_hists"):
+        if self.combined_hists is None:
             self._combine_histogram_files()
         return self.combined_hists[i_hist]["title"]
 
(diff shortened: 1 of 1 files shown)


tobiaskleiner (approved):
Looks good!

PR #656, opened 2023-11-01
Allow specifying an external nsb spectrum for the camera efficiency application

orelgueta:
For the purpose of calculating the NSB rate of each configuration planned in Prod6, we need to allow specifying the zenith, azimuth and NSB spectrum file when running `validate_camera_efficiency.py`. 

This PR adds this functionality, saves the summary of the results in a file and changes some file names to follow the new naming scheme.


simtools/applications/validate_camera_efficiency.py
```diff
@@ -75,6 +84,34 @@ def _parse(label):
...
+        help="Zenith angle in degrees (between 0 and 180).",
+        type=CommandLineParser.zenith_angle,
+        default=20,
+        required=False,
+    )
+    config.parser.add_argument(
+        "--nsb_spectrum",
+        # FIXME - describe format of the sim_telarray file or better, use a standard
+        # format and convert on the fly to sim_telarray format
+        help="File with NSB spectrum to use (following sim_telarray required format).",
```
VictorBarbosaMartins, line 110: Both here and in the docstring, it would be good to give more details on the format for non-simtel_array experts. Also, the format in simtel_array could eventually change and then this would not fit here anymore.

simtools/applications/validate_camera_efficiency.py
```diff
@@ -23,6 +23,14 @@
...
     model_version (str, optional)
         Model version (default='Current')
+    zenith_angle (float, optional)
+        Zenith angle in degrees (between 0 and 180) (default=20).
+    azimuth_angle (float, optional)
+        Telescope pointing direction in azimuth. It can be in degrees between 0 and 360 or
+        one of north, south, east or west (case insensitive). Note that North is 0 degrees
+        and the azimuth grows clockwise, so East is 90 degrees (default=0).
+    nsb_spectrum (str, optional)
+        File with NSB spectrum to use (following sim_telarray required format).
```
VictorBarbosaMartins, line 33: I would not spare on the explanation of this file. Where can we find an example? Could you give an example of it and/or its format?

simtools/camera_efficiency.py
```diff
@@ -174,6 +177,7 @@ def simulate(self, force=False):
             file_simtel=self._file_simtel,
             file_log=self._file_log,
             label=self.label,
+            nsb_spectrum=self.config.nsb_spectrum,
```
VictorBarbosaMartins, line 180: you added the nsb_spectrum here and I see that the zenith angle is already there. What about the azimuth? Doesn't it play a role in the simulation?

simtools/camera_efficiency.py
```diff
@@ -262,41 +266,54 @@ def analyze(self, export=True, force=False):
...
 
+    def results_summary(self):
+        """
+        Print a summary of the results.
+        Include a header for the zenith/azimuth settings and the NSB spectrum file which was used.
+        The summary includes the various CTAO requirements and the final expected NSB pixel rate.
+        """
+        summary = (
+            f"Results summary for {self._telescope_model.name} at "
+            f"zenith={self.config.zenith_angle:.1f}, "
```
VictorBarbosaMartins, line 285: zenith and azimuth are not appearing with units in the summary message. As the reflectivity and efficiency have no units, I would suggest adding the units to the angles

simtools/camera_efficiency.py
```diff
@@ -262,41 +266,54 @@ def analyze(self, export=True, force=False):
...
 
     # END of analyze
 
+    def results_summary(self):
+        """
+        Print a summary of the results.
+        Include a header for the zenith/azimuth settings and the NSB spectrum file which was used.
+        The summary includes the various CTAO requirements and the final expected NSB pixel rate.
+        """
+        summary = (
```
VictorBarbosaMartins, line 283: I was considering the option of saving the time stamp as well in the summary. Sometimes it is helpful to know when a report/summary was generated. Would you also consider printing out the version of the tool (testeff or sim_telarray) and version of simtools?

simtools/camera_efficiency.py
```diff
@@ -262,41 +266,54 @@ def analyze(self, export=True, force=False):
...
+            "(A-PERF-2025/B-TEL-0090): "
+            f"{self.calc_tot_efficiency(self.calc_tel_efficiency()):.4f}\n"
+            "Expected NSB pixel rate for the reference NSB: "
+            f"{self.calc_nsb_rate():.4f} [p.e./ns]\n"
+        )
+
+        return summary
+
     def export_results(self):
         """Export results to a csv file."""
```
VictorBarbosaMartins, line 303: We should change here as well. First, it should be ecsv and not csv, and we should include also the summary

simtools/simtel/simtel_runner_camera_efficiency.py
```diff
@@ -102,6 +104,8 @@ def _make_run_command(self, **kwargs):  # pylint: disable=unused-argument
             )
 
         command = str(self._simtel_source_path.joinpath("sim_telarray/bin/testeff"))
+        if self.nsb_spectrum is not None:
+            command += f" -fnsb {self.nsb_spectrum}"
```
VictorBarbosaMartins, line 108: just wondering, how did testeff handle it before, when there was no nsb file? does it have a default file somewhere?

simtools/utils/names.py
```diff
@@ -724,6 +733,8 @@ def camera_efficiency_simtel_file_name(site, telescope_model_name, zenith_angle,
         LST-1, MST-FlashCam-D, ...
     zenith_angle: float
         Zenith angle (deg).
+    azimuth_angle: float
```
VictorBarbosaMartins, line 736: I suggest using astropy units everywhere for zenith and azimuth. it is less prone to mistakes. Or even better make use of the `CommandLineParser.zenith_angle` and `CommandLineParser.azimuth_angle` as you did for the other file

tests/unit_tests/utils/test_names.py
```diff
@@ -108,3 +108,69 @@ def test_get_telescope_type():
...
+            site, telescope_model_name, zenith_angle, azimuth_angle, label
+        )
+        == "camera-efficiency-North-MST-FlashCam-D-za040deg_azm000deg_test.dat"
+    )
+    assert (
+        names.camera_efficiency_log_file_name(
+            site, telescope_model_name, zenith_angle, azimuth_angle, label
+        )
+        == "camera-efficiency-North-MST-FlashCam-D-za040deg_azm000deg_test.log"
+    )
```
VictorBarbosaMartins, line 164: I dont think it is necessary to repeat the test with other arguments

simtools/applications/validate_camera_efficiency.py
```diff
@@ -23,6 +23,14 @@
         Telescope model name (e.g. LST-1, SST-D, ...)
     model_version (str, optional)
         Model version (default='Current')
+    zenith_angle (float, optional)
```
VictorBarbosaMartins, line 26: I am commenting here but I want to address the application as a whole:
When running the application without arguments, i.e. `simtools-validate-camera-efficiency`, there is an error:
```
  File "/workdir/env/bin/simtools-validate-camera-efficiency", line 8, in <module>
    sys.exit(main())
  File "/workdir/external/simtools/simtools/applications/validate_camera_efficiency.py", line 129, in main
    tel_model = TelescopeModel(
  File "/workdir/external/simtools/simtools/model/telescope_model.py", line 59, in __init__
    self.site = names.validate_site_name(site)
  File "/workdir/external/simtools/simtools/utils/names.py", line 230, in validate_site_name
    return validate_name(name, all_site_names)
  File "/workdir/external/simtools/simtools/utils/names.py", line 274, in validate_name
    raise ValueError(msg)
ValueError: Invalid name None
```

Instead of an error, I would expect the help message such as this one:

```
(env) simtools-print-array-elements
usage: print_array_elements [-h] --array_element_list ARRAY_ELEMENT_LIST
                            [--compact {corsika,utm,mercator}] [--export {corsika,utm,mercator}]
                            [--use_corsika_telescope_height]
                            [--select_assets SELECT_ASSETS [SELECT_ASSETS ...]]
                            [--data_path DATA_PATH] [--output_path OUTPUT_PATH]
                            [--use_plain_output_path] [--model_path MODEL_PATH]
                            [--simtel_path SIMTEL_PATH] [--config CONFIG] [--env_file ENV_FILE]
                            [--test] [--label LABEL] [--log_level LOG_LEVEL] [--version]
print_array_elements: error: the following arguments are required: --array_element_list
```
Could you please have a look on that (or open an issue to solve it later)?

VictorBarbosaMartins (commented):
I am pushing only minor changes in the docstring (please undo if you dont like them).

For your changes, I made some comments. All in all it is very good, thanks for it! I will wait for your answers to approve it (no rush).

P.S.: I came across a very interesting behaviour that I didnt know about. When testing the application, which takes `--azimuth_angle` as an argument, I passed `--azimuth`  and it still worked (same for zenith). In a quick look in the code I could not find where we have this happening. How do we actually do that ?

VictorBarbosaMartins:
Ah, another comment that I forgot is to add an integration test where we pass the NSB file as argument

simtools/applications/validate_camera_efficiency.py
orelgueta: That's a good point that I was too lazy to deal with before. Now I implemented a function that converts to the sim_telarray format so that the user just has to provide two columns as expected (but still backward compatible). The description was also improved.

simtools/camera_efficiency.py
orelgueta: Added units (manually, but they are converted to degrees so it's fine)

simtools/camera_efficiency.py
orelgueta: Those are all provenance information which should be collected "automatically" by the tools running the application (like CWL). I must say also that this summary was written only for something very specific. I think this function will change later to write an ecsv file with the results and additional metadata.

simtools/camera_efficiency.py
orelgueta: I fixed the csv --> ecsv.
I don't want to combine the summary with this table since this table is not really necessary once the summary is available. We save the table as some sort of provenance, but what is actually important is the summary.

simtools/utils/names.py
orelgueta: Changing everything to use astropy units for the file names seems a bit of overkill, no? We are also sure that these are degrees because the configurator converts to the degrees, regardless of the user input.

tests/unit_tests/utils/test_names.py
orelgueta: But it doesn't repeat the test with other arguments, each time a different function is called (with the same arguments).

simtools/applications/validate_camera_efficiency.py
orelgueta: The problem here is that if you do not provide a telescope and site, you get that error instead of the help message. I added a bit of code to protect from that. It's a bit ugly because I check the arguments are provided manually, but I don't know a different way to change an already defined argument to be `required`. Do you know @GernotMaier?

orelgueta:
Hi @VictorBarbosaMartins,

Thanks a lot for the review. I implemented all of the changes you requested (and then some). Could you please review again? 

> P.S.: I came across a very interesting behaviour that I didnt know about. When testing the application, which takes --azimuth_angle as an argument, I passed --azimuth and it still worked (same for zenith). In a quick look in the code I could not find where we have this happening. How do we actually do that ?

This is actually not something we do, but argparser does. It allows to be lazy and just provide part of the argument. In fact, you can run this application with just `--z 40 --a 30` and it will parse it as 40 degrees zenith angle and 30 degrees azimuth.

simtools/applications/validate_camera_efficiency.py
orelgueta: OK, so I guess we keep my ugly solution.

VictorBarbosaMartins: Yes, I think this solution is fine!

simtools/applications/validate_camera_efficiency.py
VictorBarbosaMartins: Thanks. That is good

simtools/utils/names.py
VictorBarbosaMartins: Hmm, not really. It avoids confusion with the units. I would at least like to see a fixture such as:
https://github.com/gammasim/simtools/blob/44cbe1ff973a5d6e2127c916baa03a8d7720938e/simtools/utils/geometry.py#L93

but you decide if it is not necessary in this case

tests/unit_tests/utils/test_names.py
VictorBarbosaMartins: Yes! apologies for this comment

simtools/simtel/simtel_runner_camera_efficiency.py
```diff
@@ -194,3 +197,31 @@ def _get_one_dim_distribution(self, two_dim_parameter, weighting_distribution_pa
...
         return one_dim_file
+
+    def _validate_or_fix_nsb_spectrum_file_format(self, nsb_spectrum_file):
+        """
+        Validate or fix the nsb spectrum file format.
+        The nsb spectrum file format required by sim_telarray has three columns:
+            wavelength (nm), ignored, NSB flux [1e9 * ph/m2/s/sr/nm],
+        where the second column is ignored by sim_telarray and the third is used for the NSB flux.
+        This function makes sure the file has at least three columns,
+        by copying the second column to the third.
```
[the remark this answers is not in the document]
orelgueta: But it actually explains why we need three columns just above it. That's what sim_telarray expects.
The duplication of the second column is safe in the sense that it just adds a column for sim_telarray to actually read/use. The column is added to a new file which is saved somewhere else, so the original user input remains untouched.

simtools/applications/validate_camera_efficiency.py
```diff
@@ -24,13 +24,17 @@
...
         one of north, south, east or west (case insensitive). Note that North is 0 degrees
-        and the azimuth grows clockwise, so East is 90 degrees (default=0).
+        and the azimuth grows clockwise, so East is 90 degrees.
     nsb_spectrum (str, optional)
-        File with NSB spectrum to use (following sim_telarray required format).
+        File with NSB spectrum to use for the efficiency simulation.
+        The expected format is two columns with wavelength in nm and
+        NSB flux with the units: [1e9 * ph/m2/s/sr/nm].
+        If the file has more than two columns, the first and third are used,
+        and the second is ignored.
```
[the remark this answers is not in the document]
orelgueta: Yes, it is due to sim_telarray. I added a tiny comment about it.

simtools/simtel/simtel_runner_camera_efficiency.py
```diff
@@ -49,9 +49,20 @@ def __init__(
...
+        return self._nsb_spectrum
+
+    @nsb_spectrum.setter
+    def nsb_spectrum(self, nsb_spectrum):
+        """Setter for nsb_spectrum"""
         if nsb_spectrum is not None:
-            self.nsb_spectrum = self._validate_or_fix_nsb_spectrum_file_format(nsb_spectrum)
+            self._nsb_spectrum = self._validate_or_fix_nsb_spectrum_file_format(nsb_spectrum)
+        else:
+            self._nsb_spectrum = None
```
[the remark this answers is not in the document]
orelgueta: No, it can and should be initialised to None if no input was given. If it is None, the option is not passed along to sim_telarray.

simtools/utils/names.py
orelgueta: It's not that simple of just adding this decorator because it means I also have to make sure I pass an astropy quantity and get the value each time it is used. I think it is overkill here for the reasons explained above.
These naming functions should move into the corresponding classes anyway actually, so maybe then we can do it.

PR #662, opened 2023-11-06
more info printed to logger in plot_simtel_histograms

### LINKED ISSUE #661: simtools-plot-simtel-histograms info messages not complete
When running `simtools-plot-simtel-histograms` we get only two print out messages:
```
INFO::base(l60)::__init__::Opening new file simtel-data/North/gamma/data/run000001_gamma_za020deg_azm000deg_North_TestLayout_test-production.simtel.zst
INFO::base(l93)::__init__::Found zstd compressed file
```
I propose we increase the amount of information, especially if the script was ran successfully.

VictorBarbosaMartins:
Closes #661.
Small pull request to solve #661. It should add some info messages to the logger of `plot_simtel_histograms`.

You can test it by running e.g. ```simtools-plot-simtel-histograms --hist_file_names tests/resources/run201_proton_za20deg_azm0deg_North_TestLayout_test-prod.simtel.zst --figure_name test_hist```

### DIFF
--- a/simtools/applications/plot_simtel_histograms.py
+++ b/simtools/applications/plot_simtel_histograms.py
@@ -68,7 +68,7 @@ def main():
     logger = logging.getLogger()
     logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
     n_lists = len(args_dict["hist_file_names"])
-
+    logger.info(f"Found {n_lists} hist_file_names. Opening them.")
     histogram_files = []
     for one_file in args_dict["hist_file_names"]:
         if Path(one_file).is_file():
@@ -86,6 +86,7 @@ def main():
             raise TypeError
 
     # Building SimtelHistograms
+    logger.info(f"Histograms will be produced for {histogram_files}.")
     simtel_histograms = SimtelHistograms(histogram_files)
 
     # Checking if it is needed to add the pdf extension to the file name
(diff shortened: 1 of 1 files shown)


simtools/applications/plot_simtel_histograms.py
```diff
@@ -68,7 +68,7 @@ def main():
     logger = logging.getLogger()
     logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
     n_lists = len(args_dict["hist_file_names"])
```
GernotMaier, line 70: Suggest to condense this into one sentence: `Opening {n_lists} histogram files` (don’t call it hist_file_names, this is hard to understand)

simtools/applications/plot_simtel_histograms.py
```diff
@@ -86,14 +86,15 @@ def main():
             raise TypeError
 
     # Building SimtelHistograms
+    logger.info(f"Histograms will be produced for {histogram_files}.")
```
GernotMaier, line 89: Tiny comment, but maybe we have consistent language. My suggestion is to use the same as in line 97: `Creating pdf file`; `Creating histograms for {histogram_files}`

(Avoiding philosophical discussions about the differences between `creating` and `producing` :-) )

GernotMaier (approved):
Tiny comments, otherwise good!

simtools/applications/plot_simtel_histograms.py
VictorBarbosaMartins: Good point. I updated it to `Opening {n_lists} files` as the files can also be text files with lists of histogram files.

PR #669, opened 2023-11-10
Fix deployment of github pages

### LINKED ISSUE #653: sphynx link not working
When attempting to open simpipe documentation from the link displayed in the main page, we get the following error:

404
File not found

We believe that the link is broken.

GernotMaier:
Github pages are not deployed, this is fixed by this pull request.

The origin of the problem was this deployment, which removed all pages: https://github.com/gammasim/simtools/commit/bb952355780627e469cdddca69b13f8a6f49d673

Several issues why this stopped working:

-  non-uncode characters are included in on the of the rst files.
- bugs in docstrings
- numpy mock import for sphinx leads to a segfault. I couldn't solve this and couldn't get any good help from the web. I therefore disabled numpy mock for now in the sphinx conf.py file
- including `generate_corsika_histograms` into `docs/source/applications.rst` leads to errors (to be fixed in a separate issue; I have removed it for now)

Success has been tested (https://github.com/gammasim/simtools/commits/gh-pages and https://gammasim.github.io/simtools/) 

closes #653

### DIFF
--- a/.github/workflows/CI-docs.yml
+++ b/.github/workflows/CI-docs.yml
@@ -30,6 +30,11 @@ jobs:
         shell: bash -l {0}
         run: |
           make -C docs/ html
+          exit_code=$?
+          if [ $exit_code -ne 5 ]; then
+            echo "Make command failed with exit code $exit_code"
+            exit $exit_code
+          fi
           touch docs/build/html/.nojekyll
 
       - name: Deploy to gihub pages
--- a/.github/workflows/CI-docs.yml
+++ b/.github/workflows/CI-docs.yml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Deploy to gihub pages
         # only run when PR is merged
-        if: github.event.pull_request.merged
+        # if: github.event.pull_request.merged
         uses: JamesIves/github-pages-deploy-action@v4
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/CI-docs.yml
```diff
@@ -30,15 +30,19 @@ jobs:
         shell: bash -l {0}
         run: |
           make -C docs/ html
+          exit_code=$?
+          if [ $exit_code -ne 0 ]; then
+            echo "Make command failed with exit code $exit_code"
```
VictorBarbosaMartins, line 35: I agree it is good to include this 'if statement' here. Is it also possible to make it a github test that would fail in case we get an error in the make command?

VictorBarbosaMartins (approved):
Looks good to me. tested locally the `html make` and it works fine. Thank you

.github/workflows/CI-docs.yml
GernotMaier: This is exactly what this if statement is doing. It fails if make exits with an error code. 

Have a look at https://github.com/gammasim/simtools/actions/runs/6823947507/job/18558879126 
I have put some non-ascii for this test into the code and the github action reported it.

PR #671, opened 2023-11-10
check for non-ascii characters in docs

GernotMaier:
Deployment of github pages fails with non-ascii characters in the rst files (see #669)

Add a workflow step to check RST files and exit if there are non-ascii characters.

### DIFF
--- a/.github/workflows/CI-docs.yml
+++ b/.github/workflows/CI-docs.yml
@@ -26,6 +26,14 @@ jobs:
           python --version
           pip install '.[doc]'
 
+      - name: Check for non-ASCII characters
+        run: |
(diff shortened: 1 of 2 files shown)


VictorBarbosaMartins:
Hi Gernot.  As I do not know where to check to see whether the workflow is working, could you please provide more information in the PR?

GernotMaier:
See the test here: https://github.com/gammasim/simtools/actions/runs/6823947507/job/18558879126

I have put non-ascii characters into one rst file and the code works as expected.

VictorBarbosaMartins (approved):
Ok, thanks! then this is very good.

PR #672, opened 2023-11-10
Layout array renaming

GernotMaier:
This is a follow up to #428 to change the term `LayoutArray` to `ArrayLayout` through the code (also `layout_array` to `array_layout`). 

There is zero functional change, this is search and replace only. So I think that no code review is required (as long as all tests pass).


rnda-docker.sh
orelgueta, line 1: Should this file be included in this PR (or in general)?

rnda.sh
orelgueta, line 1: Should this file be included in this PR (or in general)?

orelgueta:
Also, there are conflicts apparently, probably cause it took me so long to review this PR...

rnda-docker.sh
GernotMaier: No!!

rnda.sh
GernotMaier: No!!

GernotMaier:
Fixed both things - remove the accidentally committed shell scripts and merged main into this branch.

Issue #681, opened 2023-11-16
Applications are painfully slow to even start.


GernotMaier:
This is really starting to get an issue, as it slows down developing, testing, everything.

Running e.g.,
```
 python simtools/applications/validate_file_using_schema.py --help
```

Take >3 sec for me. This should just print help.

Running

```
python -X importtime simtools/applications/validate_file_using_schema.py --help
```
shows that it us due to extensive importing (see attachment).

In this example, it is e.g., not clear why ctapipe is imported.

[import.txt](https://github.com/gammasim/simtools/files/13376329/import.txt)

GernotMaier:
To do further, this looks like that this is almost exclusively due to ctapipe.

I suggest to remove anything related to ctapipe from simtools.utils.general and move it into a separate module. The `general` module is used a lot, but not very often for the hdf5 functionality.

@VictorBarbosaMartins - is it fine for you if I move this out of general.py?

GernotMaier:
PR #682 addresses most of this issue.

When we discuss the cleanup of general.py, we should address the remaining issues (e.g., moving the coordinate transformations somewhere else will remove the dependency on astropy.coordinates.errors (1/3 of import time). Rest is due to astropy.units, which is something we need.

VictorBarbosaMartins:
> This looks like that this is almost exclusively due to ctapipe.
> 
> I suggest to remove anything related to ctapipe from simtools.utils.general and move it into a separate module. The `general` module is used a lot, but not very often for the hdf5 functionality.
> 
> @VictorBarbosaMartins - is it fine for you if I move this out of general.py?

Hi Gernot. That is probably a good idea, Im fine with it

Issue #665, opened 2023-11-08
Review contents of general.py


GernotMaier:
[simtools/utils/general.py](simtools/utils/general.py) starts to become a collection of "unrelated" functions which will be difficult to maintain in future.

Suggest to review the contents and if necessary restructure.

VictorBarbosaMartins:
agree. should we add this to the maintenance period?

VictorBarbosaMartins:
Oh, I see it is already there

GernotMaier:
These are my suggestions:

- add docstring at the top explaining that this should be used for very generic functions only; it is used in many places and no slow imports should be introduced
- `file_has_text` is not used anywhere in simtools (except in its own unit tests). Suggest to move it into `tests/conf.py`, where we have functions used by tests only
- move `rotate` and and `convert_2D_to_radial_distr` into a new module (transformations.py?)
- `save_dict_to_file` saves a dict to yaml. This is used only once in the code (in corsika_histograms). We should decide if we use generic functions or the simple yaml.dump syntax.

(see PR #704)

VictorBarbosaMartins:
Agree with the first two bullet points. In regard to the third one, I would call such a module `geometrical_transformations` or `geometry`. I don't have a strong opinion on the `save_dict_to_file`. When I coded this, I was expecting that this could be useful elsewhere, but a single `yaml.dump` does the job as well. I suggest to always use the yaml module from `astropy.io.misc.yaml` (not sure we are consistently using the same yaml module)

GernotMaier:
Not sure about `geometry`, these are transformations, or?

VictorBarbosaMartins:
Just to avoid more than one person working on the same thing, I am starting working on this now.

PR #682, opened 2023-11-16
Improve startup time of applications

GernotMaier:
Applications are very slow in starting up, see issue #681 .

This is mostly related to imports and this pull request is start to optimize this:

- moved hd5-related functions out of utils/general.py into a separate module (io_hdf5)
- for clarity, moved io_handler and io_hdf5 into a new directory simtools.io_operations (most changes of this PR are related to the change of the import statement; note that I have in an earlier version this was simtools.io, but I realised that io is in conflict with the python io package)

This improves the speed of calling e.g., `python simtools/applications/validate_schema_files.py --help` by a factor of 4!


simtools/io_operations/io_hdf5.py
```diff
@@ -0,0 +1,112 @@
...
+
+__all__ = [
+    "fill_hdf5_table",
+    "read_hdf5",
+]
+
+_logger = logging.getLogger(__name__)
+
+
+def fill_hdf5_table(hist, x_bin_edges, y_bin_edges, x_label, y_label, meta_data):
```
VictorBarbosaMartins, line 19: It is good to have all the hdf5 functions together, but this function is not actually an io as it does not deal with input and output. Instead, if fills the table. Maybe we should call the module `hdf5_handler` instead?

VictorBarbosaMartins (approved):
Thanks for these updates, it is improving also in the direction of #665.
I left only one comment/suggestion.

simtools/io_operations/io_hdf5.py
GernotMaier: Thanks - this is a good suggestion. I've renamed `io_hdf5` to `hd5_handler`.

Waiting for the tests and will merge when successful.

PR #684, opened 2023-11-20
Linter updates

GernotMaier:
This add linters to the CI with the goal of linting every single file type in simtools:

- [x] python (flake8)
- [x] python (black)
- [x] python (isort)
- [x] python (pylint)
- [x] docker
- [x] github action
-  rst (remove this, as no good solution was found)
- [x] yaml
- [x] markdown
- [x] bash
- [x] pyproject.toml
- [x] CITATION.cff
- [x] check for non-ascii characters
- [x] natural language
- [x] .env files
- [x] copy and paste duplications 

Most of it is achieved by using the [super-lint github action](https://github.com/marketplace/actions/super-linter).

Missing files not covered by linters:

- ecsv
- LICENSE
- Makefile
- zst
- corsikaio

For pylint, I've removed the 500 lines of default configuration which we had in pyproject.toml. I think we should list only the non-default values (otherwise readability will be hard).

Note that the extensive linting results in a run time of 10-12 min of this workflow.


GernotMaier:
Do think this is ready for production - although the code requires some changes to pass this test (see PR #690)

GernotMaier:
Attached the output from pylint.

[pylint.txt](https://github.com/gammasim/simtools/files/13558757/pylint.txt)

I think we should fix all:

- missing-function-docstring
- inconsistent-return-statements
- simplifiable-if-expression
- consider-using-from-import
- raise-missing-from
- protected-access
- logging-format-interpolation
- unused-argument
- attribute-defined-outside-init
- unspecified-encoding
- raising-bad-type

Not sure:
- fixme - not sure I like statement like this in the code.
- too-many-lines - it is good to have shorter modules and we should consider if it is possible to split the long modules up.
- invalid-name is complaining a lot about using "1D" or "2D' in variables / function names. I know that this is common, but I suggest to simply replace it by "1d" and "2d" as nothing is lost by doing so.
- don't understand the `unbalanced-tuple-unpacking`
- I think we discussed in the past the issues in simtools/visualization/legend_handlers.py ?

I suggest to change pylint defaults for:

- too-many-branches: from 12 to 15?
- too-many-locals from 25 to 50?
- too-many-instance-attributes from 7 to 15
- for invalid-name: add all complaints from camera_efficiency to allowed list

Didn't look at the duplication statements (yet).

orelgueta:
> I think we should fix all:
> 
> * missing-function-docstring

Need to see if we can ignore tests here (everywhere else I agree).

> * inconsistent-return-statements
> * simplifiable-if-expression
> * consider-using-from-import
> * raise-missing-from

This is a bit annoying and the community isn't in complete agreement about it if I remember correctly. I don't mind trying to correct them though.


> * protected-access

Here as well, need to see that we can skip the tests, cause it is inevitable there.


> * logging-format-interpolation

No, this we agreed to ignore in the past and with very good reason (which I cannot remember right now).


> * unused-argument
> * attribute-defined-outside-init
> * unspecified-encoding
> * raising-bad-type
> 
> Not sure:
> 
> * fixme - not sure I like statement like this in the code.

I don't mind them.

> * too-many-lines - it is good to have shorter modules and we should consider if it is possible to split the long modules up.

This can be very hard in some modules, so I suggest to fix where possible and add an ignore statement where we can't.


> * invalid-name is complaining a lot about using "1D" or "2D' in variables / function names. I know that this is common, but I suggest to simply replace it by "1d" and "2d" as nothing is lost by doing so.

Yes, let's change.

> * don't understand the `unbalanced-tuple-unpacking`

Me neither.

> * I think we discussed in the past the issues in simtools/visualization/legend_handlers.py ?

Yes, decided to ignore them.


> 
> I suggest to change pylint defaults for:
> 
> * too-many-branches: from 12 to 15?
> * too-many-locals from 25 to 50?
> * too-many-instance-attributes from 7 to 15
> * for invalid-name: add all complaints from camera_efficiency to allowed list

Agree (about all and in particular about camera_efficiency).


> 
> Didn't look at the duplication statements (yet).

VictorBarbosaMartins:
I don't mind any of these changes, as far as I could understand them.

GernotMaier:
I am working on this now - so please do not change / commit to this branch.

GernotMaier:
Progress here and I've reduced the number of issues significantly, see this output:

https://github.com/gammasim/simtools/actions/runs/7112049810/job/19361229437?pr=684

@orelgueta , @VictorBarbosaMartins - could you check the remaining ones? I am not 100% familiar with these bits of the code, so maybe it is easier for you.

Important:

- unit tests fail for `tests/unit_tests/corsika/test_corsika_histograms.py`. They fail with different error messages for each `pytest -n auto` run, which indicates a cross-dependency between unit tests. I still look into it, but it is quite difficult.

Please also check the pylint section in pyproject.toml. I do find some of the changes are very very large compared to the default values. Modules with almost 50 public functions or classes with 50 class variables are hard to maintain; over a 1000 lines for a module is also too much - but this is my opinion.

I am not touching anything related to pylint now, but have another look into the unit test. Please continue working on the issues listed above.

GernotMaier:
@VictorBarbosaMartins - could you have a look at [this test](https://github.com/gammasim/simtools/blob/f7994d31472bb17283dcdb2cab22fbc22b784d13/tests/unit_tests/corsika/test_corsika_histograms.py#L281) - I am not sure I understand what is going on. It reads `tests/resources/tel_output_10GeV-2-gamma-20deg-CTAO-South.corsikaio` and seems to find some histograms. Is this correct?

VictorBarbosaMartins:
Sure, I am going to have a look at this now

VictorBarbosaMartins:
> this test

In this test, we open the Corsika output file but as we do not use the `set_histograms` method, no histogram object is created. Therefore, this message should pop up. It might be that some of the variables were not replaced everywhere in the code (let me have a look if I find it)

VictorBarbosaMartins:
The error is related to the io_handler that was included in this PR as argument, although I could not find the reason yet why including the io_handler lead to this problem. Was there any update on the io_handler that could lead to this?

GernotMaier:
Not including io_handler means that output_path is not set and the test fails. I don't understand why the test ever worked without it...

But maybe it is because I am the only one running the tests locally in parallel with `-n auto`? This is actually a nice thing, as it picks up dependencies between the tests.

VictorBarbosaMartins:
> Not including io_handler means that output_path is not set and the test fails. I don't understand why the test ever worked without it...
> 
> But maybe it is because I am the only one running the tests locally in parallel with `-n auto`? This is actually a nice thing, as it picks up dependencies between the tests.

No, it also doesnt work if I run all the tests one after the other. If this was working before, I assume some change related to this PR lead to the error, but I agree it is somehow unclear where this comes from. I will keep looking for some minutes more

simtools/corsika/corsika_histograms.py
```diff
@@ -106,9 +98,15 @@ def _initialize_attributes(self):
...
-        self._allowed_2D_labels = {"counts", "density", "direction", "time_altitude"}
+        self._allowed_1d_labels = {"wavelength", "time", "altitude"}
+        self._allowed_2d_labels = {"counts", "density", "direction", "time_altitude"}
         self._header = None
+        self.hist_position = None
+        self.hist_direction = None
+        self.hist_time_altitude = None
+
+        self.read_event_information()
+        self._initialize_header()
```
VictorBarbosaMartins, line 109: These changes here are breaking the test. I suggest returning it to the previous form and change only upper case D to lower case d.

simtools/corsika/corsika_histograms.py
```diff
@@ -75,14 +75,6 @@ def __init__(self, input_file, label=None, output_path=None, hdf5_file_name=None
...
-        self.read_event_information()
-        self._initialize_header()
-
-    def _initialize_attributes(self):
-        """
-        Initializes the class attributes.
-        """
         self._telescope_indices = None
         self._telescope_positions = None
         self.num_events = None
```
VictorBarbosaMartins, line 80: lines 78 to 110 should be maintained.

These changes here are breaking the test. I suggest returning it to the previous form and change only upper case D to lower case d.

GernotMaier: Nothing has been removed, these functions are called (as before) after the init. 

The three functions are called after all variables have been initialized (class variable init should take place in the `__init__` function, this is why I have changed it.

VictorBarbosaMartins:
The order is important. Please try to run with the previous lines and see that it works

simtools/corsika/corsika_histograms.py
VictorBarbosaMartins: I dont mind putting everything in one function if the fuction is not used elsewhere. but the order seems to be important

GernotMaier:
I've merged main (after the general.py) updates and resolved the conflicts. Should be good for you to go ahead.

orelgueta:
> I've merged main (after the general.py) updates and resolved the conflicts. Should be good for you to go ahead.

Thanks!
For the record, I also use `-n auto` always, so it can't be that.

simtools/io_operations/hdf5_handler.py
```diff
@@ -106,7 +106,7 @@ def read_hdf5(hdf5_file_name):
 
     with tables.open_file(hdf5_file_name, mode="r") as file:
         for node in file.walk_nodes("/", "Table"):
-            table_path = node._v_pathname
+            table_path = node._v_pathname  # pylint: disable=protected-access
```
orelgueta, line 109: Disabled pylint error because it seems it is not actually a protected member but just a reserved prefix. At least I could not find a way to get it in another way. It should be noted that ctapipe access this variable as well.

simtools/simtel/simtel_histograms.py
```diff
@@ -92,7 +92,7 @@ def _combine_histogram_files(self):
                 for o in yield_toplevel_of_type(f, Histograms):
                     try:
                         hists = o.parse()
-                    except Exception:
+                    except Exception:  # pylint: disable=broad-except
```
orelgueta, line 95: I disabled the pylint error because there are lots of exceptions that could occur when we parse and I don't think we can catch them individually. Also, it doesn't really matter for the use case. It doesn't matter why a file cannot be read, we should skip it regardless.

simtools/visualization/visualize.py
```diff
@@ -65,7 +65,7 @@
     "#A4657A",
     "#424D9C",
     "#DC575A",
-    "#1D2D38",
+    "#1D2d38",
```
orelgueta, line 68: You should be careful with search and replace...
It doesn't make a difference in this case, but in the future...

GernotMaier: Yes, it does! I've noted the first, but not the second one! Fixed.

simtools/simtel/simtel_histograms.py
GernotMaier: Agree.

simtools/io_operations/hdf5_handler.py
GernotMaier: Agree.

simtools/corsika/corsika_histograms.py
```diff
@@ -942,21 +940,21 @@ def get_photon_radial_distr(self, bins=None, max_dist=None):
...
-        hist2D_values_list, x_position_list, y_position_list = self.get_2D_photon_position_distr()
+        hist_2d_values_list, x_position_list, y_position_list = self.get_2d_photon_position_distr()
 
         for i_hist, _ in enumerate(x_position_list):
-            hist1D, bin_edges_1D = convert_2D_to_radial_distr(
-                hist2D_values_list[i_hist],
-                x_position_list[i_hist],
+            hist_1d, bin_edges_1d = convert_2d_to_radial_distr(
+                hist_2d_values_list[i_hist],
+                x_position_list[i_hist],  # pylint: disable=unnecessary-list-index-lookup
```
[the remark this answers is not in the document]
GernotMaier: Agree.

simtools/corsika/corsika_histograms.py
```diff
@@ -980,21 +978,21 @@ def get_photon_density_distr(self, bins=None, max_dist=None):
...
-        hist2D_values_list, x_position_list, y_position_list = self.get_2D_photon_density_distr()
+        hist_2d_values_list, x_position_list, y_position_list = self.get_2d_photon_density_distr()
 
         for i_hist, _ in enumerate(x_position_list):
-            hist1D, bin_edges_1D = convert_2D_to_radial_distr(
-                hist2D_values_list[i_hist],
-                x_position_list[i_hist],
+            hist_1d, bin_edges_1d = convert_2d_to_radial_distr(
+                hist_2d_values_list[i_hist],
+                x_position_list[i_hist],  # pylint: disable=unnecessary-list-index-lookup
```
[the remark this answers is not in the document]
GernotMaier: Agree.

GernotMaier:
I've fixed the remaining one (the numpy.unravel_index can return a tuble of any length, but we definitely expect only two values here).

Merged also #714 into main / this branch to solve the corsika_parameter_file complain.

Pylint is running through without any complaints now, very good!!

(look for the super-linter complain now)

GernotMaier:
The linter works, see e.g., https://github.com/gammasim/simtools/actions/runs/7125191095/job/19400677771
(sorting of issues should be covered by the pre-commit hook, but these are all commits from myself at a day where my pre-commit hood was broken).

GernotMaier:
@orelgueta , @VictorBarbosaMartins - this pull request is now ready for final review.  Please have a look and let me know when it is ready to merge.

Have a look at the 'checks' panel, super-linter nicely lists all the linters.

Lot's of trivial changes, but some important ones. Very nice.

VictorBarbosaMartins:
> https://github.com/gammasim/simtools/actions/runs/7125191095/job/19400677771

In the output there are some errors, but I think they were corrected by the linter?
```
2023-12-07 07:47:24 [ERROR]   ERRORS FOUND in PYTHON_ISORT:[3]
2023-12-07 07:47:25 [FATAL]   Exiting with errors found!
```

GernotMaier:
Yes - this is the example to show you that it works and that the linter picks up errors.

orelgueta:
I reviewed this issue yesterday and do not have any complaints. So from my point of view, it can be merged.

VictorBarbosaMartins:
I was wondering if there is a way to correct the line breaks of the doc string as well. I see line breaks in some places that are way below the limit. And sometimes line are broken with "\" in the docstring, but most of the time not. (not too important I guess)

simtools/corsika/corsika_histograms.py
```diff
@@ -1263,71 +1261,66 @@ def _export_1D_histograms(self, overwrite=False):
...
-
         Returns
         -------
         dict:
             The dictionary with information about the 2D distributions.
         """
-        if self.__dict_2D_distributions is None:
-            self.__dict_2D_distributions = {
+        if self.__dict_2d_distributions is None:
+            self.__dict_2d_distributions = {
```
VictorBarbosaMartins, line 1274: not sure we want double underscores here. Shouldn't we use a single underscore?

simtools/data_model/model_data_writer.py
```diff
@@ -84,7 +83,7 @@ def write(self, metadata=None, product_data=None):
         if product_data is not None:
             self.write_data(product_data=product_data)
 
-    def validate_and_transform(self, metadata=None, product_data=None, validate_schema_file=None):
```
VictorBarbosaMartins, line 87: we should remove metadata from the docstring as well

VictorBarbosaMartins:
> Examples?

Too early break:
https://github.com/gammasim/simtools/blob/44cbe1ff973a5d6e2127c916baa03a8d7720938e/simtools/psf_analysis.py#L26-L28

pyproject.toml
```diff
@@ -88,543 +88,56 @@ simtools-validate-file-using-schema = "simtools.applications.validate_file_using
...
+]
 
-# List of members which are set dynamically and missed by pylint inference
-# system, and so shouldn't trigger E1101 when accessed. Python regular
-# expressions are accepted.
+# See discussion in issue 521
+# Agreed to not have module docstring (missing-module-docstring)
+# Pylint cannot handle overload (used in pyproj; unpacking-non-sequence)
+disable = [
+    "missing-module-docstring",
```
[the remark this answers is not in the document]
GernotMaier: No, this is because we generally don't have module doc strings when we have one class per module and the doc in the class header.

pyproject.toml
```diff
@@ -88,543 +88,56 @@ simtools-validate-file-using-schema = "simtools.applications.validate_file_using
...
+    "unpacking-non-sequence",
+    "logging-format-interpolation",
+    "fixme",
```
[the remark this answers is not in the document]
GernotMaier: For now, yes. But for the future: if opened an issue on this topic: #715

simtools/corsika/corsika_histograms.py
GernotMaier: I think this was because there was a function "_dict_2d_distribution" and an object "__dist...". The function is now a class function and not private anymore. Change it if you think it is better.

simtools/data_model/model_data_writer.py
GernotMaier: Yes, done.

simtools/corsika/corsika_histograms.py
VictorBarbosaMartins: Ok, I pushed a commit

VictorBarbosaMartins:
Thanks! I think we can merge after the tests pass!

PR #688, opened 2023-11-22
Pre commit flake8 options

GernotMaier:
Remove two exclude options from the pre-commit configuration. These are not consistent with those used in the CI:

- Whitespace before ':' (E203) 
- Comparison to true should be 'if cond is true:' or 'if cond:' (E712) (https://www.flake8rules.com/rules/E712.html)


VictorBarbosaMartins:
Should we rerun flake8 for the entire to correct for these missing options?

VictorBarbosaMartins:
Hi @GernotMaier. Maybe you oversaw this, but I asked whether you think it is a good idea to already run the flake8 over the entire code once. I am not sure this would make a big difference, though. If not, then I would already approve this.

GernotMaier:
My answer got lost when the internet dropped at the cta conference: our code is consistent with the changed setting. I've run flake8 over all files and no issues have been reported.

VictorBarbosaMartins (approved):
Good, thank you.

PR #689, opened 2023-11-23
Remove duplications in application default arguments printing

GernotMaier:
This pull requests deals with how we list default configuration parameters and helps to avoid mistakes by documenting "by hand" default values to docstrings and printouts:

- argparser is now configured with `formatter_class=argparse.ArgumentDefaultsHelpFormatter`, meaning the default values given for each parameters are listed when running an application with `--help`
- removed all statements about default from the description of each parameter
- removed all description of defaults in the docstrings of the applications. This is duplication and leads to mismatches.


VictorBarbosaMartins:
Ok! yes, let's merge it.

PR #692, opened 2023-11-28
Any pull requests should trigger CI tests

GernotMaier:
Any pull request should trigger the unit and integration tests. Before this change, only pull requests to main did this. This PR removes this requirement.


VictorBarbosaMartins (approved):
I think we put this at some point in order to spare some cores (as we have limited cores available from github). In any case, I dont think the number of cores could be an issue given that we are a rather small team, so I am approving

PR #700, opened 2023-11-30
Fix fixture interdependence in test_corsika_config

GernotMaier:
Units tests fail for `tests/unit_tests/corsika/test_corsika_config.py` when run with `pytest -n 8`. 

This indicated previously unknown interdependence between the tests. The issue is that `data_path` is set when calling CorsikaConfig, so I've added this fixture call to all test functions.


orelgueta (approved):
There's no way for me to confirm that this PR actually does what it's supposed to do other than testing myself before and after. That seems excessive for such a simple PR, so I simply approve.

PR #697, opened 2023-11-29
Fix docstring issues leading to sphinx errors in generate_corsika_histograms

GernotMaier:
closes #670

PR #702, opened 2023-11-30
Fix intendation in documenation to generate_corsika_histograms

GernotMaier:
Related to #697


VictorBarbosaMartins (approved):
ok! The mistake from before is gone. Thank you and apologies for missing it before. Let's wait for the tests and merge it.

PR #703, opened 2023-12-04
Remove unused parameter corsika_parameters_file in corsika_runner

### LINKED ISSUE #417: Unused parameter in corsika_runner

GernotMaier:
Carefully checked if `corsika_parameters_file` is used anywhere. It is indeed not used and safely be removed. 

closes #417

### DIFF
--- a/simtools/corsika/corsika_runner.py
+++ b/simtools/corsika/corsika_runner.py
@@ -43,10 +43,6 @@ class CorsikaRunner:
             'cscat': [10, 1500 * u.m, 0]
         }
 
-    The remaining CORSIKA parameters can be set as a yaml file, using the argument \
-    corsika_parameters_file. When not given, corsika_parameters will be loaded from \
-    data/parameters/corsika_parameters.yml.
-
     The CORSIKA output directory must be set by the data_directory entry. The following directories\
     will be created to store the logs and input file:
     {data_directory}/corsika/$site/$primary/logs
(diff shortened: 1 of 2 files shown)


VictorBarbosaMartins:
I am afraid this is the place where we would be able to change the atmosphere file. I will have a closer look tomorrow to confirm it

VictorBarbosaMartins (approved):
Good. I think now it is not a problem to remove it

orelgueta (changes requested):
docstring should be put back and corrected, will do myself.

orelgueta (changes requested):
No, I think this PR should be the other way around. We need to pass to the `CorsikaConfig` the `corsika_parameters_file` rather than removing it. 

With this PR we essentially are losing functionality and making that parameter useless in `Simulator` and `CorsikaConfig`. It is functionality we never used, but we should keep it. If you want, scrap this entire PR and I will open a new one later.

GernotMaier:
OK - let's close this.  I was trying to fix #417.

Note that this PR removes an unused variable and all tests pasts after removing. So I think something is inconsistent.

orelgueta:
Yes, that's cause there's a default option in case `None` is passed. I will fix it in a separate PR.

PR #704, opened 2023-12-04
Review of general.py

GernotMaier:
This pull requests reviews and updates the contents of [simtools/utils/general.py](simtools/utils/general.py), see also issue #665.

Updates:

- add comment to suppress pylint complain


VictorBarbosaMartins:
I think this is ready for review, given that the tests pass. I addressed all the points discussed [here](https://github.com/gammasim/simtools/issues/665#issuecomment-1837967036), so I will not duplicate the description of the PR.

VictorBarbosaMartins:
Maybe we should move `copy_as_list` as well to `transformations.py`

simtools/utils/general.py
```diff
@@ -516,29 +477,6 @@ def set_default_kwargs(in_kwargs, **kwargs):
     return in_kwargs
 
 
-def sort_arrays(*args):
```
GernotMaier, line 519: I suggest to keep sort_arrays in general.py - this is quite generic.

VictorBarbosaMartins: yet it is a transformation. You really meant geometrical transformations then

tests/unit_tests/utils/test_general.py
```diff
@@ -455,28 +312,26 @@ def test_get_log_excerpt(tmp_test_directory) -> None:
     )
 
 
-def test_file_has_text(tmp_test_directory, caplog) -> None:
+def test_file_has_text(tmp_test_directory, caplog, file_has_text) -> None:
```
GernotMaier, line 315: This is now a function in conf.py. We usually don't test these functions, but I think it is good to keep it (just to note that we don't test general.py functions here). All good!

GernotMaier (commented):
I am happy and from my point of view this can be merged. I approve!

(I can't approve per button, as I started this PR)

PR #712, opened 2023-12-06
Improve startup time of pytest by optimising imports

### LINKED ISSUE #694: Slow starting of unit tests
pytest is very slow to start, even for very simple tests (>5 seconds on my laptop). Very hard to debug with this slow response.

After investigation, it is the same issue as realized recently with general.py (#681): any including of ctapipe means that tools slow down due to a large number of slow imports.

I suggest to consider removing the following two fixtures from conftest.py:
```
@pytest.fixture
def corsika_histograms_instance(io_handler, corsika_output_file_name):
    return CorsikaHistograms(
        corsika_output_file_name, output_path=io_handler.get_output_directory(dir_type="test")
    )


@pytest.fixture
def corsika_histograms_instance_set_histograms(db, io_handler, corsika_histograms_instance):
    corsika_histograms_instance.set_histograms()
    return corsika_histograms_instance

```

Victor, is this possible without a huge amount of work? 

I will also investigate with ctapipe developers if this is a feature or an issue.

GernotMaier:
closes #694

pytest on a simple test file can take >5s to start due to the time required for imports.

This PR optimises the import by moving the `simtools.corsika.corsika_histograms` import into the corresponding pytest fixture (so it is called only when needed.

This improves startup times for tests notably (at least on my laptop).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,6 @@ from dotenv import dotenv_values, load_dotenv
 import simtools.io_operations.io_handler
 from simtools import db_handler
 from simtools.configuration.configurator import Configurator
-from simtools.corsika.corsika_histograms import CorsikaHistograms
 from simtools.layout.array_layout import ArrayLayout
 from simtools.model.telescope_model import TelescopeModel
 
@@ -279,6 +278,8 @@ def corsika_output_file_name():
 
 @pytest.fixture
 def corsika_histograms_instance(io_handler, corsika_output_file_name):
+    from simtools.corsika.corsika_histograms import CorsikaHistograms
+
     return CorsikaHistograms(
         corsika_output_file_name, output_path=io_handler.get_output_directory(dir_type="test")
     )


tests/conftest.py
```diff
@@ -279,6 +278,8 @@ def corsika_output_file_name():
 
 @pytest.fixture
 def corsika_histograms_instance(io_handler, corsika_output_file_name):
+    from simtools.corsika.corsika_histograms import CorsikaHistograms
+
```
VictorBarbosaMartins, line 282: shouldn't we add the pylint exception here or it doesnt matter for the test directory?

VictorBarbosaMartins (approved):
I already approve, but there is one comment, which is up to you to implement or not. Thanks!

tests/conftest.py
GernotMaier: This is not needed, as we do not run pylint over the test directory (we would get lots of complaints, e.g. from all the internal class functions we call in the test).

So no pylint comment required.

PR #713, opened 2023-12-06
Add tests of simtools-prod containers

GernotMaier:
Add a test for the simtools-prod container to ensure that everything is correctly installed. For this, run a single application test using the database and sim_telarray (`simtools-simulate-prod`).

This required to separate the steps of generating the container and pushing.

The successful github action workflow for the container generation do not appear below, it is:
- simtools-prod: https://github.com/gammasim/simtools/actions/runs/7210340303/job/19643252445
- simtools-dev: https://github.com/gammasim/simtools/actions/runs/7210340303/job/19643252194


GernotMaier:
And I notice now that the docker-simtools test start when pressing "read for review".

orelgueta (approved):
You are using the config file that comes in `tests/resources` for running, but do we expect that the production container comes with the `tests` directory? It probably shouldn't, no?

If we decide to remove `tests` from this container, we can just provide the arguments directly, they are not too many.

I approve anyway cause the solution works for now.

(As a side note, the amount of commits and their messages for such a small change show how much of a pain it is to work on docker stuff.)

GernotMaier:
No, the `tests` directory is not part of the container. It is available in the workflow, because we need to do a checkout of simtools to get the Dockerfile. Using the config file simplifies this - and if we change any of the command line parameters we do not have to worry about this test (unless we change the `--config` parameter.

orelgueta:
Oh yea, I forgot this is part of building the container and therefore we need to checkout simtools anyway. All good then.

PR #716, opened 2023-12-07
Add validation of ecsv file using a schema to application validate_file_using_schema

GernotMaier:
The application `validate_file_using_schema` allows to validate now:

- yaml/yml files (existing functionality)
- ecsv file with astropy tables (added functionality)

Examples are:

```
python simtools/applications/validate_file_using_schema.py \
    --file_name tests/resources/MLTdata-preproduction.meta.yml \
    --schema simtools/schemas/metadata.schema.yml
```

and

```
python simtools/applications/validate_file_using_schema.py \
    --file_name tests/resources/tests/resources/MLTdata-preproduction.ecsv \
    --schema tests/resources/MST_mirror_2f_measurements.schema.yml
```


tests/integration_tests/test_applications.py
```diff
@@ -665,16 +665,16 @@
...
-        [
-            "--input",
-            "tests/resources/telescope_positions-South-4MST.ecsv",
-            "--export",
-            "utm",
-            "--output_path",
-            "TESTOUTPUTDIR/",
-        ],
-    ],
+    #    "print_array_elements::export_utm": [
```
tobiaskleiner, line 668: Is this still needed here?

VictorBarbosaMartins:
Hi @GernotMaier. I havent checked yet the changes, but from the description I was wondering if we could point to the schema file already in the metadata of the file. Both ecsv and yaml allow metadata in the header. With that we would not need to point to a schema file, risking pointing to a wrong one.

PR #718, opened 2023-12-07
Read mirror list from ecsv

tobiaskleiner:
Reading mirror list values from ecsv file and adding a unit test.


simtools/model/mirrors.py
```diff
@@ -54,26 +56,25 @@ def _read_mirror_list_from_ecsv(self):
...
+            self.number_of_mirrors = len(self.mirror_table["focal_length"])
+
+            self._logger.debug(f"Mirror shape_type = {self.shape_type}")
+            self._logger.debug(f"Mirror diameter = {self.mirror_diameter}")
+            self._logger.debug(f"Number of Mirrors = {self.number_of_mirrors}")
         except KeyError:
-            self._logger.debug(
-                f"Missing column for mirror panel focal length (flen) in {self._mirror_list_file}"
+            self._logger.debug("Mirror shape_type not in mirror file")
+            self.number_of_mirrors = len(self.mirror_table["mirror_id"])
```
orelgueta, line 73: I wonder about this try..except. You might need to do it one by one because if the shape is not given but the diameter is, you won't read it. Also, I do not really understand why the number of mirrors is taken using a different key if the shape is no there.
Lastly, I thought we said that if the shape and diameter are not in the file, you will get them from the DB. That can be done in a separate PR though if you want.

simtools/model/mirrors.py
```diff
@@ -54,26 +56,25 @@ def _read_mirror_list_from_ecsv(self):
...
+
+        if "focal_length" not in self.mirror_table.colnames:
+            self.mirror_table["focal_length"] = (
```
orelgueta, line 76: Wait, that's not correct. The focal length (for spherical mirrors) is half the **radius of curvature**, not the mirror radius. For example, the diameter of each LST mirror is 151 cm, but their focal length is roughly 2900 cm.
Instead, you should set the focal length of each mirror to zero. This will tell sim_telarray take the nominal focal length and randomise around it with `RANDOM_FOCAL_LENGTH`

simtools/model/mirrors.py
```diff
@@ -136,23 +131,37 @@ def get_single_mirror_parameters(self, number):
...
-            X, Y positions, diameter, focal length and shape.
+        (pos_x, pos_y, mirror_diameter, focal_length, shape_type): tuple of float
+            X, Y positions, mirror_diameter, focal length and shape_type.
         """
 
-        if number > self.number_of_mirrors - 1:
-            self._logger.error("Mirror number is out range")
+        def get_mirror_table_mask():
+            if isinstance(self.mirror_table["mirror_id"][0], np.str_):
+                mask = self.mirror_table["mirror_id"] == "id=" + str(number)
```
orelgueta, line 140: Following the long discussion we had today about strings, you should use f-string here `f"id={number}"`

simtools/model/mirrors.py
```diff
@@ -136,23 +131,37 @@ def get_single_mirror_parameters(self, number):
...
+            if isinstance(self.mirror_table["mirror_id"][0], np.int32):
+                mask = self.mirror_table["mirror_id"] == number
+            return mask
+
+        mask = get_mirror_table_mask()
+        if not np.any(mask):
+            self._logger.error(f"Mirror id{number} not in table, using first mirror instead")
```
orelgueta, line 147: This should be a warning, or perhaps a debug message, not an error. For errors we usually stop running.

simtools/model/mirrors.py
```diff
@@ -136,23 +131,37 @@ def get_single_mirror_parameters(self, number):
...
+            mask[0] = True
+
+        try:
+            return (
+                self.mirror_table[mask]["mirror_x"].value[0],
+                self.mirror_table[mask]["mirror_y"].value[0],
+                self.mirror_table[mask]["mirror_diameter"].value[0],
```
orelgueta, line 154: Again, it's a shame to remove the unit here (also for focal_length)

simtools/model/mirrors.py
```diff
@@ -136,23 +131,37 @@ def get_single_mirror_parameters(self, number):
...
+                self.mirror_table[mask]["focal_length"].value[0],
+                self.mirror_table[mask]["shape_type"].value[0],
+            )
+        except KeyError:
```
orelgueta, line 158: This is not covered by tests.

simtools/model/mirrors.py
```diff
@@ -136,23 +131,37 @@ def get_single_mirror_parameters(self, number):
...
-            self._mirrors["flen"][number],
-            self._mirrors["shape"][number],
-        )
 
     def plot_mirror_layout(self):
         """
-        Plot the mirror layout (not implemented yet).
+        Plot the mirror layout.
+
+        TODO
```
orelgueta, line 166: Notice that this is an old version of the code. Probably you started your branch later and then didn't accept this merge. We removed the TODO and raise the error instead. Please revert this change.

tests/resources/MLTdata-preproduction.ecsv
```diff
@@ -1,209 +1,214 @@
 # %ECSV 1.0
 # ---
 # datatype:
-# - {name: mirror_panel_id, datatype: string, description: mirror panel ID}
-# - {name: mirror_panel_radius, unit: m, datatype: float32, description: mirror panel radius (2f)}
-# - {name: psf_opt, unit: 'mm', datatype: float32, description: Average spot size of MLT mirror panels from 2f measurement (d80), best value (not at nominal distance)}
-# - {name: psf, unit: 'mm', datatype: float32, description: Average spot size of MLT mirror panels from 2f measurement (d80) at nominal distance of 32.14 m}
+# - {name: mirror_id, datatype: string, description: mirror panel ID}
+# - {name: mirror_diameter, unit: m, datatype: float32, description: mirror panel radius (2f)}
```
orelgueta, line 5: I don't know why it says 2f here and also it's not the radius it's the diameter.

tests/resources/MLTdata-preproduction.ecsv
```diff
@@ -1,209 +1,214 @@
...
-MLT-254    32.25     12.31    12.3     90.2
-MLT-255    32.237    12.71    12.03    93.8
-MLT-256    32.312    14.98    15.78    92.4
-MLT-257    32.101    15.67    15.67    93.6
-MLT-258    32.203    12.74    12.8     93.2
-MLT-259    32.244    12.14    13.41    93.1
-MLT-260    32.284    14.06    13.56    93.3
-MLT-261    32.209    13.91    13.84    93.8
-MLT-262    32.146    15.44    15.44    92.3
+# - {name: mirror_x, unit: cm, datatype: float64}
```
orelgueta, line 10: Wait, why did you add mirror_x and mirror_y here?

tests/resources/MLTdata-preproduction.ecsv
```diff
@@ -1,209 +1,214 @@
...
+# - {name: mirror_y, unit: cm, datatype: float64}
+# - {name: shape_type, datatype: int32}
```
orelgueta, line 12: The shape can/should probably be defined in a more descriptive way than the 1/2/3 of sim_telarray

tests/resources/MLTdata-preproduction.ecsv
```diff
@@ -1,209 +1,214 @@
...
+# schema: astropy-2.0
+mirror_id mirror_diameter psf_opt psf surface_ref mirror_x mirror_y shape_type
+054 63.614 15.46 20.97 90.2 0.0 0.0 0
```
orelgueta, line 15: The name of the mirror was changed, this is probably not what we want. I wonder about the change from radius to diameter. It's probably fine, but I would like to alert @GernotMaier to this change.

tests/resources/MLTdata-preproduction.ecsv
```diff
@@ -1,209 +1,214 @@
 # %ECSV 1.0
 # ---
 # datatype:
-# - {name: mirror_panel_id, datatype: string, description: mirror panel ID}
-# - {name: mirror_panel_radius, unit: m, datatype: float32, description: mirror panel radius (2f)}
-# - {name: psf_opt, unit: 'mm', datatype: float32, description: Average spot size of MLT mirror panels from 2f measurement (d80), best value (not at nominal distance)}
-# - {name: psf, unit: 'mm', datatype: float32, description: Average spot size of MLT mirror panels from 2f measurement (d80) at nominal distance of 32.14 m}
+# - {name: mirror_id, datatype: string, description: mirror panel ID}
```
orelgueta, line 4: Why did you go from `mirror_panel_id` to `mirror_id`? I think we wanted to keep always the reference to panel when dealing with mirror panels and not full telescope mirrors.

tests/resources/MST_mirror_2f_measurements.schema.yml
```diff
@@ -14,15 +14,15 @@ base_schema_version: 0.1.0
 data:
   - type: datatable
     table_columns:
-      - name: mirror_panel_id
+      - name: mirror_id
```
orelgueta, line 17: Same comment as above, this time tagging @GernotMaier. 
Why did you go from `mirror_panel_id` to `mirror_id`? I think we wanted to keep always the reference to panel when dealing with mirror panels and not full telescope mirrors.

tests/resources/MST_mirror_2f_measurements.schema.yml
```diff
@@ -14,15 +14,15 @@ base_schema_version: 0.1.0
...
         description: |-
           Mirror panel ID.
         required: true
         units: dimensionless
         type: string
-      - name: mirror_panel_radius
+      - name: mirror_diameter
```
orelgueta, line 23: Again, probably not a problem, but would like to hear @GernotMaier opinion

tests/resources/MST_mirror_2f_measurements.schema.yml
```diff
@@ -39,3 +39,21 @@ data:
         required: false
         units: cm
         type: double
+      - name: mirror_x
```
orelgueta, line 42: I don't think this is needed for this schema.

tests/resources/MST_mirror_2f_measurements.schema.yml
```diff
@@ -39,3 +39,21 @@ data:
...
+        description: |-
+          Mirror panel position x.
+        required: false
+        units: cm
+        type: double
+      - name: mirror_y
```
orelgueta, line 48: I don't think this is needed for this schema.

tests/resources/MST_mirror_2f_measurements.schema.yml
```diff
@@ -39,3 +39,21 @@ data:
...
+        description: |-
+          Mirror panel position y.
+        required: false
+        units: cm
+        type: double
+      - name: shape_type
```
orelgueta, line 54: Same comment as above, maybe we can/should come up with a friendlier way to represent the shape

orelgueta (changes requested):
Reading is now much nicer!
There are a few small issues, so I do not approve yet, but it's close to being done.

simtools/model/mirrors.py
```diff
@@ -54,26 +56,25 @@ def _read_mirror_list_from_ecsv(self):
...
         try:
-            self._mirrors["flen"] = list(_mirror_table["mirror_panel_radius"].to("cm").value / 2.0)
-            self.number_of_mirrors = len(self._mirrors["flen"])
-            self._mirrors["number"] = list(range(self.number_of_mirrors))
-            self._mirrors["pos_x"] = [0.0] * self.number_of_mirrors
-            self._mirrors["pos_y"] = [0.0] * self.number_of_mirrors
-            self._mirrors["diameter"] = [self.diameter] * self.number_of_mirrors
-            self._mirrors["shape"] = [self.shape] * self.number_of_mirrors
+            self.shape_type = u.Quantity(self.mirror_table["shape_type"])[0].value
+            self.mirror_diameter = u.Quantity(self.mirror_table["mirror_diameter"])[0].value
```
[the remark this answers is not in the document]
tobiaskleiner: Removing this.

simtools/model/mirrors.py
tobiaskleiner: Changing.

tests/resources/MLTdata-preproduction.ecsv
tobiaskleiner: We currently read this as a mirror list in `derive_mirror_rnda.py` . I tried to add the missing columns required by the schema.
We also pass this as `psf_measurement`. Then maybe we should have two seperate lists passed here?

tests/resources/MLTdata-preproduction.ecsv
tobiaskleiner: This seems to be not consistent also in the schema files. `mirror_list.schema.yml` uses for instance `mirror_id`, wheras `MST_mirror_2f_measurements.schema.yml` uses `mirror_panel_id`. I tried to use mirror_id throughout but we can change this @GernotMaier ?

tests/resources/MST_mirror_2f_measurements.schema.yml
tobiaskleiner: Related to above comment about derive_mirror_rnda application. (Using this as a mirror_list)

tests/resources/MST_mirror_2f_measurements.schema.yml
GernotMaier: I completely agree that it should be mirror_panel_id. Not sure why this was changed.

simtools/model/mirrors.py
orelgueta: We could require them to be in the metadata. It might require some manual work for the person preparing the input file, but that's not too bad I guess. In that case, we must find a more friendly way to represent the shape then. The number 1, 2, 3 would not make sense to anyone external.

tests/resources/MLTdata-preproduction.ecsv
orelgueta: But the schema for this file has to be different than the one deriving the full telescope PSF. Are you sure the schema requires these columns also for this file?

tests/resources/MLTdata-preproduction.ecsv
orelgueta: I think the entry in `mirror_list.schema.yml` should change. I assume @GernotMaier will do it as part of the effort he mentioned today. Here we should follow the `MST_mirror_2f_measurements.schema.yml` example and use `mirror_panel_id`.

GernotMaier: Feel free to change when you work on it! Better to correct information as soon as you know that it is wrong/incomplete.

tests/resources/MST_mirror_2f_measurements.schema.yml
orelgueta: I don't understand why the `derive_mirror_rnda` application requires the positions of the mirrors. It simulates them separately to derive the mirror roughness of the mirror panel. It doesn't have anything to do with the dish.

tobiaskleiner:
@orelgueta I have adressed your comments and the PR is ready for a re-review.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
+        if "focal_length" not in self.mirror_table.colnames:
+            try:
+                self.mirror_table["focal_length"] = (
+                    self.mirror_table["mirror_panel_radius"].to("cm").value / 2
+                )
+            except KeyError:
+                self._logger.debug("mirror_panel_radius not contained in mirror list")
+                try:
+                    print("NUMBER OF MIRRORS: ", self.number_of_mirrors)
+                    self.mirror_table["focal_length"] = self.number_of_mirrors * [
```
orelgueta, line 105: These lines are not covered by tests.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
+                        u.Quantity(
+                            self.parameters["mirror_focal_length"]["Value"],
+                            self.parameters["mirror_focal_length"]["units"],
+                        )
+                    ]
+                except KeyError:
+                    self._logger.debug("mirror_focal_length not contained in db")
```
orelgueta, line 112: Again, without the focal length, we shouldn't proceed.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
-        if "focal_length" not in self.mirror_table.colnames:
-            self.mirror_table["focal_length"] = (
-                self.mirror_table["mirror_diameter"].to("cm").value / 4.0
-            )
+        try:
+            self.mirror_diameter = u.Quantity(self.mirror_table["mirror_diameter"])[0]
+            self._logger.debug(f"Mirror diameter = {self.mirror_diameter}")
+        except KeyError:
+            self._logger.debug("Mirror mirror_panel_diameter not in mirror file")
+            try:
```
orelgueta, line 88: A general comment about all of these nested `try..except`. I am not sure if they should be nested or not. I remember Gernot had an issue that the second except was never reached. Please make sure to include tests that reach all cases to make sure it works. If it doesn't, try to put the second `try..except` at the same level as the first.

simtools/model/mirrors.py
```diff
@@ -103,22 +132,30 @@ def _read_mirror_list_from_sim_telarray(self):
...
 
         self._logger.debug(f"Mirror shape_type = {self.shape_type}")
         self._logger.debug(f"Mirror diameter = {self.mirror_diameter}")
         self._logger.debug(f"Number of Mirrors = {self.number_of_mirrors}")
 
-        if self.number_of_mirrors == 0:
-            msg = "Problem reading mirror list file"
-            self._logger.error(msg)
-            raise InvalidMirrorListFile()
+    def get_mirror_table(self):
```
orelgueta, line 150: You should use the property mechanism of python instead of this classic getter (you can find lots of examples online).

simtools/model/mirrors.py
```diff
@@ -103,22 +132,30 @@ def _read_mirror_list_from_sim_telarray(self):
...
+        """
+        Get the mirror list.
+
+        Returns
+        -------
+        mirror table: astropy.table.table.Table
```
orelgueta, line 156: I know this is the actual type, but it looks funny =)
You can replace it with `astropy Table`

simtools/model/mirrors.py
```diff
@@ -136,10 +173,18 @@ def get_single_mirror_parameters(self, number):
         """
 
         def get_mirror_table_mask():
-            if isinstance(self.mirror_table["mirror_id"][0], np.str_):
-                mask = self.mirror_table["mirror_id"] == "id=" + str(number)
-            if isinstance(self.mirror_table["mirror_id"][0], np.int32):
-                mask = self.mirror_table["mirror_id"] == number
+            if isinstance(self.mirror_table["mirror_panel_id"][0], np.str_):
```
orelgueta, line 176: Why are you testing against `np.str_` and not just `str`? A `np.str_` is also a `str`, but a `str` is not an `np.str_`. I think in this case you should be more inclusive.

simtools/model/mirrors.py
```diff
@@ -136,10 +173,18 @@ def get_single_mirror_parameters(self, number):
...
+                mask = (
+                    np.array(
+                        [
+                            "".join(filter(str.isdigit, string))
+                            for string in self.mirror_table["mirror_panel_id"]
+                        ]
+                    )
+                    == f"{number}"
+                )
+            if isinstance(self.mirror_table["mirror_panel_id"][0], np.int32):
```
orelgueta, line 186: Actually, I realise now (after writing the comment above regarding the string), that we shouldn't do this at all. The id should always be a number. That is, when you read a table, you should cast the id to be a number, even if it is written as a string in the file.

simtools/model/mirrors.py
```diff
@@ -157,11 +202,11 @@ def get_mirror_table_mask():
...
-        """
-        Plot the mirror layout.
-
-        TODO
-        """
+            return_values = (
+                0,
+                0,
+                self.mirror_diameter.value,
+                self.mirror_table[mask]["focal_length"].value[0],
```
orelgueta, line 209: Shame to remove the units here no?

tests/resources/MLTdata-preproduction.ecsv
```diff
@@ -1,214 +1,209 @@
 # %ECSV 1.0
 # ---
 # datatype:
-# - {name: mirror_id, datatype: string, description: mirror panel ID}
-# - {name: mirror_diameter, unit: m, datatype: float32, description: mirror panel radius (2f)}
-# - {name: psf_opt, unit: mm, datatype: float32, description: Average spot size of MLT mirror panels from 2f measurement (d80)}
-# - {name: psf, unit: mm, datatype: float32, description: Average spot size of MLT mirror panels from 2f measurement (d80) at nominal
-#     distance of 32.14 m}
+# - {name: mirror_panel_id, datatype: string, description: mirror panel ID}
+# - {name: mirror_panel_radius, unit: m, datatype: float32, description: mirror panel radius (2f)}
```
orelgueta, line 5: The name and description here should change

tests/resources/MLTdata-preproduction.ecsv
orelgueta, line 1: Notice there are conflicts with the main branch

tests/resources/MST_mirror_2f_measurements.schema.yml
```diff
@@ -1,28 +1,26 @@
...
     table_columns:
-      - name: mirror_id
+      - name: mirror_panel_id
         description: |-
           Mirror panel ID.
         required: true
         units: dimensionless
         type: string
-      - name: mirror_diameter
+      - name: mirror_panel_radius
```
orelgueta, line 21: I think this is a mistake. I think this should stay the mirror panel diameter and on top of it you can have also the radius of curvature.

orelgueta (changes requested):
Thanks for the changes!
I think there are still a few things that need to change and/or fixed. I made some comments inline. Please also look at the previous ones and resolve the ones which you already addressed.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
+        self.number_of_mirrors = np.shape(self.mirror_table)[0]
+        self._logger.debug(f"Number of Mirrors = {self.number_of_mirrors}")
+
+        if self.number_of_mirrors == 0:
+            msg = "Problem reading mirror list file"
+            self._logger.error(msg)
+            raise InvalidMirrorListFile
+
+        try:
+            self.shape_type = u.Quantity(self.mirror_table["shape_type"])[0]
```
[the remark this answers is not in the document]
tobiaskleiner: We use `shape_type` in the tables as well. So we can use it consistent in the code.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
             self._logger.debug(f"Mirror shape_type = {self.shape_type}")
-            self._logger.debug(f"Mirror diameter = {self.mirror_diameter}")
-            self._logger.debug(f"Number of Mirrors = {self.number_of_mirrors}")
         except KeyError:
             self._logger.debug("Mirror shape_type not in mirror file")
-            self.number_of_mirrors = len(self.mirror_table["mirror_id"])
+            try:
+                self.shape_type = self.parameters["mirror_panel_shape"]["Value"]
+            except TypeError:
+                self._logger.debug("Mirror mirror_panel_shape not contained in db")
```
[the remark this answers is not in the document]
tobiaskleiner: Fixed the DB. The other part is not true. For instance we are also reading `MLTdata-preproduction.ecsv`, that does not contain `shape_type` and is also not needed in this case. We could require this as part of the schema though for all files?

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
+            self._logger.debug(f"Mirror diameter = {self.mirror_diameter}")
+        except KeyError:
+            self._logger.debug("Mirror mirror_panel_diameter not in mirror file")
+            try:
+                self.mirror_diameter = u.Quantity(
+                    self.parameters["mirror_panel_diameter"]["Value"],
+                    self.parameters["mirror_panel_diameter"]["units"],
+                )
+            except TypeError:
+                self._logger.debug("Mirror mirror_panel_diameter not contained in db")
```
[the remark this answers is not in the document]
tobiaskleiner: Same as above.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
 
-        if self.number_of_mirrors == 0:
-            msg = "Problem reading mirror list file"
-            self._logger.error(msg)
-            raise InvalidMirrorListFile()
+        if "focal_length" not in self.mirror_table.colnames:
+            try:
+                self.mirror_table["focal_length"] = (
+                    self.mirror_table["mirror_panel_radius"].to("cm").value / 2
```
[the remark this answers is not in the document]
tobiaskleiner: I changed the key now.

simtools/model/mirrors.py
```diff
@@ -60,27 +61,55 @@ def _read_mirror_list_from_ecsv(self):
...
+                )
+            except KeyError:
+                self._logger.debug("mirror_panel_radius not contained in mirror list")
+                try:
+                    print("NUMBER OF MIRRORS: ", self.number_of_mirrors)
```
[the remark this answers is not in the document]
tobiaskleiner: Removed.

tests/resources/MST_mirror_2f_measurements.schema.yml
tobiaskleiner: Changing this to radius of curvature.

simtools/model/mirrors.py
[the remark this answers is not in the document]
orelgueta: But does is `shape_type` clearer than just `shape`? If so, then no problem, we can resolve this conversation. Otherwise, we could change the tables as well (we have very few right now, so better decide now).

simtools/model/mirrors.py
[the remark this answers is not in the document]
orelgueta: It is true that `MLTdata-preproduction.ecsv` does not contain the shape, but I think that is why Gernot hardcoded the shape and diameter in the code at the time. Now you added this code that reads it from the DB if it's not in the file, which was the original plan for this PR. 
If it's not in the DB either, I don't think we can simulate anything. Any simulation that has to do with the mirrors will need the shape and diameter. Do you know of simulations we can run without it?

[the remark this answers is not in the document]
tobiaskleiner: Ok I think my comment was wrong in this regard. We would always require the shape either directly from the tables or from the DB. I added a raise error now.

simtools/model/mirrors.py
orelgueta: I thought we said you will remove this function. It's not a show stopper for this PR though.

tests/unit_tests/model/test_mirrors.py
```diff
@@ -3,26 +3,189 @@
...
-    mirror_list_file = gen.find_file(
-        test_file_name, io_handler.get_output_directory(sub_dir="model", dir_type="test")
+    logger.info(f"Using mirror list {mirror_list_file}")
+    mirrors = Mirrors(mirror_list_file)
+    assert 198 == mirrors.number_of_mirrors
+    assert 151.0 == pytest.approx(mirrors.mirror_diameter.value)
+    assert 3 == mirrors.shape_type
+
+
+def test_read_mirror_list_from_ecsv(io_handler, tmp_test_directory, caplog):
```
orelgueta, line 26: Technically each case below should have it's own test function. It simplifies hunting down issues in the future and running only specific cases when you find them. It can sometimes lead to duplications, but those are usually really minimised when using fixtures. You can look at other test files to see examples. 
Again, this isn't really a blocker for this PR. We can improve it in a separate PR if you want to merge this one first.

tests/unit_tests/model/test_mirrors.py
```diff
@@ -3,26 +3,189 @@
...
+        test=True,
+    )
+    logger.info(f"Using mirror list {mirror_list_file}")
+    mirrors = Mirrors(mirror_list_file)
+
+    report = report_diff_values(mirrors.mirror_table, mirrors.get_mirror_table())
+    assert report is True
+
+
+def test_get_single_mirror_parameters(io_handler):
```
orelgueta, line 130: Same comment as above, this should be broken up to separate tests.

simtools/model/mirrors.py
```diff
@@ -136,23 +181,47 @@ def get_single_mirror_parameters(self, number):
...
+            return_values = (
+                u.Quantity(
+                    self.mirror_table[mask]["mirror_x"].value[0],
+                    self.mirror_table[mask]["mirror_x"].unit,
+                ),
+                u.Quantity(
+                    self.mirror_table[mask]["mirror_y"].value[0],
+                    self.mirror_table[mask]["mirror_y"].unit,
+                ),
+                u.Quantity(
```
orelgueta, line 202: Do you need to explicitly build the Quantity here? If you simply return `self.mirror_table[mask]["mirror_diameter"][0]`, won't you get the quantity already?

orelgueta (commented):
Very minor comments, none of them are blockers. If you want to merge already we can probably do it and discuss the rest elsewhere.

simtools/model/mirrors.py
tobiaskleiner: Yes, this is neccessary: `self.mirror_table[mask]["mirror_diameter"][0]` would only contain the value without unit. `self.mirror_table[mask]["mirror_diameter"]` is a astropy column with Quantity entries and `type(u.Quantity(sim_t[mask]["mirror_diameter"]).value)` is a `numpy.ndarray`.

orelgueta: Ahh OK, I see. Thanks!

orelgueta (approved):
Looks good now!

PR #719, opened 2023-12-07
Configure applications from configuration file

GernotMaier:
Fix the existing functionality of configuring simtools using a configuration file, e.g., 

```
python simtools/applications/print_array_elements.py --config config.yml
```

With configurations listed in this example as

```
INPUT: ./tests/resources/telescope_positions-North-utm.ecsv
EXPORT: corsika
USE_CORSIKA_TELESCOPE_HEIGHT: True
OUTPUT_PATH: simtools-tests
SELECT_ASSETS:
  - LSTN
  - MSTN
```

Important to note that our configuration class is powerful, but with limits:

- configuration is allowed through command line, configuration file, environmental variables, and direct pass of a dictionary from the application
- difficult to determine priorities (as configuration parameters also have default values)

Also introduced a new parameter for Configurator::initialize(..., require_command_line=True/False, ...)` which means that this tools requires at least one command line parameter. If none is given, help is printed.

### DIFF
--- a/simtools/applications/print_array_elements.py
+++ b/simtools/applications/print_array_elements.py
@@ -100,7 +100,7 @@ def _parse(label=None, description=None):
     config.parser.add_argument(
         "--input",
         help="list of array element positions",
-        required=True,
+        required=False,
     )
     config.parser.add_argument(
         "--input_meta",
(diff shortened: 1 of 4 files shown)


orelgueta (approved):
All good but codecov complains about lines missing a test. I let you decide if you want to implement that test.

GernotMaier:
I've seen the codecov issue, but I think adding a test wouldn't add a lot: this is in the `initialize`, which calls a couple of sub functions. There is no logic at this steps and all of the functions called are tested.

PR #720, opened 2023-12-07
Add pylint to pre-commit

GernotMaier:
The title should be clear.

Note that this works only with a local installation of pylint (see [here](https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html)). This is why I've added pylint as a package to be installed.

Please try it out. A single file takes for me 3-6s to lint with pylint, which is quite slow but probably still accessible.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,9 +22,21 @@ repos:
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
+  - repo: local
+    hooks:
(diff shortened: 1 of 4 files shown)


orelgueta:
I have an issue that locally I get errors which I don't see on GitHub, so I have to figure this out first.

orelgueta (approved):
OK, it all works and didn't take longer for a tiny typo. Let's see if it annoys us in the future but for now I suggest we merge.

PR #721, opened 2023-12-08
Add pytest-random-order to development environment. 

GernotMaier:
This is very useful to find hidden dependencies in unit test (e.g., right now #711 is passing tests in order, but not in random order).

Run `pytest --random-order -s --no-cov tests/unit_tests/test_db_handler.py` to see the mess.

### DIFF
--- a/docs/source/developer_guidelines.rst
+++ b/docs/source/developer_guidelines.rst
@@ -77,6 +77,9 @@ This should be used to avoid duplication.
 The `pytest-xdist <https://pytest-xdist.readthedocs.io/en/latest/>`_ plugin is part of the developer environment
 and can be used to run unit and integration tests in parallel (e.g., ``pytest -n 4`` to run on four cores in parallel).
 
+Tests might pass just because they run after an unrelated test. In order to test the independence of unit tests, use the
+`pytest-random-order <https://pypi.org/project/pytest-random-order/>`_ plugin with ``pytest --random-order``.
+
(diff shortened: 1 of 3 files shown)


GernotMaier:
It is a very useful tool for developers. As mamba/conda is such a mess, I have to re-install my environments regularly and forget to add these additional packages. It is only installed for developers, so not really a dependency of simtools (prod). 

We constantly add unit tests, so we should run it regularly with the random feature (this is why I've added it to the developer documentation).

Do you mean we should add it to the CI-unittest? That probably would also be useful.

VictorBarbosaMartins:
> it tests, so we should run it regularly with the random feature (this is why I've added it to the developer documentation).

I think that was what Orel meant. And yes, I agree adding a CI-unittest to randomize the order would be nice. 

By the way, I never use the -n option. Instead, I usually first run only the tests related to the code I am developing, so no need for the -n option every time.

Also, I get an error when running the command you sent inside the container:
```
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --random-order
  inifile: /workdir/external/simtools/pyproject.toml
  rootdir: /workdir/external/simtools
```

GernotMaier:
The container doesn't know anything about this pull request! So the package is not installed.

GernotMaier:
I am merging this (Orel complained, but approved...)

GernotMaier:
@VictorBarbosaMartins : the container should be updated now and if you pull a new one in a couple of minutes, this should work.

Issue #660, opened 2023-11-06
Problems with the old production tool


VictorBarbosaMartins:
Hello. I am listing here six problems that I have found with the old production tool `production.py`. I already looked in some of these, but before spending more time on them, I would like to raise a discussion whether this is necessary. @orelgueta  suggests using his new production tool `simulate_prod.py` and discontinue this old tool. However, the new tool is not yet able to simulate corsika alone. Given that I am moving my focus away from corsika and I am concentrating more on the simtel histograms, at the moment the corsika output is indeed not needed. The problem is that we need to have a tool to simulate corsika alone, otherwise the entire corsika output (corsika histograms) module does not make sense and we would not be able to compare the results of different corsika versions. Should we fix the issues mentioned below? Should we discontinue `production.py`? Should we implement corsika simulation within `simulate_prod.py`? Please share your thoughts.

### **Problems with the production tool**
- [x] print function not used anywhere:
https://github.com/gammasim/simtools/blob/5d966190935786cba19836dc2c2d1652a17336fe/simtools/simulator.py#L596-L622
- [ ] No error message (neither printed nor passed to .err log file) when running the production with `--array-only` in case no corsika simulation was done before (to reproduce it yourself, create the directory and try ```simtools-production --productionconfig ./tests/resources/prod_config_test.yml --task simulate --array_only --submit_command local --data_directory test_sim_telarray/ --output_path test_sim_telarray/```). When running manually the bash script generated by simtools, the error is printed out. This error should go to the log file:
```
Input file: /workdir/external/simtools/test_sim_telarray/corsika-data/North/gamma/data/run000002/corsika_run000002_gamma_za020deg_azm000deg_North_TestLayout_test-production.zst
/workdir/external/simtools/test_sim_telarray/corsika-data/North/gamma/data/run000002/corsika_run000002_gamma_za020deg_azm000deg_North_TestLayout_test-production.zst: No such file or directory
No (or only empty) input files have been processed.
```
[...]

Issue #645, opened 2023-10-20
production tool produces a bunch of files in local directory


VictorBarbosaMartins:
The production tool is producing a lot of `telarray_rand.config.used-...` files in the local directory, even when using plain output path and setting the output path. I would like to understand what are these files. If they are important, then they should be written in the correctly place. On the other hand, if they are temporary files, they should be excluded on the fly.

```simtools-production --task simulate --productionconfig prod_config_test.yml --submit_command local --primary gamma --array_only --use_plain_output_path --output_path test_sim_telarray/```

**Comment by VictorBarbosaMartins** [comment]
logfiles are also not saved in the given output_path:
```
INFO::job_manager(l98)::submit::Job log stream /workdir/external/simtools/simtel-data/North/gamma/log/log_sub_run000009_gamma_za020deg_azm000deg_North_TestLayout_test-production.job
```

PR #723, opened 2023-12-08
fix zero padding in old production tool

### LINKED ISSUE #646: production tool with `--array_option` likely not simulating events with `--array_only` option
It might that I am missing something but running the production tool with `--array_only` option and running the job locally is not doing anything. I would expect sim_telarray output somewhere:

```
simtools-production --task simulate --productionconfig prod_config_test.yml --submit_command local --primary gamma --array_only --use_plain_output_path --output_path test_sim_telarray/
```

VictorBarbosaMartins:
Closes #646, #645.

Related to [660](https://github.com/gammasim/simtools/issues/660). For some time the old production tool was not working due to mismatches in the names of the files produced by CORSIKA.

It looks like a simple solution but it took me some time. First I wanted to keep the zero padding, but it was not possible to add the zero padding to the .zst file being saved by the corsika runner. There is a mess with directories, env variables and functions with the same name in the config and runner scripts. Instead, I removed the zero padding from the .zst file being read by simtel_runner. Hopefully that does not break anything in the new production tool.

To review (the test that I was doing locally): 
run the shower and then the array simulator. Check log files and more specifically the zst files in the data directories
```
mkdir test_sim_telarray
simtools-production --task simulate --productionconfig prod_config_test.yml --submit_command local --primary gamma --showers_only --output_path test_sim_telarray/
```

```
simtools-production --task simulate --productionconfig prod_config_test.yml --submit_command local --primary gamma --array_only --use_plain_output_path --output_path test_sim_telarray/
```

Edit: This is also closing #645. Now a bunch of files from the random_generator will not be written locally but only a single  file! That is already much better than before.

### DIFF
--- a/simtools/corsika/corsika_runner.py
+++ b/simtools/corsika/corsika_runner.py
@@ -354,8 +354,8 @@ class CorsikaRunner:
             zenith = self.corsika_config.get_user_parameter("THETAP")[0]
             azimuth = self.corsika_config.get_user_parameter("AZM")[0]
             file_name = (
-                f"corsika_run{kwargs['run']:06}_{kwargs['primary']}_"
-                f"za{round(zenith):03}deg_azm{round(azimuth):03}deg_"
+                f"corsika_run{kwargs['run']}_{kwargs['primary']}_"
+                f"za{round(zenith)}deg_azm{round(azimuth)}deg_"
                 f"{kwargs['site']}_{kwargs['array_name']}{file_label}"
             )
             run_dir = self._get_run_directory(kwargs["run"])
--- a/simtools/corsika_simtel/corsika_simtel_runner.py
+++ b/simtools/corsika_simtel/corsika_simtel_runner.py
@@ -156,7 +156,7 @@ class CorsikaSimtelRunner(CorsikaRunner, SimtelRunnerArray):
         command += super()._config_option(
             "output_file", self.get_file_name("output", **info_for_file_name)
         )
-        command += super()._config_option("random_state", "auto")
+        command += super()._config_option("random_state", "none")
         command += super()._config_option("show", "all")
         command += f" {kwargs['input_file']}"
         command += f" | gzip > {self.get_file_name('log', **info_for_file_name)} 2>&1 || exit"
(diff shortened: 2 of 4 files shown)


GernotMaier:
@VictorBarbosaMartins - I would like to handover this one to Orel. I am simply not familiar with the naming conventions and what is happening here. Is this ok with you?

GernotMaier:
And maybe we discuss early next week what you call a 'mess' up there so that we can fix it.

VictorBarbosaMartins:
Sounds good!

orelgueta:
Hold on, we definitely want to have zero padding. So removing it cannot be the solution. I need to figure out first why you say it cannot be added in the case of CORSIKA simulations. I will come back and comment further later.

VictorBarbosaMartins:
> Hold on, we definitely want to have zero padding. So removing it cannot be the solution. I need to figure out first why you say it cannot be added in the case of CORSIKA simulations. I will come back and comment further later.

I don't see the reason why we would need the padding in the name of the zst file, which is the final product of the simtel_array simulation, but maybe you can explain me? It costed me some hours to try to implement the missing zero padding during the writing of the file, but it is extremely non-trivial. The place in the code where we name it is based on environmental variable, so we can't deal with the string since the string is not there yet. I am happy if you find another solution, but I wouldnt dedicate too much time on that.

orelgueta:
The word "need" is a strong one. Let's say it is desirable to have zero padding because it means that the files are ordered when you list them in the terminal for example. It also means all files have the exact same structure. This requirement for zero padding came from DPPS (in particular Karl and Max). It is an easy requirement to satisfy, so I don't see a reason why not.

Regarding the exact implementation, I will open a separate PR in a few minutes. It is true that the file name is defined right now using env variables, but that's just a choice Raul made at the time. The values are defined by the user and should be available. Let me see if I can do it quickly.

orelgueta:
OK, I realise now why Raul used the env variables. It is to support the simulation of a range of runs on the same machine (or output to the same directory).

I then thought of a way to implementing the zero padding even when using these env variables (copying from Konrad's implementation in the latest run script), but realised it is not really necessary. We do not plan to simulate a range of runs sequentially or using the batch farm in this way. It is functionality we do not really need and can also be implemented in the future without relying on the internal system of CORSIKA to advance the run numbers.

Therefore, I will open a separate simple PR that fixes this specific issue. Then I will open an issue to remove the entire option of simulating a range of runs (or we retire the old production script in general).

VictorBarbosaMartins:
Closed as already addressed in #735 and #736

PR #727, opened 2023-12-08
Add `skip_output_validation` flag to allow writing without validation against a schema file.

### LINKED ISSUE #717: Coordinate transformation to UTM with writing to file in application test is broken
Pull request #716 fixes a critical bug in reading the required columns from a schema and apply this to the table validation.

This has the results that above integration test is (rightly) not working anymore. This requires probably some adaption of the schema and how the schema is found.

Until this is solved: I've commented out this tests from `tests/integration_tests/test_applications.py`.

(nice that fixing a bug reveals other bugs)

GernotMaier:
Closes #717 

Coordinate transformations from corsika to utm were broken due to the strict introduction of output validation using a schema file. The schema for UTM coordinates is not yet defined.

Added with this pull request a flag in the `output` group to be able to skip the output validation at the stage of writing (flag is `--skip_output_validation`).

This feature will be useful on the long term, however we need to prepare the schemas for all input / outputs (this is anyway an ongoing effort and I therefore don't open an issue on this point).

### DIFF
--- a/simtools/data_model/model_data_writer.py
+++ b/simtools/data_model/model_data_writer.py
@@ -17,17 +17,26 @@ class ModelDataWriter:
 
     Parameters
     ----------
+    product_data_file: str
+        Name of output file.
+    product_data_format: str
+        Format of output file.
     args_dict: Dictionary
         Dictionary with configuration parameters.
     """
 
-    def __init__(self, product_data_file=None, product_data_format=None):
+    def __init__(self, product_data_file=None, product_data_format=None, args_dict=None):
         """
         Initialize model data writer.
         """
 
         self._logger = logging.getLogger(__name__)
         self.io_handler = io_handler.IOHandler()
+        if args_dict is not None:
+            self.io_handler.set_paths(
+                output_path=args_dict.get("output_path", None),
+                use_plain_output_path=args_dict.get("use_plain_output_path", False),
+            )
         try:
             self.product_data_file = self.io_handler.get_output_file(
                 file_name=product_data_file, dir_type="simtools-result"
@@ -37,7 +46,7 @@ class ModelDataWriter:
         self.product_data_format = self._astropy_data_format(product_data_format)
 
     @staticmethod
-    def dump(args_dict, metadata=None, product_data=None, validate_schema_file=False):
+    def dump(args_dict, metadata=None, product_data=None, validate_schema_file=None):
         """
         Write model data and metadata (as static method).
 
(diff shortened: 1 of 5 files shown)


tobiaskleiner (approved):
All good.

VictorBarbosaMartins:
> Closes #717
> 
> Coordinate transformations from corsika to utm were broken due to the strict introduction of output validation using a schema file. The schema for UTM coordinates is not yet defined.
> 
> Added with this pull request a flag in the `output` group to be able to skip the output validation at the stage of writing (flag is `--skip_output_validation`).
> 
> This feature will be useful on the long term, however we need to prepare the schemas for all input / outputs (this is anyway an ongoing effort and I therefore don't open an issue on this point).

Hi @GernotMaier. I am having a quick look at these changes. There are no integration tests for the `--export` argument using `corsika`, and `mercator` as parameter. I just checked if the application runs with that and it doesnt, also because of the validation. So I suggest including two new integration tests for these options, and using the new implemented flag `--skip_output_validation`

simtools/data_model/model_data_writer.py
```diff
@@ -17,17 +17,26 @@ class ModelDataWriter:
 
     Parameters
     ----------
+    product_data_file: str
+        Name of output file.
+    product_data_format: str
+        Format of output file.
```
VictorBarbosaMartins, line 23: Could we include an example here in the docstring, what would be a valid input?

tests/unit_tests/data_model/test_model_data_writer.py
```diff
@@ -12,6 +12,79 @@
 logger.setLevel(logging.DEBUG)
 
 
+def test_write(tmp_test_directory):
+    # both none (no exception expected)
+    w_1 = writer.ModelDataWriter()
+    w_1.write(metadata=None, product_data=None)
```
VictorBarbosaMartins, line 18: According to the function `write` this is supposed to have no effect, right? But it seems to me that in this test we are not checking/asserting anything. maybe we should check that no file was written?

GernotMaier:
@VictorBarbosaMartins  - reading only now mails.

We will have schema for the validation for all cases in future. So until then this has to applied to all but the ground coordinate transformation.

tests/unit_tests/data_model/test_model_data_writer.py
GernotMaier: But this is obvious and not needed to be tested.

What important is in this case (both none) that no Exception is raised.

PR #730, opened 2023-12-11
Set explicitly output path for `simulate_prod`

### LINKED ISSUE #680: Integration tests create a local directory `directory_for_grid_upload`
Test should not leave locally any data, this directory should probably be created in tmp dir.

GernotMaier:
Closes #680

Output paths are set explicitly for `simulate_prod`. Allows to run "cleaner" integration tests without spurious files leftover. 

Requires to add `--output_path ./` to obtain original behavior (meaning: writing the production configuration into the current directory).

### DIFF
--- a/tests/integration_tests/test_applications.py
+++ b/tests/integration_tests/test_applications.py
@@ -639,7 +639,7 @@ APP_LIST = {
             "--data_directory",
             "TESTOUTPUTDIR/",
             "--output_path",
-            "./",
+            "TESTOUTPUTDIR/",
         ]
     ],
     "simulate_prod::gamma_20_deg_pack_for_grid": [
--- a/simtools/applications/simulate_prod.py
+++ b/simtools/applications/simulate_prod.py
@@ -48,7 +48,9 @@
     run (int, required)
         Run number (actual run number will be 'start_run' + 'run').
     data_directory (str, optional)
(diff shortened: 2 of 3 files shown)


tests/integration_tests/test_applications.py
```diff
@@ -638,6 +638,8 @@
             "10",
             "--data_directory",
             "TESTOUTPUTDIR/",
+            "--output_path",
+            "./",
```
orelgueta, line 642: Changed this to be the output dir as well. Otherwise it writes to the local directory (sort of).

orelgueta (commented):
I made a small change to the `output_path` which I explained inline.
Also, I modified the behaviour of `data_directory`. Notice that the actual output of the simulation is written to the `data_directory` rather than the `output_directory`. At the moment I made the default behaviour such that they are almost the same, but I am wondering if I should merge the two entirely. There's merit to keeping them separate, so I tend to keep it as is for now. 

Since I made these changes, I do not approve yet. Please let me know if you don't mind them. If you don't, we can merge.

GernotMaier:
@orelgueta - ok, these changes are fine with me. Do you approve the PR so I don't have to bypass the branch protection?

GernotMaier:
I merge main and it looks clean now. I am waiting for the tests before merging.

PR #734, opened 2023-12-11
Add codecov configuration file to suppress failed tests when coverage falls by 0.1%

GernotMaier:
(title says everything).

Save us from checking what has failed and reduces noise.

Copied from https://github.com/astropy/astropy/blob/main/codecov.yml


VictorBarbosaMartins (approved):
Looks good. Let's see in practice whether this helps

PR #735, opened 2023-12-11
Zero padding corsika output file

### LINKED ISSUE #646: production tool with `--array_option` likely not simulating events with `--array_only` option

orelgueta:
Fixes #646

### DIFF
--- a/simtools/corsika/corsika_config.py
+++ b/simtools/corsika/corsika_config.py
@@ -489,8 +489,9 @@ class CorsikaConfig:
         if file_type == "config":
             return f"corsika_config_{file_name}.input"
         if file_type == "output_generic":
+            # The XXXXXX will be replaced by the run number after the pfp step with sed
             file_name = (
-                f"corsika_run{self._user_parameters['RUNNR'][0]:06}_"
+                f"corsika_runXXXXXX_"
                 f"{self.primary}_za{int(self._user_parameters['THETAP'][0]):03}deg_"
                 f"azm{int(self._user_parameters['AZM'][0]):03}deg"
                 f"_{self.site}_{self.layout_name}{file_label}.zst"
--- a/simtools/corsika/corsika_config.py
+++ b/simtools/corsika/corsika_config.py
@@ -477,20 +477,22 @@ class CorsikaConfig:
             )
         file_name = (
             f"{self.primary}_{self.site}_{self.layout_name}_"
-            f"za{int(self._user_parameters['THETAP'][0]):d}-"
-            f"az{int(self._user_parameters['AZM'][0]):d}deg"
+            f"za{int(self._user_parameters['THETAP'][0]):03}-"
+            f"azm{int(self._user_parameters['AZM'][0]):03}deg"
             f"{view_cone}{file_label}"
         )
 
         if file_type == "config_tmp":
             if run_number is not None:
-                return f"corsika_config_run{run_number}_{file_name}.txt"
+                return f"corsika_config_run{run_number:06}_{file_name}.txt"
             raise ValueError("Must provide a run number for a temporary CORSIKA config file")
         if file_type == "config":
             return f"corsika_config_{file_name}.input"
         if file_type == "output_generic":
             file_name = (
-                "corsika_run${RUNNR}_${PRMNAME}_za${ZA}deg_azm${AZM}deg"
+                f"corsika_run{self._user_parameters['RUNNR'][0]:06}_"
+                f"{self.primary}_za{int(self._user_parameters['THETAP'][0]):03}deg_"
+                f"azm{int(self._user_parameters['AZM'][0]):03}deg"
                 f"_{self.site}_{self.layout_name}{file_label}.zst"
             )
             return file_name
--- a/simtools/corsika/corsika_runner.py
+++ b/simtools/corsika/corsika_runner.py
@@ -241,6 +241,8 @@ class CorsikaRunner:
             if use_pfp:
                 file.write("\n# Running pfp\n")
                 file.write(pfp_command)
+                file.write("\n# Replacing the XXXXXX placeholder with the run number\n")
+                file.write(f"sed -i 's/XXXXXX/{run_number:06}/g' {corsika_input_tmp_file}\n")
             else:
                 file.write("\n# Copying CORSIKA input file to run location\n")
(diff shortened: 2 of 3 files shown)


VictorBarbosaMartins (changes requested):
That is exactly one of the things I tried before and it does not work. Every test is going to work because we are not testing it correctly. Check yourself by running first the simulation for the shower (for >1 events)
```
simtools-production --task simulate --productionconfig prod_config_test.yml --submit_command local --primary gamma --showers_only --output_path test_sim_telarray/
```
and then checking the output of the zst files:
```
ls -rtlah corsika-data/North/gamma/data/run00000*/*zst
```
to see that all the zst files are saved with the same name using the run number 1:
```
-rw-r--r-- 1 root root 84M Dec 11 15:43 corsika-data/North/gamma/data/run000001/corsika_run000001_gamma_za020deg_azm000deg_North_TestLayout_test-production.zst
-rw-r--r-- 1 root root 75M Dec 11 15:44 corsika-data/North/gamma/data/run000002/corsika_run000001_gamma_za020deg_azm000deg_North_TestLayout_test-production.zst
```

That is because `self._user_parameters['RUNNR'][0]` gives you only the starting point of the simulation, hence, it cannot be used interchangeably with the environmental variable `${RUNNR}`

Feel free if you want to try to find another solution, or alternatively accept mine in #723.

orelgueta:
Yes, I know it doesn't exactly solve the problem, I mentioned it in #723. I quote my explanation below.

> OK, I realise now why Raul used the env variables. It is to support the simulation of a range of runs on the same machine (or output to the same directory).

> I then thought of a way to implementing the zero padding even when using these env variables (copying from Konrad's implementation in the latest run script), but realised it is not really necessary. We do not plan to simulate a range of runs sequentially or using the batch farm in this way. It is functionality we do not need and can also be implemented in the future without relying on the internal system of CORSIKA to advance the run numbers.

So, once again, I suggest to accept this simplified solution and then remove the entire option of simulating a range of runs (or we retire the old production script in general). We can do that in a separate PR.

VictorBarbosaMartins:
> Yes, I know it doesn't exactly solve the problem, I mentioned it in #723. I quote my explanation below.
> 
> > OK, I realise now why Raul used the env variables. It is to support the simulation of a range of runs on the same machine (or output to the same directory).
> 
> > I then thought of a way to implementing the zero padding even when using these env variables (copying from Konrad's implementation in the latest run script), but realised it is not really necessary. We do not plan to simulate a range of runs sequentially or using the batch farm in this way. It is functionality we do not need and can also be implemented in the future without relying on the internal system of CORSIKA to advance the run numbers.
> 
> So, once again, I suggest to accept this simplified solution and then remove the entire option of simulating a range of runs (or we retire the old production script in general). We can do that in a separate PR.

I think we need to discuss in person. We have not decided to discontinue the tool, which I am against, and for me it does not make sense to implement something that does not really fix the issue.

orelgueta:
But removing this functionality of simulating a range of runs does not discontinue the tool or break the possibility of simulating CORSIKA and/or sim_telarray separately. It just removes the functionality of simulating a few runs sequentially. 

This type of functionality is simply not needed (why wouldn't we run in parallel instead?). When running on the batch farm it might be useful, but we can achieve the same behaviour by copying the grid behaviour and just submitting multiple jobs with changing run numbers instead of using the internal CORSIKA (IACT actually) counting mechanism.

VictorBarbosaMartins:
Your proposal here is changing the name of the `.zst` output file of CORSIKA to include the zero padding but you are giving the very same name of the file to every output file (run) `corsika_run000001_gamma_za020deg_azm000deg_North_TestLayout_test-production.zst` and that is just not correct for the current implementation.

Independent of that, we still need to correctly retrieve the runs in `simtel_runner` one by one and that is still not possible because `simtel_runner` writes into `simtel-data/North/gamma/data/` the `.zst` output file based on the name of the CORSIKA `.zst` output file it received. If the name is the same,  `simtel_runner` is **overwriting** at each step its output file into `simtel-data/North/gamma/data/`. In the end, we have a single output file named `run000001_gamma_za020deg_azm000deg_North_TestLayout_test-production.simtel.zst`.

If you want to insist in this PR, I suggest you to implement a full working example, i.e., running the production tool with `--showers-only` and then with `--array-only`. I would only agree that it works if a number of simtel output `.zst` files are produced and if the output is not the same for every run. And again, I point you to the solution in #723 in which this is already possible!

orelgueta:
OK, let me remove the option of simulating multiple runs with one call in this PR and then the example you mentioned will work. I will not include simulating multiple runs because it is not necessary functionality.

The solution in #723 goes against the DPPS decision for file names, so it cannot be accepted.

orelgueta:
OK, further discussions made me look for another option and I found a fairly simple (but "hacky" one). I just set a place holder in the input template file and then replace it with sed. This keeps the functionality of running on a range of runs and the zero padding. It's ugly, but it works.

@VictorBarbosaMartins could you please test and say if it works for you (and if you are not horrified by the hack)? 
@GernotMaier since this is really a hack, I wouldn't mind getting also your opinion, to make sure everyone's on board with it.

VictorBarbosaMartins:
I don't mind that it is a bit uglier now, as it seems to be working. With this, the old production tool is again functional and has a zero padding (did not think this would be so important.). I tested locally as I suggested before and, therefore, I am approving.

As a side note, the tests are still not sufficient to detect whenever the tool gets broken again. So, I will wait for us to merge #726 before implementing better tests.

VictorBarbosaMartins (approved):
I am approving but maybe @GernotMaier still wants to comment?

GernotMaier:
Fine with me - and I am not making any comments about the super-long run numbers planned...

orelgueta:
> Fine with me - and I am not making any comments about the super-long run numbers planned...

Well, it's either you limit to a 100,000, which is too low (we already simulate 40,000 proton runs in each production) or to 1 million. I vote for the latter.

Merging, thanks for the quick review!

PR #736, opened 2023-12-11
Random state set to none

### LINKED ISSUE #645: production tool produces a bunch of files in local directory

orelgueta:
Set the random state to "none" in order to reduce the number of `telarray_rand.conf.used` files to just one.
Fixes #645.

### DIFF
--- a/simtools/corsika_simtel/corsika_simtel_runner.py
+++ b/simtools/corsika_simtel/corsika_simtel_runner.py
@@ -156,7 +156,7 @@ class CorsikaSimtelRunner(CorsikaRunner, SimtelRunnerArray):
         command += super()._config_option(
             "output_file", self.get_file_name("output", **info_for_file_name)
         )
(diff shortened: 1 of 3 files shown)


orelgueta:
I opened this one because I suggested to close #723 and didn't want to ask you to implement it again in a separate PR. Let's decide first what we do and then proceed with this PR.

VictorBarbosaMartins (approved):

VictorBarbosaMartins:
This is independent of the other one and can be merged

PR #739, opened 2023-12-12
Remove TODOs and FIXMEs

### LINKED ISSUE #715: Fix TODOs and FIXMEs in the code or replace by issues.
There should be no TODO or FIXME statements in the code. These should be replaced by issues.

orelgueta:
Most removals were trivial and only one required opening an issue (but it is quite broad). 

A few were removed because it is anyway planned to implement the functionality mentioned. 

In one case a note was added to the function instead to mention that it might be removed in the future. If anyone has a better idea how to do it, I am happy to hear it (I thought about marking it as deprecated, but it seems premature to do it here).

One TODO is left in the code but it should be fixed in #718.

Fixes #715.

### DIFF
--- a/simtools/applications/compare_cumulative_psf.py
+++ b/simtools/applications/compare_cumulative_psf.py
@@ -71,9 +71,6 @@
 
         d80 in cm = 3.3662565358159013
 
-    .. todo::
-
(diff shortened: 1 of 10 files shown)


GernotMaier (approved):
All good, tahnks!

PR #740, opened 2023-12-12
Switch Current to released

### LINKED ISSUE #416: Change "Current" to "Released"
Change the name "Current" version to "Released" to make the meaning clearer.

orelgueta:
This PR changes from the name "Current" for the latest "Released" telescope model to the name "Released" cause it's clearer. 
Fixes #416.

### DIFF
--- a/tests/unit_tests/model/test_telescope_model.py
+++ b/tests/unit_tests/model/test_telescope_model.py
@@ -7,7 +7,6 @@ import numpy as np
 import pytest
 from astropy import units as u
 
-import simtools.utils.general as gen
 from simtools.model.telescope_model import InvalidParameter, TelescopeModel
 
 logger = logging.getLogger()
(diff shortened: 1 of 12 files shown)


VictorBarbosaMartins (approved):
Ok, all good

tests/unit_tests/model/test_telescope_model.py
```diff
@@ -7,26 +7,16 @@
...
 from astropy import units as u
 
-import simtools.utils.general as gen
 from simtools.model.telescope_model import InvalidParameter, TelescopeModel
 
 logger = logging.getLogger()
 logger.setLevel(logging.DEBUG)
 
 
 @pytest.fixture
```
tobiaskleiner, line 16: Is this really needed just to return a string of the filepath?

orelgueta:
Thanks, I will wait for the tests to pass and then merge.

tobiaskleiner (approved):
Approved. Good for merging.

tests/unit_tests/model/test_telescope_model.py
orelgueta: pytest fixtures are very useful in general (I won't give details here, but it's worth reading about). In this case it is used as a nice way to define a constant for all tests in the module that need it. It is not really the intended use case and It is much more powerful than that, but since we use it everywhere for other things, we might as well use it here as well.

orelgueta:
OK the integration test failed because I just replaced a file in the DB while it was running (stupid of me). I will rerun it just to be safe and then merge.

PR #742, opened 2023-12-12
Add mirror shape diameter to db

### LINKED ISSUE #419: Add mirror shape and diameter to DB
https://github.com/gammasim/gammasim-tools/blob/8cdb37a97cf9c4fcf37ed0380927c28e4735cb3a/simtools/model/mirrors.py#L60

orelgueta:
Fixes #419.
The issue was not really related to the code but to add parameters to the DB. However, I took advantage and improved the way we add parameters to the DB (even though we do not plan to do it often).
This resulted also in the addition of two general functions in `general.py`. Since I have seen somewhat similar code in @GernotMaier PR lately, I mention you explicitly and ask you to review. Perhaps you see a way to use these functions elsewhere or modify them such that they are useful (I tried to look quickly, but it's not immediately obvious to me).


simtools/utils/general.py
```diff
@@ -808,3 +809,59 @@ def sort_arrays(*args):
         _, value = zip(*sorted(zip(order_array, arg)))
         new_args.append(list(value))
     return new_args
+
+
+def extract_type_of_value(value) -> str:
```
GernotMaier, line 814: Do we need this as well for astropy Quantities? Do we want to keep them in the long style?

```
>>> type(3.*u.m)
<class 'astropy.units.quantity.Quantity'>
>>> type(u.m)
<class 'astropy.units.core.IrreducibleUnit'>
```

Or is the class always connect to get_value_unit_type which takes Quantities apart? If yes, maybe add a remark in the docstring.

GernotMaier (approved):
All good - minor comment only. Approved.

simtools/utils/general.py
orelgueta: I don't see how it would be useful for astropy quantities. This is really only useful in order to save in the DB the type in a way that `type()` can then be applied directly.
It doesn't have to be related to `get_value_unit_type`, so a comment in the docstring is probably not necessary. However, I could add some sort of "protection" that if a Quantity is provided, it raises an error. What do you think?

GernotMaier: This would be fine with me - but you will have to judge if this is useful. If not, just leave it and merge.

orelgueta: OK, I added a not implemented error for astropy units. Since tests passed before on Github and now they passed again locally, I won't wait for them to pass and just merge.

PR #747, opened 2023-12-15
Old production tool real test

### LINKED ISSUE #741: improve testing of old production tool after merging #726
I don't mind that it is a bit uglier now, as it seems to be working. With this, the old production tool is again functional and has a zero padding (did not think this would be so important.). I tested locally (merging #740 before) as I suggested before and, therefore, I am approving.

As a side note, the tests are still not sufficient to detect whenever the tool gets broken again. So, I will wait for us to merge #726 before implementing better tests.

_Originally posted by @VictorBarbosaMartins in https://github.com/gammasim/simtools/issues/735#issuecomment-1852322878_

VictorBarbosaMartins:
Closes #741.

After getting the old production tool (`simtools-production`) to life, now it is time to test it properly. In this PR, I made some changes in the tool to allow it running simtel_array for specific input CORSIKA files. Before, only a rather non-trivial structure of sub-directories was allowed, which made difficult to test the tool with `--array_only` option. 

Major changes:
- It is now possible to run e.g. `simtools-production --productionconfig ./tests/resources/prod_config_test.yml --task simulate --array_only --submit_command local --output_path . --corsika_files tests/resources/tel_output_10GeV-2-gamma-20deg-CTAO-South.zst` and check the output in `ls simtel-data/North/proton/data/run000001_proton_za020deg_azm000deg_North_TestLayout_test-production.simtel.zst`
- This new feature is tested by checking whether the output file exists (expanded a bit the test_applications_from_config). This is good and we should use for all the applications which generate output files and are not ecsv.

Remaining point, which I deem for unimportant is whether we want or not to change the name of the test files to include the zero padding. We have a couple of files in resources still in the old format, e.g. `tests/resources/run1_proton_za20deg_azm0deg_North_1LST_test-lst-array.corsika.zst`.


tests/unit_tests/corsika/test_corsika_runner.py
```diff
@@ -43,7 +43,7 @@ def corsika_runner(corsika_config_data, io_handler, simtel_path, db_config):
 @pytest.fixture
 def corsika_file(io_handler):
     corsika_file = io_handler.get_input_data_file(
-        file_name="run1_proton_za20deg_azm0deg_North_1LST_test-lst-array.corsika.zst", test=True
+        file_name="run1_proton_za20deg_azm0deg_north_1lst_test-lst-array.corsika.zst", test=True
```
orelgueta, line 46: Please do not change the name of the file, it is produced with the specific naming and capitalisation and should not change unless there is good reason for it. In this case it seems to be a bug that changes the name of the file, not an actual good reason.

tests/unit_tests/simtel/test_simtel_runner_array.py
```diff
@@ -49,7 +49,7 @@ def simtel_runner(array_model, simtel_path):
 @pytest.fixture
 def corsika_file(io_handler):
     corsika_file = io_handler.get_input_data_file(
-        file_name="run1_proton_za20deg_azm0deg_North_1LST_test-lst-array.corsika.zst", test=True
+        file_name="run1_proton_za20deg_azm0deg_north_1lst_test-lst-array.corsika.zst", test=True
```
orelgueta, line 52: Same comment as above:
> Please do not change the name of the file, it is produced with the specific naming and capitalisation and should not change unless there is good reason for it. In this case it seems to be a bug that changes the name of the file, not an actual good reason.

orelgueta (changes requested):
A few minor issues and the file name point which points to a bug somewhere.

simtools/applications/production.py
```diff
@@ -270,7 +286,24 @@ def main():
...
+            else:
+                if not isinstance(args_dict["corsika_files"], list):
+                    args_dict["corsika_files"] = [args_dict["corsika_files"]]
+                input_list = args_dict["corsika_files"]
+
+            for corsika_file in input_list:
+                if not Path(corsika_file).exists():
+                    msg = (
+                        f"CORSIKA file {corsika_file} does not exist. Please run the "
+                        f"production with the `--showers_only` option first or point the "
```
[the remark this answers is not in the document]
VictorBarbosaMartins: There is one variable: `{corsika_file}` maybe you didn't see it because it is an iterator

VictorBarbosaMartins:
Hi Orel. Thanks for the comments. I addressed your comments and returned the name to what it was before. I had not seen before the `.lower()` in the argument of the new parameter, I think that might solve the problem I was having before.

orelgueta (approved):

PR #750, opened 2023-12-19
Metaschema updates

GernotMaier:
This is one of several pull requests coming up in relation to schema and data validation (probably not this week). This pull request renames and adds a file without changes in functionality.

- renaming of `metadata.schema.yml` to `metadata.metaschema.yml`: learned from Karl this week that the schemas describing other schemas are called "metaschema". Renaming will clarify this and hopefully resolves some of the confusion on the naming

- move the metaschema describing the schema for input data and model parameters from [workflows/schemas/jsonschema.yml](https://github.com/gammasim/workflows/blob/main/schemas/jsonschema.yml) into [simtools/schemas/data.metaschema.yml](simtools/schemas/data.metaschema.yml).
   - this way metaschema for data and metadata are in one central place
   - will remove the file in the workflow repository to avoid duplication (as soon as this one has merged)
   - this file will be used in future for the data validation (next step)

Changed also the examples, unit, and integration tests which read the metadata.metaschema.yml file.

### DIFF
--- a/simtools/applications/generate_default_metadata.py
+++ b/simtools/applications/generate_default_metadata.py
@@ -8,7 +8,7 @@
     ----------------------
     schema (str, optional)
         Schema file describing the input data
-        (default: simtools/schemas/metadata.schema.yml)
+        (default: simtools/schemas/metadata.metaschema.yml)
     output_file (str, optional)
         Output file name.
 
@@ -17,7 +17,7 @@
     .. code-block:: console
 
         simtools-generate-default-metadata
-            --schema simtools/schemas/metadata.schema.yml
+            --schema simtools/schemas/metadata.metaschema.yml
             --output_file default_metadata.yml
 
 
--- a/simtools/schemas/data.metaschema.yml
+++ b/simtools/schemas/data.metaschema.yml
@@ -1,0 +1,395 @@
+---
+$schema: http://json-schema.org/draft-06/schema#
+$ref: '#/definitions/SimtoolsDataAndModelParameters'
+title: SimPipe Data and Model Parameter Metaschema
+description: YAML representation of data and model parameter metaschema
+version: 0.1.0
+name: data.metaschema
+type: object
+additionalProperties: false
+
+
+definitions:
+  SimtoolsDataAndModelParameters:
+    type: object
+    additionalProperties: false
+    properties:
+      title:
+        type: string
+        description: "title for schema file"
+      version:
+        type: string
+        description: "version of this schema file (semver)"
+      base_schema:
+        type: string
+        description: "name of base schema (e.g. simpipe-schema)"
+      base_schema_url:
+        type: string
+        description: "url to base schema definition"
+      base_schema_version:
+        type: string
+        description: "version of base schema applied to this file (semver)"
+      name:
+        type: string
+        description: "name of model parameter"
+      developer_note:
+        type: string
+        description: "notes for developers"
+      short_description:
+        type: string
+        description: "short description of model parameter"
+      description:
+        type: string
+        description: "description of model parameter"
+      instrument:
+        $ref: '#/definitions/Instrument'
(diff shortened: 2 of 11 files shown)


simtools/schemas/data.metaschema.yml
GernotMaier, line 1: Note that this file has been moved without changes (apart from the title description) from https://github.com/gammasim/workflows/blob/main/schemas/jsonschema.yml

simtools/applications/generate_default_metadata.py
```diff
@@ -8,7 +8,7 @@
     ----------------------
     schema (str, optional)
         Schema file describing the input data
-        (default: simtools/schemas/metadata.schema.yml)
+        (default: simtools/schemas/metadata.metaschema.yml)
```
VictorBarbosaMartins, line 11: By doing a `grep -r "metadata.schema"` in the main directory I still see some references to `metadata.schema` in `tests/unit_tests/data_model/test_metadata_collector.py` and `tests/unit_tests/data_model/test_metadata_model.py`. I believe we want to change those too.

VictorBarbosaMartins (approved):
I have only one comment but I am already approving. As the metadata schema was simply moved from one place to another I have not checked in depth its content.

simtools/applications/generate_default_metadata.py
GernotMaier: Which one do you mean? Those like `array_coordinates.schema.yml` needs to stay, as they are actual schemas (and not schemas describing schemas).

VictorBarbosaMartins: No, I meant these lines which you can get if you run `grep -r "metadata.schema"`. These lines still refer to schema of the metadata, and we now know that they are metaschema, so it is appropriate to change the names there too:

- https://github.com/gammasim/simtools/blob/8bfda9f58c77f76f52a195ca8d6a6b9fa4fa73ef/tests/unit_tests/data_model/test_metadata_collector.py#L62
- https://github.com/gammasim/simtools/blob/8bfda9f58c77f76f52a195ca8d6a6b9fa4fa73ef/tests/unit_tests/data_model/test_metadata_collector.py#L66
- https://github.com/gammasim/simtools/blob/8bfda9f58c77f76f52a195ca8d6a6b9fa4fa73ef/tests/unit_tests/data_model/test_metadata_model.py#L22-L24

GernotMaier: I think you have a different or outdated branch on disk, I cannot find these issues. Can you check again?

VictorBarbosaMartins:
Ok, the grep I did in the right branch but I copied the lines above from main. Then let me be clearer about my suggestion:

I would replace all the references to `schema` in [simtools/data_model/metadata_collector.py](https://github.com/gammasim/simtools/blob/bc06d57910525a252b9abe9bd45f5f25aa0a6899/simtools/data_model/metadata_collector.py) to `meta_schema`. These names are in attributes and methods. For example, 

https://github.com/gammasim/simtools/blob/bc06d57910525a252b9abe9bd45f5f25aa0a6899/simtools/data_model/metadata_collector.py#L49-L50

should become
```
        self.meta_schema_file = None
        self.meta_schema_dict = None
```

These changes would have to be propagated to the tests as well.

The same would happen to `simtools/data_model/metadata_model.py`, for example, in :

https://github.com/gammasim/simtools/blob/bc06d57910525a252b9abe9bd45f5f25aa0a6899/simtools/data_model/metadata_model.py#L39

would become

```
    meta_schema, meta_schema_file = _load_meta_schema(meta_schema_file)

```

GernotMaier:
The variable should stay as `schema` only, without the `meta` - this code is working for both schema and metaschema files and therefore kept as it is. Is this ok?

VictorBarbosaMartins:
Ok, if you prefer this way. But note the name of files are referring to metadata only.

PR #752, opened 2023-12-19
Reading of metadata from table header; validation of input data; introduction of data_reader

GernotMaier:
Improves reading of input data and metadata and validation of schema:

- metadata can now be read as an additional option from an astropy table header (table.meta); before: metadata had to be given as separate yaml file.
- reading of astropy table now through a generic module [simtools/data_model/data_reader.py](simtools/data_model/data_reader.py), which provides both reading of input data and metadata plus validation. In the simplest case, this command will return a table which has been verified against a schema listed in the table.meta of the ecsv file:

```
table = data_reader.read_table_from_file(file_name=telescope_list_file, validate=True)
```

This will be extended in future to reading of json-type structures (as given for single values from the DB), so don't worry about a single-function module in this case of simtools/data_model/data_reader.py.

This brought up the follow issue:

- the schema validation of the input files revealed that many of the test files don't follow the expected standards (`pos_x` vs `position_x` etc.). This is fixed now, but quite a number of layout files have been touched. **You only have to look at one during the review!**

This pull requests tries also to improve the coverage of the unit tests for the modules touched.


simtools/data_model/metadata_collector.py
```diff
@@ -35,7 +38,7 @@ class MetadataCollector:
 
     """
 
-    def __init__(self, args_dict, data_model_name=None):
+    def __init__(self, args_dict, metadata_file_name=None, data_model_name=None):
```
GernotMaier, line 41: This covers the case where the meatdata_collector is not called from an application with all in the required information in `args_dict`. Use `metadata_file_name`, we can simple read a file and extract the metadata.

simtools/data_model/metadata_model.py
```diff
@@ -66,9 +66,7 @@ def get_default_metadata_dict(schema_file=None, observatory="CTA"):
 
     """
```
GernotMaier, line 68: This debug statement was duplicated, as exactly the same one is printed in _load_schema

GernotMaier:
@VictorBarbosaMartins - ignore the failing test for docker-simtools. This is related to #755 and is solved elsewhere.

simtools/data_model/metadata_collector.py
```diff
@@ -228,23 +248,46 @@ def _read_input_metadata_from_file(self):
...
+                    in_yaml=metadata_file_name, in_dict=None
+                )
+            except (gen.InvalidConfigData, FileNotFoundError):
+                self._logger.error("Failed reading metadata from %s", metadata_file_name)
+                raise
+        # metadata from table meta in ecsv file
+        elif Path(metadata_file_name).suffix == ".ecsv":
+            try:
+                _input_metadata = {observatory: Table.read(metadata_file_name).meta[observatory]}
+            except (FileNotFoundError, KeyError):
```
VictorBarbosaMartins, line 283: In general my impression is that we are too acceptive with the errors without distinguishing between them. For example, FileNotFoundError and KeyError are pretty different errors, and it would help to deal with these cases separately, as we are not limited to one `except` statement.

simtools/data_model/metadata_model.py
```diff
@@ -43,7 +43,7 @@ def validate_schema(data, schema_file):
     except jsonschema.exceptions.ValidationError:
         _logger.error(f"Failed using {schema}")
         raise
-    _logger.debug(f"Succeeded using {schema_file}")
+    _logger.debug(f"Successful validation of data using schema from {schema_file}")
```
VictorBarbosaMartins, line 46: not too important. One could point to which data was validated as well

simtools/data_model/model_data_writer.py
```diff
@@ -113,33 +95,40 @@ def validate_and_transform(self, product_data=None, validate_schema_file=None):
...
         Raises
         ------
         FileNotFoundError
-            if data writing was not sucessfull.
+            if data writing was not successful.
 
         """
 
+        if product_data is None:
+            return
```
VictorBarbosaMartins, line 117: I would prefer to include a warning here to raise the awareness that nothing was written (although one used a `write` function). To be honest, I did not understand the reasoning behind it. Why not making `product_data` a required argument?

simtools/layout/array_layout.py
```diff
@@ -559,61 +568,42 @@ def _load_telescope_list(self, table):
...
         """
-        Read list of telescopes from a ecsv file.
+        Initialize the Layout array from a telescope list file.
 
         Parameters
         ----------
         telescope_list_file: str or Path
             Path to the telescope list file.
+        telescope_list_metadata_file: str or Path
+            Path to the telescope list metadata file.
```
VictorBarbosaMartins, line 589: We are missing the validate parameter in the docstring

tests/integration_tests/config/simulate_prod_gamma_20_deg_pack_for_grid.yml
```diff
@@ -14,3 +14,4 @@ CTA_SIMPIPE:
     DATA_DIRECTORY: simtools-output
     OUTPUT_PATH: simtools-output
     PACK_FOR_GRID_REGISTER: true
+    LOG_LEVEL: DEBUG
```
VictorBarbosaMartins, line 17: probably we don't need this anymore, right?

tests/resources/telescope_positions-North-utm-withoutmeta.ecsv
```diff
@@ -0,0 +1,54 @@
...
+# - EPSG: 32628
+# - comments: |-
+#       following CTA-SPE-PSC-000000-0001 1c
+#       Subset of positions defined in reference repository
+#       https://gitlab.cta-observatory.org/cta-science/array-element-positions/-/tree/main
+#       Array center coordinates defined as centre of the 4 LST
+#       position (this is slightly different to the definition
+#       in CTA-SPE-PSC-000000-0001 1c)
+#       Never change positions here. This should always be
+#       synchronized with the reference repository
```
VictorBarbosaMartins, line 28: There is still a metadata, just not as defined by the schema, though I am not sure this comment makes a difference.

VictorBarbosaMartins (commented):
Hi @GernotMaier. Thank you for this implementation. This is going to help us standardise more and more input and output files. I provide some comments below and will wait with the approval until you address/answer them. 

I had a rather quick look through the tests but as they are increasing coverage, I am fine with them.

GernotMaier:
@VictorBarbosaMartins - I tried to look and answer / follow up all your comments. Please let me know.

And thanks for the review!

VictorBarbosaMartins:
Hi @GernotMaier, thanks for looking into the comments. I closed the conversations that I could see you already addressed, but I did not see the answers to the many remaining comments. Is this still coming?

GernotMaier:
@VictorBarbosaMartins - don't see this, which conversations are not addressed?

simtools/data_model/metadata_collector.py
VictorBarbosaMartins: Ok, thanks.

simtools/data_model/data_reader.py
```diff
@@ -0,0 +1,68 @@
+import logging
+
+from astropy.io.registry.base import IORegistryError
+from astropy.table import QTable
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Ok, and what about the `Table` instances in the rest of the code? Don't we want to have a consistency use of astropy Table throughout? (wouldn't need to be in this PR though)

simtools/data_model/data_reader.py
```diff
@@ -0,0 +1,68 @@
...
+
+from simtools.data_model import validate_data
+from simtools.data_model.metadata_collector import MetadataCollector
+
+__all__ = ["read_table_from_file"]
+
+_logger = logging.getLogger(__name__)
+
+
+def read_table_from_file(file_name, schema_file=None, validate=False, metadata_file=None):
```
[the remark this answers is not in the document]
VictorBarbosaMartins: ok, makes sense.

simtools/data_model/data_reader.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Read astropy table from file and validate against schema.
```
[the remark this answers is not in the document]
VictorBarbosaMartins: Ok, this is good then.

simtools/data_model/data_reader.py
```diff
@@ -0,0 +1,68 @@
...
+    Metadata is read from metadata file or from the metadata section of the data file.
+    Schema for validation can be given as argument, or is determined
+    from the metadata associated to the file.
```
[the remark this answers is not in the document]
VictorBarbosaMartins: ok!

simtools/data_model/data_reader.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Returns:
+    --------
+    astropy Table
+        Table read from file.
+
+    Raises
+    ------
+    FileNotFoundError
+        If file does not exist.
```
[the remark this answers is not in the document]
VictorBarbosaMartins: ok!

simtools/data_model/metadata_collector.py
```diff
@@ -8,6 +8,9 @@
 import datetime
 import getpass
 import logging
+from pathlib import Path
+
+from astropy.table import Table
```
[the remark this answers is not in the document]
VictorBarbosaMartins: But the QTable also allows one to access the metadata through `.meta`, so I still don't see a reason to use both classes. Maybe the performance could be a bit worse for QTable but I believe this wouldn't matter too much.

simtools/data_model/metadata_collector.py
```diff
@@ -228,23 +248,46 @@ def _read_input_metadata_from_file(self):
...
         """
 
-        if self.args_dict is None or self.args_dict.get("input_meta", None) is None:
+        try:
+            metadata_file_name = (
+                self.args_dict.get("input_meta", None)
+                if metadata_file_name is None
+                else metadata_file_name
+            )
+        except TypeError:
```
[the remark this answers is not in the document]
VictorBarbosaMartins: ok!

simtools/data_model/metadata_collector.py
VictorBarbosaMartins: That is fine, the code does what it is supposed to, and I guess each case inside the if statements is already being tested, so all good.

simtools/data_model/metadata_model.py
VictorBarbosaMartins: ok, then let's leave it like it is.

simtools/data_model/model_data_writer.py
VictorBarbosaMartins: One wouldn't need an if statement if the argument is required, it would be actually less code than the current version. Personally, I wouldn't like if I run a function that is supposed to write something and I don't get something written in the end without being told about it. But if you think this is style, then we can go ahead.

simtools/data_model/data_reader.py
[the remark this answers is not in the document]
GernotMaier: The plan is to replace all (maybe most - let's see if all is possible) Table reader calls by `data_reader` class calls. This is something to be done in future (this PR is already big), this way we automatically get consistency.

simtools/data_model/metadata_collector.py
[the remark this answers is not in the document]
GernotMaier: Yes, but QTable signals that we want to have quantities. This is what we don't want to have in this case, so I think it is less confusing to call Tables here.

Issue #642, opened 2023-10-18
Create application to calculate the event rate


orelgueta:
It would be good to have an application that calculates the trigger rate of the arrays (and sub-arrays). An example table with such rates can be found in https://www.mpi-hd.mpg.de/hfm/CTA/MC/Prod3/Config/rates-post-prod3b.pdf
Some of the calculation details are provided in the table in the link. Further details would probably require asking KB.

VictorBarbosaMartins:
Is it ok to copy Konrad's answer here just to make sure we document it and relate to this issue?

orelgueta:
Yes of course. And let me add one more piece of information that we realised later, the transition from read out pixels to data rate is done by assuming that each sample read is 2 Bytes long. This is marked in the R1 document as unsigned short.

VictorBarbosaMartins:
## Important email from Konrad to help developing this tool:
```Hi Orel,

That would require very small productions for each site,
a) single-telescope for each telescope type, and
b) the intended layout, without any extras,
for all relevant primaries.

You only need the histogram files, adding up all the files
for the same configuration (telescope-type or layout, primary)
with the 'add_histograms' tool into a single histogram file.
(For statistical errors on the evaluated rates, you can add
histograms from multiple sub-groups of runs and evaluate each
of them the same way as indicated below. Get the r.m.s. of the
sub-group rates and divide it by the square root of the
number of sub-groups.)

I think histogram IDs 1006 and 1007 is what you need, for CORSIKA
compiled with the usual VOLUMEDET flag (or whatever it is called).
IDs 6 and 7 would also work, if just looking at the ratio - with the
usual 20 bins per decade the difference between weighted and
unweighted is tiny. Just don't mix the pairs up.
The radius into which events were generated can be evaluated
from the X projection of histogram 1006 (or 6), the fraction of triggered
events from the ratio of the Y projection of 1007 (or 7) over the Y
projection of 1006 (or 6), in bins of lg(E). Multiply with the area thrown
and, in each energy bin, the integral of the assumed primary
particle flux from bin lower edge to bin upper edge (in terms
of energy E instead of lg(E)).
Add up all the resulting bin-wise rates. Done.
I used to get the numbers with the help of the 'rht' tool
(in hessio(xxx); needs the 'iactio' library; not built by default;
use 'make IACTIOPROG=rht rht' to build)
but the idea behind it should be simple enough that you don't
need that tool. Without the area and flux part of the calculation,

[...]

PR #754, opened 2023-12-21
Simtel histogram operations

VictorBarbosaMartins:
This PR addresses #642. Here I start the development of a tool to estimate the trigger rate for CTA. In this very first step, I only expand the `simtel_histograms` module, following Konrad's recipe as described in #642.

Main changes:
- Expanded integration tests of `simtel_histograms` to check the existence of the expected output file;
- Implemented functions in `simtel_histograms` to calculate the trigger rate for all the histograms passed to the Class.

Overall progress and next steps can be found in  #642.


simtools/simtel/simtel_histograms.py
```diff
@@ -67,50 +69,184 @@ def get_histogram_title(self, i_hist):
...
+        for key_to_test in [
+            "lower_x",
+            "upper_x",
+            "n_bins_x",
+            "title",
+        ]:
+            if first_hist_file[key_to_test] != second_hist_file[key_to_test]:
+                msg = "Trying to add histograms with inconsistent dimensions"
+                self._logger.error(msg)
+                raise BadHistogramFormat(msg)
```
GernotMaier, line 121: Tiny comment on the naming and you are free to ignore "BadHistogramFormat" indicates that something is wrong with the format of a histogram, while here you are testing for inconsistent dimensions. Maybe `InconsistentHistogramFormat`?

simtools/simtel/simtel_histograms.py
```diff
@@ -67,50 +69,184 @@ def get_histogram_title(self, i_hist):
...
-                            )
-
-                    n_files += int(count_file)
+                    this_combined_hist["data"] = np.add(this_combined_hist["data"], hist["data"])
+
+            n_files += int(count_file)
 
         self._logger.debug(f"End of reading {n_files} files")
 
+    def _derive_trigger_rate_histograms(self, livetime=1):
```
GernotMaier, line 146: Can you help with with the default value for the livetime of 1 s. Why? Seems also that _derive_trigger_rate_histograms is called always with livetime set, so maybe this is not necessary?

simtools/simtel/simtel_histograms.py
```diff
@@ -67,50 +69,184 @@ def get_histogram_title(self, i_hist):
...
+            integrated_hist = np.zeros_like(radius_axis)
+            for i_radius, _ in enumerate(radius_axis):
+                integrated_hist[i_radius] = np.sum(
+                    hist["data"][:-1, i_radius].value * np.diff(energy_axis)
+                )
+
+            list_of_integrated_hists.append(np.sum(integrated_hist) * hist["data"][0, 0].unit)
+        return list_of_integrated_hists
+
+    def trigger_rate_per_histogram(self, livetime=1):
```
GernotMaier, line 237: (same comment as above on `livetime=1`)

tests/integration_tests/test_applications_from_config.py
```diff
@@ -161,11 +161,21 @@ def validate_application_output(config):
...
         if "OUTPUT_FILE" in integration_test:
             # In this case the output file is the file generated by the application
-            assert (
-                Path(config["CONFIGURATION"]["DATA_DIRECTORY"])
-                .joinpath(integration_test["OUTPUT_FILE"])
-                .exists()
-            )
+            # First check if the output is in the data directory (simtel_array related),
+            # Then check if the file is in the output directory (remaining tools).
+            print(Path(config["CONFIGURATION"]["OUTPUT_PATH"]))
```
GernotMaier, line 166: This is something we need to solve - the writing into output paths should be clarified. I remember you complained about it in Dec - did you open an issue?

GernotMaier (approved):
Two very minor comments, otherwise all good!

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: Ok, I changed the name of the raise statement as suggested.

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: Hmm, it doesn't really need to be one as default. I am making it a required input for now but I am going to check whether we can get it directly from the histogram file (therefore I am leaving this conversation open)

tests/integration_tests/test_applications_from_config.py
VictorBarbosaMartins: Ok, I just opened an issue for that https://github.com/gammasim/simtools/issues/760

The case here is a bit different. When running CORSIKA and simtel_array, the outputs are saved into the data_directory (corsika-data and simtel-data directories), while the remaining applications save their outputs into output_path. Perhaps we should reconsider saving CORSIKA and simtel_array results into the output directory or moving the output files to the output directory in the application code.

GernotMaier: We had in Dec a discussion on this (not sure if you where around) and we decided that we want to keep the separation between the output of simtools and those of corsika/sim_telarray. I don't like that it is called 'data' (as it is simulations), but this is to stay consistent with many of the historically grown scripts.

VictorBarbosaMartins: This discussion is older than December as I raised it a couple of times before. Ok, then we focus on the use of `use_plain_output_path` in the new issue and deal with the output directory only

VictorBarbosaMartins: One question remains though: what should we do with the output of simtel_array that is also the output of an application?

GernotMaier: Do we want to move this discussion somewhere else - I think it is clear to us now that we need to revisit how we write any data/output to disk.

VictorBarbosaMartins: Ok, moved it to #760

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: As far as I could find out, there is no such thing as "livetime" in the simtel_array output files. That is understandable, as the simulations are conducted based on the number of events and not on livetime and some reference flux. Therefore, we will leave as it is now and implement the remaining features in the next steps.

VictorBarbosaMartins:
Thanks for the review @GernotMaier ! I adjusted the docstring once more, and will wait for the tests to pass before merging. Please let me know if something is still missing.

simtools/simtel/simtel_histograms.py
GernotMaier: But is this a reasonable default? 1s seems to be weird, why not simply having no default and call the function `def _derive_trigger_rate_histograms(self, livetime)`? It is quite reasonable to request that lifetime must be given as an argument when calling this function.

GernotMaier:
@VictorBarbosaMartins - have a quick look at my comment above; merge if you want.

simtools/simtel/simtel_histograms.py
GernotMaier: OK, this is good! Sorry, didn't see that - it would really be nice if the github interface would indicate also changes to lines one has commented on. So all good!!

VictorBarbosaMartins: agreed!

Issue #755, opened 2024-01-02
Simtools-prod uses always main for image generation


GernotMaier:
The docker container simtools-prod is always generated using the main branch, see [here](https://github.com/gammasim/simtools/blob/c1b42a652bab8638e3e69e7e0d16a17b961b30d7/docker/Dockerfile-prod#L11).

This might lead to failure of the integration test during the action workflow to generate and deploy containers (see [this test](https://github.com/gammasim/simtools/blob/c1b42a652bab8638e3e69e7e0d16a17b961b30d7/.github/workflows/build-docker-images.yml#L79)) for cases in which notable changes are introduced in a branch (note: this is the integration test used to test the container, not our general integration testing).

Suggest to replace 'main' with the current branch. This should be fine, as we push only images for release and merge-to-main to the package repository.

orelgueta:
> Suggest to replace 'main' with the current branch. This should be fine, as we push only images for release and merge-to-main to the package repository.

That sounds good. We just have to make sure that we indeed do not push those "interim" images. We sometimes had trouble with that in the past.

GernotMaier:
The current settings are:

```
push:
    ${{ github.event_name == 'release' || github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true }}
```

I think we agree that 'release' and 'merge-to-main' are good triggers. 'workflow dispatch' is the manual trigger of the workflow, which is needed for testing. I suggest to remove 'workflow dispatch' from the push line (and add it only if needed for testing). OK?

orelgueta:
> The current settings are:
> 
> ```
> push:
>     ${{ github.event_name == 'release' || github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true }}
> ```

Yes I remember that these are the settings, but I think in the past I had trouble with `github.event.pull_request.merged`, so hopefully it was just a stupid mistake I had then.


> I think we agree that 'release' and 'merge-to-main' are good triggers. 'workflow dispatch' is the manual trigger of the workflow, which is needed for testing. I suggest to remove 'workflow dispatch' from the push line (and add it only if needed for testing). OK?

That's fine.

PR #756, opened 2024-01-03
Set branch for simtools-prod testing

GernotMaier:
Addresses issue #755 

For testing, the container simtools-prod used until always the main branch, even if another branch should be tested. This merge request introduces a build arg for the container building which is set in the github workflow to the current branch.

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -70,7 +70,7 @@ jobs:
         if: matrix.type == 'prod'
         with:
           context: .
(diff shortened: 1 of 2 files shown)


GernotMaier:
As a confirmation that this works, I am attaching here the screen shot of the corresponding build (or scroll through [here](https://github.com/gammasim/simtools/actions/runs/7395995637/job/20120271399).

Screenshot 2024-01-03 at 11 16 21

orelgueta (approved):

PR #757, opened 2024-01-03
Addition of data type check to schema validation 

GernotMaier:
Introduce data type checking for data tables, making sure that columns have the correct type.

Introduce at the same time clearer numpy data types (float32, float64), as we had issues in the past with getting this right.


simtools/data_model/validate_data.py
```diff
@@ -293,6 +294,33 @@ def _get_reference_unit(self, column_name):
 
         return u.Unit(reference_unit)
 
+    def _check_data_type(self, col, column_name):
```
VictorBarbosaMartins, line 297: the remaining functions usually return the final corrected value. I suggest doing the same for this function.

simtools/data_model/validate_data.py
```diff
@@ -293,6 +294,33 @@ def _get_reference_unit(self, column_name):
...
+
+        reference_dtype = self._get_reference_data_column(column_name).get("type", None)
+
+        if not np.issubdtype(col.dtype, reference_dtype):
+            self._logger.error(
+                f"Invalid data type in column '{column_name}'. "
+                f"Expected type '{reference_dtype}', found '{col.dtype}'"
+            )
+            raise TypeError
+
```
VictorBarbosaMartins, line 323: I think we could include either a debug or a info logger message here to inform the user that the column is validated.

VictorBarbosaMartins: and then test in the unit tests for a successful verification of the column catching the logger message

VictorBarbosaMartins (approved):
Good, just minor suggestions.

simtools/data_model/validate_data.py
GernotMaier: The difference with this function is that this is an 'either-the-type-is-correct-or-raise-error-function'.

For the other ones, we can convert (e.g., if a value is given in "km", and we convert it to "m".

simtools/data_model/validate_data.py
GernotMaier: Added.

simtools/data_model/validate_data.py
GernotMaier: Added a DEBUG statement.

VictorBarbosaMartins: sounds good!

VictorBarbosaMartins:
> @VictorBarbosaMartins - I've followed your suggestions, thanks for this.
> 
> Waiting for the tests and will then merge.

Thanks for considering my comments, I think we are good to go!

PR #763, opened 2024-01-03
Add json reader

GernotMaier:
Add the possibility to read json files (simulation model parameters are stored in json in the database). 

This is a straightforward implementation following the yaml reader routines. `general.collect_data_from_yaml_or_dict` has not been renamed (although it can read now as well json files). My feeling is that the file name file be too long if we do `collect_data_from_yaml_or_json_or_dict`.

Additional adopted `simtools/applications/validate_file_using_schema.py` to validate json files (integration test added).


simtools/utils/general.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import copy
+import json
```
VictorBarbosaMartins, line 6: I think this is something we need to decide on. Should we agree to always use the astropy json package instead? There are some issues regarding astropy units if one reads a json file with the normal json package (though one can test the impact more closely).

simtools/utils/general.py
```diff
@@ -316,7 +317,7 @@ def is_url(url):
     try:
         result = urlparse(url)
         return all([result.scheme, result.netloc])
-    except ValueError:
```
VictorBarbosaMartins, line 319: This change seems to come out of nowhere (not related to the PR). In any case, it is good to test for this error in the unit test (I could not find it)

simtools/utils/general.py
```diff
@@ -348,7 +349,14 @@ def collect_data_from_http_yaml(url):
     try:
         with tempfile.NamedTemporaryFile() as tmp_file:
             urllib.request.urlretrieve(url, tmp_file.name)
-            data = yaml.load(tmp_file)
+            if url.endswith("yml") or url.endswith("yaml"):
```
VictorBarbosaMartins, line 352: general comment for the function: I suggest changing the name of the function to `collect_data_file_from_http` and adding to the docstring the allowed types. This would reflect the fact that the function also allows getting json file. same comment for the `collect_data_from_yaml_or_dict`, which could change to `collect_data_from_file_or_dict`

tests/integration_tests/config/validate_file_using_schema_json_input.yml
```diff
@@ -0,0 +1,7 @@
+CTA_SIMPIPE:
+  APPLICATION: simtools-validate-file-using-schema
+  TEST_NAME: yml_input
+  CONFIGURATION:
+    SCHEMA: >
+      https://raw.githubusercontent.com/gammasim/workflows/main/schemas/altitude.schema.yml
```
VictorBarbosaMartins, line 6: should we add a second test to test `.yaml` extension or would that be overkill?

VictorBarbosaMartins (approved):
Thanks for these changes. I included some suggestions but I am already approving it. I am also opening an issue to deal with the `Unchanged files with check annotations` which github is complaining (not related to this PR)

simtools/utils/general.py
GernotMaier: Note entirely sure - we use the astropy yaml package because of some extensions (which is actually not ideal - as astropy is using pyaml which does not allows follows the yaml standard), but we do not use any of those extensions for json. I would advocate the simpler solution (using the base package), unless anything else is required.

simtools/utils/general.py
GernotMaier: This line has been reported by codecov to be not tested - and while adding tests it became clear that the wrong error is given in the exception statement.

simtools/utils/general.py
GernotMaier: I agree, but I suggest to leave it for now as I see a very similar function in general.py (collect_dict_from_file). I think we can remove this one and make the code simpler - let me do this in a separate PR (I have a branch already, so no need to start an issue)

GernotMaier:
@VictorBarbosaMartins - I've followed your suggestions, thanks for this.

Waiting for the tests and will then merge.

simtools/utils/general.py
VictorBarbosaMartins: ok! I think I mixed up, since astropy has a yaml package but doesn't seem to have a json similar

simtools/utils/general.py
VictorBarbosaMartins: Ok!

tests/integration_tests/config/validate_file_using_schema_json_input.yml
VictorBarbosaMartins: Ok, then unless strictly necessary, we could stop supporting `.yaml` and remove the if statement in the read function that deals with it.

PR #765, opened 2024-01-04
Remove collect_dict_from_file function from general.py; Rename collect_data_from_yaml_or_dict

GernotMaier:
Follow suggestion from @VictorBarbosaMartins during the PR #763 discussion to rename `collect_data_from_yaml_or_dict` to `collect_data_from_file_or_dict`. This reflects better the added functionality of reading also json files.

Remove function `collect_dict_from_file` from general.py, as it provided almost the same functionality as `collect_data_from_yaml_or_dict`.

Removes also the temporary url forgotten in tests/unit_tests/utils/test_general.py (from PR 763).

### DIFF
--- a/simtools/utils/general.py
+++ b/simtools/utils/general.py
@@ -369,13 +369,13 @@ def collect_data_from_http(url):
     return data
 
 
-def collect_data_from_file_or_dict(in_yaml, in_dict, allow_empty=False):
+def collect_data_from_file_or_dict(file_name, in_dict, allow_empty=False):
     """
     Collect input data that can be given either as a dict or as a yaml/json file.
 
     Parameters
     ----------
-    in_yaml: str
+    file_name: str
         Name of the yaml/json file.
     in_dict: dict
         Data as dict.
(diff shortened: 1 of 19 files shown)


simtools/utils/general.py
```diff
@@ -369,7 +369,7 @@ def collect_data_from_http(url):
     return data
 
 
-def collect_data_from_yaml_or_dict(in_yaml, in_dict, allow_empty=False):
+def collect_data_from_file_or_dict(in_yaml, in_dict, allow_empty=False):
```
VictorBarbosaMartins, line 372: I suggest replacing the argument `in_yaml` by `file` or `file_name` as the file is not necessarily a yaml file anymore.

VictorBarbosaMartins (changes requested):
Thanks for these changes. Just one comment on the argument of `collect_data_from_file_or_dict` which should also be changed, otherwise all good

simtools/utils/general.py
GernotMaier: Good point - changed it to `file_name`.

VictorBarbosaMartins (approved):
All good. let's merge when the tests pass.

PR #767, opened 2024-01-05
Reading of metadata from json

GernotMaier:
Add functionality to read CTA metadata from json files (as the database files are in json). All implementations are similar to the yaml implementation.

- add functionality to read metadata from json file. This expects the format as provided by the database:
```
value:
type:
unit:
version:
metadata: { }
```
- generate json-style metadata with the application `generate_default_metadata.py`
- integration tests to test both json and yml for `generate_default_metadata.py`
- (and as bonus: added a file type test to the integration testing)

(note that the data model url in the metadata of altitude.json will only be valid after merging of [this pull request](https://github.com/gammasim/simulation_model/pull/3), which will happen in the next days).


simtools/applications/generate_default_metadata.py
```diff
@@ -83,14 +85,25 @@ def main():
...
-            yaml.dump(
-                default_values,
-                file,
-                default_flow_style=False,
-                sort_keys=False,
-            )
+        _io_handler = io_handler.IOHandler()
+        _out_file = _io_handler.get_output_file(args_dict["output_file"])
+        _logger.info(f"Writing default values to {_out_file}")
+        if args_dict["output_file"].endswith((".yml", ".yaml")):
```
VictorBarbosaMartins, line 91: I would rather use `_out_file` and not `args_dict["output_file"]` in these two next if statements. It seems safer to me, as  `_out_file` already passed through the io_handler.

simtools/data_model/metadata_collector.py
```diff
@@ -268,13 +268,18 @@ def _read_input_metadata_from_file(self, metadata_file_name=None, observatory="C
...
             try:
                 self._logger.debug("Reading meta data from %s", metadata_file_name)
                 _input_metadata = gen.collect_data_from_file_or_dict(
                     file_name=metadata_file_name, in_dict=None
                 )
+                _json_type_metadata = {"Metadata", "metadata", "METADATA"}.intersection(
+                    _input_metadata
+                )
+                if len(_json_type_metadata) > 0:
+                    _input_metadata = _input_metadata[_json_type_metadata.pop()]
```
VictorBarbosaMartins, line 282: It is not totally clear to me the idea behind these lines here. If there are at least one "metadata" key in `_input_metadata` we only take the last one? Could you please explain me?

simtools/data_model/metadata_collector.py
```diff
@@ -462,7 +467,7 @@ def _process_metadata_from_file(self, meta_dict):
             meta_dict["cta"]["product"]["description"] = self._remove_line_feed(
                 meta_dict["cta"]["product"]["description"]
             )
-        except KeyError:
+        except (KeyError, AttributeError):
```
VictorBarbosaMartins, line 470: None of the errors are in the docstring yet. Despite being an internal function, I prefer the docstring to be correct.

VictorBarbosaMartins: I also don't see when an attribute error could be raised by this function. Can you please clarify?

tests/integration_tests/config/generate_default_metadata_to_json_file.yml
```diff
@@ -0,0 +1,9 @@
+CTA_SIMPIPE:
+  APPLICATION: simtools-generate-default-metadata
+  TEST_NAME: test-json
+  CONFIGURATION:
+    OUTPUT_FILE: test.json
+    OUTPUT_PATH: simtools-tests
+  INTEGRATION_TESTS:
+    - OUTPUT_FILE: test.json
+    - FILE_TYPE: json
```
VictorBarbosaMartins, line 9: I tested this implementation by running the application in the container with 
```
simtools-generate-default-metadata --output_file tst.json --output_path .
```
but I get an error that 
```
generate_default_metadata: error: the following arguments are required: --schema
```

I remember that with the config files we can not assure that all necessary arguments were passed, but I suggest adding the `--schema` argument into the config file to really test the application.

tests/integration_tests/config/generate_default_metadata_to_yml_file.yml
```diff
@@ -1,6 +1,9 @@
 CTA_SIMPIPE:
   APPLICATION: simtools-generate-default-metadata
-  TEST_NAME: test
+  TEST_NAME: test-yaml
```
VictorBarbosaMartins, line 3: That made realized that the same issue with the json test is present here (missing --schema argument). I would suggest including it here too.

tests/integration_tests/config/generate_default_metadata_to_yml_file.yml
```diff
@@ -1,6 +1,9 @@
...
   CONFIGURATION:
     OUTPUT_FILE: test.yml
     OUTPUT_PATH: simtools-tests
+  INTEGRATION_TESTS:
+    - OUTPUT_FILE: test.yml
+    - FILE_TYPE: yml
```
VictorBarbosaMartins, line 9: I was thinking that the file type test is redundant because of the suffix already being present in the output_file test, but after looking into `test_applications_from_config.py` I understand the importance of this test. However, `file_type` is actually testing whether opening the output file works (not only the type of the output). So, I suggest changing the name to `OPEN_OUTPUT_FILE` for example.

tests/integration_tests/test_applications_from_config.py
```diff
@@ -134,6 +135,42 @@ def compare_ecsv_files(file1, file2, tolerance=1.0e-5):
             assert np.allclose(table1[col_name], table2[col_name], rtol=tolerance)
 
 
+def assert_file_type(file_type, file_name):
```
VictorBarbosaMartins, line 138: Should we include ecsv file here as well?

tests/integration_tests/test_applications_from_config.py
```diff
@@ -134,6 +135,42 @@ def compare_ecsv_files(file1, file2, tolerance=1.0e-5):
...
+    """
+    Assert that the file is of the given type.
+
+    Parameters
+    ----------
+    file_type: str
+        File type (json, yaml).
+    file_name: str
+        File name.
+
```
VictorBarbosaMartins, line 148: I would include the raise statements as well

tests/integration_tests/test_applications_from_config.py
```diff
@@ -134,6 +135,42 @@ def compare_ecsv_files(file1, file2, tolerance=1.0e-5):
...
+            with open(file_name, "r", encoding="utf-8") as file:
+                yaml.safe_load(file)
+            return True
+        except (yaml.YAMLError, FileNotFoundError):
+            return False
+
+    # no dedicated tests for other file types, checking suffix only
+    logger.info(f"File type test is checking suffix only for {file_name} (suffix: {file_type}))")
+    if file_name.suffix[1:] == file_type:
+        return True
```
VictorBarbosaMartins, line 169: This is the redundant part I mentioned above. There is no need for this part, given that the existence of the output file (with suffix) is already checked by the other test.

tests/unit_tests/data_model/test_metadata_collector.py
```diff
@@ -252,6 +255,9 @@ def test_process_metadata_from_file():
     meta_dict_3 = {"cta": {"PRODUCT": {"description": "This is a sample\n description"}}}
     _dict_test_1 = _collector._process_metadata_from_file(meta_dict_3)
 
+    meta_dict_3 = {"cta": {"product": {"description": None}}}
```
VictorBarbosaMartins, line 258: meta_dict_4 ?

VictorBarbosaMartins (changes requested):
Hi @GernotMaier, thanks for this implementation! I leave a couple of minor comments here but just in case I mark "request changes" such that I remember to have a look again.

simtools/applications/generate_default_metadata.py
GernotMaier: This is checking for the file suffix only, so we can use directly the args_dict value. io_handler is doing the work on the full path, which is not needed here.

simtools/data_model/metadata_collector.py
GernotMaier: We cannot expect at this point that the metadata keys are all lower case (as they will be later). So these four lines make sure that any key from `{"Metadata", "metadata", "METADATA"}` is accepted.

simtools/data_model/metadata_collector.py
GernotMaier: These exceptions are passed. Maybe I am wrong, and but don't we document only Exceptions which a module raises?

On AttributeError: self._remove_line_feed(None) is giving this Exception. This is also tested in https://github.com/gammasim/simtools/blob/d9920d2536bf3371a8052a7d96c82e8b9ecd41b8/tests/unit_tests/data_model/test_metadata_collector.py#L258

tests/integration_tests/config/generate_default_metadata_to_json_file.yml
GernotMaier: We are absolutely sure that all parameters are passed with the `--config` parameters. Do you have an example where this is not the case?

Are you sure you did also a `pip install -e` after checking the branch out? It works for me and the integration tests on github are also working.

tests/integration_tests/config/generate_default_metadata_to_yml_file.yml
GernotMaier: see previous comment.

tests/integration_tests/config/generate_default_metadata_to_yml_file.yml
GernotMaier: The important point for yaml and json is that it uses the corresponding libraries to open it. So if the file is not of type yaml/json, the tests fails. I therefore think INTEGRATION_TEST:FILE_TYPE is a good name. For opening a file, one could use a very generic file-open, which does not knows anything about the type. 

Let me know if this makes sense.

tests/integration_tests/test_applications_from_config.py
GernotMaier: Yes, and probably for any other file types we are using. I think this is beyond this pull request (it is already a 'bonus' on this one).

tests/integration_tests/test_applications_from_config.py
GernotMaier: This function does not raise Exceptions, it catches them and adopts the return value accordingly. So I think no need to document it, as these exceptions will not be visible.

tests/integration_tests/test_applications_from_config.py
GernotMaier: Not really - any file type which is not json or yaml is tested by this (too) simplistic test of comparing the file suffix. For yaml/json, these lines are not reached. 

I agree with you that as discussed above this should be replaced by explicit tests for each file type. Something to be added in future.

tests/unit_tests/data_model/test_metadata_collector.py
GernotMaier: OK.

GernotMaier:
@VictorBarbosaMartins - thanks for the review! I've tried to reply to all your comments and integrated the fixes.

simtools/applications/generate_default_metadata.py
VictorBarbosaMartins: ok.

simtools/data_model/metadata_collector.py
VictorBarbosaMartins: Any key, but it always takes the last one (that is what `.pop()` does) if there are more than one key among `{"Metadata", "metadata", "METADATA"}`. My point is that I do not see the reason why we want to take the last one specifically. I think we should indeed accept any of the keys but if there are more than one, we should raise a warning or e.g. take the metadata from both keys, no?

simtools/data_model/metadata_collector.py
VictorBarbosaMartins: Yes, you are right. I oversaw this. If we `pass` we don't need the statements.

tests/integration_tests/config/generate_default_metadata_to_yml_file.yml
VictorBarbosaMartins: Ok, it is just the name and I don't have a strong opinion on that. It would be nice in the future (as you commented below) if we could test opening all kinds of output file in this test, but obviously not on this PR.

tests/integration_tests/test_applications_from_config.py
VictorBarbosaMartins: I agree this is not redundant for yaml and json files, but for all the other output files in which we have both `OUTPUT_FILE` and `FILE_TYPE` tests, there will be a repetition. I agree that we can leave it like this for now because we only use `FILE_TYPE` to check json and yaml files, but something to keep in mind when expanding the function.

VictorBarbosaMartins (approved):
Thanks for taking my comments into consideration. I am fine with the changes, hence, I am approving. One last comment to clarify a very minor thing but I think we can merge it as soon as possible.

simtools/data_model/metadata_collector.py
GernotMaier: OK, this is a very unlikely event!

Anyway, I've added a check for this now (and a unit test).

VictorBarbosaMartins: Indeed. Ok, looks good, thanks!

PR #769, opened 2024-01-08
Simplify usage of simtools-validate-file-using-schema

GernotMaier:
Simplify the usage of the application `simtools-validate-file-using-schema`: read `schema` file from metadata if this parameter is not given on the command line. This could be either the usual CTA metadata convention, or from a `meta_schema` field for validation of schema using meta_schema.

Add functionality to test the metadata header of data files. Use the flag `validate_metadata` for this (see integration tests [tests/integration_tests/config/validate_file_using_schema_ecsv_input_test_meta.yml](tests/integration_tests/config/validate_file_using_schema_ecsv_input_test_meta.yml) and [tests/integration_tests/config/validate_file_using_schema_json_input_test_meta.yml](tests/integration_tests/config/validate_file_using_schema_json_input_test_meta.yml)).

Other changes:

- corrected `base_schema` to `meta_schema` in `tests/resources/reference_position_mercator.schema.yml` (this was forgotten earlier)

Important to note that there is a preliminary entry in [tests/resources/altitude.json](tests/resources/altitude.json) (required [this one](https://github.com/gammasim/simulation_model/pull/3) to be merged).

### DIFF
--- a/simtools/applications/validate_file_using_schema.py
+++ b/simtools/applications/validate_file_using_schema.py
@@ -61,7 +61,9 @@ def _parse(label, description):
 
 def _validate_yaml_or_json_file(args_dict, logger):
     """
-    Validate a yaml or json file
+    Validate a yaml or json file.
+    Schema is either given as command line argument, read from the meta_schema_url entry
+    in the data, or read from the metadata.
 
     """
 
@@ -71,8 +73,15 @@ def _validate_yaml_or_json_file(args_dict, logger):
         logger.error(f"Input file {args_dict['file_name']} not found")
         raise
 
-    _collector = metadata_collector.MetadataCollector(args_dict)
-    print("FFF", _collector.get_data_model_schema_file_name())
+    if args_dict.get("schema", None) is None and "meta_schema_url" in data:
+        args_dict["schema"] = data["meta_schema_url"]
+        logger.debug(f'Using schema from meta_schema_url: {args_dict["schema"]}')
+    if args_dict.get("schema", None) is None:
+        _collector = metadata_collector.MetadataCollector(
+            None, metadata_file_name=args_dict["file_name"]
+        )
+        args_dict["schema"] = _collector.get_data_model_schema_file_name()
+        logger.debug(f'Using schema from meta_data_url: {args_dict["schema"]}')
 
     metadata_model.validate_schema(data, args_dict.get("schema", None))
 
--- a/simtools/applications/validate_file_using_schema.py
+++ b/simtools/applications/validate_file_using_schema.py
@@ -32,7 +32,7 @@ from pathlib import Path
 
 import simtools.utils.general as gen
 from simtools.configuration import configurator
-from simtools.data_model import metadata_model, validate_data
+from simtools.data_model import metadata_collector, metadata_model, validate_data
 
 
 def _parse(label, description):
@@ -55,7 +55,7 @@ def _parse(label, description):
 
     config = configurator.Configurator(label=label, description=description)
     config.parser.add_argument("--file_name", help="file to be validated", required=True)
-    config.parser.add_argument("--schema", help="json schema file", required=True)
+    config.parser.add_argument("--schema", help="json schema file", required=False)
     return config.initialize(paths=False)
 
 
@@ -71,7 +71,10 @@ def _validate_yaml_or_json_file(args_dict, logger):
         logger.error(f"Input file {args_dict['file_name']} not found")
         raise
 
-    metadata_model.validate_schema(data, args_dict["schema"])
+    _collector = metadata_collector.MetadataCollector(args_dict)
+    print("FFF", _collector.get_data_model_schema_file_name())
+
+    metadata_model.validate_schema(data, args_dict.get("schema", None))
 
 
 def _validate_ecsv_file(args_dict):
--- a/simtools/applications/validate_file_using_schema.py
+++ b/simtools/applications/validate_file_using_schema.py
@@ -56,6 +56,12 @@ def _parse(label, description):
     config = configurator.Configurator(label=label, description=description)
     config.parser.add_argument("--file_name", help="file to be validated", required=True)
     config.parser.add_argument("--schema", help="json schema file", required=False)
+    config.parser.add_argument(
+        "--validate_metadata",
+        help="validate metadata",
+        action="store_true",
+        default=False,
+    )
     return config.initialize(paths=False)
 
 
@@ -84,9 +90,10 @@ def _validate_yaml_or_json_file(args_dict, logger):
         logger.debug(f'Using schema from meta_data_url: {args_dict["schema"]}')
 
     metadata_model.validate_schema(data, args_dict.get("schema", None))
+    logger.debug("Successful validation of yaml/json file")
 
 
-def _validate_ecsv_file(args_dict):
+def _validate_ecsv_file(args_dict, logger):
     """
     Validate an ecsv file
 
@@ -97,6 +104,18 @@ def _validate_ecsv_file(args_dict):
         data_file=args_dict["file_name"],
     )
     data_validator.validate_and_transform()
+    logger.debug("Successful validation of escv file")
+
+
+def _validate_metadata(args_dict, logger):
+    """
+    Validate metadata.
+
+    """
+
+    # metadata_collector runs the metadata validation by default, no need to do anything else
+    metadata_collector.MetadataCollector(None, metadata_file_name=args_dict["file_name"])
+    logger.debug("Successful validation of metadata")
 
 
 def main():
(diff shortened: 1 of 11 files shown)


GernotMaier:
@VictorBarbosaMartins  - this one is now ready for review.

This is a nice one, as it removes duplication: for most (all?) data files we have the underlying schema in the metadata. So tools should retrieve the schema url from the metadata and use that one.

simtools/applications/validate_file_using_schema.py
```diff
@@ -85,6 +104,18 @@ def _validate_ecsv_file(args_dict):
...
+    logger.debug("Successful validation of escv file")
+
+
+def _validate_metadata(args_dict, logger):
+    """
+    Validate metadata.
+
+    """
+
+    # metadata_collector runs the metadata validation by default, no need to do anything else
```
VictorBarbosaMartins, line 116: An idea: I think it would be better if the other functions (`_validate_yaml_or_json_file`) could get also as compact as this one. Ideally, we move the functionalities (of `_validate_yaml_or_json_file` in this case) to the modules and keep the application clean.

simtools/applications/validate_file_using_schema.py
```diff
@@ -94,10 +125,12 @@ def main():
     logger = logging.getLogger()
     logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
-    if any(args_dict["file_name"].endswith(ext) for ext in (".yml", ".yaml", ".json")):
+    if args_dict["validate_metadata"]:
+        _validate_metadata(args_dict, logger)
+    elif any(args_dict["file_name"].endswith(ext) for ext in (".yml", ".yaml", ".json")):
```
VictorBarbosaMartins, line 130: I was actually expecting we would be able to validate both data and metadata in a single shot. But this `elif` means that if the metadata was already validated, the data will not be validated. I suggest allowing both data and metadata to be validated together, even if the metadata validated is the one that comes from the data file (or url). For that, I think that simply replacing the `elif`s by `if`s would do the work.

tests/integration_tests/config/validate_file_using_schema_json_input_test_meta.yml
VictorBarbosaMartins, line 1: Based on the name of the test, I believe you want the schema file as well, no?

```
 SCHEMA: >
      https://raw.githubusercontent.com/gammasim/workflows/main/schemas/altitude.schema.yml
```

VictorBarbosaMartins (commented):
Hi @GernotMaier, thanks for this PR. The code is slowly getting self-consistent in the schema validation. I provide some comments below, I hope I understood the code such that the comments make sense, please let me know if that is not the case.

Non-related, I have this warning for missing test, probably reminiscent from the last PR merged (@orelgueta)? 
Screenshot 2024-01-10 at 11 12 35

simtools/applications/validate_file_using_schema.py
```diff
@@ -71,10 +79,21 @@ def _validate_yaml_or_json_file(args_dict, logger):
...
 
-    metadata_model.validate_schema(data, args_dict["schema"])
+    if args_dict.get("schema", None) is None and "meta_schema_url" in data:
+        args_dict["schema"] = data["meta_schema_url"]
+        logger.debug(f'Using schema from meta_schema_url: {args_dict["schema"]}')
+    if args_dict.get("schema", None) is None:
+        _collector = metadata_collector.MetadataCollector(
+            None, metadata_file_name=args_dict["file_name"]
+        )
+        args_dict["schema"] = _collector.get_data_model_schema_file_name()
```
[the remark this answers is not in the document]
GernotMaier: I really don't see an issue with that. The alternative is to assign a new variable and introduce decisions in the code to decide with `args_dict` or the new variable is used. I think this approach here is much simpler.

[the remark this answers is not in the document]
orelgueta: I don't think this is an issue (I think I do that in the simulation production application as well).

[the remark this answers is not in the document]
VictorBarbosaMartins: ok! then it is fine

orelgueta:
> 
> Non-related, I have this warning for missing test, probably reminiscent from the last PR merged (@orelgueta)? Screenshot 2024-01-10 at 11 12 35

Yes, this line wasn't tested (cause it's trivial). I added a test for it now that will be included in a future PR that expands the CORSIKA default config class.

VictorBarbosaMartins:
> > Non-related, I have this warning for missing test, probably reminiscent from the last PR merged (@orelgueta)? Screenshot 2024-01-10 at 11 12 35
> 
> Yes, this line wasn't tested (cause it's trivial). I added a test for it now that will be included in a future PR that expands the CORSIKA default config class.

Ok, thanks for looking into this. (nothing to worry about in this PR)

simtools/applications/validate_file_using_schema.py
```diff
@@ -71,10 +79,21 @@ def _validate_yaml_or_json_file(args_dict, logger):
         logger.error(f"Input file {args_dict['file_name']} not found")
         raise
 
-    metadata_model.validate_schema(data, args_dict["schema"])
+    if args_dict.get("schema", None) is None and "meta_schema_url" in data:
+        args_dict["schema"] = data["meta_schema_url"]
+        logger.debug(f'Using schema from meta_schema_url: {args_dict["schema"]}')
```
orelgueta, line 84: The quotes here (and below) should be reversed, `"` should be the outer ones and `'` should be the inner. 

(Unfortunately, despite me switching jobs, you still have to put up with such annoying comments sometimes...)

tests/resources/altitude.json
```diff
@@ -25,7 +25,7 @@
                     "MODEL": {
                         "NAME": "altitude",
                         "VERSION": "0.1.0",
-                        "URL": "https://raw.githubusercontent.com/gammasim/simulation_model/main/schema/altitude.schema.yml"
+                        "URL": "https://raw.githubusercontent.com/gammasim/simulation_model/site-parameters/schema/altitude.schema.yml"
```
orelgueta, line 28: Should you add a comment somewhere to remember to change it to main later? I know you can't add a comment in json, so maybe an issue would help.

orelgueta (commented):
I do not approve because Victor actually reviewed and made suggestions. I just read quickly through.

tests/integration_tests/config/validate_file_using_schema_json_input_test_meta.yml
GernotMaier: No - we don't need to give the schema when testing metadata. This schema is fixed and known, as our metadata has to follow the CTAO metadata requirements (so in this case we always use simtools/schemas/metadata.metaschema.yml)

simtools/applications/validate_file_using_schema.py
GernotMaier: OK, this was anyway planned (and already implemented in another context and branch on my local disk). So I've moved now a function into the data_reader called `def read_value_from_file(file_name, schema_file=None, validate=False, metadata_file=None):`, which does exactly the same as `read_table_from_file` (with the obvious difference that it is for a value). It also includes the validation.

This simplifies that method.

simtools/applications/validate_file_using_schema.py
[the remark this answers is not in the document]
GernotMaier: (and this code is not there anymore given that I have moved this into data_reader.

simtools/applications/validate_file_using_schema.py
GernotMaier: Thanks - you keep being annoying :-)

(this code has been removed; so no issue anymore)

GernotMaier:
@VictorBarbosaMartins - could you please have another look now (after the tests passed):

As mentioned above, your comment on simplification of the validation step triggered that I've moved some development into this pull request - I think it logically fits very well and it addresses your suggestion (and I know that this slightly validates our approach on pull requests, hope this is ok).

tests/resources/altitude.json
GernotMaier: Done

tests/integration_tests/config/validate_file_using_schema_json_input_test_meta.yml
VictorBarbosaMartins: Ok!

simtools/data_model/data_reader.py
```diff
@@ -66,3 +68,62 @@ def read_table_from_file(file_name, schema_file=None, validate=False, metadata_f
         return _validator.validate_and_transform()
 
     return data_table
+
+
+def read_value_from_file(file_name, schema_file=None, validate=False):
```
VictorBarbosaMartins, line 73: Not so important by I suggest adding this function to the __all__ list for consistency

VictorBarbosaMartins (approved):
Much nicer now the compact application, thanks for implementing my suggestion. I am already approving, but there is one open comment, which I deem for important too. Please have a look at my suggestion to test both metadata and data together by replacing two `elif`s by `if`s. Thanks!

simtools/applications/validate_file_using_schema.py
VictorBarbosaMartins: This is one remaining comment. Otherwise all good from my side

simtools/applications/validate_file_using_schema.py
VictorBarbosaMartins: ok! Thanks. That is perfectly fine.

simtools/applications/validate_file_using_schema.py
GernotMaier: I think I want to keep it how it is, this is expected behavior (it least in my view): it is an application to validate data or metadata, and I would find it confusing when trying to work on validating data in a file and would get errors on the meta data. Note also that this is an application for developers, used in CI workflows or with interfaces, and not by the typical simtools user.

VictorBarbosaMartins: ok, maybe it is indeed simpler to keep it separate. My point was due to the fact that the tool is called `validate file`, but when we pass a data file, the tool only validates the data of that file and not the metadata in there (if present). But let's then keep it like this, I won't object and I think we can merge.

PR #771, opened 2024-01-08
Default CORSIKA simulation config class

orelgueta:
This PR introduces a class with default CORSIKA simulation config. It is far from complete, missing many parameters still. Also, it is not actually used yet in the code. However, I would like to get first feedback before I move forward with this implementation.

For the record, including it in the code should be fairly simple and harmless, by adding one line to the snippet below to complement any dictionary from the user with the default dictionary returned by this new class.

https://github.com/gammasim/simtools/blob/7cfbe595c005bbdaa2f39ffb1b23e2698d83b250/simtools/simulator.py#L192-L205

### DIFF
--- a/simtools/corsika/corsika_default_config.py
+++ b/simtools/corsika/corsika_default_config.py
@@ -1,0 +1,217 @@
+from collections import defaultdict
+
+import astropy.units as u
+from scipy.interpolate import interp1d
+
+
+class CorsikaDefaultConfig:
+    """
+    This class contains the default configuration for CORSIKA parameters for
+    the various primary particles. It includes all basic dependencies on zenith angles, etc.
+    """
+
+    def __init__(self, primary=None, zenith_angle=None):
+        """
+        Initialize CorsikaDefaultConfig.
+        """
+        self._primary = primary
+        self._zenith_angle = zenith_angle
+        self._energy_slope = -2.0
+        self.energy_ranges = self._define_hardcoded_energy_ranges()
+        self.number_of_showers = self._define_hardcoded_number_of_showers()
+
+    @property
+    def primary(self):
+        """
+        Primary particle.
+        """
+        return self._primary
+
+    @primary.setter
+    def primary(self, primary):
+        """
+        Set primary particle.
+        """
+        supported_primaries = [
+            "gamma",
+            "gamma_diffuse",
+            "electron",
+            "proton",
+            "muon",
+            "helium",
+            "nitrogen",
+            "silicon",
+            "iron",
+        ]
+        if primary not in supported_primaries:
+            raise ValueError(
+                f"Invalid primary particle: {primary}. Must be one of {supported_primaries}"
+            )
+        self._primary = primary
+
+    @property
+    def zenith_angle(self):
+        """
+        Zenith angle.
+        """
+        return self._zenith_angle
+
+    @zenith_angle.setter
+    @u.quantity_input(zenith_angle=u.deg)
+    def zenith_angle(self, zenith_angle):
+        """
+        Set zenith angle.
+        """
+        allowed_zenith_angle_interval = [20.0, 60.0] * u.deg
+        if (
+            zenith_angle < allowed_zenith_angle_interval[0]
+            or zenith_angle > allowed_zenith_angle_interval[1]
+        ):
+            raise ValueError(
+                f"The zenith angle, {zenith_angle:.1f}, "
+                f"is outside of the allowed interval, {allowed_zenith_angle_interval}."
+                "This interval is enforced because values can only be interpolated "
+                "between zenith angles that were manually optimized."
+            )
+
+        self._zenith_angle = zenith_angle
+
+    @property
+    def energy_slope(self):
+        """
+        Energy slope.
+        """
+        return self._energy_slope
+
+    def _define_hardcoded_energy_ranges(self):
+        """
+        Define the hardcoded energy ranges for the various primaries.
+        """
+
+        energy_ranges = defaultdict(dict)
+        energy_ranges["gamma"][20] = [3 * u.GeV, 330 * u.TeV]
+        energy_ranges["gamma"][40] = [6 * u.GeV, 660 * u.TeV]
+        energy_ranges["gamma"][60] = [12 * u.GeV, 990 * u.TeV]
+
+        for zenith_angle in energy_ranges["gamma"]:
+            energy_ranges["gamma_diffuse"][zenith_angle] = energy_ranges["gamma"][zenith_angle]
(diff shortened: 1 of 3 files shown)


simtools/corsika/corsika_default_config.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        return self._zenith_angle
+
+    @zenith_angle.setter
+    @u.quantity_input(zenith_angle=u.deg)
+    def zenith_angle(self, zenith_angle):
+        """
+        Set zenith angle.
+        """
+        allowed_zenith_angle_interval = [20.0, 60.0] * u.deg
```
GernotMaier, line 65: Why is the default allowed zenith angle interval 20-60, and not 0-60 (or even >60, as the required elevation range of CTA is larger)? Is it because energy range, etc are only hardwired for these values?

simtools/corsika/corsika_default_config.py
```diff
@@ -0,0 +1,217 @@
...
+    @property
+    def energy_slope(self):
+        """
+        Energy slope.
+        """
+        return self._energy_slope
+
+    def _define_hardcoded_energy_ranges(self):
+        """
+        Define the hardcoded energy ranges for the various primaries.
```
GernotMaier, line 88: Do you want to add in the docstring (or in the module docstring) that these default values have been derived to cover the full CTAO energy range and assume that all three telescopes types are simulated?

simtools/corsika/corsika_default_config.py
```diff
@@ -0,0 +1,217 @@
...
+        Primary particle.
+        """
+        return self._primary
+
+    @primary.setter
+    def primary(self, primary):
+        """
+        Set primary particle.
+        """
+        supported_primaries = [
```
GernotMaier, line 35: We agree to use strings for particle names and not e.g., PDG names (would not be sure about gamma_diffuse then).

simtools/corsika/corsika_default_config.py
```diff
@@ -0,0 +1,217 @@
...
+
+        for zenith_angle in number_of_showers["gamma"]:
+            number_of_showers["gamma_diffuse"][zenith_angle] = number_of_showers["gamma"][
+                zenith_angle
+            ]
+            number_of_showers["electron"][zenith_angle] = number_of_showers["gamma"][zenith_angle]
+
+        number_of_showers["proton"][20] = 15000
+        number_of_showers["proton"][40] = 15000
+        number_of_showers["proton"][60] = 7500
```
GernotMaier, line 140: Why less at 60? The increased energy threshold?

simtools/corsika/corsika_default_config.py
```diff
@@ -0,0 +1,217 @@
...
+    def primary(self):
+        """
+        Primary particle.
+        """
+        return self._primary
+
+    @primary.setter
+    def primary(self, primary):
+        """
+        Set primary particle.
```
GernotMaier, line 33: Here and below: no description of parameters in the docstring.

tests/unit_tests/corsika/test_corsika_default_config.py
```diff
@@ -0,0 +1,84 @@
...
+    config = CorsikaDefaultConfig(primary="gamma", zenith_angle=40.0 * u.deg)
+    number_of_showers = config.number_of_showers_for_primary()
+    assert number_of_showers == 5000
+
+
+def test_view_cone_for_primary():
+    config = CorsikaDefaultConfig(primary="gamma")
+    view_cone = config.view_cone_for_primary()
+    expected_view_cone = [0.0, 0.0]
+    compare_lists_ignoring_units(view_cone, expected_view_cone)
```
GernotMaier, line 59: This is trivial, but I expected to see a test of `gamma_diffuse` - as this is the closest one to `gamma`.

GernotMaier (approved):
Have some relatively minor comments, therefore already approve. Feel free to address them (or ignore them).

GernotMaier:
@orelgueta - maybe you can explain again the motivation of not using configuration files. These are values you don't expect to change ever? And all of them will still be configurable, these are 'just' defaults?

simtools/corsika/corsika_default_config.py
orelgueta: Yes, I do not have optimised values of energy ranges and number of events for anything below 20 degrees or above 60 degrees. Below 20 degrees I could perhaps "guess" because it is not much different to 20 degrees, but I don't see a need for it right now. Above 60 degrees would require an actual (short) study.

simtools/corsika/corsika_default_config.py
orelgueta: Yes, I will add it to the docstring in both places. BTW, the energy range does not change if SSTs are simulated or not (only the area).

simtools/corsika/corsika_default_config.py
orelgueta: Yes, less events can be simulated in 24 hours (the limit on the grid), because they are higher energy events.

simtools/corsika/corsika_default_config.py
GernotMaier: There was a "?" missing - we agreed to use those strings and not PDG codes? Fine with me, just want to mention it.

simtools/corsika/corsika_default_config.py
orelgueta: You are right, added documentation everywhere.

tests/unit_tests/corsika/test_corsika_default_config.py
orelgueta: No problem, I added one for gamma_diffuse as well.

orelgueta:
> @orelgueta - maybe you can explain again the motivation of not using configuration files. These are values you don't expect to change ever? And all of them will still be configurable, these are 'just' defaults?

Yes, these are just defaults, but they are expected to be used often. Putting it in a configuration file would probably lead to very cumbersome and long files. It is easier and shorter to do it in dictionaries and code. We can revisit this point once this class is complete though. Maybe I am wrong and we could switch to files later.

orelgueta:
Oh and I think I answered everything. Maybe take a second look and we resolve the question about the particle names before we merge?

simtools/corsika/corsika_default_config.py
orelgueta: Oh, for the configuration we can use whatever we want essentially and we should use these strings as they are easy and standard. I don't think the discussion about PDG usage is relevant for this specific case, but more about the output we pass to the analysis software.

GernotMaier: OK!

GernotMaier:
All good, please merge!

PR #777, opened 2024-01-11
Improve configuration of make_regular_arrays

GernotMaier:
Some improvements and changes to`make_regular_arrays`:

- allow to configure simulation model version on the command line (before: prod5 was hardwired)
- do not generate for North and Site sites the arrays, but allow to configure the site on the command line (`--site` as usual) and generate regular arrays for one site only
      - required to introduce a `site_model` flag in the configurator class
- fix a bug introduced when changing the output file writing using `writer.ModelDataWriter.dump`. This used always `args_dict[out_file]`, meaning that all subarray telescope lists are overwritten during the process of writing
   - required to add a `output_file` parameter to  `writer.ModelDataWriter.dump` to handle this case

### DIFF
--- a/simtools/applications/make_regular_arrays.py
+++ b/simtools/applications/make_regular_arrays.py
@@ -13,7 +13,7 @@
     Command line arguments
     ----------------------
     site (str, required)
-        North or South.
+        observatory site (e.g., North or South).
     model_version (str, optional)
         Model version to use. If not provided, the latest version is used.
     verbosity (str, optional)
--- a/simtools/applications/make_regular_arrays.py
+++ b/simtools/applications/make_regular_arrays.py
@@ -27,15 +27,6 @@
 
         simtools-make-regular-arrays --site=North
 
-    The output is saved in simtools-output/make_regular_arrays.
-
-    Expected final print-out message:
-
-    .. code-block:: console
-
-        INFO::array_layout(l608)::export_telescope_list::Exporting telescope list to /workdir/exter\
-        nal/simtools/simtools-output/make_regular_arrays/layout/telescope_positions-North-4LS\
-        T-corsika.ecsv
 """
 
 import logging
--- a/simtools/applications/make_regular_arrays.py
+++ b/simtools/applications/make_regular_arrays.py
@@ -15,7 +15,7 @@
     site (str, required)
         observatory site (e.g., North or South).
     model_version (str, optional)
-        Model version to use. If not provided, the latest version is used.
+        Model version to use (e.g., prod6). If not provided, the latest version is used.
     verbosity (str, optional)
         Log level to print.
 
--- a/simtools/applications/make_regular_arrays.py
+++ b/simtools/applications/make_regular_arrays.py
@@ -12,6 +12,10 @@
 
     Command line arguments
     ----------------------
+    site (str, required)
+        North or South.
+    model_version (str, optional)
+        Model version to use. If not provided, the latest version is used.
     verbosity (str, optional)
         Log level to print.
 
@@ -21,7 +25,7 @@
 
     .. code-block:: console
 
-        simtools-make-regular-arrays
+        simtools-make-regular-arrays --site=North
 
     The output is saved in simtools-output/make_regular_arrays.
 
@@ -36,6 +40,7 @@
 
 import logging
 from pathlib import Path
+import os
 
 import astropy.units as u
 
@@ -57,7 +62,7 @@ def main():
             "The array layout files created should be available at the data/layout directory."
         ),
     )
-    args_dict, db_config = config.initialize(db_config=True, output=True)
+    args_dict, db_config = config.initialize(db_config=True, site_model=True, output=True)
 
     label = "make_regular_arrays"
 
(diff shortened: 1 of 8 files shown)


simtools/applications/make_regular_arrays.py
```diff
@@ -12,6 +12,10 @@
 
     Command line arguments
     ----------------------
+    site (str, required)
+        North or South.
```
VictorBarbosaMartins, line 16: It is usually clear when we talk about site we mean the CTA site but in other places in the code we started specifying the observatory as well. Not that I prefer to specify the observatory (I don't) but I want to call your attention to this.

simtools/applications/make_regular_arrays.py
```diff
@@ -12,6 +12,10 @@
...
+    model_version (str, optional)
+        Model version to use. If not provided, the latest version is used.
```
VictorBarbosaMartins, line 18: I would still include an example here in the docstring, "e.g. prod5, prod6". That is simply because one might not understand immediately what model_version means

simtools/applications/make_regular_arrays.py
```diff
@@ -72,101 +68,92 @@ def main():
...
     db = db_handler.DatabaseHandler(mongo_db_config=db_config)
+    site_pars_db = db.get_site_parameters(
+        site=args_dict["site"], model_version=args_dict["model_version"]
+    )
 
-    site_pars_db = {}
     layout_center_data = {}
+    layout_center_data["center_lat"] = float(site_pars_db["ref_lat"]["Value"]) * u.deg
+    layout_center_data["center_lon"] = float(site_pars_db["ref_long"]["Value"]) * u.deg
+    layout_center_data["center_alt"] = float(site_pars_db["altitude"]["Value"]) * u.m
```
VictorBarbosaMartins, line 78: I don't remember, but isn't it possible to get the unit from the DB as well? That would be ideal, since the unit might change with time and the code here gets outdated.

simtools/applications/make_regular_arrays.py
```diff
@@ -72,101 +68,92 @@ def main():
...
+                yy=-telescope_distance[tel_size],
+                tel_corsika_z=0 * u.m,
+            )
+
+        layout.convert_coordinates()
+        layout.print_telescope_list()
+        output_file = str(args_dict.get("output_file", None))
+        if output_file is not None:
+            base_name, file_extension = os.path.splitext(output_file)
+            output_file = f"{base_name}-{args_dict['site']}-{array_name}{file_extension}"
```
VictorBarbosaMartins, line 147: Didn't resist f-strings after all? :)

simtools/configuration/commandline_parser.py
```diff
@@ -46,6 +47,8 @@ def initialize_default_arguments(
             Add output file configuration to list of args.
         telescope_model: bool
             Add telescope model configuration to list of args.
+        site_model: bool
```
VictorBarbosaMartins, line 50: Just a comment (no action needed): I tend to think that flags have to come with a verb in their names, otherwise I expect a model from an argument called `site_model` but I see that all the other parameters here also don't have verbs so I won't suggest changing everywhere.

simtools/configuration/commandline_parser.py
```diff
@@ -237,7 +242,10 @@ def initialize_telescope_model_arguments(self, add_model_version=True, add_teles
             Set to allow a telescope name argument.
         """
 
-        _job_group = self.add_argument_group("telescope model")
+        if add_telescope:
+            _job_group = self.add_argument_group("telescope model")
+        else:
+            _job_group = self.add_argument_group("site model")
```
VictorBarbosaMartins, line 248: if add_telescope is hard wired in l.61 we don't need this check here. I suggest either propagating the option to pass add_telescope to the application or get rid of the if statement.

VictorBarbosaMartins: After looking at the remaining changes, getting rid of this if statement sounds like a better option to me.

VictorBarbosaMartins (commented):
Hi @GernotMaier. Thanks for this clean up of the make_regular_array application. I provide a couple of minor comments.

simtools/applications/make_regular_arrays.py
GernotMaier: Good points, I've improved the docstring.

simtools/applications/make_regular_arrays.py
GernotMaier: OK, added `e.g., prod6`

simtools/applications/make_regular_arrays.py
GernotMaier: Yes. This will be fixed in one of the next pull request (already done on a local branch)

simtools/applications/make_regular_arrays.py
GernotMaier: It is not a log message.

simtools/configuration/commandline_parser.py
GernotMaier: Yes, here I followed all the other parameters to be consistent.

simtools/configuration/commandline_parser.py
GernotMaier: But we still have line 59:

- with `add_telescope=True` we allow to specify `--telescope` and `--site`  on the command line
- with `add_telescope=False`, we allow only `site` (this is important for applications which depend on the site only

Or did I misunderstand your comment?

GernotMaier:
@VictorBarbosaMartins - thanks for the quick review. I've improved the docstring in question and answered on the other issue. Let me know how to continue.

simtools/configuration/commandline_parser.py
VictorBarbosaMartins: Ah, that is all good (I looked at a different function).

VictorBarbosaMartins (approved):
All good from my side. We can merge once tests pass. Thanks!

simtools/applications/make_regular_arrays.py
```diff
@@ -72,101 +68,92 @@ def main():
...
+                telescope_name=tel_size + "-04",
+                crs_name="ground",
+                xx=-telescope_distance[tel_size],
+                yy=-telescope_distance[tel_size],
+                tel_corsika_z=0 * u.m,
+            )
+
+        layout.convert_coordinates()
+        layout.print_telescope_list()
+        output_file = str(args_dict.get("output_file", None))
```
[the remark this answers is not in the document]
GernotMaier: Correct - and it is not needed!

See another short PR: https://github.com/gammasim/simtools/pull/780

PR #782, opened 2024-01-15
add test for InconsistentHistogramFormat (and change combined_hists)

### LINKED ISSUE #764: github complains about missing tests in PR-unrelated file
Screenshot 2024-01-04 at 14 12 42

Github seems to be complaining about lines not tested which are not related to the PRs being worked on. We should add tests for these lines or add a comment that these lines should be ignored

VictorBarbosaMartins:
Adding tests to solve #764.

I had to change one function which was a method before to become a property and allow manually changing an attribute to be able to test the consistency. I hope that is not too tricky.

Ready for review.

Closes #764.

### DIFF
--- a/simtools/simtel/simtel_histograms.py
+++ b/simtools/simtel/simtel_histograms.py
@@ -80,11 +80,7 @@ class SimtelHistograms:
             for file in self._histogram_files:
                 with EventIOFile(file) as f:
                     for o in yield_toplevel_of_type(f, Histograms):
-                        try:
-                            hists = o.parse()
-                        except Exception as error_msg:  # pylint: disable=broad-except
-                            self._logger.warning(f"Problematic file {file}: {error_msg}")
-                            continue
+                        hists = o.parse()
                         self._list_of_histograms.append(hists)
         return self._list_of_histograms
 
--- a/simtools/simtel/simtel_histograms.py
+++ b/simtools/simtel/simtel_histograms.py
@@ -123,11 +123,8 @@ class SimtelHistograms:
         # Processing and combining histograms from multiple files
         if self._combined_hists is None:
             self._combined_hists = []
-            n_files = 0
-            for hists_one_file in self.list_of_histograms:
-                count_file = True
-
-                if len(self._combined_hists) == 0:
+            for i_hist, hists_one_file in enumerate(self.list_of_histograms):
+                if i_hist == 0:
                     # First file
                     self._combined_hists = copy.copy(hists_one_file)
 
@@ -139,9 +136,7 @@ class SimtelHistograms:
                             this_combined_hist["data"], hist["data"]
                         )
 
-                n_files += int(count_file)
-
-            self._logger.debug(f"End of reading {n_files} files")
+            self._logger.debug(f"End of reading {len(self.list_of_histograms)} files")
         return self._combined_hists
 
     @combined_hists.setter
--- a/simtools/simtel/simtel_histograms.py
+++ b/simtools/simtel/simtel_histograms.py
@@ -41,14 +41,12 @@ class SimtelHistograms:
         self._histogram_files = histogram_files
         self._is_test = test
         self._list_of_histograms = None
-        self.combined_hists = None
+        self._combined_hists = None
         self.__meta_dict = None
 
     @property
     def number_of_histograms(self):
         """Returns number of histograms."""
-        if self.combined_hists is None:
-            self.combine_histogram_files()
         return len(self.combined_hists)
 
     def get_histogram_title(self, i_hist):
@@ -65,8 +63,6 @@ class SimtelHistograms:
         str
             Histogram title.
         """
-        if self.combined_hists is None:
-            self.combine_histogram_files()
         return self.combined_hists[i_hist]["title"]
 
     @property
@@ -79,17 +75,17 @@ class SimtelHistograms:
         list:
             List of histograms.
         """
-
-        self._list_of_histograms = []
-        for file in self._histogram_files:
-            with EventIOFile(file) as f:
-                for o in yield_toplevel_of_type(f, Histograms):
-                    try:
-                        hists = o.parse()
-                    except Exception:  # pylint: disable=broad-except
-                        self._logger.warning(f"Problematic file {file}")
-                        continue
-                    self._list_of_histograms.append(hists)
+        if self._list_of_histograms is None:
+            self._list_of_histograms = []
+            for file in self._histogram_files:
+                with EventIOFile(file) as f:
+                    for o in yield_toplevel_of_type(f, Histograms):
+                        try:
+                            hists = o.parse()
+                        except Exception:  # pylint: disable=broad-except
+                            self._logger.warning(f"Problematic file {file}")
+                            continue
+                        self._list_of_histograms.append(hists)
         return self._list_of_histograms
 
     def _check_consistency(self, first_hist_file, second_hist_file):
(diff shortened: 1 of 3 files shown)


simtools/simtel/simtel_histograms.py
```diff
@@ -41,14 +41,12 @@ def __init__(self, histogram_files, test=False):
         self._histogram_files = histogram_files
         self._is_test = test
         self._list_of_histograms = None
-        self.combined_hists = None
+        self._combined_hists = None
```
tobiaskleiner, line 44: I believe it should be self.combined_hists here.

simtools/simtel/simtel_histograms.py
```diff
@@ -79,17 +75,17 @@ def list_of_histograms(self):
...
-                        continue
-                    self._list_of_histograms.append(hists)
+        if self._list_of_histograms is None:
+            self._list_of_histograms = []
+            for file in self._histogram_files:
+                with EventIOFile(file) as f:
+                    for o in yield_toplevel_of_type(f, Histograms):
+                        try:
+                            hists = o.parse()
+                        except Exception:  # pylint: disable=broad-except
```
tobiaskleiner, line 85: Do you know what type of Exceptions are these? Maybe add a test as well when this Exception is triggered.

simtools/simtel/simtel_histograms.py
```diff
@@ -79,17 +75,17 @@ def list_of_histograms(self):
...
+                            self._logger.warning(f"Problematic file {file}")
```
tobiaskleiner, line 86: I don't know about the severity of this, but maybe raising an error here would be better?

simtools/simtel/simtel_histograms.py
```diff
@@ -79,17 +75,17 @@ def list_of_histograms(self):
...
+                            continue
```
tobiaskleiner, line 87: Related to previous comment about adding a test here.

simtools/simtel/simtel_histograms.py
```diff
@@ -120,28 +116,45 @@ def _check_consistency(self, first_hist_file, second_hist_file):
...
         """Add the values of the same type of histogram from the various lists into a single
         histogram list."""
         # Processing and combining histograms from multiple files
-        self.combined_hists = []
-        n_files = 0
-        for hists_one_file in self.list_of_histograms:
-            count_file = True
+        if self._combined_hists is None:
+            self._combined_hists = []
+            n_files = 0
```
tobiaskleiner, line 126: Can this counting here maybe be simplified and instead using the length of the list of histograms?

simtools/simtel/simtel_histograms.py
```diff
@@ -120,28 +116,45 @@ def _check_consistency(self, first_hist_file, second_hist_file):
...
+                        this_combined_hist["data"] = np.add(
+                            this_combined_hist["data"], hist["data"]
+                        )
 
-            n_files += int(count_file)
+                n_files += int(count_file)
 
-        self._logger.debug(f"End of reading {n_files} files")
+            self._logger.debug(f"End of reading {n_files} files")
+        return self._combined_hists
```
tobiaskleiner, line 145: Is this return here needed/useful?

simtools/simtel/simtel_histograms.py
```diff
@@ -120,28 +116,45 @@ def _check_consistency(self, first_hist_file, second_hist_file):
...
+    def combined_hists(self, new_combined_hists):
+        """
+        Setter for combined_hists.
+
+        Parameters
+        ----------
+        new_combined_hists:
+            Combined histograms.
+        """
+        self._combined_hists = new_combined_hists
```
tobiaskleiner, line 157: This seems to be not used somewhere in the code and codecov complains about missing test. This setter might need additional checks to only allow actual combined histograms in the right format.

tobiaskleiner (changes requested):
Looks in general good. There are possibly some issues with the property defined and the requirements for the setter. Also it would be good to add a test for the Exception and specifiy its kind.

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: `except Exception` catches the exceptions dealt with in pyeventio but not `BaseException` as I understood from [here](https://stackoverflow.com/questions/18982610/difference-between-except-and-except-exception-as-e). 

The `# pylint: disable=broad-except` lets us not test this line and get away with it, but indeed it is at least important to pass on the information, hence I added the message to the logger now.

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: I was also wondering about it (did not add those lines now). But I could not come up with a test for this. I guess we would have to mess up the simtel output file such that the parser from eventio does not work. That is not supposed to be done in simtools though. Not raising an error above means we want to access the file even if the parser does not work. If we know what we are doing in the following, I think that is fine.

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: Indeed, I simplified it now.

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: Yes. Every property has to have a return statement. Also, we are not supposed to use the internal attribute `self._combined_hists` elsewhere. Instead, we should always use `self.combined_hists` and get the correct value returned.

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: I only added this setter to be able to test an invalid format in the unit tests, which is necessary since we dont have invalid histograms to start with. I added a line to test the setter though

VictorBarbosaMartins:
Hi @tobiaskleiner, thanks for the review and the comments. I addressed/answered them accordingly. Please let me know if there is something missing.

tobiaskleiner (approved):
Looks all good now. Thanks for the adjustments.

simtools/simtel/simtel_histograms.py
GernotMaier: Do I see it correctly the EventIOFile raises on ValueErrors? If this is correct, you could be a bit more specific.

See here for the function definition: https://github.com/cta-observatory/pyeventio/blob/16a6061102512e6db5f1fa3f81886fb95c7f327e/src/eventio/base.py#L57

VictorBarbosaMartins: Why is being tested above is the parsers (`o.parse`) inside `yield_toplevel_of_type`, which does not raise anything but a warning:

https://github.com/cta-observatory/pyeventio/blob/16a6061102512e6db5f1fa3f81886fb95c7f327e/src/eventio/search_utils.py#L20C1-L26C1

In principle we could even remove this try and except statement. I have not written these lines of code in the PR but I assume there was a reason for Raul including it, therefore I prefer to keep them.

VictorBarbosaMartins: Also, I am not in favor of discussing here topics that are not strictly related to this PR. Let's keep it simple.

GernotMaier: Sorry, wrong link: is this histgram.parse? Does it raises any exceptions? I actually cannot see any in https://github.com/cta-observatory/pyeventio/blob/16a6061102512e6db5f1fa3f81886fb95c7f327e/src/eventio/histograms.py#L10

Just wondering - I think this is one of the few places in the code (only?) where we introduce a broad exception.

VictorBarbosaMartins: I removed now the try and except

simtools/simtel/simtel_histograms.py
VictorBarbosaMartins: But all attributes should be defined as None is the init. So what you mean is that the init should already call the setter function? I think that is strange.

orelgueta: What I mean is that if you follow Tobias' suggestion and do instead `self.combined_hists = None`, you will be using the setter function to set it to `None`. That is the preferred method and should always be used. 

That is why codecov complained before btw, because you weren't using the setter function when you ran your tests.

VictorBarbosaMartins: I tried to address your issue but this is quite annoying. `combined_hists ` is a property which is estimated from the list of histograms, hence, it is not supposed to be changed. Allow me to explain the chain of unnecessary changes triggered by a initial comment:

- First, to test the `InconsistentHistogramFormat` error, I needed to implement this setter such that we manually change a histogram property. In fact, we are not supposed to be able to change a histogram like this. But now we are able to test the error at least;
- Second, `self.combined_hists` is estimated only once, i.e. when `self._combined_hists` is None (initial state). By setting `self.combined_hists = None` in the init, pylint gets all confused later when we refer to e.g. `self.combined_hists[i_hist]["title"]` which it thinks is still a None, but we actually get the value calculated by the property function. Tests work but pylint does not like it.

image

Using the setter in the init, especially for a function that should not be set by the user is not straightforward and demands unnecessary complication.

I refer you to the best answer in https://stackoverflow.com/questions/55800218/setting-default-empty-attributes-for-user-classes-in-init and also to the chatgpt final answer to my question "should one use the setter to initialize an attribute to None in python classes?":
>In general, the choice between direct initialization and using a setter depends on the specific requirements of your class and whether you need the additional capabilities that a setter provides. If you just need to set an attribute to None without any special logic, direct initialization in the __init__ method is often more straightforward.

VictorBarbosaMartins: Also: https://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters

The pythonic way does not use setter in the init.

orelgueta: Definitely didn't mean to cause so much trouble =) 
It's perfectly fine to keep the original `self._combined_hists = None`. 

Let me just explain why I made the comment. A few weeks ago I found on Stack Overflow an opposite example to what you provided here. Then I had the issue of a line uncovered by tests and I found a post that said I should always use the setter. In my trivial case it worked fine. However, it is clear from your links and experience that it is not true and in case of initialising to None, we can bypass it. So we can forget about it.

VictorBarbosaMartins: Well, the trouble was already there, you didn't cause it :)

ok, Stack Overflow probably not always points to a definitive solution, it seems. I agree we could try to use the setter whenever possible in the init, but sometimes it is better to keep it safer and simpler by using the internal attribute instead.

Issue #658, opened 2023-11-06
Wrong naming of coordinate systems in `print_array_elements`


GernotMaier:
The naming of the coordinate system `CORSIKA` is not correct in `print_array_elements` and classes associated with it.

- CORSIKA coordinate systems are aligned with geomagnetic North
- this coordinate system in simtools is aligned with geographic North (the difference is handled in CORISKA using the `ARRANG` parameter)

Find a better naming "ground or simtools coordinates?" and replace it accordingly in the code.

orelgueta:
I would vote for `ground` or `local` coordinates. `sim_telarray` coordinates (not `simtools`) would also be reasonable, but is a bit too specific I think.

PR #786, opened 2024-01-17
Bug fix in mercator coordinator transformation

GernotMaier:
Fixed call to set coordinates of a telescopes when input is in Mercator in [simtools/layout/array_layout.py](simtools/layout/array_layout.py).

Includes also 

- improvements in both units and integration tests in this context. 
- renaming of `alt` to `altitude` in some remaining places (these had been missed before)
- renaming of 'CORSIKA' to 'Ground' in some remaining places (had also been missed before)

Note that we hardly use the longitude/latitude system (essentially only to get the coordinates of the reference positions, and we will not use them for telescope coordinates. Still, for testing we are doing the conversion - to allow this to work I introduced a flag `skip_input_validation`, as we don't want to introduce a schema for mercator just for testing.

### DIFF
--- a/simtools/applications/print_array_elements.py
+++ b/simtools/applications/print_array_elements.py
@@ -162,7 +162,7 @@ def main():
     layout = array_layout.ArrayLayout(
         telescope_list_file=args_dict["input"],
         telescope_list_metadata_file=args_dict["input_meta"],
-        validate=True,
+        validate=False,
     )
     layout.select_assets(args_dict["select_assets"])
     layout.convert_coordinates()
--- a/simtools/applications/print_array_elements.py
+++ b/simtools/applications/print_array_elements.py
@@ -162,7 +162,7 @@ def main():
     layout = array_layout.ArrayLayout(
         telescope_list_file=args_dict["input"],
         telescope_list_metadata_file=args_dict["input_meta"],
-        validate=False,
+        validate=True,
     )
     layout.select_assets(args_dict["select_assets"])
     layout.convert_coordinates()
--- a/simtools/applications/print_array_elements.py
+++ b/simtools/applications/print_array_elements.py
@@ -145,6 +145,13 @@ def _parse(label=None, description=None):
         default=None,
         nargs="+",
     )
+    config.parser.add_argument(
+        "--skip_input_validation",
+        help="skip input data validation against schema",
+        default=False,
+        required=False,
+        action="store_true",
+    )
     return config.initialize(output=True, require_command_line=True)
 
 
@@ -162,7 +169,7 @@ def main():
     layout = array_layout.ArrayLayout(
         telescope_list_file=args_dict["input"],
         telescope_list_metadata_file=args_dict["input_meta"],
-        validate=True,
+        validate=not args_dict["skip_input_validation"],
     )
     layout.select_assets(args_dict["select_assets"])
     layout.convert_coordinates()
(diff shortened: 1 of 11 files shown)


tests/unit_tests/layout/test_telescope_position.py
```diff
@@ -55,15 +55,15 @@ def test_str(crs_wgs84, crs_local, crs_utm):
...
+    _test_string = "LST-01\t Ground x(->North): 50.00 y(->West): -25.00"
     assert _tcors == (_test_string + "\t Alt: 2158.00")
     tel.convert_all(crs_local=crs_local, crs_wgs84=crs_wgs84)
     _tcors = tel.__str__()
     _test_string += "\t Longitude: 28.76262 Latitude: -17.89177"
     assert _tcors == (_test_string + "\t Alt: 2158.00")
     tel.convert_all(crs_local=crs_local, crs_wgs84=crs_wgs84, crs_utm=crs_utm)
     _tcors = tel.__str__()
-    _test_string = "LST-01\t CORSIKA x(->North): 50.00 y(->West): -25.00"
+    _test_string = "LST-01\t Ground x(->North): 50.00 y(->West): -25.00"
```
VictorBarbosaMartins, line 66: I don't remember whether we discussed about replacing CORSIKA by Ground but I am afraid Ground does not pass the same information as CORSIKA. CORSIKA already comes with the definition of the coordinate system, where X points towards North and Ground is somewhat vague. If we manage to defined Ground somewhere to consider the coordinate system as CORSIKA I would agree with this chance.
Screenshot 2024-01-23 at 13 40 55

simtools/applications/print_array_elements.py
```diff
@@ -145,6 +145,13 @@ def _parse(label=None, description=None):
         default=None,
         nargs="+",
     )
+    config.parser.add_argument(
```
VictorBarbosaMartins, line 148: Not sure it justifies adding a parameter to skip the validation, which is necessary only in the very specific case of longitude/latitude? Isn't it better to implement an if statement in the code and exclude the validation when longitude/latitude is used?

VictorBarbosaMartins (commented):
Thanks for implementing this (and increasing coverage of both integration and unit tests). I leave two main comments proposing discussions.

tests/unit_tests/layout/test_telescope_position.py
GernotMaier: The name CORSIKA is wrong, as this is not the CORSIKA coordinate system. Ours is aligned to geographic north, while CORSIKA is aligned with magnetic north.

This was discussed in https://github.com/gammasim/simtools/issues/658 and we have changed it everywhere else in the code.

simtools/applications/print_array_elements.py
GernotMaier: I do think this is useful for other cases. We use this application a lot with input from quickly generated files. In many cases, the metadata is not complete (which is fine, as these are temporary fine) and maybe not all columns are there. So this is useful to have for this application.

tests/unit_tests/layout/test_telescope_position.py
VictorBarbosaMartins: ok!

simtools/applications/print_array_elements.py
VictorBarbosaMartins: ok. this should not become default though.

VictorBarbosaMartins (approved):

GernotMaier:
@VictorBarbosaMartins - thanks for the quick review!

PR #787, opened 2024-01-24
Add CONTEXT:NOTES to metadata

GernotMaier:
Context fields are useful to add additional information for a data product. Adding here a CONTEXT:NOTES filed to add any associated comments / notes to the metadata.  As all CONTEXT field, this one is also optional.

In detail, this is a list:
```
CONTEXT:
   NOTES:
       - TITLE: "..."  (optional)
          TEXT: "..."  (required)
          CREATION_TIME: "date field"  (optional)
```

### DIFF
--- a/simtools/schemas/metadata.metaschema.yml
+++ b/simtools/schemas/metadata.metaschema.yml
@@ -462,6 +462,39 @@ definitions:
         type: object
         additionalProperties: false
         properties:
+          NOTES:
+            title: Context notes.
+            description: |-
+              Notes that provide additional context for this data product.
+            type: array
+            items:
+              type: object
+              additionalProperties: false
+              properties:
+                TITLE:
+                  description: |-
+                    Title of note.
+                  anyOf:
+                    - type: string
+                    - type: "null"
+                  default: null
+                TEXT:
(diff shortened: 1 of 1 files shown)


tobiaskleiner (approved):

orelgueta:
Do you have a test that checks that this input is actually read properly with all the different allowed options?

GernotMaier:
> Do you have a test that checks that this input is actually read properly with all the different allowed options?

We run the json schema validation each time we read the metadata from a file. This also checks the list of required options and would complain if something is missing. As this is an optional field, it will not complain if it doesn't exist.

Note that we define our schemas with `additionalProperties: false`, so that no new fields which are not in the schema are allowed.

orelgueta:
Sure, but my question is whether you have one file in the tests which has this content and you can therefore make sure that the schema is written properly (correct indentation, list where you expect it, etc.). I am asking just because it is hard to validate those points by reading the schema and example data that is validated is the safest way.

GernotMaier:
Good point - we have `tests/resources/altitude.json`, which I will adapt accordingly. Will come soon.

PR #791, opened 2024-01-26
Cosmic ray spectrum

VictorBarbosaMartins:
This is not yet the application that calculates the trigger rate, but it is an application that estimates the number of protons that reach a certain area in a certain observation time and given a certain cone in the sky. This is a tool that might help deciding how many protons we would like to simulate for a production. However, the trigger ratio application (yet to come) will be more important.

After discussions with Orel, Max and Karl, we have decided to detach the spectral information from pyirf and place it into a separate module. Hence, I started a new repository under my account (**_temporary_**) as [ctao_cosmic_ray_spectra](https://github.com/VictorBarbosaMartins/ctao_cosmic_ray_spectra/tree/main) and copied [spectral.py](https://github.com/cta-observatory/pyirf/blob/0059d3bceea126d014ba9efc953cb6470c37e7b0/pyirf/spectral.py) and [simulations.py](https://github.com/cta-observatory/pyirf/blob/0059d3bceea126d014ba9efc953cb6470c37e7b0/pyirf/simulations.py) from pyirf into this new repository.

I adapted the code in `ctao_cosmic_ray_spectra` and included new functions to allow the integration of the proton spectrum in time and area. Once this is all ready and agreed on, we might move it to the `cta_observatory` account and remove the duplicated functions from pyirf. At the moment, the package is not published anywhere, so in order to test the implementation in `simtools`, one needs to run `pip install git+https://github.com/VictorBarbosaMartins/ctao_cosmic_ray_spectra.git` beforehand. This works both inside the container and outside the container (if testing locally).

I would like to get comments from the group, both on the general idea, on the changes proposed here and on the separate package created. Please note that for the moment the tests will fail because `ctao_cosmic_ray_spectra` is not yet being installed by default. Please test the application with:

```
pip install git+https://github.com/VictorBarbosaMartins/ctao_cosmic_ray_spectra.gitpython simtools/applications/calculate_proton_events.py --config tests/integration_tests/config/calculate_proton_events.yml
```

Thank you!


VictorBarbosaMartins:
This is embedded (in a different version) in #801 and therefore not necessary as a separate PR.

PR #796, opened 2024-01-30
Fix workflow to generate packages after release

GernotMaier:
Workflows to generate packages don't run through for releases, see e.g., [here](https://github.com/gammasim/simtools/actions/runs/7712878687/job/21021437303#step:8:335)

Reason is that we are using

`git clone -b $BUILD_BRANCH https://github.com/gammasim/simtools.git`

to clone a certain package. This works for pull requests with $BUILD_BRANCH is the actual branch name.

For releases, $BUILD_BRANCH is set e.g., to `refs/tags/v0.6.0`, which does not work with `git clone`. Added a sed command to remove `refs/tags/` so that in the example we do `git clone -b v0.6.0 ...`

I think the final test can only be done with a release. So after review and approval, I will release v0.6.1.

### DIFF
--- a/docker/Dockerfile-prod
+++ b/docker/Dockerfile-prod
@@ -9,7 +9,8 @@ RUN microdnf update -y && microdnf install -y \
     python3.11-devel && \
     microdnf clean all
 
-RUN git clone -b $BUILD_BRANCH https://github.com/gammasim/simtools.git --depth 1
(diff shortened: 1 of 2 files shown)


tobiaskleiner (approved):

GernotMaier:
To confirm: this worked, see e.g. 0.6.1 package at https://github.com/gammasim/simtools/pkgs/container/simtools-prod

PR #805, opened 2024-02-19
Add CI for daily backups of the simulation model DB

GernotMaier:
Add a CI for daily backups of the simulation model DB:

- runs every night at 2 am
- backup to sync and share folder at DESY (in GM area)
- runs md5sum over all files and moves the checksums also to sync and share
- all 7 databases are backed up (not that the matrix workflow is forced to run non-sequential by setting `max-parallel: 1` (no need to query the DB in parallel)
- mongodump uses the gzip option (important to note for restoring)

Uses the official mongo container. There is one inefficiency in the script: curl is installed for each matrix job, I simply could find out how to do this out of the loop. This takes 20s, and I don't think we are performance limited.

Quite important: I backup is only verified when it has been used to restore the data. This is what we haven't done and I am not sure if we should add this with a sandbox as part of this CI.

### DIFF
--- a/.github/workflows/DB-Backup.yml
+++ b/.github/workflows/DB-Backup.yml
@@ -5,8 +5,6 @@ on:
   schedule:
     - cron: '0 2 * * *'  # Run every night at 2 am
   workflow_dispatch:     # Manual trigger
-  pull_request:
-    types: [opened, synchronize]
 
 jobs:
   backup:
--- a/.github/workflows/DB-Backup.yml
+++ b/.github/workflows/DB-Backup.yml
@@ -1,0 +1,62 @@
+name: Simulation model DB Backup
+env:
+  SIMTOOLS_DB_SERVER: ${{ secrets.DB_SERVER }}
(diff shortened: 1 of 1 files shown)


GernotMaier:
This is how the backup folder looks like:

Screenshot 2024-02-19 at 11 09 17

Will remove the following trigger before merging:
```
 pull_request:
    types: [opened, synchronize]
```
(Otherwise each pull request will lead to a backup)

tobiaskleiner (approved):

orelgueta (approved):
I thought you were going to do it with a single script that loops over the databases instead of using the CI matrix. 
No problem with using the CI matrix obviously, just assume it was a bit more complicated than a bash script. 

All good anyway, approving. Please do not forget to remove the `types: [opened, synchronize]` before merging.

GernotMaier:
Thanks! I didn't wanted to add an additional script. Of course could have added as well a for loop in the workflow at one step, but let's keep it for now this way.

I removed trigger of workflow by pull requests now.

PR #807, opened 2024-02-19
Add application to add a new parameter to the database from a json file

GernotMaier:
This PR adds a new application `add_new_parameter_from_json_file_to_db.py` to read a DB-style json file and add this as new parameter to the DB.

**IMPORTANT**

- this PR merges in to the new-simulation-model branch
- nothing is added to the DB yet when running the application (don't want to mess up the database).

## Example 1 - adding to the `telescopes` collection

The json files looks like:

```
{
    "parameter": "telescope_sphere_radius",
    "instrument": "LSTN-design",
    "site": "North",
    "version": "2024-02-01",
    "value": 12.5,
    "unit": "m",
    "type": "numpy.float64",
    "applicable": true,
    "file": false
}
```

Running:

```
python simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py --file_name ../simulation_model/verified_model/LSTN-design/telescope_sphere_radius.json --db_collection "telescopes"
 ```
 
will add the following new parameter to the DB:

```
{'instrument': 'LSTN-design', 'version': '2024-02-01', 'parameter': 'telescope_sphere_radius', 'site': 'North', 'value': 12.5, 'type': 'numpy.float64', 'file': False, 'applicable': True, 'unit': 'm'}
```

## Example - adding to the `sites` collection

Example is:

```
{
    "parameter": "epsg_code",
    "instrument": "North",
    "site": "North",
    "version": "2024-02-01",
    "value": 32628,
    "type": "int",
    "applicable": true,
    "file": false
}
```

Application:

```
python simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py --file_name ../simulation_model/verified_model/Site/North/epsg_code.json --db_collection "sites"
```

will add the new parameter:

```
{'instrument': 'North', 'version': '2024-02-01', 'parameter': 'epsg_code', 'site': 'North', 'value': 32628, 'type': 'int', 'file': False, 'applicable': True, 'unit': None}
```

## Missing

Added a parameter with `file=True` and the associated file. As this PR is really needed for #761, my suggestion is to move this functionality to a future update of the application.

## Unit and integration tests

Are failing....and this will be all fixed in #761.

### DIFF
--- a/simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
+++ b/simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
@@ -14,17 +14,16 @@
 import logging
 
 import simtools.utils.general as gen
-from simtools.db import db_handler
 from simtools.configuration import configurator
+from simtools.db import db_handler
 
 
 def main():
-    config = configurator.Configurator(
-        description=("Add a new parameter to the DB.")
-    )
+    config = configurator.Configurator(description=("Add a new parameter to the DB."))
     config.parser.add_argument("--file_name", help="file to be added", required=True)
     config.parser.add_argument(
-        "--db_collection", help="DB collection file will be added", required=True)
+        "--db_collection", help="DB collection file will be added", required=True
+    )
     args_dict, db_config = config.initialize(db_config=True)
 
     logger = logging.getLogger()
@@ -35,8 +34,7 @@ def main():
     if args_dict["db_collection"] not in ("telescopes", "sites"):
         raise ValueError(f"DB collection {args_dict['db_collection']} not recognized")
 
-    par_dict = gen.collect_data_from_file_or_dict(
-        file_name=args_dict["file_name"], in_dict=None)
+    par_dict = gen.collect_data_from_file_or_dict(file_name=args_dict["file_name"], in_dict=None)
 
     logger.info(f"Adding the following parameters to the DB: {par_dict['parameter']}")
 
--- a/simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
+++ b/simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
@@ -1,0 +1,60 @@
+#!/usr/bin/python3
+
+"""
+    Summary
+    -------
+    This application is used to add a new parameter to the sites collection in the DB
+    using a json file as input.
+
+    This application should not be used by anyone but expert users and not often.
+    Therefore, no additional documentation about this applications will be given.
+
+"""
+
+import logging
+
+import simtools.utils.general as gen
+from simtools.db import db_handler
+from simtools.configuration import configurator
+
+
+def main():
+    config = configurator.Configurator(
+        description=("Add a new parameter to the DB.")
+    )
+    config.parser.add_argument("--file_name", help="file to be added", required=True)
+    config.parser.add_argument(
+        "--db_collection", help="DB collection file will be added", required=True)
+    args_dict, db_config = config.initialize(db_config=True)
+
+    logger = logging.getLogger()
+    logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
+
+    db = db_handler.DatabaseHandler(mongo_db_config=db_config)
+
+    if args_dict["db_collection"] not in ("telescopes", "sites"):
+        raise ValueError(f"DB collection {args_dict['db_collection']} not recognized")
+
+    par_dict = gen.collect_data_from_file_or_dict(
+        file_name=args_dict["file_name"], in_dict=None)
+
+    logger.info(f"Adding the following parameters to the DB: {par_dict['parameter']}")
+
+    db.add_new_parameter(
+        db_name=db.DB_CTA_SIMULATION_MODEL,
+        telescope=par_dict["instrument"],
+        parameter=par_dict["parameter"],
+        version=par_dict["version"],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by GernotMaier** [comment]
@tobiaskleiner & @orelgueta 

- small variation of `add_new_parameter_to_db.py` to read directly from a json file
- the temporary line is explicitly left to make sure not to add to the DB

**Comment by GernotMaier** [comment]
note that the failing unit test is expected: I've switched off the addition of a new parameter to the DB for now.

File: simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+import simtools.utils.general as gen
+from simtools.configuration import configurator
+from simtools.db import db_handler
+
+
+def main():
+    config = configurator.Configurator(description=("Add a new parameter to the DB."))
+    config.parser.add_argument("--file_name", help="file to be added", required=True)
+    config.parser.add_argument(
+        "--db_collection", help="DB collection file will be added", required=True
```
orelgueta, line 25: Description should be "DB collection to which to add the parameter".

simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    args_dict, db_config = config.initialize(db_config=True)
+
+    logger = logging.getLogger()
+    logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
+
+    db = db_handler.DatabaseHandler(mongo_db_config=db_config)
+
+    if args_dict["db_collection"] not in ("telescopes", "sites"):
```
orelgueta, line 34: Why are you limiting to those specific ones? Because they are only ones which the specific schema you are using? 
In general I am not crazy about hardcoding these names here. Maybe you can get the available names from `db`? Or perhaps at that check to the `add_new_parameter` function? 
Ahh it's actually already in that function, so this test can be removed from here.

simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
```diff
@@ -0,0 +1,64 @@
...
+            telescope=par_dict["instrument"],
+            parameter=par_dict["parameter"],
+            version=version,
+            value=value,
+            site=par_dict["site"],
+            type=par_dict["type"],
+            collection_name=args_dict["db_collection"],
+            applicable=par_dict["applicable"],
+            file=par_dict["file"],
+            unit=par_dict["unit"] if "unit" in par_dict and par_dict["unit"] is not None else None,
```
orelgueta, line 58: Notice that if the value is given with a unit like `10 m`, then the unit will be taken from there (after the change of `units` to `unit`). I think that if you give both, the `par_dict["unit"]` would take priority. Probably worth checking for that case (or at least notify the user).

orelgueta (approved):
A couple of minor comments. Otherwise all good so I am approving already.

Two general comments,
- I thought you planned to validate the schema using json schema instead of the db_handler function. Did you decide this is better?
- In the PR description you show the entry that would be added to the DB, but you forgot the ObjectID which mongodb adds automatically. Not important of course, but for completeness, since the PR description is so clear, it's probably worth mentioning.

simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
GernotMaier: Fixed.

simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
GernotMaier: Fixed.

simtools/applications/db_development_tools/add_new_parameter_from_json_file_to_db.py
GernotMaier: This is actually a bug: the `unit` argument is not used at all in add_new_parameter.

I've changed it now that (mostly in general.get_value_unit_type):
- if value is a astropy quantity and unit is given, it is expected that they can be converted to each other. Value and unit is given as "unit" (so value might be multiplied  e.g. from km to m). Added the corresponding unit test.

GernotMaier:
@orelgueta , thanks a lot. Could you have another quick look, especially at the unit conversion?

> A couple of minor comments. Otherwise all good so I am approving already.
> 
> Two general comments,
> 
>     * I thought you planned to validate the schema using json schema instead of the db_handler function. Did you decide this is better?

Yes, I've started to do this but discarded it as:

- PR #761 introduces a lot of changes to db_handler which are not reviewed. I didn't want to add on top of that many changes
- we probably want to changes the name validation in names.py to use json schema as well. We otherwise have to define valid site and telescope names in names.py and in a json schema. I guess we don't want that duplication.

> 
>     * In the PR description you show the entry that would be added to the DB, but you forgot the ObjectID which mongodb adds automatically. Not important of course, but for completeness, since the PR description is so clear, it's probably worth mentioning.

Not sure I understand. As the object_ID is added by the DB, why does it need to be in the json file? Or do you simply want to add a sentence above that this id is automatically added?

orelgueta:
All good now.

> simply want to add a sentence above that this id is automatically added

Yes, that's what I meant.

PR #819, opened 2024-03-01
Fix failing raytracing unit test 

GernotMaier:
Unclear why this worked before - setting `caplog.at_level(logging.INFO)` uses the correct logging level.

Further minor changes:

- remove a spurious debug print out in a test
- slightly rearrange the check for KeyError in raytraing:plot (asking for forgiveness..)

### DIFF
--- a/simtools/ray_tracing.py
+++ b/simtools/ray_tracing.py
@@ -408,16 +408,17 @@ class RayTracing:
         KeyError
             If key is not among the valid options.
         """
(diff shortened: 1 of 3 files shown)


VictorBarbosaMartins (approved):
All good. I also have not figured out why the the level of caplog was changed, as it is supposed to be always come back to its original state when the `with` statement ends.

GernotMaier:
OK, thanks for the review!!

PR #827, opened 2024-03-06
Re enable camera efficiency and ray tracing tests; files from test-data

### LINKED ISSUE #804: Move files from test-data database to tests/resources directory
Currently unit tests use files from tests/resources and from the test-data database. For consistency, all files should be moved to test/resources.

These are:

- [x] photons-North-LST-1-d10.0-za20.0-off0.000_validate_optics.lis.gz (which needs to be named correctly with LSTN-01); Unit tests using these files are skipped (with and after PR #761) . Fixed with #827
- [x] camera-efficiency-table-North-LST-1-za020deg_azm000deg_validate_camera_efficiency.ecsv
- [x] Transmission_Spectrum_PlexiGlass.dat accessed in test_visualize:test_plot_table
- [x] corsikaConfigTest accessed in test_corsika_config:corsikaConfigTest.yml

GernotMaier:
This PR re-enables the unit tests for the camera efficiency and ray-tracing module and moves files from the test-data DB to ./test/resources

- the test file for the camera efficiency test is moved from the test-data database to test/resources.
- for ray tracing, a renaming of the test files to the new telescope naming ('LSTN-01') fixed the issues.
- files Transmission_Spectrum_PlexiGlass.dat and corsikaConfigTest.yml are added

During unit tests, no files from test-data are used anymore. Closes #804


tests/unit_tests/test_camera_efficiency.py
```diff
@@ -35,21 +36,17 @@ def camera_efficiency_sst(telescope_model_sst, simtel_path):
...
         sub_dir="camera-efficiency",
         dir_type="test",
     )
-    db.export_file_db(
-        db_name="test-data",
-        dest=output_directory,
-        file_name=test_file_name,
-    )
-
+    shutil.copy(test_file_name, output_directory)
```
tobiaskleiner, line 49: Probably not needed here, but copy2 also preserves metadata.

tobiaskleiner (approved):
Looks good.

tests/unit_tests/test_camera_efficiency.py
GernotMaier: OK - here not needed, as we just need to read the file and metadata like permissions are not important.

(not note the 'red warning' in the following URL that even copy2 does not preserve all metadata: https://docs.python.org/3/library/shutil.html)

GernotMaier:
Thanks for the review!

PR #831, opened 2024-03-07
Address conda warning about pip

GernotMaier:
Conda / mamba warns:

```
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
```

Added a one liner to add pip explicitly to the environment.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - matplotlib-base
   - numpy
   - numpydoc
+  - pip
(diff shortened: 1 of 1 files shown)


tobiaskleiner (approved):
All good!

PR #832, opened 2024-03-12
Add `-n auto` options to pytest CI

GernotMaier:
Add `-n auto` for parallel execution of pytests in CI (after seeing it in the ctapipe workflows).

Unit test time is reduced typically by a factor of 2 using this (typically we get 2 workers per workflow).

This revealed in a issue with the unit test `tests/unit_tests/layout/test_array_layout.py::test_array_layout_empty()` - success / failure seems to depend on other tests. I've spend an hour on trying to solve this (see also PR #833) but without success. I have now looked how we use the `ArrayLayout` class and I think there is no need to test the case of an constructor without arguments. I therefore removed the test.

### DIFF
--- a/.github/workflows/CI-integrationtests.yml
+++ b/.github/workflows/CI-integrationtests.yml
@@ -42,4 +42,4 @@ jobs:
           source /workdir/env/bin/activate
           pip install '.[tests,dev,doc]'
           export PYTHONPATH=$PYTHONPATH:.
-          pytest --color=yes --durations=10 --no-cov tests/integration_tests/
+          pytest --color=yes --durations=10 -n auto --dist loadscope --no-cov tests/integration_tests/
--- a/.github/workflows/CI-unittests.yml
+++ b/.github/workflows/CI-unittests.yml
@@ -97,7 +97,7 @@ jobs:
       - name: Unit tests
         shell: bash -l {0}
         run: |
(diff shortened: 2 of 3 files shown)


.github/workflows/CI-integrationtests.yml
```diff
@@ -42,4 +42,4 @@ jobs:
           source /workdir/env/bin/activate
           pip install '.[tests,dev,doc]'
           export PYTHONPATH=$PYTHONPATH:.
-          pytest --color=yes --durations=10 --no-cov tests/integration_tests/
+          pytest --color=yes --durations=10 -n auto --dist loadscope --no-cov tests/integration_tests/
```
tobiaskleiner, line 45: Is there a reason why you group the tests by module and not in random order using load?

GernotMaier: No strong reasons - it guarantees that all tests in a module run in the same process. I have used the ctapipe CI as a guidance here. Do you think we should change this?

tobiaskleiner: I don't have a strong optionion on this at the moment. We can let it like this and add the new CI we just discussed.

GernotMaier: see also issue #834

Issue #798, opened 2024-02-02
Cumbersome file name extraction in names.py


GernotMaier:
There is a list of methods in names.py to get file names:

```
   "camera_efficiency_log_file_name",
    "camera_efficiency_results_file_name",
    "camera_efficiency_simtel_file_name",
    "layout_telescope_list_file_name",
    "ray_tracing_file_name",
    "ray_tracing_plot_file_name",
    "ray_tracing_results_file_name",
    "simtel_array_config_file_name",
    "simtel_single_mirror_list_file_name",
    "simtel_telescope_config_file_name",
```

Each of them is a method on its own, with very similar code. Suggest to merge this into a single method - this will shorten this part of the code significantly.

PR #835, opened 2024-03-16
Simplify file naming in names.py

GernotMaier:
This PR addresses issue #798:

- file name retrieval in `names.py` for output files from especially camera_efficiency and ray tracing is very repetitive. Improved this a bit, but a condensation into a single function is hardly possible. I've simplified as much as possible. 
- introduced consistent naming patterns (e.g., should a value in a file name has a unit or not (yes, it always should))


simtools/camera_efficiency.py
```diff
@@ -193,7 +183,7 @@ def analyze(self, export=True, force=False):
         """
         self._logger.info("Analyzing CameraEfficiency")
 
-        if self._file_results.exists() and not force:
+        if self._file["results"].exists() and not force:
```
tobiaskleiner, line 186: Better to use `if "results" in self._file` for dict. Also codecov complains about missing test. Maybe you can add tests? Also for the following two lines 225 and 303.

simtools/ray_tracing.py
```diff
@@ -195,15 +197,16 @@ def simulate(self, test=False, force=False):
...
-                photons_file_name = names.ray_tracing_file_name(
-                    self._telescope_model.site,
-                    self._telescope_model.name,
-                    self._source_distance,
-                    self.config.zenith_angle,
-                    this_off_axis,
-                    this_mirror if self.config.single_mirror_mode else None,
-                    self.label,
-                    "photons",
+                photons_file_name = names.generate_file_name(
```
tobiaskleiner, line 200: Line not covered by test.

simtools/utils/names.py
```diff
@@ -557,210 +534,62 @@ def layout_telescope_list_file_name(name, label):
...
     """
-    File name for files required at the RayTracing class.
+    Generate a file name for output, config, or plotting.
+    Used e.g., to generate camera-efficiency and ray-tracing output files.
 
     Parameters
     ----------
+    file_type: str
+        Type of file (e.g., config, output, plot)
+    key: str
```
tobiaskleiner, line 558: This should be named suffix. Description is not matching.

tests/unit_tests/test_psf_analysis.py
```diff
@@ -12,7 +12,7 @@
 
 def test_reading_simtel_file(args_dict, io_handler):
     test_file = io_handler.get_input_data_file(
-        file_name="photons-North-LSTN-01-d10.0-za20.0-off0.000_validate_optics.lis.gz",
+        file_name="photons-North-LSTN-01-d10.0km-za20.0deg-off0.000deg_validate_optics.lis.gz",
```
tobiaskleiner, line 15: Why do we use here `0.000deg` for 0 offset, but above ` 000deg ` for 0 azimuth angle?

tests/unit_tests/utils/test_names.py
```diff
@@ -170,102 +170,144 @@ def test_get_telescope_type_from_telescope_name():
...
+            "camera-efficiency-table",
+            ".ecsv",
+            site,
+            telescope_model_name,
+            zenith_angle,
+            azimuth_angle,
+            label=label,
         )
-        == "camera-efficiency-table-North-MST-FlashCam-D-za040deg_azm000deg_test.ecsv"
+        == "camera-efficiency-table-North-MSTN-za40.0deg_azm000deg_test.ecsv"
```
tobiaskleiner, line 235: Related to previous comment, also seen later a couple of times.

tobiaskleiner (commented):
Thanks for the PR. It looks good in general and improves readability of the code a lot. I have added few minor comments.

simtools/ray_tracing.py
GernotMaier: See comment above, same thing.

simtools/utils/names.py
GernotMaier: Exactly. Changed it.

tests/unit_tests/test_psf_analysis.py
GernotMaier: This is motivated by:

- zenith and offset angle needs to be set on a sub-degree accuracy
- azimuth not - it is probably impossible to see a difference between a shower coming from 180 or 185 deg. 

I am generally not a friend of putting metadata in the file name. This information should be in a metdata header. We should fix this at some point in the future.

tests/unit_tests/utils/test_names.py
GernotMaier: Same reply - I've commented on zenith vs azimuth accuracy. Let me know if this makes sense, easy to change it to `azm000.0deg`, but I don't think it makes sense (and python is horrible in getting the accuracy right)

GernotMaier:
@tobiaskleiner - thanks for the comments and suggestions! Added them or commented on the issues. Let me know if this is ok.

simtools/camera_efficiency.py
tobiaskleiner: Ok I agree. This would then probably go in a seperate PR. If you agree we can open an issue to adress this.

tests/unit_tests/test_psf_analysis.py
tobiaskleiner: Yes this makes sense, but do we really need milli degree accuracy for offset? Agree to put this in the metadata in the future.

simtools/camera_efficiency.py
GernotMaier: Yes and no. We have all the tools like pycoverage and codecov to tell us the missing tests. Do we need to open additionally issues on that? I am really not sure.

tests/unit_tests/test_psf_analysis.py
GernotMaier: We actually have simulated at 0.125 deg offset at some point, so this is not completely wrong.

tobiaskleiner: Ok thanks, everything is clear then. Going to mark it as approved.

tobiaskleiner (approved):
Good to merge.

GernotMaier:
@tobiaskleiner - thank you!

PR #836, opened 2024-03-20
Improvements to model parameter schema validation

GernotMaier:
This pull requests adds the following improvement:

- schema file search path includes now for a pip-package installation the simtools.schema directory - meaning that if a schema file is not found, we search for it in the simtools/schema directory.
- improved naming of metaschema file used for data and model parameters - a bit lengthy but hopefully clearer: `model_parameter_and_data_schema.metaschema.yml` (I constantly got confused by the old name)
- `validate_file_using_schema` has a new command line option called `--require_exact_data_type`. This requires that values have exact type (e.g., an int is not allowed when a float is required in a field - although an int can converted to a float). We usually want to be relaxed and the default is therefore 'off'.


tobiaskleiner (approved):
All good.

GernotMaier:
@tobiaskleiner - thank you!

PR #838, opened 2024-03-20
Allow to push packages for manual workflow

GernotMaier:
`.github/workflows/build-docker-images.yml` builds the images `simtools-dev` and `simtools-prod`, but pushes them to the package registry only for the events 'merge to main' and 'release'. 

This PR allows to push an image to the package registry using the manual workflow `workflow dispatch`.

Fixes also a bug that all images have been generated for main only (which was not an issue really before, but would be now).

Tests:

- this workflow was triggered manually: https://github.com/gammasim/simtools/actions/runs/8362792450
- this package was uploaded by the workflow: https://github.com/gammasim/simtools/pkgs/container/simtools-prod/193601007?tag=20240320-163811

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -47,6 +47,7 @@ jobs:
           else
             echo "BUILD_BRANCH=${GITHUB_REF#refs/heads/}" >> "$GITHUB_ENV"
           fi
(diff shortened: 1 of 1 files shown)


GernotMaier:
@tobiaskleiner - thank you!

PR #842, opened 2024-03-24
Updates to get correct site and default values into DB

GernotMaier:
This PR is a follow up to #836 and includes updates relevant for the export of simtel model parameters using the applications `convert_all_model_parameters_from_simtel.py` and `convert_model_parameter_from_simtel.py`. Note that getting the handling of the format from simtel right is hard and difficult to write in a very clean code style.

We also expect that these applications are needed only in the short-to-medium term, until all model parameter are exported correctly to the databases.

Changes:

1. (hopefully) correct handling of `all:` variants in the simtel configuration files output (this is unfortunately complicated; e.g. "all:0", "all: 0", "all:0, 5:3" (the latter means all all values are set to zero with the exception of the 5th)
2. `convert_all_model_parameters_from_simtel.py` prints differences between default and set values for parameters which are not covered by schemas. This is important to debug the schema files in the [model_parameters](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters) and make sure that schemas are there for all model parameters and that defaults / limits are correct.
3. add `require_exact_data_type` command line option to `convert_all_model_parameters_from_simtel`: if set (not default), we require exactly the same data types in the validation (meaning 'int' as 'int64' is not accepted). In this context allow as well int values as floats (meaning: a '5' is allowed as value when the required data type is float)
4. add 'prod6' as allowed production (not ideal that these are hardwired in `names.py`)
5. print list of file-type model parameters and values (this is useful to find those files)


simtools/simtel/simtel_config_reader.py
```diff
@@ -159,27 +168,57 @@ def export_parameter_dict_to_json(self, file_name, dict_to_write):
...
+            # ignore limits checks for boolean
+            if data_type == "limits" and self.parameter_dict.get("type") == "bool":
+                continue
+            try:
+                if data_type == "limits":
+                    _from_schema = [
+                        self.schema_dict["data"][0]["allowed_range"].get("min"),
+                        self.schema_dict["data"][0]["allowed_range"].get("max"),
+                    ]
+                    _from_schema = _from_schema[0] if _from_schema[1] is None else _from_schema
```
tobiaskleiner, line 187: This means for parameters where we only set lower limits. We don't have/expect to have upper limits?

simtools/simtel/simtel_config_reader.py
```diff
@@ -159,27 +168,57 @@ def export_parameter_dict_to_json(self, file_name, dict_to_write):
...
-            self._logger.info("  from schema: None")
+            try:
+                if (
+                    not isinstance(_from_schema, (list, np.ndarray))
+                    and _from_simtel == _from_schema
+                ):
+                    self._logger.debug(f"Values for {data_type} match")
+                    continue
+            except ValueError:
+                pass
```
tobiaskleiner, line 207: No raise here or warning?

tobiaskleiner: Ok I see now that you add warnings later.

tobiaskleiner (commented):
Thanks, I have addded few comments.

simtools/applications/convert_all_model_parameters_from_simtel.py
```diff
@@ -138,15 +140,94 @@ def get_list_of_simtel_parameters(simtel_config_file, logger):
...
+            simtel_telescope_name=args_dict["simtel_telescope_name"],
+        )
+        _camera_pixel = simtel_config_reader.parameter_dict.get(args_dict["simtel_telescope_name"])
+    except FileNotFoundError:
+        logger.warning(
+            "Camera pixel schema file not found. Using default value for number of camera pixels."
+        )
+        _camera_pixel = None
+    logger.info(f"Number of camera pixels: {_camera_pixel}")
+    return _camera_pixel
```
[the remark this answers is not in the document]
GernotMaier: Yes, returning None is intended and good. It should indicate that the number of camera pixels are not read from file. I don't like this function at all, but it is necessary as simtel uses in some cases e.g., "all: 0." without explicitly writing an array-type value. We always write it explicitly and therefore need to know the number of pixels (at this point, this only helps for a small number of parameters).

I've improved the docstring.

simtools/simtel/simtel_config_reader.py
GernotMaier: Some parameters have only lower limits (e.g., QE_VARIATION needs to be >=0., but could take any value; but obviously not any large value is useful)

simtools/simtel/simtel_config_reader.py
GernotMaier: Good, I think this is resolved.

GernotMaier:
@tobiaskleiner - thanks for the review. I've improved the docstring to make it clearer that 'none' is ok as return value for the camera pixels.

tobiaskleiner (approved):
Thanks for the adjustment. All good then.

PR #843, opened 2024-03-28
Simplify application to add values to DB from json file; move to main application directory

GernotMaier:
This PR simplifies to application used to add values to the DB using json files:

- consider it a more frequently used application and move it to the main simtools/application folder (as e.g., add_file_to_db.py already is)
- upload explicitly the content of the json file to the DB and remove the temporarily introduced hardwired prod6 and default versions
- move `user_confirm()` function (asking for y/n on the command line) to simtools.general

code_cov keeps complaining about an untested line in general.py - but I don't find a way on how to make pylint and code_cov happy.


tobiaskleiner (approved):
Good for merging.

Issue #848, opened 2024-04-05
Issue with boost histograms


GernotMaier:
Installation fails with a boost histogram error, see e.g., https://github.com/gammasim/simtools/actions/runs/8570990185/job/23490220634#step:10:1455

```
387.4   Stored in directory: /tmp/pip-ephem-wheel-cache-5yhnjyy8/wheels/0e/67/1b/36aa2c0cdecd2c0db98a601d88d56ac0054c3af66db547bf71
387.5   Building wheel for psutil (pyproject.toml): started
409.1   Building wheel for psutil (pyproject.toml): finished with status 'done'
409.1   Created wheel for psutil: filename=psutil-5.9.8-cp311-abi3-linux_aarch64.whl size=288740 sha256=dc0636f72fcd568a364eca7dae4e5d00b168cb59838c3dcae582c5c68c037375
409.1   Stored in directory: /tmp/pip-ephem-wheel-cache-5yhnjyy8/wheels/a6/1e/65/fb0ad37886dca3f25a0aa8e50f4903c5bdbde4bb8a9b1e27de
409.1 Successfully built eventio psutil
409.1 Failed to build boost-histogram
409.1 ERROR: Could not build wheels for boost-histogram, which is required to install pyproject.toml-based projects
------
Dockerfile-dev:12
--------------------
  11 |     
  12 | >>> RUN wget --quiet https://raw.githubusercontent.com/gammasim/simtools/main/pyproject.toml && \
  13 | >>>     python3.11 -m venv env && \
  14 | >>>     source env/bin/activate && \
  15 | >>>     pip install -U pip && \
  16 | >>>     pip install toml-to-requirements && \
  17 | >>>     toml-to-req --toml-file pyproject.toml --include-optional && \
  18 | >>>     pip uninstall -y toml-to-requirements && \
  19 | >>>     pip install --no-cache-dir -r requirements.txt
  20 |     
--------------------
ERROR: failed to solve: process "/bin/bash -c wget --quiet https://raw.githubusercontent.com/gammasim/simtools/main/pyproject.toml &&     python3.11 -m venv env &&     source env/bin/activate &&     pip install -U pip &&     pip install toml-to-requirements &&     toml-to-req --toml-file pyproject.toml --include-optional &&     pip uninstall -y toml-to-requirements &&     pip install --no-cache-dir -r requirements.txt" did not complete successfully: exit code: 1
```

PR #849, opened 2024-04-08
fix boost-histogram

VictorBarbosaMartins:
This issue provides a temporary fix for #848 by disabling the generating of a docker image compatible with arm64 (e.g., Mac M1) architectures.

The issue has been reported to the boost-histogram maintainers already in January by others (see [here](https://github.com/scikit-hep/boost-histogram/discussions/910)).

This is issue fixes also an error by `toml-to-req` due to a change in API in the newer versions.

See issue #861 for the follow up work to do.


GernotMaier:
https://github.com/gammasim/simtools/actions/runs/8646092546 is an example that the workflow is running no successfully.

tobiaskleiner (approved):

PR #856, opened 2024-04-10
No need of test-data DB, as files are now in tests/resources

GernotMaier:
The test-data DB is not used anymore, as we have moved all files to tests/resources.

This PR does some minor maintenance:

- remove an integration test still using the test-data (the functionality of getting files from a DB is tested with two other tests, so no loss in testing)
- removal of `-test-data` from list of DBs.

Also dropped the test-data mongo DB.

### DIFF
--- a/simtools/applications/add_file_to_db.py
+++ b/simtools/applications/add_file_to_db.py
@@ -86,7 +86,6 @@ def main():
             _db_tmp.DB_TABULATED_DATA,
             _db_tmp.DB_DERIVED_VALUES,
             "sandbox",
-            "test-data",
         ],
         help=("The database to insert the files to."),
     )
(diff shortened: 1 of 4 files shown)


GernotMaier:
Attached the dump of the test-data DB (for the unlikely case we need to recover it at some point).

[mongodump-test-data.gz](https://github.com/gammasim/simtools/files/14931199/mongodump-test-data.gz)

tobiaskleiner (approved):
Good to merge.

GernotMaier:
@tobiaskleiner  - thank you!

@VictorBarbosaMartins - no need to look at at. I have asked Tobias to review it, as tests in other branches started to fail (as test-db has been removed).

Issue #858, opened 2024-04-10
Drop support for python 3.9.


GernotMaier:
Suggest to drop support for python 3.9.

Looks like ctapipe has dropped the support for 3.9 and we have failures of unit tests using ctapipe (see https://github.com/gammasim/simtools/actions/runs/8632455850/job/23663893773). 

I've tested this locally and everything works fine on 3.11.

GernotMaier:
OK - this is actually not related but to an issue with mac os packages, see https://github.com/gammasim/simtools/actions/runs/8633310074/job/23666068631#step:11:1328

GernotMaier:
Done with #860 (forgot to connect this issue with the PR).

PR #860, opened 2024-04-10
Remove python 3.9 and add 3.12 from CI workflow

GernotMaier:
see #858 

3.12 is definitely more important than 3.9.


VictorBarbosaMartins (approved):
Ok, agree!

PR #864, opened 2024-04-12
Use astropy units in jsonschema validation

GernotMaier:
Units for the schema and data file validation were hardwired until now in the schema files requiring explicit listing of all units.

This PR introduces a `format_checker` to allow testing that a given unit is a valid astropy unit (or `dimensionless`).

### DIFF
--- a/tests/unit_tests/data_model/test_metadata_model.py
+++ b/tests/unit_tests/data_model/test_metadata_model.py
@@ -50,7 +50,7 @@ def test_validate_schema(tmp_test_directory):
         metadata_model.validate_schema(invalid_data, schema_file)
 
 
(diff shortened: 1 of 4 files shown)


tests/unit_tests/data_model/test_metadata_model.py
```diff
@@ -49,6 +50,53 @@ def test_validate_schema(tmp_test_directory):
         metadata_model.validate_schema(invalid_data, schema_file)
 
 
+def test_validate_schema_astro_py_units(caplog):
```
tobiaskleiner, line 53: astropy

tobiaskleiner (approved):

PR #866, opened 2024-04-15
Units with spaces

GernotMaier:
#865 introduces an issue for combined units including spaces: e.g., `m s` is resolved to `['m','s']`. 

This is solved by applying this only when `value` is also of list type. Note that this is still not perfect, but works for the current model parameters.

Solves also the missing correct treatment of `int` type for the string-to-list step (where the `type` field is now used to identify float entries).

### DIFF
--- a/tests/unit_tests/data_model/test_validate_data.py
+++ b/tests/unit_tests/data_model/test_validate_data.py
@@ -566,6 +566,13 @@ def test_prepare_model_parameter():
     data_validator._prepare_model_parameter()
     assert all(item == "" for item in data_validator.data_dict["unit"])
 
+    data_validator.data_dict["value"] = "1000. 2000. 3000."
+    data_validator.data_dict["unit"] = "ct mV, m/s, N /m**2"
+    data_validator._prepare_model_parameter()
+    assert data_validator.data_dict["unit"][0] == "ct mV"
+    assert data_validator.data_dict["unit"][1] == "m/s"
+    assert data_validator.data_dict["unit"][2] == "N /m**2"
+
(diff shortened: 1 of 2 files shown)


tests/unit_tests/data_model/test_validate_data.py
```diff
@@ -554,16 +554,24 @@ def test_prepare_model_parameter():
...
     data_validator.data_dict["unit"] = ", , "
     data_validator._prepare_model_parameter()
     assert all(item == "" for item in data_validator.data_dict["unit"])
 
+    data_validator.data_dict["value"] = "1000 2000 3000"
+    data_validator.data_dict["unit"] = "ct"
+    data_validator.data_dict["type"] = "int64"
+    data_validator._prepare_model_parameter()
+    assert isinstance(data_validator.data_dict["value"][0], int)
+
```
tobiaskleiner, line 574: Maybe add a test for space separated units? I.e. 
`data_validator.data_dict["unit"] = "c t"` 
`assert data_validator.data_dict["unit"] == "ct"`

GernotMaier: Good point - added a test for this.

But I don't think your assumption is right: in astropy units `c t` would not resolve to `ct` but to `c*t`. So we should to the same thing. But note that this function is just separating the unit string; no other changes are applied.

tobiaskleiner: Okey true, thanks for adding the further tests. Looks all good now.

tobiaskleiner (approved):

GernotMaier:
Thank you @tobiaskleiner

PR #869, opened 2024-04-16
Reduce number of calls to DB pass plus some maintenance in db_handler

GernotMaier:
Reduce the number of database calls.

What we do is that we store any parameters retrieved in a dictionary as a class variable. I should say that I am not firm enough in python to know if using class variables is a good solution for problem. Maybe one of you knows better. I have been looking also into the cache-functionality from functools, but this is not working the way we have integrated the  model parameter classes.

Tests are passing. Locally, this reduces the type required for unit tests by almost 40%.

From the discussion below, we want to make sure that we notice when parameters are locally changed:

- in general, the parameter dicts in db_handler should not be changed in any other way than reading from DB

Additional maintenance:

- removed `db_handler.add_parameter`. This is nowhere used, we generally use `add_new_parameter`
- removed `db_handeler.update_parameter`. This is nowhere used, we use `update_parameter_field`.


VictorBarbosaMartins:
> Potential issues raised by Tobias:
> 
> * Memory issues when the parameters include lots of data.
> 
> Don't think this will be an issue. We have roughly 200 parameters per telescope, most of them are dicts with single values (the typical DB dict for a parameter), a few (<10%) are tables (e.g., reflectivity). A single telescope right now is roughly 500kb, so 100 Telescopes x 500 kB is still not a lot to keep in memory (one could imaging that we will have at some point reflectivity tables per mirror panel or quantum efficiencies per channel - but this is many years away).
> 
> * Maybe potential outdated values in the cache/making sure the cache is read only.
> 
> This is more important. Does one of you know how to protect a variable from being written over in python?
> 
> db_handler by itself is already a bit protected, as it it called directly only by `model_parameters`. There are three functions in db_handler we need to look at:
> 
> * update_parameter
> * update_parameter_field
> * add_parameter
> * add_new parameter (hmm, this is an interesting naming...what is the difference to add_parameter?)
> 
> Let's discuss this in person to see if it is worth continuing.

Hi @GernotMaier,

I will look into this solution. Regarding your question, I think a good way to protect the variable is to use it as a @property and set the variable only in the first use with something like:
```
def __init__(self):
    self._variable = None
# code
# code ....

@property
def variable(self):
    if self._variable is None:
        # calculate or retrieve variable
    return self._variable
```
But I am not sure about the amount of variables you are talking about. Defining trillions of properties doesnt sound like a good idea either.

simtools/db/db_handler.py
```diff
@@ -144,14 +155,15 @@ def get_model_parameters(
...
                 parameters=_pars,
                 site=_site,
                 parameter_collection="telescopes",
                 telescope_name=_telescope_model_name,
                 model_version=_model_version,
                 db_simulation_model_url=self.mongo_db_config.get("db_simulation_model_url", None),
             )
+        DatabaseHandler.model_parameters[_array_elements_key] = _pars
 
         return _pars
```
VictorBarbosaMartins, line 168: for consistency I would return `DatabaseHandler.model_parameters[_array_elements_key]` instead, although I think it really does not make a big difference.

simtools/db/db_handler.py
```diff
@@ -144,14 +155,15 @@ def get_model_parameters(
 
         # update using simulation model repository
         if self.mongo_db_config.get("db_simulation_model_url", None) is not None:
-            return db_from_repo_handler.update_model_parameters_from_repo(
+            _pars = db_from_repo_handler.update_model_parameters_from_repo(
```
VictorBarbosaMartins, line 158: This line is not tested (codecov). I would suggest we make sure everything is tested in such a core module

simtools/db/db_handler.py
```diff
@@ -382,14 +403,15 @@ def get_site_parameters(
         )
         # update simulation model using repository
         if self.mongo_db_config.get("db_simulation_model_url", None) is not None:
-            return db_from_repo_handler.update_model_parameters_from_repo(
+            _pars = db_from_repo_handler.update_model_parameters_from_repo(
```
VictorBarbosaMartins, line 406: codecov: line not tested

simtools/db/db_handler.py
VictorBarbosaMartins: similar to the other functions below

VictorBarbosaMartins (commented):
Hi @GernotMaier. I liked this solution and I think it will work as we expect. I am not very used to the module, so I have to ask the question: in case we want to update a parameter after getting it from the DB, will we still be able to change the variable to the new value before updating it back to the DB? And what about the collections ? In `update_parameter` and other functions we still always get the collection each time directly from the DB.

simtools/db/db_handler.py
GernotMaier: Which is actually very weird: all what has been changed is 'return' to an assignment, and suddenly the line is not covered anymore?

GernotMaier: I've looked at it and decided not to add a dedicated test for two reasons:

- reading the telescope parameters from repo is slow (idea is to use this for local development)
- exactly the same mechanism is implemented in get_site_parameters. Here a test has been added (there are a few site parameters vs 100s of telescope parameters).

simtools/db/db_handler.py
GernotMaier: Fixed.

GernotMaier:
@VictorBarbosaMartins - thanks again for the quick feedback. I've updated the code now (plus removed some unnecessary code) and added tests. Could you have another look?

simtools/db/db_handler.py
VictorBarbosaMartins: ok!

VictorBarbosaMartins (approved):
Thanks for the updates. Weird that we had two functions not being used like that. But the tests are passing and they indeed looked very similar to the other ones you pointed out, so I am approving.

GernotMaier:
@VictorBarbosaMartins - thank you!

PR #870, opened 2024-04-17
allow to add to calibration_devices

tobiaskleiner:
This PR allows to add json schemas to the collection of calibration_devices. I have added a TODO to modify the code later to be more general and to remove the naming of telescope and adjust the validation methods.

### DIFF
--- a/simtools/db/db_handler.py
+++ b/simtools/db/db_handler.py
@@ -1039,8 +1039,14 @@ class DatabaseHandler:
             db_entry["instrument"] = names.validate_telescope_name(telescope)
         elif "sites" in collection_name:
             db_entry["instrument"] = names.validate_site_name(site)
+        # TODO: generalize later
+        elif "calibration_devices" in collection_name:
+            db_entry["calibration_devices"] = names.validate_telescope_name(site)
         else:
-            raise ValueError("Can only add new parameters to the sites or telescopes collections")
+            raise ValueError(
+                "Can only add new parameters to the sites, \
+                telescopes or calibration_devices collections"
+            )
 
         db_entry["version"] = version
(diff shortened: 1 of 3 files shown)


simtools/db/db_handler.py
```diff
@@ -1039,8 +1039,14 @@ def add_new_parameter(
             db_entry["instrument"] = names.validate_telescope_name(telescope)
         elif "sites" in collection_name:
             db_entry["instrument"] = names.validate_site_name(site)
+        # TODO: generalize later
```
GernotMaier, line 1042: Suggest to remove this comment and open an issue regarding the generalization.

simtools/db/db_handler.py
```diff
@@ -1039,8 +1039,14 @@ def add_new_parameter(
...
+        elif "calibration_devices" in collection_name:
```
GernotMaier, line 1043: Do you want to add one single test for this in test_adding_new_parameter_db in test_db_handler? A single line is probably enough.

GernotMaier (approved):
Two minor comments - but fine as it is here.

simtools/db/db_handler.py
tobiaskleiner: Removed, opening an issue.

simtools/db/db_handler.py
tobiaskleiner: Added.

GernotMaier (approved):
Approved (again - all good!)

simtools/db/db_handler.py
tobiaskleiner: Issue is here: #871

PR #878, opened 2024-04-19
Allow `model_version` command line configuration of integration tests

GernotMaier:
This PR adds an option to allow to run the integration tests for different model versions and adds a `--model_version` to the integration tests.

Example:

```
pytest --no-cov --model_version="prod5" \
    "tests/integration_tests/test_applications_from_config.py::test_applications_from_config[simtools-print-array-elements_utm_to_ground_meta_in_yml]"
```

Adds also a matrix job to the integration test CI to run over prod6 (default, labeled as 2024-02-01) and prod5.

Note that the additional tests (e.g., file content comparison) are disabled for the non-default version (which is given in conftest.py through the `model_version` fixture. Reason for that is that we otherwise would have to keep track and add comparison tables in test/resources for all model versions.

### DIFF
--- a/tests/integration_tests/test_applications_from_config.py
+++ b/tests/integration_tests/test_applications_from_config.py
@@ -270,7 +270,7 @@ def validate_application_output(config):
     return 0
 
 
-def prepare_configuration(config, output_path):
+def prepare_configuration(config, output_path, model_version=None):
     """
     Prepare configuration. This means either to write a temporary config file
     or to return a single string for single boolean options (e.g., --version).
@@ -282,6 +282,8 @@ def prepare_configuration(config, output_path):
         Dictionary with the configuration for the application test.
     output_path: str
         Output path.
+    model_version: str
+        Model versions (default: use those given in config files)
 
     Returns
     -------
@@ -294,6 +296,8 @@ def prepare_configuration(config, output_path):
         return None, "--" + list(config.keys())[0].lower()
 
     tmp_config_file = output_path / "tmp_config.yml"
+    if model_version is not None and "MODEL_VERSION" in config:
+        config.update({"MODEL_VERSION": model_version})
     if "OUTPUT_PATH" in config:
         config.update({"OUTPUT_PATH": str(Path(output_path).joinpath(config["OUTPUT_PATH"]))})
         config.update({"USE_PLAIN_OUTPUT_PATH": True})
@@ -312,7 +316,7 @@ def prepare_configuration(config, output_path):
     get_list_of_test_configurations(),
     ids=get_list_of_test_configurations(get_test_names=True),
 )
-def test_applications_from_config(tmp_test_directory, config, monkeypatch):
+def test_applications_from_config(tmp_test_directory, config, monkeypatch, request):
     """
     Test all applications from config files found in the config directory.
 
@@ -340,9 +344,12 @@ def test_applications_from_config(tmp_test_directory, config, monkeypatch):
 
     tmp_output_path.mkdir(parents=True, exist_ok=True)
     logger.info(f"Temporary output path: {tmp_output_path}")
+    logger.info(f"Model version: {request.config.getoption('--model_version')}")
     if "CONFIGURATION" in config:
         config_file, config_string = prepare_configuration(
-            config["CONFIGURATION"], output_path=tmp_output_path
+            config["CONFIGURATION"],
+            output_path=tmp_output_path,
+            model_version=request.config.getoption("--model_version"),
         )
     else:
         config_file = None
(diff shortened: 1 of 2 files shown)


.github/workflows/CI-integrationtests.yml
```diff
@@ -22,6 +22,9 @@ jobs:
     container:
       image: ghcr.io/gammasim/simtools-dev:latest
       options: --user 0
+    strategy:
+      matrix:
+        type: ['2024-02-01', 'prod5']
```
VictorBarbosaMartins, line 27: I was expecting this to run over prod6 as well. Is there a reason why it is not?

simtools/simtel/simtel_runner_camera_efficiency.py
```diff
@@ -82,16 +82,7 @@ def _make_run_command(self, **kwargs):  # pylint: disable=unused-argument
...
         # Processing focal length
-        focal_length = 0.0
-        try:
-            focal_length = self._telescope_model.get_parameter_value_with_unit(
-                "effective_focal_length"
-            )[0]
-        except IndexError:
-            pass
-        if focal_length == 0.0:
-            self._logger.warning("Using focal_length because effective_focal_length is 0")
```
VictorBarbosaMartins, line 93: the warning did not make it into the other function. Perhaps a good idea to have a warning there too if we keep this implementation.

tests/integration_tests/config/convert_model_parameter_from_simtel_num_gains.yml
```diff
@@ -5,7 +5,7 @@ CTA_SIMPIPE:
     SIMTEL_CFG_FILE: ./tests/resources/simtel_config_test_la_palma.cfg
     SIMTEL_TELESCOPE_NAME: CT1
     TELESCOPE: LSTN-01
-    MODEL_VERSION: "2024-03-06"
+    MODEL_VERSION: "2024-02-01"
```
VictorBarbosaMartins, line 8: I dont keep track of these dates. What is actually changed here?

tests/integration_tests/test_applications_from_config.py
```diff
@@ -294,6 +296,8 @@ def prepare_configuration(config, output_path):
         return None, "--" + list(config.keys())[0].lower()
 
     tmp_config_file = output_path / "tmp_config.yml"
+    if model_version is not None and "MODEL_VERSION" in config:
+        config.update({"MODEL_VERSION": model_version})
```
VictorBarbosaMartins, line 300: I am not entirely sure these lines are doing what we expect. We prioritize here the parameter passed over the one in the config, but perhaps we gave the version in the config file because that is the one we expect it to work with, no?

VictorBarbosaMartins (commented):
Hi @GernotMaier. I leave some comments here, but it was a bit hard to grasp some of the changes, as I am not very used to these configs. Let me know if my comments make sense. I will continue with the remaining PRs tomorrow

.github/workflows/CI-integrationtests.yml
GernotMaier: Oh, 2024-02-01 is a prod6 prototype implementation. The model needs some more testing and some update, so better not to call it prod6.

simtools/simtel/simtel_runner_camera_efficiency.py
GernotMaier: OK, done.

tests/integration_tests/test_applications_from_config.py
GernotMaier: This replaces the `model_version` for all tools which have a `model_version` set via the config file. 

This is exactly what we want to do and allow us to run the integration test using a command line parameter (e.g., `--model_version=prod5` to run all integration tests with prod6.

.github/workflows/CI-integrationtests.yml
VictorBarbosaMartins: ok

tests/integration_tests/test_applications_from_config.py
VictorBarbosaMartins: Ok, so that means all integration tests should work with both versions? For instance, we don't have integration tests that should work only with prod5 ?

.github/workflows/CI-integrationtests.yml
VictorBarbosaMartins: I tested now running the test for model_version="prod6" `pytest --no-cov --model_version="prod6" "tests/integration_tests/test_applications_from_config.py::test_applications_from_config[simtools-print-array-elements_utm_to_ground_meta_in_yml]"` and it works (the test also passes). I guess this is because we already have the model parameters in the DB?

VictorBarbosaMartins (approved):
Thanks @GernotMaier. I have only two minor questions, so I am already approving

tests/integration_tests/test_applications_from_config.py
GernotMaier: All tests should work with both versions (and they do). I do think we should aim that this is also true in the future.

The only difference we have to make if we check for output files. At this point, we only check for the version which is labeled in the configuration files (file naming is a bit complicated and adding this is probably not necessary)

.github/workflows/CI-integrationtests.yml
GernotMaier: Exactly. There is no difference between 2024-02-01 and prod6 (this is resolved in `names.py`)

GernotMaier:
Thank you again.

PR #884, opened 2024-04-22
Add unix find to development docker

GernotMaier:
This PR adds the standard unix 'find' utility to the development container.

This is needed for debugging and finding e.g., a certain generated configuration file in the simtools output directories.


VictorBarbosaMartins (approved):
Although `find` could be instead used locally, I think this is a good way to help us testing implementations within the container without taking too many storage in the container

GernotMaier:
You can use only the local find on mounted directories. Pytest is e.g., writing to tmp which is only internal to the container.

GernotMaier:
Thanks for the review!

PR #875, opened 2024-04-17
Simulate prod test

VictorBarbosaMartins:
**Done:**
Here I am including some missing tests to check whether the files for either corsika or simtel_array output were generated by the production tool. We are constantly breaking this tool because the tests are not checking whether the output files exist. The integration tests for simulate_prod.py will fail at the moment.

I removed the test flag for `sim_showers_for_trigger_rates` because it was preventing us to simulate and de facto test the tool. To minimize computational time of the simulation, I reduced the number of events to 2.

**To do:**
The best would be to make the corrections to the tool before merging this and have the integration tests passing.

PR #887, opened 2024-04-24
remove if statement -> return testing all the output files in integra…

VictorBarbosaMartins:
After discussions in #875, I have realized that the tests for the output files of integrations tests were turned off. I turn them on again, but I have to be honest that I do not know about the impact on the choice for the model version. I will see the results of the integration tests here and evaluate the situation.

Usually the name of output files from integration tests are not related to the model parameter version (or at least it shouldnt). If that is the case, I hope that only applications that are really not working will fail the tests


VictorBarbosaMartins:
> As discussed in person: this if statement is not correct, but should not be removed entirely. We should do the tests for the output files only for the version outline in the config file.

Ok, I am waiting for your corrections to proceed with my tests. Thanks!

GernotMaier:
I've adjust the code now that it runs only for those model_versions given in the config file.

Unfortunately a bit more complicated as expected, but I've tested all options (without command line, with command line model_version as/as not in config file) and it works as expected.

VictorBarbosaMartins:
> I've adjust the code now that it runs only for those model_versions given in the config file.
> 
> Unfortunately a bit more complicated as expected, but I've tested all options (without command line, with command line model_version as/as not in config file) and it works as expected.

Ok, thanks! it looks like it works now. So, I expect only problematic tools to fail the integration tests now.

VictorBarbosaMartins:
> > I've adjust the code now that it runs only for those model_versions given in the config file.
> > Unfortunately a bit more complicated as expected, but I've tested all options (without command line, with command line model_version as/as not in config file) and it works as expected.
> 
> Ok, thanks! it looks like it works now. So, I expect only problematic tools to fail the integration tests now.

Ups, it doesn't seem to be the case

PR #889, opened 2024-04-24
Unify array elements in names.py and json schema

GernotMaier:
This PR removes the duplicated listing of all allowed array elements (e.g., names of telescopes or calibration devices). Introduces a yaml file with the list of all allowed array elements and necessary properties.

This list is:

- ready in names.py using the new `load_array_elements` function. The dictionary read in is cached, so reading of the yaml file is happening only once.
- this dict is now returned instead of `names.array_element_names`
- this is added to the enum field of allowed array elements in the json schema field `InstrumentTypeElement` in `simtools/schemas/model_parameter_and_data_schema.metaschema.yml`

### DIFF
--- a/simtools/utils/names.py
+++ b/simtools/utils/names.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from functools import lru_cache
+from pathlib import Path
 
 import yaml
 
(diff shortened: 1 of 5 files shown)


simtools/utils/names.py
```diff
@@ -18,39 +22,33 @@
...
+    -------
+    dict
+        Array elements.
+    """
+    base_path = Path(__file__).parent
+    with open(base_path / "../schemas/array_elements.yml", "r", encoding="utf-8") as file:
+        return yaml.safe_load(file)["data"]
+
+
+def array_elements():
```
tobiaskleiner, line 41: What is the reason for this function and not calling directly `load_array_elements()`?

tobiaskleiner (commented):
All good, only a minor comment.

simtools/utils/names.py
GernotMaier: Good point, this was a leftover from an earlier version.

Fixed this.

PR #893, opened 2024-04-25
Name of simulation model database as configuration entry

GernotMaier:
This PR adds functionality to configure the name of the simulation model using the environmental variables or the command line. This is especially useful now in the period of substantial updates and modifications to the model databases.

Before this PR: name of the simulation model DB is hardwired as class variable in the db_handler class.

This PR removes also the database DB_TABULATED_DATA, which is not used anymore.

Note that a surprisingly large number of changes are necessary, as this class variable has been accessed quite frequently.

There are two database names still hardwired in the db_handler class:

- db.DB_CTA_SIMULATION_MODEL_DESCRIPTIONS - can be removed as soon as we have integrated the model schema files into simtools
- db.DB_DERIVED_VALUES will be replaced in future by a to-be-discussed functionality (but I am pretty sure that we won't keep it this way)

### DIFF
--- a/simtools/applications/db_development_tools/add_unit_to_parameter_in_db.py
+++ b/simtools/applications/db_development_tools/add_unit_to_parameter_in_db.py
@@ -35,14 +35,13 @@ def main():
     for site in ["North", "South"]:
         for par_now, unit_now in zip(pars_to_update, units):
             all_versions = db.get_all_versions(
-                db_name=db.DB_CTA_SIMULATION_MODEL,
                 site=site,
                 parameter=par_now,
                 collection_name="sites",
             )
             for version_now in all_versions:
                 db.update_parameter_field(
-                    db_name=db.DB_CTA_SIMULATION_MODEL,
+                    db_name=None,
                     site=site,
                     version=version_now,
                     parameter=par_now,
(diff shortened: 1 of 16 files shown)


simtools/applications/db_development_tools/add_unit_to_parameter_in_db.py
```diff
@@ -35,14 +35,13 @@ def main():
...
             all_versions = db.get_all_versions(
-                db_name=db.DB_CTA_SIMULATION_MODEL,
                 site=site,
                 parameter=par_now,
                 collection_name="sites",
             )
             for version_now in all_versions:
                 db.update_parameter_field(
-                    db_name=db.DB_CTA_SIMULATION_MODEL,
+                    db_name=None,
```
VictorBarbosaMartins, line 44: I dont understand this None here. We only allow this application to work with the default staging DB?

VictorBarbosaMartins (approved):
Generally, looks good to me.

simtools/applications/db_development_tools/add_unit_to_parameter_in_db.py
GernotMaier: We actually might never need this application again (as it was used to add units to each parameters in the DB) - we mostly keep it as a reference of how we add fields.

But in general, we need to be flexible with the database names, especially now during the development. To get the model parameter items correct, tests with different databases are necessary and flexibility is therefore important.

Issue #852, opened 2024-04-08
Suggestion to move all model_parameter schema files into simtools


GernotMaier:
We currently store the knowledge about the model parameters in schema files in the [model_parameters gitlab repository](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/tree/main/schema?ref_type=heads).

I suggest to move them into simtools and make them part of the package. This would allow to consistently use the schema files for the name transformation between simtools and sim_telarray/CORSIKA plus remove the (not so nice) complicated data dicts in [names.py](https://github.com/gammasim/simtools/blob/df26c4aad57c28ef963db1f081759524f1da106e/simtools/utils/names.py#L95). The information in these dicts is duplicated information which is already in the schema files.

GernotMaier:
#894 addressed this issue.

PR #894, opened 2024-04-25
Move model parameter schemas from gitlab into simtools

GernotMaier:
This PR adds the model parameter schemas from the [model parameter gitlab](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters) into simtools (see issue #852).

This will allow to get rid of some of the not-so-nice entries in names.py and also to consistently treat the model parameters with all information in simtools.

There is **no** review (no!) of the contents of the files necessary. As a reviewer, you can disagree with moving all files into simtools, which means we have to discuss again the approach. None of the files has been changed and they have been reviewed before in the gitlab instance (any changes, CI updates, updates to names.py will come in follow up PRs).

### DIFF
--- a/simtools/schemas/model_parameters/adjust_gain.schema.yml
+++ b/simtools/schemas/model_parameters/adjust_gain.schema.yml
@@ -1,0 +1,37 @@
+%YAML 1.2
+---
+title: Schema for adjust_gain model parameter
+version: 0.1.0
+meta_schema: simpipe-schema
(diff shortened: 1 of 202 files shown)


VictorBarbosaMartins:
I agree with moving the parameter schemas to simtools but I am unsure about keeping the gitlab repo as well. Do we need it or is it going to be removed and by that we lose all the updates there?

GernotMaier:
The gitlab repo will stay, but the schema directory will go (obviously, we always avoid duplication). We don't loose history if we delete the schema files there.

VictorBarbosaMartins (approved):
Ok, so I am approving. But perhaps this is something @orelgueta would like to have a look first?

PR #895, opened 2024-04-25
Integrate and use model parameter schema to determine sim_telarray and CORSIKA parameters

GernotMaier:
This PR adds functionality to use the model parameter schema (introduced in PR #894) and removes hardwired lists of sim_telarray and CORSIKA parameters in names.py.

We require that any model parameter used has a corresponding schema. 

Having the strict control now with the schema parameters, this PR 'unveiled' some issues:

- several database issue with missing or wrong parameter entries; moved therefore the 'clean' simulation model database `sandbox_20240422` (I know, this is not a good name)
- duplication possibility to define model-version in simulate_prod (use now always the command line argument; remove the one from the array model configuration file
- duplication of resolving telescopes to design types (e.g., if SST-25 is not defined, it should be SST-design). This is done now consistently in model_parameter and therefore removed from the array model configuration (and the treatment of this in the array_model module)
- removed default value for the name of the simulation model database. We don't give default values for the other database parameter, having this one differently is not useful
- removed 'sim_telarray' in the simulation software section for a couple of parameters. These are all parameters which are not sim-telarray parameters.
- changed the name for simtools in the schema as simulation_software entry from `gamma-sim-tools` to `simtools` (we don't use gammasimtools anymore)

Because of this, adding new parameters without schema is not wanted and we remove the functionality in the model_parameter module.


simtools/simtel/simtel_config_writer.py
```diff
@@ -71,35 +71,38 @@ def write_telescope_config_file(self, config_file_path, parameters, config_param
...
                         value = 1 if value else 0
                     elif isinstance(value, (list, np.ndarray)):
                         value = gen.convert_list_to_string(value)
                     file.write(f"{_simtel_name} = {value}\n")
+            _config_meta = self._add_simtel_metadata("telescope")
+            for _simtel_name, value in _config_meta.items():
+                file.write(f"{_simtel_name} = {value}\n")
 
-    def _add_simtel_metadata(self, parameters, config_type):
+    def _add_simtel_metadata(self, config_type):
```
tobiaskleiner, line 89: Maybe this should be renamed to `get_simtel_metadata`. Similar to `get_simtel_parameters()` in the following.

tobiaskleiner (approved):
Thanks Gernot! One minor cosmetic comment, nothing that would prevent merging.

simtools/simtel/simtel_config_writer.py
GernotMaier: Good point, changed this to _get_simtel_metadata().

GernotMaier:
Thanks for the review. I made the minor update. I am waiting for the tests now and will then merge.

PR #897, opened 2024-04-25
CI schema validation

GernotMaier:
Move the model parameter schema validation from the gitlab CI to here (as these schemas are now part of simtools, see PR #894 ).

The CI is running as it should and fails at this point (good - shows that it works!): https://github.com/gammasim/simtools/actions/runs/8836327205

I will fix those schemas after this PR is merged (as only then a manual workflow run is possible).


tobiaskleiner (approved):
All good.

PR #898, opened 2024-04-26
Housekeeping: pylint and location of schema files

GernotMaier:
This is a short PR with changes to fix issues with the CI pipelines:

- solved in issue of pylint with astropy units by adding ignored-classes=["astropy.units"] to pyproject.toml
- changed urls pointing to schemas on gitlab to github (as we moved them with pull request #894); this required changes in several places

### DIFF
--- a/simtools/constants.py
+++ b/simtools/constants.py
@@ -5,6 +5,5 @@ METADATA_JSON_SCHEMA = "schemas/metadata.metaschema.yml"
 
 # URL to the schema repository
 SCHEMA_URL = (
-    "https://gitlab.cta-observatory.org/cta-science/"
-    "simulations/simulation-model/model_parameters/-/raw/main/schema/"
+    "https://raw.githubusercontent.com/gammasim/simtools/main/simtools/schemas/model_parameters/"
 )
--- a/tests/unit_tests/data_model/test_data_reader.py
+++ b/tests/unit_tests/data_model/test_data_reader.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+from pathlib import Path
 
 import astropy.units as u
 import jsonschema
@@ -97,11 +98,8 @@ def test_read_value_from_file_and_validate(caplog, tmp_test_directory):
         assert "Successful validation of yaml/json file" in caplog.text
 
     # schema explicitly given
-    schema_file = (
-        "https://gitlab.cta-observatory.org/cta-science/simulations/"
-        "simulation-model/model_parameters/-/raw/main/"
-        "schema/reference_point_altitude.schema.yml"
-    )
+    schema_dir = Path(__file__).parent / "../../../simtools/schemas/model_parameters/"
+    schema_file = str(schema_dir) + "/reference_point_altitude.schema.yml"
     with caplog.at_level(logging.DEBUG):
         data_reader.read_value_from_file(
             "tests/resources/reference_point_altitude.json",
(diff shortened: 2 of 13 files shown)


tests/unit_tests/data_model/test_data_reader.py
```diff
@@ -97,11 +98,8 @@ def test_read_value_from_file_and_validate(caplog, tmp_test_directory):
         assert "Successful validation of yaml/json file" in caplog.text
 
     # schema explicitly given
-    schema_file = (
-        "https://gitlab.cta-observatory.org/cta-science/simulations/"
-        "simulation-model/model_parameters/-/raw/main/"
-        "schema/reference_point_altitude.schema.yml"
-    )
+    schema_dir = Path(__file__).parent / "../../../simtools/schemas/model_parameters/"
```
VictorBarbosaMartins, line 101: I think the "../../../" is not very elegant and can easily be disrupted by any change on the path. I suggest we create a variable for schema_dir in the configurator, as we have for simtel_path. This would also solve the problem I pointed out above

simtools/constants.py
```diff
@@ -5,6 +5,5 @@
 
 # URL to the schema repository
 SCHEMA_URL = (
-    "https://gitlab.cta-observatory.org/cta-science/"
-    "simulations/simulation-model/model_parameters/-/raw/main/schema/"
+    "https://raw.githubusercontent.com/gammasim/simtools/main/simtools/schemas/model_parameters/"
```
VictorBarbosaMartins, line 8: Do we still want to give it as a URL even if it is part of the project now? This way we are getting the parameter schema from the released version only. I think it would be better if we point it to the directory of the current version of simtools (whichever version one is working on). This would allow us to test implementations of the schema files on the spot. Same for all the links.

VictorBarbosaMartins (commented):
Thanks for this update. I have only two comments and a proposal to replace the URL by a simtools directory.

tests/unit_tests/data_model/test_data_reader.py
GernotMaier: This should be relative paths - remember that we distribute simtools also as pip package and those files need to be findable. I don't think this is so bad (of course it breaks when we move the model_parameters; we will notice immediately).

There is another solution for package data suggested in https://setuptools.pypa.io/en/latest/userguide/datafiles.html . I still not find time to understand exactly the details (beyond this PR).

simtools/constants.py
GernotMaier: We need to keep url, as this is what is listed in the metadata of the model parameters (model parameters are not part of simtools).

What is correct is that we do not need this URL to read the schema files, as they are part of the package. This is related to the previous point and I've opened an issue on this: #899

GernotMaier:
@VictorBarbosaMartins - thanks for the review. See above comment. Let me know if it would be absolute essential to address this in this PR (note that this PR fixes the issues that all tests are failing right now).

VictorBarbosaMartins (approved):

VictorBarbosaMartins:
> @VictorBarbosaMartins - the solution of the env parameter would require an explicit setting of the value outside of simtools (in the shell). We definitely don't want that.
> 
> This is different for the path to sim_telarray, as this is anyway a software separate from simtools.

simtools does not work at the moment without a DB and the DB is already configured in the .env file (outside simtools). Adding an extra line to .env is not a problem in my view but I also understand if you prefer to have it self consistent in simtools.

GernotMaier:
No,  schemas are part of the package and this is completely internal to the package. 

You really don't want to add this somewhere else.

PR #905, opened 2024-04-29
Set schema paths using importlib

GernotMaier:
Follow the guidelines for python packages to set package internal directories (see [here]( https://setuptools.pypa.io/en/latest/userguide/datafiles.html)) with `importlib.resources.files`.

Remove `SCHEMA_URL` as all schemas are now part of the simtools package (we have set `include-package-data = true` in pyproject.toml, which means that all non-python files in the simtools directory are included as package data).

### DIFF
--- a/simtools/data_model/metadata_collector.py
+++ b/simtools/data_model/metadata_collector.py
@@ -9,6 +9,7 @@ implementation of the metadata model.
 import datetime
 import getpass
 import logging
+from importlib.resources import files
 from pathlib import Path
 
 from astropy.table import Table
@@ -115,7 +116,7 @@ class MetadataCollector:
         # from data model name
         if self.data_model_name:
             self._logger.debug(f"Schema file from data model name: {self.data_model_name}")
-            return f"{simtools.constants.SCHEMA_URL}{self.data_model_name}.schema.yml"
+            return f"{files('simtools')}/schemas/model_parameters/{self.data_model_name}.schema.yml"
 
         # from input metadata
         try:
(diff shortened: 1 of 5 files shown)


simtools/data_model/metadata_collector.py
```diff
@@ -115,7 +116,7 @@ def get_data_model_schema_file_name(self):
         # from data model name
         if self.data_model_name:
             self._logger.debug(f"Schema file from data model name: {self.data_model_name}")
-            return f"{simtools.constants.SCHEMA_URL}{self.data_model_name}.schema.yml"
+            return f"{files('simtools')}/schemas/model_parameters/{self.data_model_name}.schema.yml"
```
VictorBarbosaMartins, line 119: I have a question in regard to this `files`. Does it really access the directory in simtools or it accesses the directory in the released (main) package online? I didn't look deeply at the guidelines but if the last, then I would a bit concerned, because in that case, it wouldnt be straightforward to test updating the schemas in a development branch.

VictorBarbosaMartins (approved):
One comment, otherwise good

simtools/data_model/metadata_collector.py
GernotMaier: setuptools is used when running `pip install` - the package directory known by pip to access code is the one which is access by importlib.resources.files.

The system doesn't know anything about online resources, so definitely uses the local ones.

GernotMaier:
@VictorBarbosaMartins - thanks for the review!

PR #907, opened 2024-05-02
Separate plot_camera from camera module

GernotMaier:
This PR separates the camera-related plotting routines from simtools.model.camera into a new module called simtools.visualization.plot_camera.

No functional changes has been applied (also no tests removed or added; this is necessary at some point), so there is no line-by-line review required.

Main motivation is that the camera routine is part of the telescope model. The import of plotting-related modules (..matplotlib) is slow and has an impact on the the usability of tools. This PR removes all plotting and improves therefore the rather slow startup time of some simtools (e.g., print_array_elements).

I do think separate plotting should be a general design feature.

### DIFF
--- a/simtools/applications/validate_camera_fov.py
+++ b/simtools/applications/validate_camera_fov.py
@@ -58,6 +58,7 @@ import simtools.utils.general as gen
 from simtools.configuration import configurator
 from simtools.io_operations import io_handler
 from simtools.model.telescope_model import TelescopeModel
+from simtools.visualization import plot_camera
 
 
 def main():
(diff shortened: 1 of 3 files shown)


GernotMaier:
@VictorBarbosaMartins - I am opening this one now without adding missing tests. The PR only re-organized code by moving the plotting routines into a separate module. I don't have time now to read myself into the module to be able to add the missing tests. We otherwise keep the PR open (but it won't get easier with time to review...).

simtools/applications/validate_camera_fov.py
```diff
@@ -58,6 +58,7 @@
 from simtools.configuration import configurator
 from simtools.io_operations import io_handler
 from simtools.model.telescope_model import TelescopeModel
+from simtools.visualization import plot_camera
```
VictorBarbosaMartins, line 61: I would rather simply call it **camera**. since it is inside visualization, the **plot** is implicit. Their names used to be different when they are in the same module but as they are separate now, it is not necessary to change the name (just an opinion)

VictorBarbosaMartins (approved):
Only one comment on the name. I did not check line by line as you suggested to leave the coverage to follow-up PRs. Thanks for these changes!

simtools/applications/validate_camera_fov.py
GernotMaier: Hesitating - having modules with the same name lead to confusion, even with they are in different subfolders.

Also thing that plot_camera describes well what the module is doing, while visualization.camera could also mean something related to e.g, a web page.

PR #909, opened 2024-05-02
Improve import efficiency for UnitsErrors

GernotMaier:
The more directory import of `from astropy.units import UnitsError`  instead of `from astropy.coordinates.errors import UnitsError` improves the import efficiency.


VictorBarbosaMartins (approved):
all good

PR #921, opened 2024-05-16
Unit tests in random order to make sure that unit tests are independent of each other

GernotMaier:
We had a couple of times the issue that unit tests depend on the order of execution. This is obviously wrong, as unit tests by definition should be testing 'units'.

This PR adds a CI step for the nightly test of main and runs the unit tests 5x in random order.


GernotMaier:
Interestingly, it actually fails before correcting the tests in test_general.py, see https://github.com/gammasim/simtools/actions/runs/9107587029/job/25036730200

I removed this test, as this one tests for an executable without giving the full path. This is not what the function to be tested is supposed to handle.

GernotMaier:
One more failure fixed: `test_model_parameter` was modifying a parameter of a TelescopeModel given by a fixture. This was used by other tests, which failed depending on the test sequence. Fixed now (and this is exactly why this PR is important).

GernotMaier:
There is one last test failing: `tests/unit_tests/model/test_array_model.py::test_get_single_telescope_info_from_array_config`

This has been fixed in PR #910 and I will wait for this PR to be merged. 

(leave current PR in draft in the mean while)

tests/unit_tests/model/test_model_parameter.py
```diff
@@ -31,7 +32,7 @@ def test_get_parameter_dict(telescope_model_lst):
 
 
 def test_get_parameter_value(telescope_model_lst):
-    tel_model = telescope_model_lst
+    tel_model = copy.deepcopy(telescope_model_lst)
```
VictorBarbosaMartins, line 35: We usually uses copy.copy but I checked now in the reference API and deepcopy is probably better. Should we worry that `copy.copy` is not doing what we think it is in other parts of the code?

.github/workflows/CI-unittests.yml
```diff
@@ -101,6 +101,12 @@ jobs:
         run: |
           pytest --durations=10 --color=yes -n auto --dist loadscope
 
+      - name: Random order
```
VictorBarbosaMartins, line 104: One question: is this going to be applied only to the main or to each branch?

VictorBarbosaMartins (approved):
Thanks. Good to have this ensuring the tests are not depending on each other.

tests/unit_tests/model/test_model_parameter.py
GernotMaier: My assumption is always that if the test does not fail, then `copy.copy` is enough. In this case, we required a deepcopy.

.github/workflows/CI-unittests.yml
GernotMaier: Yes, only to main. In 99% of all cases, tests don't depend on their sequence. For the rare cases for which they do, it is better good enough to find this out after merge-to-main.

To add: this is only used for the scheduled runs which runs automatically every night on the main branch.

GernotMaier:
@VictorBarbosaMartins - thanks for the review!

PR #922, opened 2024-05-16
Add metaschema and tests for integration tests config files

GernotMaier:
Integration tests are configured using the yaml files in tests/integration_tests/config . The configuration starts getting more complex and this PR adds two features to address this:

- complete description of all fields in the configuration file with a `metaschema simtools/schemas/integration_tests_config.metaschema.yml`
- testing of configuration files in the CI `.github/workflows/CI-schema-validation.yml`

Note that the 'CONFIGURATION' branch of the test configuration is not part of the metaschema, as this lists the command line configuration parameters for each simtool.


simtools/schemas/integration_tests_config.metaschema.yml
```diff
@@ -0,0 +1,74 @@
...
+    type: object
+    additionalProperties: false
+    properties:
+      OUTPUT_FILE:
+        description: |
+          "Path of output file tested to be generated by integration test."
+        type: string
+      FILE_TYPE:
+        description: |
+          "Expected file type of output file generated by integration test."
```
VictorBarbosaMartins, line 65: I think we actually don't need this anymore, do we? If we check whether the file exists (with the type in the file name), we will not need to test its type again.

VictorBarbosaMartins (approved):
Looks good! We will have to remember to update the meta schema whenever we introduce new tests (we should include that in the developers guidelines)

GernotMaier:
Thanks for the review!

We only have to update the schema if we introduce changes to the structure of the test configuration files. We will immediately notice this, as we a CI running validating all files.

simtools/schemas/integration_tests_config.metaschema.yml
GernotMaier: Yes - although I would keep it as a weaker version of the exact file name test ('whatever the output file name is, the type should be e.g., json').

PR #928, opened 2024-05-17
Read site names from list of array elements

GernotMaier:
This is a follow-up PR to reduce the number of hard-wired values in `names.py` - this time removing the list of sites and replaces it by reading the sites from the array elements file in `simtools/schemas/array_elements.yml`. 

Remove also the ambiguity in site definitions (it is South/North (or whatever is defined in  simtools/schemas/array_elements.yml) - not CTAO-South, not CTA-South, Paranal, paranal etc).


simtools/utils/names.py
```diff
@@ -39,10 +39,23 @@ def array_elements():
...
-    "South": ["paranal", "south", "cta-south", "ctao-south", "s"],
-    "North": ["lapalma", "north", "cta-north", "ctao-north", "n"],
-}
+@lru_cache(maxsize=None)
+def site_names():
+    """
+    Site names from reference file.
+
+    The list of sites is derived from location of available array elements
+    Return a dictionary for compatibility with the validation routines.
```
VictorBarbosaMartins, line 48: I found this explanation a bit confusing. Could you please improve it a bit?
E.g. "The list of sites is derived from the array element schema file."
"Returns a dictionary that is compatible with..."

tests/resources/telescope_positions-North-ground.ecsv
```diff
@@ -34,7 +34,7 @@
 #       - {CLASS: null, DESCRIPTION: null, ID: null, SITE: null, SUBTYPE: null, TYPE: null}
 #       DOCUMENT:
 #       - {CREATION_TIME: null, ID: null, TITLE: null, TYPE: null, URL: null}
-#     INSTRUMENT: {CLASS: null, DESCRIPTION: null, ID: null, SITE: CTA-North, SUBTYPE: null, TYPE: null}
```
VictorBarbosaMartins, line 37: This made me realize that we still have a lot of references to "CTA" in the code, whereas it should now be "CTAO". Should we open an issue to change that?

tests/unit_tests/db/test_db_handler.py
```diff
@@ -347,6 +347,18 @@ def test_update_parameter_field_db(db, random_id, db_cleanup, io_handler):
     )
     assert pars["camera_pixels"]["applicable"] is False
 
+    with pytest.raises(ValueError):
+        db.update_parameter_field(
+            db_name=f"sandbox_{random_id}",
```
VictorBarbosaMartins, line 352: With the other big PR merged we have now to replace the DB to the Staging one, right? We should remind both Orel and Tobias to do that as well once they return.

tests/unit_tests/utils/test_names.py
```diff
@@ -38,8 +38,8 @@ def test_get_list_of_telescope_types():
 
 def test_validate_name():
     with_lists = {
-        "South": ["paranal", "south", "cta-south", "ctao-south", "s"],
-        "North": ["lapalma", "north", "cta-north", "ctao-north", "n"],
+        "South": ["south"],
+        "North": ["north"],
```
VictorBarbosaMartins, line 42: Do we really need such a validation now that only one string name is accepted?

VictorBarbosaMartins (commented):
Hi @GernotMaier, I provide a couple of comments but in general all good here.

simtools/utils/names.py
GernotMaier: Good point, improved it.

tests/resources/telescope_positions-North-ground.ecsv
GernotMaier: I wouldn't spend time on it now but only when a new version of the metadata documents is released. There might be other changes to be done.

tests/unit_tests/utils/test_names.py
GernotMaier: The validation still works with a dict, as the general validation routine in names.py (names._validate_name) requires a dict.

GernotMaier:
@VictorBarbosaMartins - thanks for looking into it. I've improved the docstring, let me know if I should do anything else.

tests/resources/telescope_positions-North-ground.ecsv
VictorBarbosaMartins: Ok, I just created an issue such that we don;t forget about it. No need to work on that now

VictorBarbosaMartins (approved):
All good.

PR #935, opened 2024-05-23
Trigger rate stacked files

VictorBarbosaMartins:
This PR is a follow-up of #801. It expands the tool with a new functionality (which was partially implemented before): stack files. It is now possible to pass a text file with a list of simtel files and stack the histograms together before estimating the trigger rate. This is especially useful when we have many simulation results in separate files but we are interested in increase the event statistics of the trigger rate estimates.


tests/unit_tests/simtel/test_simtel_histogram.py
```diff
@@ -173,6 +173,22 @@ def test_compute_system_trigger_rate_and_table(simtel_array_histogram_instance):
     assert table["Energy (TeV)"].value[0] == 1.00000000e-03
 
 
+def test_compute_system_trigger_rate_with_input(simtel_array_histogram_instance):
+    from astropy import units as u
```
GernotMaier, line 177: Any particular reason why doing the import here and not as usual at the top of the test module?

GernotMaier (approved):
All good, thank you.

One minor question on the import statements in the tests.

tests/unit_tests/simtel/test_simtel_histogram.py
VictorBarbosaMartins: This was to avoid having a very long import at the beginning but since many of the functions anyways need the module, I now put it on top

VictorBarbosaMartins:
Thanks for the quick feedback @GernotMaier. I merged main and if tests pass I will merge.

PR #939, opened 2024-05-25
Add pre-commit hook to avoid large files updates

GernotMaier:
The reason why the simtools repo is so large is that we have uploaded at some point accidentally some very large files.

To prevent this, introduced a pre-commit to prevent this:

Set the max value to 500 kB. We have a couple (4-5 files) above this value; in this case I am expected that we consciously bypass the pre-commit hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,8 @@ repos:
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
+      - id: check-added-large-files
+        args: ['--maxkb=500']
   - repo: local
     hooks:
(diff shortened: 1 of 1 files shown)


VictorBarbosaMartins:
Why introducing a pre-commit just to by-pass it? Is there a way to automatically by-pass the files that are already in the project such that we don't need to do that by hand in every commit?

GernotMaier:
The other way around: we introduce a pre-commit to avoid committing large files by accident. 

If we actually want to do it, we have to consciously bypass the pre-commit stage.

VictorBarbosaMartins (approved):

PR #938, opened 2024-05-24
Step 1 of ruff integration: auto fixes

GernotMaier:
This is the first of probably several pull request dealing with the integration of the ruff linter.

This PR:

- updates the versions of the pre-commit tools using `pre-commit autoupdate`
- adds ruff options corresponding to our project (e.g., to the pylint rules we are using) to pyproject.toml
- add ruff to add package

All other changes are the result of running `ruff check --fix ./` - so all automatic fixes.

(#913 is probably obsolete - will be deleted after we have worked on some of the issues)

Issue #886, opened 2024-04-23
Suggest to add pre-commit for docstrings (numpy.doc)


GernotMaier:
I've noticed that a lot of good pull review comments are regarding the docstrings. 

As we are using the numpydoc style: to make reviews easier, I suggest to add the numpy doc pre-commit hook (see https://numpydoc.readthedocs.io/en/latest/validation.html#).

I've checked the comments Tobias had in PR #885 , and the hook picked all of them up (and some others):

```
+----------------------------------------------------+----------------------------------------------------+---------+--------------------------------------------+
| file                                               | item                                               | check   | description                                |
+====================================================+====================================================+=========+============================================+
| simtools/applications/db_development_tools/add_mod | add_model_parameters_from_repository_to_db.add_val | PR01    | Parameters {'db_name'} not documented      |
| el_parameters_from_repository_to_db.py:75          | ues_from_json_to_db                                |         |                                            |
+----------------------------------------------------+----------------------------------------------------+---------+--------------------------------------------+
| simtools/applications/db_development_tools/add_mod | add_model_parameters_from_repository_to_db._add_me | PR01    | Parameters {'logger', 'db', 'db_name'} not |
| el_parameters_from_repository_to_db.py:110         | tadata_to_db                                       |         | documented                                 |
+----------------------------------------------------+----------------------------------------------------+---------+--------------------------------------------+
| simtools/applications/db_development_tools/add_mod | add_model_parameters_from_repository_to_db._add_me | PR02    | Unknown parameters {'db: DatabaseHandler'} |
| el_parameters_from_repository_to_db.py:110         | tadata_to_db                                       |         |                                            |
+----------------------------------------------------+----------------------------------------------------+---------+--------------------------------------------+
```

Note that depending on how many of the validation checks we allow, this might be a bit painful in the beginning. But I think it is worth to do it. Any objections?

(in general we don't spend enough time on the documentation, so this maybe forces us to do so)

VictorBarbosaMartins:
If it helps you I have no objection. I think even private functions should be included (and documented). Unfortunately, there is no way to generate the documentation automatically ...

GernotMaier:
Ruff has pydocstyle integrated and allows and option `lint.pydocstyle.convention = "numpy"`. 

Looking through the documentation, it seems that this is enough and we do not need the additional numpydoc validation. I am therefore closing this issues (and open one on ruff documentation checks).

PR #941, opened 2024-05-27
Step 2 of ruff integration: fix issues

GernotMaier:
**Needs to be reviewed after merging #938**

This is step 2 of the ruff integration which fixes all warnings with the following exceptions:

- simtools/camera_efficiency.py used variables called N1, N2, etc, which are used in this way in the DB and in scripts of sim-telarray. Disabled here ruff (we had it disabled in pyling)
- temporarily disabled these two issues regarding tests:
	- PT011 `pytest.raises(ValueError)` is too broad (I think we should fix those)
	- PT012 `pytest.raises()` block should contain a single simple statement (Not sure if this is absolutely necessary; but easy to fix)

This is with "D" (documentation checks) switched off. Ruff suggest a lot (thousands) of changes to the docstrings and we need to discuss how we do this (also in light of #886). I think we first want to understand how we want our docstrings to look like (I don't like all ruff suggestions.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,6 +139,8 @@ lint.ignore = [
   "D215",   # section-underline-not-over-indented
   "G004",   # Logging statement uses f-string
   "ISC001", # incompatible with ruff format
+  "PT011",  # TODO - fix
+  "PT012",  # TODO - fix
 ]
 lint.per-file-ignores."**/tests/**" = [
   "D",
--- a/simtools/applications/plot_array_layout.py
+++ b/simtools/applications/plot_array_layout.py
@@ -40,7 +40,7 @@
 import logging
 from pathlib import Path
 
-import matplotlib
+import matplotlib as mpl
 import matplotlib.pyplot as plt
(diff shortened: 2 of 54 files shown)


GernotMaier:
I mark this ready for review - codecov complaints about missing tests, but this PR is not about test coverage and I suggest to ignore these.

VictorBarbosaMartins:
Having a broad test is still better than having no test. I think it is strange to remove tests only to have ruff not complaining about them.

GernotMaier:
As discussed in person, PT011  and PT012 is something we need to discuss if we want to integrate them. We are not obliged to use all options proposed by ruff. I am opening an issue on it.

Note also that these two are very independent of the rest of the code.

VictorBarbosaMartins (approved):
All good, actually I have no comment on this PR. It seems that this could have been part of an automatic update as well.

VictorBarbosaMartins:
As @GernotMaier prefer to ignore the failing coverage test and deal with it later, I am merging this.

PR #943, opened 2024-05-27
Step 3 of ruff integration: add ruff to CI (plus cleanup of different linter stages)

### LINKED ISSUE #942: Linter cleanup
There is unnecessary duplication in the linters as implemented in the workflows (positive: better to test twice than not at all). This is amplfied by the recent introduction of added pre-commit runs over all files as CI and the introduction of ruff (not yet added to pre-commit, but planned).

This is the suggestion:

1. Review pre-commit and make sure that any workflows which run reasonable fast are included. 
2. Remove duplications (e.g., flake8 is running in the pre-commits, in CI-linter and CI-unittests! Additionally, ruff is almost 100% compatible with flake8 tests)
3. Replace pylint by ruff in pre-commit. Continue to run pylint in CI-linter
4. Review if we need the super-linter workflow in CI-linter. Given that most linters run using the pre-commit workflows, it might be better to run the few which are not explicitly.
5. If possible, configure tools in pyproject.toml (some tools recently introduced this; e.g., markdownlint)

GernotMaier:
Integrate ruff as linter and add it to pre-commit. 

Cleanup of linters, see #942.

Pre-commit changes:

- removed flake8 from pre-commit (covered by ruff, see [here](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-flake8))
- removed pylint from pre-commit (this is too slow for pre-commit; most of it is covered by ruff - anything not picked by ruff will by shown by the CI-linter workflow)
- removed isort from pre-commit (covered by ruff, see [here](https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort))
- did not remove black (and do not use the `ruff format` as this introduces some changes)

Github action changes:

- remove flake8 from CI-unittest

Others

- removed flake8 from list of required packages
- removed actionlint, black from superlinter (already in pre-commit)
- removed flake8, isort from superlinter (already in ruff in pre-commit)
- removed markdownlint from superlinter (added to pre-commit in PR #940)
- replaced superlint yamllint by pre-commit yamllint

Closes #942.

### DIFF
--- a/.github/workflows/CI-linter.yml
+++ b/.github/workflows/CI-linter.yml
@@ -47,33 +47,16 @@ jobs:
 
       - name: install packages not included in super-linter
         run: |
-          pip install validate-pyproject restructuredtext-lint
+          pip install validate-pyproject
 
       - name: pyproject.toml
         run: |
           validate-pyproject pyproject.toml
 
-      # RST linter
-      # Note: unclear how to suppress error messages
-      # (use grep -v in this case)
-    #  - name: restructuredtext-lint
-    #    run: |
-    #      rst-lint README.rst docs/source |
-    #       grep -v "Unknown directive type" |
-    #       grep -v "Unknown interpreted text role" |
-    #       grep -v "Cannot analyze code. Pygments package not found."
-
       - name: Check whether the citation metadata from CITATION.cff is valid
         uses: citation-file-format/cffconvert-github-action@2.0.0
         with:
           args: "--validate"
-
-      - name: yaml_config used by super-linter
-        run: |
-          # TODO - very large line length
-          echo 'rules:' > yaml_config.yaml
-          echo '  line-length:' >> yaml_config.yaml
-          echo '    max: 250' >> yaml_config.yaml
 
       # Dependencies required to avoid errors
       # reported by linters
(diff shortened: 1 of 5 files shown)


GernotMaier:
@VictorBarbosaMartins - please review only after #941 is merged to main (otherwise you review the same changes again)

VictorBarbosaMartins:
I think we should talk to other groups in DPPS to check what is their approach in this regard. Perhaps repeating checks in the pre-commit hooks and ci linters is not a bad idea.

.github/workflows/CI-linter.yml
```diff
@@ -131,5 +98,5 @@ jobs:
           LOG_FILE: superlinter.log
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # exclude link to markdown files
+          # exclude markdown file with links
           FILTER_REGEX_EXCLUDE: "docs/source/docker_files.md"
```
VictorBarbosaMartins, line 102: what is actually the problem with the links?

.pre-commit-config.yaml
```diff
@@ -58,3 +39,9 @@ repos:
     rev: v2.3.0
     hooks:
       - id: codespell
+  # yamllint
+  - repo: https://github.com/adrienverge/yamllint.git
```
VictorBarbosaMartins, line 43: this is not included in ruff?

GernotMaier:
> I think we should talk to other groups in DPPS to check what is their approach in this regard. Perhaps repeating checks in the pre-commit hooks and ci linters is not a bad idea.

Wait - pre-commit is running as linter in CI-linter over all files (this is also how DPPS is doing it). So we are still using all the tools and do a full linting.

.pre-commit-config.yaml
GernotMaier: Ruff is for python (and this is yaml)

.github/workflows/CI-linter.yml
GernotMaier: There is a markdown rule which requires that the first line is a header line (‘# This is a Header’). This is why the linter would fail.

GernotMaier: But we actually don't need this anymore; we have moved markdownlint to pre-commit and disabled this check (`"--disable", "MD041"]` see PR 940. 

I remove this line now.

GernotMaier:
Just to make the last point clearer: precommit is running over all files in Ci-linter.yml, see this line: https://github.com/gammasim/simtools/blob/6d10f5bf47956f9722d7a41062a223799944a9f7/.github/workflows/CI-linter.yml#L78

VictorBarbosaMartins (approved):
Ok! thank you for the answers. Looks good.

GernotMaier:
Thanks @VictorBarbosaMartins

PR #946, opened 2024-05-29
Markdown badges

GernotMaier:
Fix links in main README.md

### DIFF
--- a/.github/workflows/CI-linter.yml
+++ b/.github/workflows/CI-linter.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Pre-commit
         run: |
-          pre-commit run --all-files --verbose
+          pre-commit run --all-files
 
       - name: Lint Code Base
         uses: super-linter/super-linter@v6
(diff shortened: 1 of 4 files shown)


VictorBarbosaMartins:
Hi @GernotMaier, thanks for this fix. It indeed fixed the issue. but I found another similar one. The link to the code in the documentation is broken because there is no `docs/index.md` (see two screenshots below). I think we can fix that here too:

Action (click on **show source**):
Screenshot 2024-05-29 at 14 11 57


Error message (same for main and this branch):
Screenshot 2024-05-29 at 14 12 05

GernotMaier:
Could you have a look and fix it?

VictorBarbosaMartins:
I looked into this and it seems that this is because `sphinx-book-theme` expects an `index.html` file inside `./docs` and not inside `./docs/source`. I couldn't find in simtools how to change the expected path (I imagine this is encoded in the theme). In any case I could move `index.html` from `./docs/source` to `./docs` but this would motivate us to move all the other files as well and get rid of the source directory. This goes beyond my decision alone and I need your feedback.

GernotMaier:
Fixed this by setting the `"path_to_docs":` as outlined in https://sphinx-book-theme.readthedocs.io/en/stable/components/source-files.html#add-a-button-to-the-page-source

VictorBarbosaMartins (approved):
Thanks. Looks good now.

PR #948, opened 2024-05-30
Add output_path to integration tests with file checks

GernotMaier:
Interesting issue:

- integration tests are working fine in the CI on github
- some fail when running locally, e.g. ` pytest  --no-cov "tests/integration_tests/test_applications_from_config.py::test_applications_from_config[simtools-calculate-trigger-rate_file_list]"`
- they don't fail when adding the model version, e.g., ` pytest --model_version="2024-02-01" --no-cov "tests/integration_tests/test_applications_from_config.py::test_applications_from_config[simtools-calculate-trigger-rate_file_list]"`

Failing are tests which do not have `OUTPUT_PATH` configured, but are using the functionality to locate or compare output files. I tried to understand what happens in `test_applications_from_config.py`, but it is not absolutely clear to me. The easiest fix is to configure explicitly `OUTPUT_PATH` in the test configuration to make sure that tests are written into the temp_test_directory (otherwise tests are writing locally).

### DIFF
--- a/tests/integration_tests/config/calculate_trigger_rate_file_list.yml
+++ b/tests/integration_tests/config/calculate_trigger_rate_file_list.yml
@@ -5,6 +5,7 @@ CTA_SIMPIPE:
     SIMTEL_FILE_NAMES:
       - './tests/resources/simtel_output_files.txt'
     SAVE_TABLES: true
+    OUTPUT_PATH: simtools-tests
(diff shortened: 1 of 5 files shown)


VictorBarbosaMartins (approved):
Interesting. It makes indeed sense to define an output path given that we check for the existence of the output files. Thanks for finding this out and proposing the solution.

PR #950, opened 2024-05-31
Replace ArrayLayout by ArrayModel for retrieving telescope positions in light emission package

GernotMaier:
The ArrayLayout model is used primarily for coordinate transformations and should not be used to read telescope positions from a file. Replace ArrayModel therefore in the light emission package by ArrayModel. 

ArrayModel has the additional functionality of reading telescope positions from the database.

- [x] unit tests fail - need @tobiaskleiner for help

(also removed an unused fixed in one of the tests related to the array_layout module)


tobiaskleiner (approved):
Thanks, looks all good!

PR #953, opened 2024-06-04
Set logging level for pymongo module to warning

GernotMaier:
Debugging is hard due to extensive and many debug statements printed by the pymongo module. This is especially true when debugging unit tests.

Suggest to set the logging level for the pymongo module to `WARNING`, independent of the general logging level.


GernotMaier:
Commit accidentally the removal of `names.validate_array_layout_name` related to #487 to this PR.

Probably still fine to review this here without trying to revert the commit. Apologies!

PR #954, opened 2024-06-05
Disable fail fast for integration tests

GernotMaier:
We run integration tests for several (two) different model versions. 

This PR disables `fail fast` (meaning that all matrix jobs are canceled if one fails), so that we understand better if integration tests generally fail or for a certain model version only.

This workflow setting is described [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast).


VictorBarbosaMartins (approved):
This is a good point. Approved.

PR #957, opened 2024-06-06
First version of  CTAO DPPS sonarqube workflow

GernotMaier:
Integration of simtools into the CTAO-DPPS test environment using sonarqube. **This is not about fixing the issues reported by sonarqube but to get it running.**

This includes:
- adding the sonarqube app
- add access tokens
- add step in CI-unit test workflow to upload coverage report
- add file with sonarque properties

In the current setting, even when sonarque is reporting 'Failed', the github workflows are fine. This is the intention until we have gained experience, and then see how we proceed. At this point, everything (!) is switched at sonarque and lots of issues are reported.

The report can be found here: https://sonar-cta-dpps.zeuthen.desy.de/dashboard?id=gammasim_simtools_AY_ssha9WiFxsX-2oy_w&pullRequest=957

Note that the badges on the readme page will only work after main has been analyzed once.


tobiaskleiner (approved):
All good

PR #964, opened 2024-06-07
Add a small tool to inspect databases

GernotMaier:
Add a simple tool to quickly list the available databases and its collections. I've need this now a couple of times and it is simply faster than opening robot 3T (or something similar). Also means that we reduce the necessity that developers install robot 3T. And the last point is that this is a tool to quickly check the mongoDB status.

This gives e.g. (with my limited local database instance and not the full one running at DESY)

```
python simtools/applications/db_inspect_databases.py --db_name all
Database: Staging-CTA-Simulation-Model-v0-3-0
   Collections: ['configuration_sim_telarray', 'telescopes', 'metadata', 'sites', 'calibration_devices', 'fs.files', 'fs.chunks']
```

### DIFF
--- a/database_scripts/README.md
+++ b/database_scripts/README.md
@@ -14,4 +14,41 @@ For testing and development, it might be useful to work with a local copy of the
 
 ### Prerequisites
 
-Access to a database dump of the production database is required.
+Access to a database dump of the production database is required. It is assumed that the dumps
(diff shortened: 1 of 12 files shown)


GernotMaier:
Close this, as I've accidentally added other files. New correct PR is #967.

PR #967, opened 2024-06-10
Add a small tool to inspect databases 

GernotMaier:
Add a simple tool to quickly list the available databases and its collections. I've need this now a couple of times and it is simply faster than opening robot 3T (or something similar). Also means that we reduce the necessity that developers install robot 3T. And the last point is that this is a tool to quickly check the mongoDB status.

This gives e.g. (with my limited local database instance and not the full one running at DESY)

```
python simtools/applications/db_inspect_databases.py --db_name all
Database: Staging-CTA-Simulation-Model-v0-3-0
   Collections: ['configuration_sim_telarray', 'telescopes', 'metadata', 'sites', 'calibration_devices', 'fs.files', 'fs.chunks']
```


VictorBarbosaMartins:
I am starting reviewing this, but I already have a question. What do you mean by local database instance? These are DBs hosted locally in your laptop? this is quite strange to me, as the very concept of DB is to make it remotely accessible in the most efficient way possible. Perhaps I missed something, but can you please clarify this to me?

GernotMaier:
Exactly. You can run the DB locally, which is especially useful for heaving development and debugging, tests. etc without having any impact on the remove DB which is used by others.

simtools/applications/db_inspect_databases.py
```diff
@@ -0,0 +1,48 @@
...
+
+    for db_name in databases:
+        if args_dict["db_name"] != "all" and db_name != args_dict["db_name"]:
+            continue
+        # missing admin rights; skip config and admin
+        if db_name in ("config", "admin", "local"):
+            continue
+        print("Database:", db_name)
+        collections = db.get_collections(db_name=db_name)
+        print("   Collections:", collections)
```
VictorBarbosaMartins, line 44: there is an extra space in the beginning of the string, which we should probably remove

simtools/applications/db_inspect_databases.py
VictorBarbosaMartins, line 1: shouldnt this tool be inside the db applications directory?

VictorBarbosaMartins (approved):
Some minor comments only.

simtools/applications/db_inspect_databases.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/python3
+
+"""
+Inspect databases and print (available database names and collections).
```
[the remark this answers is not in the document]
GernotMaier: Added.

simtools/applications/db_inspect_databases.py
```diff
@@ -0,0 +1,48 @@
...
+import simtools.utils.general as gen
+from simtools.configuration import configurator
+from simtools.db import db_handler
+
+
+def main():
+    config = configurator.Configurator(description="Inspect databases")
+    config.parser.add_argument(
+        "--db_name",
+        help="Inspect a specific database (use all to print all databases)",
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

simtools/applications/db_inspect_databases.py
```diff
@@ -0,0 +1,48 @@
...
+    logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
+
+    db = db_handler.DatabaseHandler(mongo_db_config=db_config)
+
+    databases = db.db_client.list_database_names()
+
+    for db_name in databases:
+        if args_dict["db_name"] != "all" and db_name != args_dict["db_name"]:
+            continue
+        # missing admin rights; skip config and admin
```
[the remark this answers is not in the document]
GernotMaier: Improved comment.

simtools/applications/db_inspect_databases.py
GernotMaier: Fixed.

simtools/applications/db_inspect_databases.py
GernotMaier: Not really - we put tools which we do not expect anyone except absolute db experts to use in to the db directory. This one is a tool the general user might find useful.

GernotMaier:
Thanks for the review!

PR #968, opened 2024-06-10
Improve position coordinate transformation and add json reader / writer

### LINKED ISSUE #924: Add coordinate conversion of single telescope positions in the DB format.
The application `print_array_elements.py` allows to convert positions between different coordinate systems given as astropy tables.

Add functionality to read (from DB or file) a telescope position and convert it to a requested system.

Consider renaming this application to `transform_array_element_position.py` (it does far more than simply printing).

GernotMaier:
This PR adds a couple of improvements to the position coordinate transformations:

- add possibility to do a coordinate conversion of single array element positions in the  simtools-db-json format (application will understand the generic json format and write it also; see `tests/integration_tests/config/convert_geo_coordinates_of_array_elements_ground_to_utm_json.yml`
- generalized the writing of simtools-db-json format-style files and move the function from `simtel_config_reader` to `model_data_writer` (simple move; no change in functionality
- renamed application to convert coordinates (from misleading `print_array_elements` to `convert_geo_coordinates_of_array_elements`
- improvement of docstrings to follow numpydoc standards

Add a small improvement to the integration tests to the function of `compare_json_files`: for simtools-db-type json files where a list of float is encoded as string (similar to sim_telarray; e.g., `5.5 3.2  7.6`). Used np.all_close to compare those floats.

Closes #924


simtools/applications/convert_geo_coordinates_of_array_elements.py
```diff
@@ -1,58 +1,52 @@
...
-    coordinate systems.
-
+    Convert array element positions in different CTAO coordinate systems.
     Available coordinate systems are:
 
     1. UTM system
-    2. ground system (similar to sim_telarray system with x-axis pointing toward \
-        geographic north and y-axis pointing towards the west); altitude relative \
-        to the CORSIKA observation level.
+    2. ground system (similar to sim_telarray system with x-axis pointing toward geographic north
```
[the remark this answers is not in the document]
GernotMaier: Good point. I've added one sentence to make this clearer:

`Altitude is the height of the elevation rotation axis (plus some possible mirror offset).`

simtools/applications/convert_geo_coordinates_of_array_elements.py
```diff
@@ -1,58 +1,53 @@
...
+       and y-axis pointing towards the west); altitude relative to the CORSIKA observation level.
+       Altitude is the height of the elevation rotation axis (plus some possible mirror offset).
```
VictorBarbosaMartins, line 13: It is not clear to me what "plus some possible mirror offset" means. I think this refers to a possible shift due to the mirror shape, but it does not inform whether the shift is already included in the model, or how much is the shift? Should the user worry about this?

simtools/data_model/model_data_writer.py
```diff
@@ -11,6 +16,23 @@
 __all__ = ["ModelDataWriter"]
 
 
+class JsonNumpyEncoder(json.JSONEncoder):
+    """Convert numpy to python types as accepted by json.dump."""
```
VictorBarbosaMartins, line 20: I know this was in the code before, but I googled it and there seems to be a better way to do it, at least for the python types (using the `.item()` method). Please check the best answer here: 
[https://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types](https://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types)

simtools/layout/array_layout.py
```diff
@@ -524,6 +546,63 @@ def export_telescope_list_table(self, crs_name):
 
         return table
 
+    def export_telescope_list_as_json(self, crs_name):
+        """
+        Return a single-telescope list as simtools-DB-style json.
```
VictorBarbosaMartins, line 551: I am not sure what "single-telescope list" means. A list with a single telescope in it? or a single list with all the telescopes?

simtools/layout/array_layout.py
```diff
@@ -524,6 +546,63 @@ def export_telescope_list_table(self, crs_name):
...
+            Name of coordinate system to be used for export.
+
+        Returns
+        -------
+        dict
+            Dictionary with array element information.
+        """
+        table = self.export_telescope_list_table(crs_name)
+        if len(table) != 1:
+            raise ValueError("Only one telescope can be exported to json")
```
VictorBarbosaMartins, line 565: If the function is supposed to work only with a single telescope, I would make that clearer in the name of the function. Or alternatively, I would expand the function to check the size of table and loop through the telescopes, exporting a json file to each one.

VictorBarbosaMartins (commented):
Hi @GernotMaier. I provide some comments here. In general I think it is already in a good state.

simtools/applications/convert_geo_coordinates_of_array_elements.py
```diff
@@ -152,12 +146,22 @@ def main():
     logger = logging.getLogger()
     logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
 
-    metadata = MetadataCollector(args_dict=args_dict, data_model_name=data_model_name)
+    json_type = args_dict.get("input", "").endswith(".json")
```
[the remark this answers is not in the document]
VictorBarbosaMartins: We actually do not get an error message telling you that `--input` should not be `None`. On the contrary, we get an error that a method we tried to use does not exist, which is an issue. These messages are not the same. A simple try and except statement with the message "`--input` should not be `None`" replacing the error message would already correct this.

simtools/applications/convert_geo_coordinates_of_array_elements.py
VictorBarbosaMartins: My better suggestion is to remove `(plus some possible mirror offset)`. If this is an information the user needs, then I would pass the complete information, otherwise I think this piece of information is not needed here. This confuses more than adds to the docstring.

simtools/data_model/model_data_writer.py
VictorBarbosaMartins: I disagree. Creating an encoder itself is not clearer than a simple one-line method. But as this is not necessarily part of the PR, I will leave it as it is then.

VictorBarbosaMartins (approved):
Ok!

simtools/data_model/model_data_writer.py
GernotMaier: To explain why I think the current way is better: `if isinstance(o, np.floating):` is understandable without knowing even python. `type(np.float64(0).item()) ` requires knowledge about the internals of  the numpy float64 class - I would have to look it up what it means. The encodes is required by the json writer.

PR #969, opened 2024-06-10
Sonarqube fixes

tobiaskleiner:


simtools/ray_tracing.py
```diff
@@ -307,7 +307,6 @@ def analyze(
                     containment_diameter_cm = image.get_psf(containment_fraction, "cm")
                     containment_diameter_deg = image.get_psf(containment_fraction, "deg")
                     centroid_x = image.centroid_x
```
GernotMaier, line 309: Why do you need to keep centroid_x?

simtools/simtel/simtel_config_reader.py
GernotMaier, line 1: This I don't understand. What is better?

simtools/simtel/simtel_runner.py
```diff
@@ -168,7 +168,7 @@ def run(self, test=False, force=False, input_file=None, run_number=None):
             self._logger.error(msg)
             raise RuntimeError(msg)
 
-        if not self._shall_run(run_number=run_number) and not force:
+        if not self._shall_run() and not force:
```
GernotMaier, line 171: And you are sure that the run number is not needed? This seems to be a functional change.

GernotMaier (changes requested):
Could you have another look at the changes in simtel-runner? Not sure why you don't need anymore the run number.

simtools/simtel/simtel_runner.py
tobiaskleiner: Don't really understand what the purpose of shall_run is here: https://github.com/gammasim/simtools/blob/b010eee27884d85353cac010f0ae17e19ed291a0/simtools/simtel/simtel_runner.py#L231
It only returns False. Do you know?

simtools/ray_tracing.py
tobiaskleiner: @GernotMaier did you mean the centroid_y? This one could be removed probably.

GernotMaier: I was wondering why we need only centroid x (and not centroid y). I can see that is used for the effective focal length calculation.

Doesn't that depend where in the FOV we put the simulated optical point source? I am wondering if this calculation is still correct when using another source position (isn't this a configurable option in the program and we can easily get a wrong effective focal length)

simtools/simtel/simtel_runner.py
GernotMaier: I think we need this function in case this is not re-implemented in the child classes (as e.g., in simulated_light_emission).

simtools/ray_tracing.py
GernotMaier: And we can move this discussion also into an issue, as it is not related to the sonarqu reports.

tobiaskleiner: Isse is here #981

simtools/simtel/simtel_runner.py
tobiaskleiner: Ok this seems indeed needed. The `run_number` however is not required to be passed.

GernotMaier (approved):
Great, thanks for the changes. Approved!

PR #976, opened 2024-06-12
Improved naming of simtel modules plus docstring ruff formatting.

GernotMaier:
**Do NOT review every single change in this PR. Review only the name changes of the modules in the simtools/simtel directory.**

Improve naming of classes in simtel directory to have a clearer separation of those modules responsible for simulations and those for io. This is the change to review: are the new names good?

- renaming done with search and replace
- applied ruff check --fix to fix docstrings in simtel modules
- applied trivial changes by hand (period at end of line. One-line docstring should fit on one line. 1 blank line required between summary line and description.


tobiaskleiner (approved):
Thanks @GernotMaier . The new naming is good and browsing through I didn't spot any obvious problem in the changes.

PR #977, opened 2024-06-12
Fix complexity issue in simtel_config_writer and add unit test

GernotMaier:
Address the complexity issue: https://sonar-cta-dpps.zeuthen.desy.de/project/issues?resolved=false&pullRequest=969&id=gammasim_simtools_AY_ssha9WiFxsX-2oy_w&open=AZALqF5Y-Rsd-Q9KQnor

Actually a good point - moving this into a separate function is actually a good point and allows as well to test the string treatment.


tobiaskleiner (approved):
Thanks Gernot, good to merge.

PR #988, opened 2024-06-14
improve docstring

### LINKED ISSUE #987: Improve docstrings in simtel_io_histogram and simtel_io_histograms modules
This is mostly about the first sentence:

"Handle a single histogram (or simtel_array output) file."
"Handle histograms sim_telarray files especially for trigger rate calculations."

Could we replace 'Handle'? Really not clear what it mean (does it mean read, write, remove, copy, analyse, ...?). Would be good to be a bit clearer.

VictorBarbosaMartins:
Closes #987.

I hope that is enough. Please let me know if more information is needed.


GernotMaier (approved):
Good! Much better than 'handle'. Thanks!

PR #991, opened 2024-06-17
improve guidelines for testing

VictorBarbosaMartins:
I included in the guidelines some information about how to test a single application at a time (and a single integration test). I always forget and had to look it up somewhere, so better we have that in the guidelines as well.


GernotMaier (approved):
Thanks - looks good!

PR #998, opened 2024-06-18
adjust arguments legends_handler

tobiaskleiner:


simtools/visualization/legend_handlers.py
```diff
@@ -176,12 +166,7 @@ class HexOffPixelHandler:
...
     @staticmethod
-    def legend_artist(*args, **kwargs):
-        handlebox = kwargs.pop("handlebox", None)
-        if handlebox is None:
-            # handlebox is expected, return None if not provided
-            return None
-
+    def legend_artist(_, __, ___, handlebox):
         x0, y0 = (
             handlebox.xdescent + handlebox.width / 3,
```
GernotMaier, line 171: This line is repeated quite frequently. Is there an easy way to avoid it?

GernotMaier (approved):
Thanks for this.

Can you check on the one comment I have. Otherwise ready to merge.

simtools/visualization/legend_handlers.py
tobiaskleiner: I have added a function to return the center values.

PR #1000, opened 2024-06-19
Allow hist file into trigger rate application

### LINKED ISSUE #934: Allow the trigger rate application to take the histogram file .hdata as input (and not only .simtel)
This is specially useful to access the additional histograms that are available in the hdata file, for example potentially the histograms for individual telescopes (to be checked)

VictorBarbosaMartins:
In this PR, I implemented some improvements to the trigger rate tool that allows it to calculate the trigger rate for a given an input .hdata.zst file (and not only an .simtel.zst file as it is in main now). However, the configuration used for the simulation is not propagated to the .hdata.zst as it is to the .simtel.zst, which means we do not have the view cone and the energy range used, which are both necessary for the operation of this tool. Therefore, if passing a .hdata to the tool, `--energy_range` and `--view_cone` also have to be passed. One can test this with the new integration test `simtools-calculate-trigger-rate --config tests/integration_tests/config/calculate_trigger_rate_one_file_hdata.yml`. Not passing `--energy_range` or `--view_cone` for an input .hdata file will lead to error, which is tested in the unit tests.

I believe the low coverage by sonarqube is due to the changes in the application itself, which we do not test with unit tests at the moment, so I suggest to ignore it.

Closes #934.

### DIFF
--- a/simtools/applications/calculate_trigger_rate.py
+++ b/simtools/applications/calculate_trigger_rate.py
@@ -88,6 +88,25 @@ def _parse(label, description):
         action="store_true",
     )
 
+    config.parser.add_argument(
+        "--energy_range",
+        help="Energy range (TeV) passed as a list of floats, the minimum and maximum, "
+        "respectively.",
+        required=False,
+        nargs=2,
+        type=str,
+        default=None,
+    )
+
+    config.parser.add_argument(
+        "--view_cone",
+        help="View cone (deg) passed as two floats, the minimum and maximum, respectively.",
+        nargs=2,
+        required=False,
+        type=str,
+        default=None,
+    )
+
     config_parser, _ = config.initialize(db_config=False, paths=True)
 
     return config_parser
@@ -120,8 +139,24 @@ def main():
             logger.error(msg)
             raise FileNotFoundError from exc
 
+    if config_parser["energy_range"] is not None:
+        emin = float(config_parser["energy_range"][0])
+        emax = float(config_parser["energy_range"][1])
+        energy_range = [emin, emax]
+    else:
+        energy_range = None
+    if config_parser["view_cone"] is not None:
+        cone_min = float(config_parser["view_cone"][0])
+        cone_max = float(config_parser["view_cone"][1])
+        view_cone = [cone_min, cone_max]
+    else:
+        view_cone = None
+
     histograms = SimtelIOHistograms(
-        simtel_array_files, area_from_distribution=config_parser["area_from_distribution"]
+        simtel_array_files,
+        area_from_distribution=config_parser["area_from_distribution"],
+        energy_range=energy_range,
+        view_cone=view_cone,
     )
 
     logger.info("Calculating simulated and triggered event rate")
--- a/simtools/applications/calculate_trigger_rate.py
+++ b/simtools/applications/calculate_trigger_rate.py
@@ -112,6 +112,39 @@ def _parse(label, description):
     return config_parser
 
 
+def _get_simulation_parameters(config_parser):
+    """
+    Get the energy range and view cone in the correct form to use in the simtel classes.
+
+    Parameters
+    ----------
+    CommandLineParser:
+        Command line parser object as derived by the _parse function.
+
+    Returns
+    -------
+    list:
+        The energy range used in the simulation.
+    list:
+        The view cone used in the simulation.
+
+    """
+    if config_parser["energy_range"] is not None:
+        emin = float(config_parser["energy_range"][0])
+        emax = float(config_parser["energy_range"][1])
+        energy_range = [emin, emax]
+    else:
+        energy_range = None
+    if config_parser["view_cone"] is not None:
+        cone_min = float(config_parser["view_cone"][0])
+        cone_max = float(config_parser["view_cone"][1])
+        view_cone = [cone_min, cone_max]
+    else:
+        view_cone = None
+
+    return energy_range, view_cone
+
+
 def main():
     label = Path(__file__).stem
     description = (
@@ -139,18 +172,7 @@ def main():
             logger.error(msg)
             raise FileNotFoundError from exc
 
-    if config_parser["energy_range"] is not None:
-        emin = float(config_parser["energy_range"][0])
-        emax = float(config_parser["energy_range"][1])
-        energy_range = [emin, emax]
-    else:
-        energy_range = None
-    if config_parser["view_cone"] is not None:
-        cone_min = float(config_parser["view_cone"][0])
-        cone_max = float(config_parser["view_cone"][1])
-        view_cone = [cone_min, cone_max]
-    else:
-        view_cone = None
+    energy_range, view_cone = _get_simulation_parameters(config_parser)
 
     histograms = SimtelIOHistograms(
         simtel_array_files,
--- a/simtools/applications/calculate_trigger_rate.py
+++ b/simtools/applications/calculate_trigger_rate.py
@@ -109,7 +109,7 @@ def _parse(label, description):
 
     config.parser.add_argument(
         "--view_cone",
-        help="View cone (deg) passed as two floats, the minimum and maximum, respectively.",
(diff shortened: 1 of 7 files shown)


simtools/applications/calculate_trigger_rate.py
```diff
@@ -97,11 +97,63 @@ def _parse(label, description):
         action="store_true",
     )
 
+    config.parser.add_argument(
+        "--energy_range",
+        help="Energy range (TeV) passed as a list of floats, the minimum and maximum, "
+        "respectively.",
+        required=False,
+        nargs=2,
+        type=str,
```
GernotMaier, line 106: Type = 'str', and help requests floats? Not clear if this is consistent.

simtools/applications/calculate_trigger_rate.py
```diff
@@ -97,11 +97,63 @@ def _parse(label, description):
...
+        default=None,
+    )
+
+    config.parser.add_argument(
+        "--view_cone",
+        help="View cone (deg) passed as two floats, the minimum and maximum, respectively.",
```
GernotMaier, line 112: Maybe 'View cone radius (deg)' (I assume it is the radius, or?)

simtools/applications/calculate_trigger_rate.py
```diff
@@ -97,11 +97,63 @@ def _parse(label, description):
...
 
 
+def _get_simulation_parameters(config_parser):
+    """
+    Get the energy range and view cone in the correct form to use in the simtel classes.
+
+    Parameters
+    ----------
+    CommandLineParser:
+        Command line parser object as derived by the _parse function.
```
GernotMaier, line 131: 'derived' --> 'defined'

simtools/simtel/simtel_io_histogram.py
```diff
@@ -50,26 +50,35 @@ class SimtelIOHistogram:
         calculating trigger rate for individual telescopes.
         If false, the area thrown is estimated based on the maximum distance as given in
         the simulation configuration.
-
+    energy_range: list
+        The energy range used in the simulation. It must be passed as a list of floats and the
+        energy must be in TeV (as in the CORSIKA configuration).
```
GernotMaier, line 55: Important: CORSIKA usually works in GeV. Are you sure about the TeV here?

simtools/simtel/simtel_io_histogram.py
```diff
@@ -78,6 +87,11 @@ def __init__(self, histogram_file, area_from_distribution=False):
         self.radius_axis = None
         self.area_from_distribution = area_from_distribution
 
+        self.view_cone = view_cone
```
GernotMaier, line 90: I would find this clearer with `self.view_cone = self._set_view_cone(viewcone)`: in the current implementation, the first assignment is immediately overwritten by the follow up statement.

Same comment for the energy range in lines 93/94.

simtools/simtel/simtel_io_histogram.py
```diff
@@ -207,19 +218,25 @@ def fill_event_histogram_dicts(self):
...
-            self._view_cone = self.config["viewcone"] * u.deg
-        return self._view_cone
+        ValueError:
+            if input parameter is missing.
+        """
+        if self.view_cone is None:
+            try:
+                self.view_cone = self.config["viewcone"] * u.deg
+            except TypeError as exc:
+                msg = "view_cone needs to be passed as argument (a list of cone in deg)."
```
GernotMaier, line 234: `a list of cone` is unclear. "Minimum and maximum of the viewcone radius in deg"

simtools/simtel/simtel_io_histogram.py
```diff
@@ -248,44 +265,50 @@ def total_area(self):
...
+            if input parameter is missing.
+        """
+        if self.energy_range is None:
+            try:
+                self.energy_range = [
+                    self.config["E_range"][0] * u.TeV,
+                    self.config["E_range"][1] * u.TeV,
+                ]
+            except TypeError as exc:  # E_range not in self.config
+                msg = "energy_range needs to be passed as argument (a list of energies in TeV)."
```
GernotMaier, line 306: Same here, 'list of energies' is not what we want - we want min/max of the energy range.

simtools/simtel/simtel_io_histograms.py
```diff
@@ -51,14 +51,33 @@ class SimtelIOHistograms:
         calculating trigger rate for individual telescopes.
         If false, the area thrown is estimated based on the maximum distance as given in
         the simulation configuration.
+    energy_range: list
+        The energy range used in the simulation. It must be passed as a list of floats and the
```
GernotMaier, line 55: Same comment here. GeV in Corsika?

GernotMaier (changes requested):
Thanks for this. See my minor comments below.

simtools/applications/calculate_trigger_rate.py
VictorBarbosaMartins: This is a good point. For some reason the configurator did not manage to add a parameter, which is a list of two floats, so I decided to get a list of str and transform to float in the code. If you happen to know the reason why type=float does not work here, please me know.

simtools/applications/calculate_trigger_rate.py
VictorBarbosaMartins: True. I adjusted the text.

simtools/applications/calculate_trigger_rate.py
VictorBarbosaMartins: Done.

simtools/simtel/simtel_io_histogram.py
VictorBarbosaMartins: Good point. The docstring is confusing, hence I removed the parentheses. But the code is correct. E_range is already given in TeV (this is dealt with in pyeventio).

simtools/simtel/simtel_io_histogram.py
VictorBarbosaMartins: I understand your point. I improved it now.

simtools/simtel/simtel_io_histogram.py
VictorBarbosaMartins: Done.

simtools/simtel/simtel_io_histogram.py
VictorBarbosaMartins: Done.

simtools/simtel/simtel_io_histograms.py
VictorBarbosaMartins: As pointed out before, yes, this is correct. We are not directly dealing with CORSIKA parameters here (they are already dealt with in eventio).

VictorBarbosaMartins:
Hi @GernotMaier, thanks for the comments. I think I addressed all of them. Please let me know what to do in regard to the type=str or type=float. Also please have a look again once tests pass. If they fail, I will deal with them later.

simtools/applications/calculate_trigger_rate.py
GernotMaier: Don't know. This probably requires to look a bit deeper into argparse.

simtools/simtel/simtel_io_histograms.py
GernotMaier: Please remove then "as in CORSIKA configuration", as this is not correct (CORSIKA is strictly in GeV).

GernotMaier (approved):
Thanks for this.

Please correct the one bit on the energy / CORSIKA configuration. Otherwise ready to merge.

simtools/applications/calculate_trigger_rate.py
VictorBarbosaMartins: Yes, definitely not for this PR. I think it is fine as it is but if we really want to understand the reason for it we would have to look into argparse. I opened an issue for that #1005.

VictorBarbosaMartins:
Thanks again @GernotMaier. Will wait until the tests pass again to merge.

PR #1008, opened 2024-06-24
Step 1 of ruff fixes:  autofixes

GernotMaier:
This PR adds the "D" option (for docstring linter) to the ruff options and includes all automatic fixes (using the `--fixes` option).

I don't think there is anything to review...it essentially removes obsolete lines.


tobiaskleiner (approved):
Thanks @GernotMaier, I approve already since it is only removing lines.

PR #1013, opened 2024-06-24
Step 6 of ruff fixes: D100 no docstring in public module

GernotMaier:
Fixes undocumented public module: https://docs.astral.sh/ruff/rules/undocumented-public-module/


simtools/corsika/corsika_histograms.py
```diff
@@ -1,3 +1,5 @@
+"""Extracts Cherenkov photons information from a CORSIKA IACT file."""
```
tobiaskleiner, line 1: Extract

tobiaskleiner (approved):

PR #1017, opened 2024-06-24
Step 10 of ruff fixes: ruff list comprehension

GernotMaier:
Fixes list comprehension, see https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

(also removes some unnecessary settings from a previous PR)

Includes also autofixes changes to tests and docs/conf.py (sorry, was in the wrong branch to do that)

### DIFF
--- a/simtools/db/db_handler.py
+++ b/simtools/db/db_handler.py
@@ -1102,7 +1102,7 @@ class DatabaseHandler:
         if _cache_key not in DatabaseHandler.model_versions_cached:
             db_collection = DatabaseHandler.db_client[self._get_db_name()][collection]
             DatabaseHandler.model_versions_cached[_cache_key] = list(
-                set(post["version"] for post in db_collection.find(query))
+                {post["version"] for post in db_collection.find(query)}
             )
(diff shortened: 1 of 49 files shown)


simtools/db/db_handler.py
```diff
@@ -1102,7 +1102,7 @@ def get_all_versions(
         if _cache_key not in DatabaseHandler.model_versions_cached:
             db_collection = DatabaseHandler.db_client[self._get_db_name()][collection]
             DatabaseHandler.model_versions_cached[_cache_key] = list(
-                set(post["version"] for post in db_collection.find(query))
+                {post["version"] for post in db_collection.find(query)}
```
tobiaskleiner, line 1105: removal of set could lead to two identical entries here depending on the query?

tobiaskleiner (approved):

simtools/db/db_handler.py
GernotMaier: The two lines are identical. `{a for b in c}` is set comprehension (I didn't know that before ruff pointed it out).

PR #1019, opened 2024-06-24
Tidying documentation into sections

GernotMaier:
The number of files in docs/sources increased over the past months significantly (which is good!). But this makes it a bit hard to get an overview and one gets easily lost finding the correct documentation file.

This PR addresses this issue and introduces a typical organisation as found in most python projects (and also in e.g. ctapipe) with main sections for

- user guide
- API doc
- developer guide

No content has been changed. As a reviewer you have to agree/disagree with this reorganisation  and that's it.


VictorBarbosaMartins (approved):
Looks good. I tested also building the documentation and clicking on the links. There is no index inside the api-reference (not sure we need one though).

GernotMaier:
Wait, what about this one: https://github.com/gammasim/simtools/blob/tidying-documentation/docs/source/api-reference/index.md ?

PR #1026, opened 2024-06-27
Improve simulation configuration for corsika and simtel simulations.

### LINKED ISSUE #1005: Look deeper into argparse and Configurator
Passing a list of floats to an application does not work. The work around we have been using (see e.g. `calculate-trigger-rate` application) is to require a list of strings and then convert to a list of floats in the application. Although this works, we need to understand why a list of float does not work in the first place. 

https://github.com/gammasim/simtools/pull/1000#discussion_r1647715407

GernotMaier:
Improve simulation configuration for corsika and simtel simulations.

Added `simulation_configuration=["software", "corsika_configuration"]` to allow for the configuration of simulation software (e.g., corsika/simtel) or corsika configuration parameters (e.g. number of showers to be simulated).

Added `array_layout` model option to add especially `array_layout_name` to command line configuration.

Added functionality to the configuration model required to configure corsika parameters like `erange`, `viewcone`, `cscatter`:

1. for `erange`, `viewcone`: CommandLineParser.parse_quantity_pair requires a single string with a pair of quantities, e.g., "10 GeV 50 PeV" or "10GeV 50PeV" or "0deg 1.55deg"
2. for `cscatter`: CommandLineParser.parse_integer_and_quantity requires a single with with a pair consisting of an integer and a astropy.quantity (as needed for the core scatter keyword). This mean e.g., "5 1500m" or "5 1500 m".

Minor changes:

- simplified handling of azimuth angle configuration
- improved unit testing

Closes #1005

### DIFF
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -369,14 +369,14 @@ class CommandLineParser(argparse.ArgumentParser):
         )
         shower_config.add_argument(
             "--erange",
-            help="Energy range of the primary particle (min/max value).",
+            help="Energy range of the primary particle (min/max value, e'g', '10 GeV 5 TeV').",
             type=CommandLineParser.parse_quantity_pair,
             required=False,
             default=["3 GeV 330 TeV"],
         )
         shower_config.add_argument(
             "--viewcone",
-            help="Viewcone for primary arrival directions (min/max value in degrees).",
+            help="Viewcone for primary arrival directions (min/max value, e.g. '0 deg 5 deg').",
             type=CommandLineParser.parse_quantity_pair,
             required=False,
             default=["0 deg 0 deg"],
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -331,7 +331,7 @@ class CommandLineParser(argparse.ArgumentParser):
         )
         _configuration_group.add_argument(
             "--nshow",
-            help="Number of showers to simulate.",
+            help="Number of showers per run to simulate.",
             type=int,
             required=False,
         )
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -383,7 +383,7 @@ class CommandLineParser(argparse.ArgumentParser):
         )
         shower_config.add_argument(
             "--core_scatter",
-            help="Scatter area for shower cores (number of use; scatter radius).",
+            help="Scatter radiusfor shower cores (number of use; scatter radius).",
             type=CommandLineParser.parse_integer_and_quantity,
             required=False,
             default=["10 1400 m"],
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -656,7 +656,7 @@ class CommandLineParser(argparse.ArgumentParser):
         """
         Parse a string representing an integer and a quantity with units.
 
-        This si e.g., used for the 'core_scatter" argument.
+        This is e.g., used for the 'core_scatter' argument.
 
         Parameters
         ----------
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -383,7 +383,7 @@ class CommandLineParser(argparse.ArgumentParser):
         )
         shower_config.add_argument(
             "--core_scatter",
-            help="Scatter radiusfor shower cores (number of use; scatter radius).",
+            help="Scatter radius for shower cores (number of use; scatter radius).",
             type=CommandLineParser.parse_integer_and_quantity,
             required=False,
             default=["10 1400 m"],
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -2,6 +2,7 @@
 
 import argparse
 import logging
+import re
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


simtools/configuration/commandline_parser.py
```diff
@@ -231,26 +236,218 @@ def initialize_simulation_model_arguments(self, model_options):
...
+        _software_group.add_argument(
+            "--simulation_software",
+            help="Simulation software steps.",
+            type=str,
+            choices=["corsika", "simtel", "corsika_simtel"],
+            required=True,
+            default="corsika_simtel",
+        )
+
+    def _initialize_simulation_configuration(self):
```
VictorBarbosaMartins, line 296: These arguments are all very specific to the simulation applications. By putting them here we do not have an argument anymore in the future against putting all the other parameters such as the magnetic field strength, particle interaction model,... Maybe this is ok for now, but I think we should be more flexible in the sense that we should not need to list all possible parameters manually in the code.

VictorBarbosaMartins: Perhaps a better solution would be to use the DB and check whether a passed parameter exists and use it in case positive?

simtools/configuration/commandline_parser.py
```diff
@@ -231,26 +236,218 @@ def initialize_simulation_model_arguments(self, model_options):
...
+        )
+        _configuration_group.add_argument(
+            "--number_of_runs",
+            help="Number of runs to be simulated.",
+            type=int,
+            required=True,
+            default=1,
+        )
+        _configuration_group.add_argument(
+            "--event_number_first_shower",
```
VictorBarbosaMartins, line 353: first shower event number?

simtools/configuration/commandline_parser.py
```diff
@@ -231,26 +236,218 @@ def initialize_simulation_model_arguments(self, model_options):
...
+        shower_config.add_argument(
+            "--erange",
+            help="Energy range of the primary particle (min/max value, e'g', '10 GeV 5 TeV').",
+            type=CommandLineParser.parse_quantity_pair,
+            required=False,
+            default=["3 GeV 330 TeV"],
+        )
+        shower_config.add_argument(
+            "--viewcone",
+            help="Viewcone for primary arrival directions (min/max value, e.g. '0 deg 5 deg').",
```
VictorBarbosaMartins, line 379: Perhaps, as before, we should replace Viewcone by "opening angle" or "inner opening angle" and "outer opening angle"

simtools/configuration/commandline_parser.py
```diff
@@ -231,26 +236,218 @@ def initialize_simulation_model_arguments(self, model_options):
...
+            type=CommandLineParser.parse_quantity_pair,
+            required=False,
+            default=["0 deg 0 deg"],
+        )
+        shower_config.add_argument(
+            "--core_scatter",
+            help="Scatter area for shower cores (number of use; scatter radius).",
```
VictorBarbosaMartins, line 386: We pass the distance, not the area. So I would call it "X and Y distances that define the scatter area. Scatter radius would also be wrong, because the distance in each direction might be different.

simtools/configuration/commandline_parser.py
```diff
@@ -231,26 +236,218 @@ def initialize_simulation_model_arguments(self, model_options):
...
+        )
+        shower_config.add_argument(
+            "--viewcone",
+            help="Viewcone for primary arrival directions (min/max value, e.g. '0 deg 5 deg').",
+            type=CommandLineParser.parse_quantity_pair,
+            required=False,
+            default=["0 deg 0 deg"],
+        )
+        shower_config.add_argument(
+            "--core_scatter",
```
VictorBarbosaMartins, line 385: I think "scatter_distances" would be a better name as "core_scatter" does not implicitly define a quantity.

VictorBarbosaMartins (approved):
Hi @GernotMaier, thanks for these changes. I have just one major comment but this can also be dealt with in the future. the remaining comments are minor.

simtools/configuration/commandline_parser.py
GernotMaier: We separate those parameters which might 

1. change per production (e.g., the energy range and the number of showers)
2.  those which do not change (e.g., the hadronic interaction model or geomagnetic field value)

1. is configured via the command line and 2. is part of the simulation model (to be added; add this point in data data/parameters/corsika_parameters.yml

I do think this makes sense.

simtools/configuration/commandline_parser.py
GernotMaier: I prefer it this way, as the important point is the event number.

simtools/configuration/commandline_parser.py
GernotMaier: Correct, fixed.

simtools/configuration/commandline_parser.py
GernotMaier: I think this would not be entirely correct: core_scatter is both the re-use number for shower and the scatter distance, so we can't call it scatter distance only.

GernotMaier:
Thanks for the review!

simtools/configuration/commandline_parser.py
VictorBarbosaMartins: I had in mind the docstring (not the parameter name). but ok, this should be clear for experts.

PR #1027, opened 2024-06-27
Use quantity pairs in calculate_trigger_rate.py

GernotMaier:
Use the `CommandLineParser.parse_quantity_pair` for the configuration of viewcone and energy range in calculate_trigger_rate.py.


VictorBarbosaMartins (approved):
All good here, thanks.

PR #1028, opened 2024-07-01
Exclude main from coverage

GernotMaier:
Exclude `__main__` from coverage calculation.


GernotMaier:
(apologies for merging without review - would like to test this now to understand sonar)

PR #1029, opened 2024-07-01
Apply simplifications and fixes to command line configuration.

GernotMaier:
PR #1026 introduced a couple of improvements for the command line configuration. This PR applies these improvements to a couple of applications (especially those required to configure layouts).

This PR also fixes a bug in the core scatter configuration of type "<int> <astropy quantitiy>" (e.g., `5 1400 m`). The command line configuration needs to understand also the tuble "(5, <Quantity 1500 m>)", as this might be set such one of the configuration steps (remember that the configurator steps through command line, configuration file, environment variables).

Note that I have also removed the unecessary functionality in plot_array to loop over a list of arrays for plotting.


simtools/configuration/commandline_parser.py
```diff
@@ -612,18 +612,20 @@ def azimuth_angle(angle):
...
+                "north": 0 * u.deg,
+                "south": 180 * u.deg,
+                "east": 90 * u.deg,
+                "west": 270 * u.deg,
+            }
+            azimuth_angle = angle.lower()
+            if azimuth_angle in azimuth_map:
+                return azimuth_map[azimuth_angle]
+            raise argparse.ArgumentTypeError(
+                "The azimuth angle can only be one of " f"(north, south, east, west), not {angle}"
```
tobiaskleiner, line 626: Is this clear here? We allow strings and floats in this function, the last check here is separate from above. Maybe better to have a seperate function or Change the message to something as "If the angle is pased as a string only ... is allowed and not ..".

tobiaskleiner (commented):
Thanks @GernotMaier. One suggestion, otherwise all good.

simtools/configuration/commandline_parser.py
GernotMaier: Good point - I've improved the error message.

tobiaskleiner (approved):

PR #1043, opened 2024-07-09
Fix bug in validating dimensionless values

GernotMaier:
This PR fixes a bug in validating values without dimensions. The following case was accepted but should raise an Exception:

- input data is of unit `dimensionless`
- requested unit is not `dimensionless`

Added a unit test (in `tests/unit_tests/data_model/test_validate_data.py::test_check_and_convert_units_dimensionless`) to test this.


simtools/data_model/validate_data.py
```diff
@@ -402,6 +402,23 @@ def _check_for_not_a_number(self, data, col_name):
 
         return False
 
+    @staticmethod
+    def _is_dimensionless(unit):
+        """
+        Check is unit is dimensionless, None, or empty.
```
tobiaskleiner, line 408: if

tobiaskleiner (approved):

simtools/data_model/validate_data.py
GernotMaier: Fixed, thanks!

PR #1046, opened 2024-07-09
Minor bug fix to allow to run db_add_model_parameters_from_repository_to_db with SIMTOOLS_DB_SIMULATION_MODEL defined

GernotMaier:
The application `db_add_model_parameters_from_repository_to_db` uploads model parameters from a repository to the DB into a new DB with a name given on the command line `--db_name`.

Noticed that this application is not working when the environmental variable `SIMTOOLS_DB_SIMULATION_MODEL` is not defined. This is obviously dangerous, as we have to be sure that only data from the repository and not from another DB is uploaded.

This PR fixes this and `db_add_model_parameters_from_repository_to_db` is working now without `SIMTOOLS_DB_SIMULATION_MODEL` defined.

### DIFF
--- a/simtools/db/db_handler.py
+++ b/simtools/db/db_handler.py
@@ -911,7 +911,9 @@ class DatabaseHandler:
         kwargs.pop("type", None)
         db_entry.update(kwargs)
 
-        self._logger.info(f"Will add the following entry to DB:\n{db_entry}")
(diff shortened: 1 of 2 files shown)


tobiaskleiner (approved):
Thanks for the fix @GernotMaier. Looks good.

Issue #981, opened 2024-06-13
Centroid_x and centroid_y for effective focal length calculation


tobiaskleiner:
In the current implementation in `ray_tracing` only `centroid_x` is used for the calculation of the effective focal length. This should be updated to correctly calculate it depending on the position of the simulated light source in the field of view.

tobiaskleiner:
In #1047 we have removed the unused centroid_y variable (temporary) until we generalize it further.

PR #1047, opened 2024-07-09
Remaining sonar issues

GernotMaier:
Removed one more issue reported by sonar.

All others are todo in related to the light emission package. @tobiaskleiner - you will have to judge if we want to leave them (we can).


tobiaskleiner:
I have adressed the remaining issues in  #1052. 

This one here is probably related to #981. We can remove it here until we adress the issue. (Added a comment there).

tobiaskleiner (approved):

PR #1053, opened 2024-07-10
Fix in database documentation; improve upload from repository

GernotMaier:
Two small items:

- minor improvements in the database user's guide
- improvement of script to setup a local DB for testing. The environment variables (`.env`) are now set automatically in the script.

(we see how and if we develop those scripts further; for now, they are very useful for developers when adding / removing things to the DB).


tobiaskleiner (approved):
All good.

PR #1058, opened 2024-07-11
Data field in schema file should be a list

GernotMaier:
This fixes a bug in the schema file definition.

Absolutely puzzling that the CI ([here](https://github.com/gammasim/simtools/actions/runs/9890868431/job/27320057258)) did run successful. 

I tried to reproduce this locally, without success. Very very weird.


tobiaskleiner (approved):

tobiaskleiner:
Indeed very strange that this was not catched.

PR #1073, opened 2024-07-15
Flexible command line

GernotMaier:
This introduces a more flexible configuration of the command line parameters implement on the example of the option `simulation_configuration`.  Motivation is that there are applications which require all command line parameters are set, while others only require a submit. Guidance here is:

- avoid duplication (e.g., currently we have 'zenith_angle' defined several times as command line parameter) 
- keep parameter groups the same (even if the set of parameter per group is sometimes not complete).

The PR allows for these options (examples):

- use all parameters as in simulate_prod: `simulation_configuration={"software": None, "corsika_configuration": ["all"]},`
- use a subset of parameters, as in calculate_trigger_rate: `simulation_configuration={"corsika_configuration": ["energy_range", "view_cone"]},`

(both are parameters in the `config.initialize` call.)

Noticed and fixed inconsistencies (e.g. energy range was erange and energy_range, view cone was viewcone and view_cone)


simtools/configuration/commandline_parser.py
```diff
@@ -295,102 +303,126 @@ def _initialize_simulation_software(self):
...
-            "--event_number_first_shower",
-            help="Event number of first shower",
-            type=int,
-            required=False,
-            default=1,
-        )
+    @staticmethod
+    def _get_dictionary_with_corsika_configuration():
+        """Return dictionary with CORSIKA configuration parameters."""
+        return {
```
tobiaskleiner, line 309: I would suggest to keep the previous way of defining the dictionary to have this consistent in the code. Defining a dictionary manually with all the entries here seems more error prone.

tobiaskleiner (commented):
Thanks @GernotMaier, I have added a comment.

simtools/configuration/commandline_parser.py
GernotMaier: I have introduced the dictionaries to allow for the simple configuration of all parameters with the `all` flag - then we simply loop over all parameters and add all of them to argpars. If we don't need all, we only configure those given in the list of input parameters.

I think this solution avoids a lot of if statements. Do you have another suggestion on how to do this?

tobiaskleiner: Wouldn't it be possible to get all added parameters from `_configuration_group`?

GernotMaier: Not sure I get it - `configuration_group = self.add_argument_group(group_name)` is when it is defined an empty group. We have to add arguments depending on the request in the function call.

tobiaskleiner: Yes, so the same what we did previously with `_configuration_group.add_argument(`. After this can we access the added parameters to this group?

tobiaskleiner: After talking, we decided now to keep it this way.

tobiaskleiner (approved):

GernotMaier:
Thanks a lot!

PR #1081, opened 2024-07-18
Improve unit tests for model_parameters; add read parameters from file

GernotMaier:
Add the following related to the model_parameters module:

- allow to change model_parameters read from a file. This is clearly an experimental feature (for now) and a corresponding warning is printed out. I would keep it as it is for now, until we find time to either replace or improve it (this functionality has been there before, see the replace in `validate_cumulative_psf.py`
- improve unit tests for model_parameters and get coverage to 100%.


tobiaskleiner (approved):
All good, thanks. Do we need to open an issue to remove this in the future or to add additional validation steps?

GernotMaier:
Thanks for the review! 

I've added a issue #1084

Issue #1054, opened 2024-07-11
Use more mocking and less the database in the unit tests


GernotMaier:
A majority of the unit tests depend on a database connections. This was a conscious decision, as we said that the DB is an integral part of simtools.

I think we learnt in the meanwhile that we can replace the DB connections in most cases using the unittest/pytest mocking mechanism. We should apply this wherever possible.

Given the large number of unittests, implementing this for all existing tests is a huge task.

Suggest to:

- use mocking when implementing new tests
- change existing tests whenever we touch the test code

An ideal solution would be to have no / very few DB connections in unit tests. Those which require a connection should get a pytest label.

GernotMaier:
I suggest to close this issue - this is a standing task and not a single issue and I think (hope) we are all aware of it.

PR #1516 adds a section on mocking to the developer documentation.

orelgueta:
Fine with me to close this.

PR #1082, opened 2024-07-18
Improve unit test execution time

GernotMaier:
This PR does not add or remove any tests or any functionality. It addresses the issue "Use more mocking and less the database in the unit tests" #1054 and introduces more mocking to avoid time consuming calls to the DB.

Given the large number of database calls, this does not replace all but is a start to address this. 

Strategy was to run pytest with `--durations 10` and always look at a slow one which appeared and easy target for mocking.


tobiaskleiner (approved):
Thanks @GernotMaier, good replacements with mock instances.

PR #1085, opened 2024-07-22
Bug fixes related to  pytest-raises-with-multiple-statements (PT012)

GernotMaier:
This is actually interesting...every correction in the PR is a bug fix of a unit test!

We did very often:

```
with pytest.raises(some error):
    call_a_a_function_raising_some_error()
    assert statement
```

This is wrong - the assert is never called when doing this, it should be:

```
with pytest.raises(some error):
    call_a_a_function_raising_some_error()
assert statement
```

 So flake8-pytest-style/ruff [PT012](https://docs.astral.sh/ruff/rules/pytest-raises-with-multiple-statements/) is actually a very useful test.

This PR corrects all PT012 and fixed 2-3 unit tests (mostly wrong tests on the error messages). PT012 is now also switched on in pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,7 +145,6 @@ lint.ignore = [
   "G004",   # Logging statement uses f-string
   "ISC001", # incompatible with ruff format
   "PT011",  # TODO - fix
-  "PT012",  # TODO - fix
   "RUF012", #  Mutable class attributes should be annotated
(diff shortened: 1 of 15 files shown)


tobiaskleiner (approved):
@GernotMaier good fix, thanks!

PR #1086, opened 2024-07-23
Addresses errors of type "pytest-raises-too-broad" from ruff

GernotMaier:
Addresses "pytest-raises-too-broad" (PT011) error from ruff, see https://docs.astral.sh/ruff/rules/pytest-raises-too-broad/

Some smaller bug fixes are included learned from switching this on: in some cases the same error was thrown at a different then expected place in the code.

Note the discussion in issue #1087 : this PR implements the common way of issuing error messages (e.g. ValueError(msg)).


tobiaskleiner (approved):
Thanks @GernotMaier, looks like a good update.

GernotMaier:
@tobiaskleiner - thanks for the review!

PR #1095, opened 2024-07-25
Use array_element instead of telescope in names, model_parameter, db_handler

### LINKED ISSUE #871: Generalize db_handler
The db_handler contains around 140 entries `telescope`. Going forward to include further array elements in the DB (as for instance `calibration_devices`, we should generalize some methods. This might require just a renaming in most cases.

GernotMaier:
Mostly a scripted name change from telescope to array_element in the modules mention above. This generalizes the functions and names to take into account the we simulate telescopes and calibration devices.

Closes #871

### DIFF
--- a/simtools/utils/names.py
+++ b/simtools/utils/names.py
@@ -11,16 +11,16 @@ import yaml
 _logger = logging.getLogger(__name__)
 
 __all__ = [
-    "get_site_from_telescope_name",
-    "get_telescope_type_from_telescope_name",
+    "get_site_from_array_element_name",
+    "get_array_element_type_from_name",
     "generate_file_name",
     "layout_telescope_list_file_name",
     "sanitize_name",
     "simtel_single_mirror_list_file_name",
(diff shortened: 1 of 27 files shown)


GernotMaier:
One thing to note regarding this pull request: we are still using two different names for the same thing: `instrument` and `array_element`. `instrument is used mostly in database entries.

(I am not suggesting to change this, just a remark)

simtools/utils/names.py
```diff
@@ -167,14 +167,14 @@ def _validate_name(name, all_names):
...
+def validate_array_element_name(name):
     """
-    Validate telescope name (e.g., MSTN-design, MSTN-01).
+    Validate array element name (e.g., MSTN-design, MSTN-01).
 
     Parameters
     ----------
     name: str
-        Telescope name.
+        array element name.
```
tobiaskleiner, line 177: Array element

tobiaskleiner (approved):
All good.

PR #1102, opened 2024-08-27
Remove obsolete db development scripts

GernotMaier:
Remove scripts used to manipulate database entries. This is not needed anymore, as we now update first the model parameter repository and then copy these values to the database.


tobiaskleiner (approved):
Good to remove this.

Issue #1070, opened 2024-07-15
Review and replace derived value DB


GernotMaier:
Review derived values DB and merge these files if necessary into the general model. Goal should be to make the derived values DB obsolete.

This should also allow to remove the code related to derived values in the db_handler and model_parameter modules.

GernotMaier:
Note that the version entry in the derived database is out of date (not semantic, old dates). Not fixing this, as we anyway want to find a different solution.

PR #1103, opened 2024-08-27
Require semantic version for model parameter version string

GernotMaier:
Model versions have changed to semantic versioning with this [model_parameters PR](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/merge_requests/50).

Add here:

- a test to validate_data that the version string follows semantic versioning.
- change of "2024-02-01" to prod6 and "2020-06-28" to prod5
- run `validate_camera_efficiency_ssts` test for prod6 only (as Derived-DB is not updated with the new versioning; see issue #1070)


simtools/data_model/validate_data.py
```diff
@@ -708,3 +710,25 @@ def _prepare_model_parameter(self):
...
+
+        Raises
+        ------
+        ValueError
+            if version string does not follow semantic versioning
+
+        """
+        if version is None:
+            return
+        semver_regex = r"^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$"
```
tobiaskleiner, line 731: Not perfectly familiar with regex. Does this here allow also number.number.number-(additional expression)?

tests/unit_tests/data_model/test_validate_data.py
```diff
@@ -695,3 +696,41 @@ def test_prepare_model_parameter():
...
+
+
+def test_check_version_string(caplog):
+    data_validator = validate_data.DataValidator()
+
+    valid_versions = [
+        "1.0.0",
+        "0.1.0",
+        "2.3.4",
+        "1.0.0-alpha",
```
tobiaskleiner, line 708: Related to above question. So we really want to allow this kind of version structure?

GernotMaier: This is how semantic versioning is defined, see https://semver.org/

tobiaskleiner (commented):
I put this comment here that this would also accept pre release versions.

tests/unit_tests/data_model/test_validate_data.py
tobiaskleiner: Ok all good then.

tobiaskleiner (approved):

simtools/data_model/validate_data.py
GernotMaier: Yes - but see also #1115

PR #1113, opened 2024-08-28
Bunch size should be an integer

orelgueta:


GernotMaier (changes requested):
Following the definitions from the CORSIKA manual, and there it is `(A6, F)`. I think it is a float, although we always use integer values.

Screenshot 2024-08-29 at 09 23 17

orelgueta:
You are right, it does accept a float. Not entirely sure why and reading the manual doesn't make it clearer. So even if we always use an integer, I don't see a reason to limit it to integers on our side. Therefore I am closing this PR.

PR #1118, opened 2024-08-29
Improve speed of validate file using schema

### LINKED ISSUE #1104: validate_file_using_schema.py is incredibly slow
Stunning how slow the CI on gitlab is: e.g. https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/jobs/216303 

This loops over all json files in the model (2140 files) and it takes 30-50 min!!  If would be 1 min using bash, this needs investigation.

orelgueta:
First commit that improves things (~20% faster), will continue studying.
- Import particle package only if necessary since apparently it is somewhat slow

Fixes #1104

### DIFF
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -8,7 +8,6 @@ from pathlib import Path
 import astropy.units as u
 
 import simtools.version
-from simtools.corsika.primary_particle import PrimaryParticle
 from simtools.utils import names
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


orelgueta:
I assume you profiled the application just like I did.

I was also planning to change the application to allow it to accept a list of files/schema files as well. That will probably reduce the run time by quite a bit. However, If you are already working on it, then I will leave it to you. You are more familiar with this application than I am.

GernotMaier (approved):

orelgueta:
This was marked as a draft...

PR #1119, opened 2024-08-30
Improve import efficiency of general.py 

GernotMaier:
Startup times for all tools are slow due to slow imports. In all cases the import of astropy modules dominates the import times (in most cases this is >80% of the imports).

Improve the import efficiency of general.py by moving all astropy related functions into a new module called `simtools.utils.value_conversion`. These functions are used less often and it is reasonable to have them not in general.py

No code has been changed (apart from a unlogical error type) and no functionality added but moved only.

These changes improve the import efficiency by 25% (I was hoping for more...).


tobiaskleiner (commented):
Thanks @GernotMaier, this is a good improvement. However I am not sure if this is the way to go if we really want to optimize this to have a much better performance. I am including @orelgueta here, as he had a suggestion to change the way we use `validate_file_using_schema.py` to accept also lists of schema files. Then the loading would be only once and the tests likely would be much faster.

GernotMaier:
@tobiaskleiner - this is independent of validate_file_using_schema (the acceptance of lists will be added there).

This is a general improvement which we can feel at many places, as we use general.py quite often.

tobiaskleiner (approved):

PR #1121, opened 2024-08-30
Ignore pre-releases of the model parameter database

### LINKED ISSUE #1115: SIMTOOLS_DB_SIMULATION_MODEL="..LATEST" should not select pre-releases
Setting `SIMTOOLS_DB_SIMULATION_MODEL='CTAO-Simulation-Model-LATEST'` chooses automatically the most recent DB version defined by the largest version number.

Do not user pre-releases indicated by e.g. "5.0.0-alpha.1" in this case. 

(might motivate the use of a dedicated package like semver instead of regex expressions)

GernotMaier:
Minor changes to skip pre-releases when selecting the latest version of the model parameter database.

Using the definition from semantic versioning, where pre-releases are added as additional string (e.g., 5.0.0-alpha.1).

Closes #1115

### DIFF
--- a/simtools/db/db_handler.py
+++ b/simtools/db/db_handler.py
@@ -108,10 +108,11 @@ class DatabaseHandler:
 
     def _update_db_simulation_model(self):
         """
-        Find the latest version (if requested) of the simulation model and update the DB config.
+        Find the latest released version of the simulation model and update the DB config.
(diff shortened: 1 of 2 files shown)


tobiaskleiner (approved):
All good.

PR #1122, opened 2024-09-02
Bug fix in reading model versions

GernotMaier:
Fix a bug noticed while running `simtools-db-add-model-parameters-from-repository-to-db` in the CI of the gitlab model_parameter repository (see [here](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/jobs/218577)).

Issue:

- `db_handler::_get_all_versions()` requires the existence of no collections or at least the telescope collection to be available when reading the model versions from the DB
- gitlab CI run starts by adding a file to the collection `calibration_devices`. This leads to a failure when trying to add a second file.

Solution:

`db_handler::_get_all_versions()` default collection is now 'None', in which case we query all collections to get the list of available model parameter versions.

This is a solution for now and probably works only as long as we have all model parameters for each model parameter version in the repository / DB (in future, we probably have only changes relative to a baseline model).

Unrelated minor change: improved naming by renaming `update_db_simulation_model`. to `_find_latest_simulation_model_db`.

### DIFF
--- a/simtools/db/db_handler.py
+++ b/simtools/db/db_handler.py
@@ -1193,7 +1193,11 @@ class DatabaseHandler:
             If key to collection_name is not valid.
 
         """
-        _cache_key = f"model_versions_{self._get_db_name()}-{collection}"
+        db_name = self._get_db_name()
+        if not db_name:
+            self._logger.warning("No database name defined to determine list of model versions")
+            return []
+        _cache_key = f"model_versions_{db_name}-{collection}"
 
         query = {}
         if parameter is not None:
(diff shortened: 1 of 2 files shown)


simtools/db/db_handler.py
```diff
@@ -1208,10 +1208,19 @@ def get_all_versions(
...
+                        db_collection = DatabaseHandler.db_client[self._get_db_name()][
+                            collection_name
+                        ]
+                        all_versions.update(post["version"] for post in db_collection.find(query))
+                    DatabaseHandler.model_versions_cached[_cache_key] = list(all_versions)
+                else:
+                    db_collection = DatabaseHandler.db_client[self._get_db_name()][collection]
+                    DatabaseHandler.model_versions_cached[_cache_key] = list(
+                        {post["version"] for post in db_collection.find(query)}
+                    )
```
orelgueta, line 1223: These lines became a bit hard to read, but unfortunately I cannot come up with an idea on how to improve it. Perhaps `_db_name = self._get_db_name()` and then using `_db_name` everywhere would at least make the lines shorter, but it doesn't help with the series of ifs and for loops.

simtools/db/db_handler.py
```diff
@@ -1420,11 +1429,19 @@ def get_collections(self, db_name=None):
         -------
         list
             List of collection names
+        model_collections_only: bool
+            If True, only return model collections (i.e. exclude fs.files, fs.chunks)
```
orelgueta, line 1433: please include "metadata" in this example

tests/unit_tests/db/test_db_handler.py
```diff
@@ -613,6 +622,12 @@ def test_get_collections(db, db_config):
     collections_from_name = db.get_collections(db_config["db_simulation_model"])
     assert isinstance(collections_from_name, list)
     assert "telescopes" in collections_from_name
+    assert "fs.files" in collections_from_name
+
+    collections_no_model = db.get_collections(db_config["db_simulation_model"], True)
+    assert isinstance(collections_no_model, list)
+    assert "telescopes" in collections_no_model
+    assert "fs.files" not in collections_no_model
```
orelgueta, line 630: test also for metadata

orelgueta (approved):
very minor comments. approving in the meantime.

simtools/db/db_handler.py
GernotMaier: Had another look and simplified it significantly.

simtools/db/db_handler.py
GernotMaier: Done.

tests/unit_tests/db/test_db_handler.py
GernotMaier: I consciously didn't do that, as the open PR #1123 is removing the metadata collection. But as we might want to use old DBs, it maybe good to test it. Therefore followed your suggestion.

GernotMaier:
Thanks for the review - I've followed all your suggestion. Will wait for the tests before merging.

simtools/db/db_handler.py
orelgueta: Looks much better thanks!

I am a bit surprised none of the linters complained about `if collection`. I thought they require to check explicitly `if collection is not None`. I don't mind it personally, so if it passes, it's fine.

PR #1123, opened 2024-09-02
Move to strictly using semantic versions for models (remove e.g., 'prod5', 'prod6', 'Latest', 'Released')

GernotMaier:
Move to strictly using semantic versions for models - e.g., 5.0.0 or 6.0.0. This makes the code clearer and it is less confusing to use one (and only one) way to name model parameter versions.

This PR:

- removes the metadata collection in the database (not needed anymore; this was used to map e.g., "Latest" to "prod5")
- removed references to 'prod5', 'prod6', 'Latest', 'Released' in the code
- try to simplify future changes of model versions in the unit tests by strictly using a pytest.fixture `model_version` and `model_version_prod5` (we test also with prod5, as some of the parameters are different or only available in prod5)
- simplify the naming of the sandbox DBs by adding the random id to the DB name only and not additionally to the collections


GernotMaier:
I hope we can allow for -0.19% of coverage (which is probably due to additionally line breaks)

tobiaskleiner (approved):

tobiaskleiner:
> I hope we can allow for -0.19% of coverage (which is probably due to additionally line breaks)

Yes this should be ok.

Issue #1125, opened 2024-09-03
Invalid CORSIKA7 ID 6 when trying to create muon


maxnoe:
```
>>> PrimaryParticle(corsika7_id=6)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workdir/env/lib64/python3.11/site-packages/simtools/corsika/primary_particle.py", line 34, in __init__
    self.corsika7_id = corsika7_id
    ^^^^^^^^^^^^^^^^
  File "/workdir/env/lib64/python3.11/site-packages/simtools/corsika/primary_particle.py", line 59, in corsika7_id
    raise ValueError(f"Invalid CORSIKA7 ID: {value}")
ValueError: Invalid CORSIKA7 ID: 6
```

Issue #1126, opened 2024-09-03
PrimaryParticle doesn't validate consistency of arguments


maxnoe:
```
>>> PrimaryParticle(corsika7_id=5, pdg_id=-12)
<simtools.corsika.primary_particle.PrimaryParticle object at 0x7fac3564cb10>
```

the correct pdg_id for the mu+ is -13.

GernotMaier:
This is a bad API design - the intention is that you declare PrimaryParticle with either the corsika_id or the pdg_id or the name. I will change this.

maxnoe:
I think it is in general bad design to have a constructor that takes mutually exclusive arguments.

A better pattern would be to have the constructor take only one argument and have alternative constructors for the other options, e.g.:

```
PARTICLE_NAMES_TO_PDGID = {
    "gamma": Particle.from_name("gamma").pdgid,
    "proton": Particle.from_nucleus_info(a=1, z=1).pdgid,
...
}


class PrimaryParticle:
    def __init__(self, pdg_id):
        self.particle = Particle.from_pdgid(pdg_id)

   @classmethod
   def from_corsika7_id(cls, corsika7_id):
        pdg_id = Corsika7ID(corsika7_id).to_pdgid()
        return cls(pdg_id)

    def from_name(cls, name):
        ...
```

PR #1127, opened 2024-09-03
Improve primary particle API

GernotMaier:
Address two issues regarding the PrimaryParticle class:

1. PrimaryParticle should be declared with either the corsika_id or the pdg_id or the name. This is the intended use case, but the constructor allowed to define mismatching particle IDs (see #1126 ).
2. PrimaryParticle relies the the particle actually appears in a dict (`particle_names`). This is required as we want to allow for the somewhat different naming in CTA compared to PDG (e.g., a proton is a `proton` in the CTA simulation language but a `p` in PDG). Changed the functionality now that particles which are not in the dict use the name provided from PDG.
3. Fixed issue #1125 and allow to create `muon-` and `muon+`.

@maxnoe suggested a slightly different implementation of 1. The reason for choosing `PrimaryParticle(particle_id_type, particle_id)` is that this allows a clean configuration of this class with no if statements in corsika_config.

### DIFF
--- a/simtools/corsika/corsika_config.py
+++ b/simtools/corsika/corsika_config.py
@@ -328,7 +328,6 @@ class CorsikaConfig:
         """
         if not args_dict or args_dict.get("primary_id_type") is None:
             return PrimaryParticle()
-        print("FFFF", args_dict.get("primary_id_type"), args_dict["primary"])
         return PrimaryParticle(
             particle_id_type=args_dict.get("primary_id_type"), particle_id=args_dict.get("primary")
         )
--- a/simtools/corsika/corsika_config.py
+++ b/simtools/corsika/corsika_config.py
@@ -326,15 +326,12 @@ class CorsikaConfig:
             Primary particle.
 
         """
-        if not args_dict:
+        if not args_dict or args_dict.get("primary_id_type") is None:
             return PrimaryParticle()
-        if args_dict.get("primary_id_type") == "common_name":
-            return PrimaryParticle(name=args_dict.get("primary"))
-        if args_dict.get("primary_id_type") == "corsika7_id":
-            return PrimaryParticle(corsika7_id=int(args_dict.get("primary")))
-        if args_dict.get("primary_id_type") == "pdg_id":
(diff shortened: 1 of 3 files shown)


simtools/corsika/primary_particle.py
```diff
@@ -14,28 +14,33 @@ class PrimaryParticle:
...
         self._pdg_id = None
         self._pdg_name = None
 
-        if corsika7_id is not None:
-            self.corsika7_id = corsika7_id
-        if name is not None:
-            self.name = name
-        if pdg_id is not None:
-            self.pdg_id = pdg_id
+        if bool(particle_id_type) != bool(particle_id):
```
maxnoe, line 31: Why not just make both arguments required?

simtools/corsika/corsika_config.py
```diff
@@ -326,15 +326,11 @@ def _set_primary_particle(self, args_dict):
             Primary particle.
 
         """
-        if not args_dict:
+        if not args_dict or args_dict.get("primary_id_type") is None:
             return PrimaryParticle()
```
maxnoe, line 330: This just raises an error, right?

GernotMaier: No - `bool(None)` is `False`, this simply sets all variables to None in the constructor.

simtools/corsika/primary_particle.py
GernotMaier: We are using `PrimaryParticle()` without any arguments in the code (although at least for the corsika configuration we need to have this value set).

maxnoe: I'd probably would prefer using `None` instead of an instance of particle that contains `None`s to represent "no particle configured".

GernotMaier: Yes, I understand that motivation. However, as it is used right now we rely on that a PrimaryParticle object is defined. So changing this year, would mean to introduce if statements to check for None. Would prefer for now to leave it as it is to allow to have a 'default' PrimaryParticle object.

orelgueta (approved):

PR #1134, opened 2024-09-11
Run upload_from_model_repo from within the container

orelgueta:
Added a note to remind the user (me...) to run the script from within the container.

### DIFF
--- a/docs/source/user-guide/databases.md
+++ b/docs/source/user-guide/databases.md
@@ -114,6 +114,8 @@ uploads its contents to the local database instance.
 
 Note that repository branches are hardcoded in the scripts and need to be adjusted accordingly.
 
+**This should be run from within the simtools container**
+
 ### Purge the local database instance and all networks, images, containers
 
 The script `purge_local_db.sh` stops and removes the container and deletes all networks, images, and containers.
--- a/docs/source/user-guide/databases.md
+++ b/docs/source/user-guide/databases.md
@@ -114,7 +114,7 @@ uploads its contents to the local database instance.
 
 Note that repository branches are hardcoded in the scripts and need to be adjusted accordingly.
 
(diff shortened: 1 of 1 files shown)


GernotMaier (changes requested):
Hmm, I don't need to do this within a container and can run the upload script in my conda environment. Are you sure about this?

orelgueta:
When I do it from outside the container (with the conda environment activated), the connection times out. Perhaps there's something I am missing, but I spent a good 20 minutes yesterday trying to figure it out.

Either way, I have done it so many times now that I probably won't forget anymore, so this PR can be skipped if you wish.

GernotMaier:
I see the issue now - you need to have a `.env` which points to the local DB. It should look like:

```
SIMTOOLS_DB_API_PORT=27017 #Port on the MongoDB server
SIMTOOLS_DB_SERVER='localhost'
SIMTOOLS_DB_API_USER='api' # username for MongoDB
SIMTOOLS_DB_API_PW='password' # Password for MongoDB
SIMTOOLS_DB_API_AUTHENTICATION_DATABASE='admin'
SIMTOOLS_DB_SIMULATION_MODEL='CTAO-Simulation-Model-LATEST'
```

I guess this is something we should add to the documentation.

orelgueta:
I actually tried that and it still didn't work. Maybe I did that wrong though. 

Either way, I agree we should figure out what's the easiest way where it always works and add it to the documentation. Running from within the container is easy, but running from outside the full list chain of
- ./purge_local_db.sh
- ./setup_local_db.sh
- ./upload_from_model_repository_to_db.sh CTAO-Simulation-Model-v0-10-0

is probably easier/clearer.

GernotMaier (approved):

PR #1135, opened 2024-09-16
Test for two new lines at end of sim_telarray configuration file writing

### LINKED ISSUE #455: Make sure sim_telarray configuration files end with a newline
Actually a `\n\n` (an empty line plus newline). We should add a unit test for this as well.
sim_telarray would complain from now on, but with older versions it doesn't and simply ignores that last line which can be a real issue sometimes (c.f. divergent pointing production...)

GernotMaier:
Addresses #455 :

- added a comment into to code that this is required at the end of a configuration file
- added unittest to test that the written simtel configuration file ends with `\n\n`  

Closes #455

### DIFF
--- a/tests/unit_tests/simtel/test_simtel_config_writer.py
+++ b/tests/unit_tests/simtel/test_simtel_config_writer.py
@@ -41,7 +41,7 @@ def test_write_array_config_file(
     # simtel configuration files need to end with two new lines
     with open(file) as f:
         lines = f.readlines()
-        assert lines[-2].endswith("#endif \n")
+        assert lines[-2].endswith("\n")
         assert lines[-1] == "\n"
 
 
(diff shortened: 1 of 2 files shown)


tests/unit_tests/simtel/test_simtel_config_writer.py
```diff
@@ -38,6 +38,12 @@ def test_write_array_config_file(
     )
     assert file_has_text(file, "TELESCOPE == 1")
 
+    # simtel configuration files need to end with two new lines
+    with open(file) as f:
+        lines = f.readlines()
+        assert lines[-2].endswith("#endif \n")
```
orelgueta: I would actually just check for `\n` instead of the exact `#endif \n` because in the future we might not have/need the `#endif`. 
I added a commit to that effect, if you agree, please merge, otherwise revert and merge.

orelgueta (approved):
Thanks for this, see minor comment.

tests/unit_tests/simtel/test_simtel_config_writer.py
GernotMaier: Good, thanks!

PR #1138, opened 2024-09-18
Introduce simtools testing modules; move tests from integration tests into this module

### LINKED ISSUE #970: Added functionality to integration test: move those function to a separate module and add unit testing
The application `test_applications_from_config.py` is controlling the application tests and includes functionality to compare output from those tests to reference files (e.g., after a coordinate transformation). 

Suggest to move these functionality into a separate module to lower complexity in `test_applications_from_config.py` and to allow to add unit tests to those functions.

GernotMaier:
Move the functions to compare files and assert file types from `tests/integration_tests/test_applications_from_config.py` into separate modules in `simtools/testing`:

- allows to reuse those functions and have them unit tested
- expected to be extended significantly in future (it looks a bit empty right now)

There is no change in functionality, I've moved simply the code into the modules and:

- added unit tests
- improved the code itself (as a result of the unit tests)

Please have a look at the naming of the modules and provide feedback if this is good. Naming new modules is often difficult.

Closes #970

### DIFF
--- a/simtools/testing/assertions.py
+++ b/simtools/testing/assertions.py
@@ -32,7 +32,6 @@ def assert_file_type(file_type, file_name):
         try:
             with open(file_name, encoding="utf-8") as file:
                 yaml.safe_load(file)
-            print("FFFF safe load", file_name)
             return True
         except (yaml.YAMLError, FileNotFoundError):
             return False
--- a/simtools/testing/assertions.py
+++ b/simtools/testing/assertions.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+from pathlib import Path
 
 import yaml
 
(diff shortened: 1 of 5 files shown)


simtools/testing/assertions.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    if file_type == "json":
+        try:
+            with open(file_name, encoding="utf-8") as file:
+                json.load(file)
+            return True
+        except (json.JSONDecodeError, FileNotFoundError):
+            return False
+    if file_type in ("yaml", "yml"):
+        if Path(file_name).suffix[1:] not in ("yaml", "yml"):
```
orelgueta, line 32: I thought that if the file has no suffix, `file.suffix[1:]` would raise an error of the index being out of bound, but apparently that doesn't happen with strings and this is safe.

orelgueta (approved):
Minor change of docstring and a lesson learned.

Issue #829, opened 2024-03-06
Hardwired atmosphere IDs for backwards compatibility


GernotMaier:
https://github.com/gammasim/simtools/blob/5fe76f74387e9781a47291d4f726897cad9afcf9/simtools/model/site_model.py#L84

This probably can/should go at some point.

PR #1139, opened 2024-09-19
Check output of simulations

### LINKED ISSUE #1129: Production tool seems broken
Running the production tool reports a successful run but no simulation is actually performed. I didn't look into it deeply yet, but at minimum I noticed so far that
- no telescopes are actually defined in CORSIKA. 
- There's an error due to zeroth telescope line below:
https://github.com/gammasim/simtools/blob/6484aa889c5fa196ecf6e7ec871c7ec568afeeee/simtools/simtel/simtel_config_writer.py#L179
- The names of the submission scripts contain None.
- etc.

I will try to fix these issues soon, but I thought I should open this issue already to make sure everyone is aware.

P.S.
I discovered these issues because I am writing an integration test that will discover such cases, so at least we should be covered in the future.

orelgueta:
The goal of this PR is to introduce an additional step to the integration tests of the simulation productions that checks the output and makes sure it is reasonable.

The goal above required various changes throughout the code: 
- allow setting random seeds in order to produce reproducible results
- add parameters and schemas that are necessary to reproduce the Prod6 results
- change hardcoded values that did not comply with the Prod6 model

There are also a few random bits of changes introduced as a part of this PR which are not directly related to the subject, but I tried to make sure the commit messages clarify those in case they need clarification.

This PR is dependent on the [corresponding MR](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/merge_requests/60) in the model parameters repo. Tests will fail until that one is merged.

Fixes #1129.


simtools/corsika/corsika_config.py
```diff
@@ -52,7 +52,8 @@ def __init__(self, array_model, args_dict, db_config=None, label=None):
         self.azimuth_angle = None
         self._run_number = None
         self.config_file_path = None
-        self.primary_particle = self._set_primary_particle(args_dict)
+        # The following uses the setter defined below, that is why the args_dict is passed
+        self.primary_particle = args_dict
```
GernotMaier, line 56: This is correct but is without reading the functions below very confusing. Not nice that an additional comment is needed.

simtools/model/model_parameter.py
GernotMaier, line 1: Do you actually remember why we had this 'or' statement? There must have been a reason.

simtools/model/array_model.py
```diff
@@ -108,6 +108,11 @@ def _initialize(self, site, array_elements_config):
             array_elements = self._get_array_elements_from_list(
                 site_model.get_array_elements_for_layout(self.layout_name)
             )
+        if array_elements == {}:
+            raise ValueError(
+                "No array elements found. "
+                "Did you provide a valid layout name or list of telescopes?"
```
GernotMaier, line 114: I don't like the 'you' in the message - it implies that the user is a person. Maybe "Possibly missing valid layout name or missing telescope list"?

simtools/model/site_model.py
```diff
@@ -68,20 +68,22 @@ def get_corsika_site_parameters(self, config_file_style=False):
...
         if config_file_style:
-            _atmosphere_id = 26 if self.site == "North" else 36
             return {
-                "OBSLEV": [self.get_parameter_value("corsika_observation_level")],
-                "ATMOSPHERE": [_atmosphere_id, "Y"],
+                "OBSLEV": [
+                    self.get_parameter_value_with_unit("corsika_observation_level").to_value("cm")
+                ],
+                # We always use a custom profile by filename, so this has to be set to 99
+                "ATMOSPHERE": [99, "Y"],
```
GernotMaier, line 85: Good. This closes #829

simtools/schemas/model_parameters/save_pe_with_amplitude.schema.yml
```diff
@@ -0,0 +1,38 @@
...
+meta_schema_url: https://raw.githubusercontent.com/gammasim/simtools/main/simtools/schemas/model_parameter_and_data_schema.metaschema.yml
+meta_schema_version: 0.1.0
+name: save_pe_with_amplitude
+description: |-
+  True if photo-electron data is to be stored with the resulting amplitudes (in mean-p.e. units) in addition to the arrival times.
+  Storing of photo-electron data should be activated separately (see save_photons (bit 1) and store_photoelectrons parameters)
+  but non-zero values of this parameter will also activate save_photons bit 1.
+  Storing p.e. in internal calibration events is controlled separately by save_calibration_pe.
+data:
+  - type: int
```
GernotMaier, line 15: Would `type: boolean` be more appropriate? We have done that with all (most?) other parameters with this type of range (I think Konrad uses this because bool was not available in his C version when he started sim_telarray).

simtools/schemas/model_parameters/save_pe_with_amplitude.schema.yml
```diff
@@ -0,0 +1,38 @@
...
+    unit: byte
```
GernotMaier, line 16: It this unit correct? I understood it as boolean.

simtools/schemas/model_parameters/store_photoelectrons.schema.yml
```diff
@@ -0,0 +1,41 @@
...
+  While the save_photons=2 configuration option will save photo-electron data for all events,
+  this parameter allows to set a lower limit on the number of photo-electrons
+  which must get registered before deciding to store them.
+  Setting it to a value of zero or bigger (default: -1) will automatically enable the corresponding save_photons bit.
+  It should be cross-checked that the min_photons parameter is unlikely to result
+  in fewer than the given number of photo-electrons, as no attempt is made to trace photons through the telescope
+  and produce any photo-electrons if there are too few photons in the first place.
+data:
+  - type: int
+    unit: byte
```
GernotMaier, line 19: Is this unit correct?

simtools/simtel/simulator_array.py
```diff
@@ -111,3 +114,30 @@ def _check_run_result(self, run_number=None):
             raise InvalidOutputFileError(msg)
         self._logger.debug(f"simtel_array output file {output_file} exists.")
         return True
+
+    @staticmethod
```
GernotMaier, line 118: Can you explain the choice of having those values hardwired? Any easy way to make them configurable?

simtools/testing/assertions.py
```diff
@@ -41,3 +43,109 @@ def assert_file_type(file_type, file_name):
     # no dedicated tests for other file types, checking suffix only
     _logger.info(f"File type test is checking suffix only for {file_name} (suffix: {file_type}))")
     return Path(file_name).suffix[1:] == file_type
+
+
+def assert_n_showers_and_energy_range(file):
+    """
+    Assert #showers and energy range.
```
GernotMaier, line 50: What does `#` mean?

GernotMaier: And maybe say that you look into a log file of corsika or simtelarray here?

GernotMaier (approved):
Thanks a lot for this big update - really hard to keep up with all those little details, I wouldn't have known them.

My comments are mostly minor, this is why I approve already.

GernotMaier:
@orelgueta - I guess you have seen the sonarqu complaint about duplications:

https://sonar-cta-dpps.zeuthen.desy.de/component_measures?id=gammasim_simtools_AY_ssha9WiFxsX-2oy_w&metric=new_duplicated_lines_density&pullRequest=1139&view=list

simtools/corsika/corsika_config.py
orelgueta: What do you suggest we do then? The other option I see to make it more readable is to change to
```
self.primary_particle = {"primary_id_type": args_dict.get("primary_id_type"), "primary": args_dict.get("primary")}
```

But I think this is worse than the comment explaining it to future developers.

simtools/model/array_model.py
orelgueta: No problem.

simtools/schemas/model_parameters/save_pe_with_amplitude.schema.yml
orelgueta: Changed to boolean

simtools/schemas/model_parameters/save_pe_with_amplitude.schema.yml
orelgueta: changed to boolean

simtools/schemas/model_parameters/store_photoelectrons.schema.yml
orelgueta: No, it's in p.e., but that's not a valid unit, so I set it to dimensionless.

simtools/simtel/simulator_array.py
orelgueta: I hardcoded them in the same way they are hardcoded in sim_telarray. Since we do not use the sim_telarray histograms, I didn't see a reason to make it configurable. However, if you think it's worth it, please make this comment into an issue and I will implement it after this PR (and use this function to set default values).

simtools/testing/assertions.py
orelgueta: `#` means number and I think I wrote it because I didn't want to repeat myself with the word "number". I changed it back to number.
This is one of those cases in which I had to reformulate because ruff wouldn't let me have one long sentence which took more than one line...


I do not look into the log files of CORSIKA and sim_telarray here but instead read the sim_telarray output file itself and extract the headers and count the `#` of events and energies produced.

simtools/model/model_parameter.py
orelgueta: Yes, it was there in case no label was provided to the application. That caused an issue actually because only in the case of the model parameter the output went to a directory with the name of the telescope instead of a directory under the date (`d-19-09-24`). Then the application would fail because the model parameters were not found. 

This fixes that (though I know you don't like the date thing.

orelgueta:
Thanks for the detailed review! 

I fixed the comments and answered a couple of points. I won't merge yet because of the point regarding the primary particle.

I also fixed the duplication, so I will wait for sonarcube to run again. I didn't know it is so pedantic, but it's not bad I guess.

orelgueta:
The integration tests are failing because of the change to the pixel rate that is missing in the db. The old style of the long string doesn't work anymore, it is replaced with a real list of values instead (which is shortened in the cfg file). We could wait for the model parameter MR to create a new DB, rerun the tests and only then merge this one. That would be the safest.

simtools/corsika/corsika_config.py
GernotMaier: No suggestion - I was just complaining. Should have made it clearer that no changes are expected.

simtools/model/model_parameter.py
GernotMaier: No. I don’t like the date thing - `use_plain_output_path` is my friend. Thanks for the explanation!

simtools/simtel/simulator_array.py
GernotMaier: I think it is easy to miss that this is for histogramming only. Could we either change the function call or at least say in the first line of the docstring `..used for histogramming in sim_telarray`? Quick readers might otherwise thing we hardwire the spectral index for the simulations.

simtools/testing/assertions.py
GernotMaier: Still don’t understand why you can’t use `number`. The first line is not that long.

GernotMaier:
And yes, sonarq is pedantic and our ruff settings are strict. This is good as it really simplifies reviews and we are not bothered with this stuff.

simtools/schemas/model_parameters/sampled_output.schema.yml
```diff
@@ -0,0 +1,35 @@
...
+title: Schema for sampled_output model parameter
+version: 0.1.0
+meta_schema: simpipe-schema
+meta_schema_url: https://raw.githubusercontent.com/gammasim/simtools/main/simtools/schemas/model_parameter_and_data_schema.metaschema.yml
+meta_schema_version: 0.1.0
+name: sampled_output
+description: |-
+  Select a specific output format (if several available). A value of 0 means to write ADC sums, a value of 1 to write ADC samples.
+data:
+  - type: int
```
GernotMaier, line 12: Should this be boolean?

simtools/schemas/model_parameters/sampled_output.schema.yml
```diff
@@ -0,0 +1,35 @@
...
+    unit: byte
```
GernotMaier, line 13: Should this be null?

simtools/simtel/simulator_array.py
orelgueta: Changed the name of the function to get_power_law_for_sim_telarray_histograms.

simtools/testing/assertions.py
orelgueta: I changed it already to be `Assert the number of showers and the energy range.`

My complaint was about the original version you didn't see which was 
```
Assert the number of showers is consistent with the required one (up to 1% tolerance) 
and that the energies simulated are within the configured ones.
```

This is too long to be one line and ruff doesn't allow a summary line with a line break in it, so I had to break it into two separate statements with some repetition. I find it a bit annoying, so I complained.

simtools/schemas/model_parameters/sampled_output.schema.yml
orelgueta: Yes, changed to boolean, I missed it yesterday.

simtools/schemas/model_parameters/sampled_output.schema.yml
orelgueta: Removed since it is a boolean

orelgueta:
> Two minor comments, but as before already approved.
> 
> This is what I usually do for testing (and you might have done it):
> 
> * upload the branch in question of the model parameter repository to a local database
> * Run the integration tests locally (remember to run them with `—model_version=“6.0.0”` and `5.0.0`
> 
> If this is fine, we probably can first merge the model_parameter changes; release a new version there (to trigger the DB upload) and then re-run the integration tests here on github.

Yes I have done all of that locally during the development process. Otherwise there's no way to test anything. It should all work now after the merging of both branches.

orelgueta:
I resolved the conflicts with the main branch. I will wait for the tests to pass (or perhaps fail if the DB is not updated yet) and then merge.

PR #1140, opened 2024-09-19
Add EXPECTED_OUTPUT field to schema for integration test configuration

GernotMaier:
Required for schema validation in PR #1139


orelgueta (approved):
Didn't know this is necessary.

GernotMaier:
The rule is no input without a schema validation.

For these config file, we had different keywords for the same thing.

PR #1143, opened 2024-09-20
Improve robustness of database scripts

GernotMaier:
Improve robustness of `simtools-db-add-model-parameters-from-repository-to-db` application and `upload_from_model_repository_to_db.sh` scripts. No functional changes.

This is triggered by:

- a leading space in the argument of `upload_from_model_repository_to_db.sh` for the database name is removed now (I have been puzzled quite a while why I can't read from a newly generated local database of name "CTAO-Simulation-Model-v0-0-1" (it was uploaded as " CTAO-Simulation-Model-v0-0-1")
- ask user for confirmation before `upload_from_model_repository_to_db.sh` is uploading to a remote DB (because I did this accidentally...). This requires `upload_from_model_repository_to_db.sh` to read the `.env` file and therefore the local DB variables in the script had to be renamed.

### DIFF
--- a/simtools/configuration/commandline_parser.py
+++ b/simtools/configuration/commandline_parser.py
@@ -194,7 +194,7 @@ class CommandLineParser(argparse.ArgumentParser):
         _job_group.add_argument(
             "--db_simulation_model",
             help="name of simulation model database",
-            type=self.strip_string,
+            type=str.strip,
             required=False,
             default=None,
         )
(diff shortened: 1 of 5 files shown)


simtools/configuration/commandline_parser.py
```diff
@@ -194,7 +194,7 @@ def initialize_db_config_arguments(self):
         _job_group.add_argument(
             "--db_simulation_model",
             help="name of simulation model database",
-            type=str,
+            type=self.strip_string,
```
orelgueta, line 197: Can't you just do `type=str.strip` to achieve the same?

GernotMaier: Yes, exactly - this is simpler. Changed it (and changed also the corresponding unit test)

(I've done the function because I tried with with `type=str.strip()` and got an error. Clearly didn't think enough...)

GernotMaier:
@orelgueta - this one should be ready for final review. Thanks for your comments.

PR #1148, opened 2024-09-23
Address numpy warnings in psf_analysis; fix minor bug

GernotMaier:
PR includes the following items:

1. Addresses numpy warnings by removing the numpy.core calls in `simtools/psf_analysis.py`.

Warnings were:

```
 /__w/simtools/simtools/simtools/psf_analysis.py:365: DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.records.
    return np.core.records.fromarrays(np.c_[x_pos_data, y_pos_data].T, dtype=d_type)

tests/integration_tests/test_ray_tracing.py::test_integral_curve
  /__w/simtools/simtools/simtools/psf_analysis.py:430: DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.records.
    return np.core.records.fromarrays(np.c_[radius_all, intensity].T, dtype=d_type)
```

2. Increase coverage (which was below 60% and is now above 80%)
3. Fix a bug in the find_psf function (it counts photons and does a `s0 < target_number < s1` comparison. This fails when `target_number==s1` (or s0). Fixed this. Also increased the number of iterations for the fast (smart) psf finder from 100 to 1000. 

Might also fix issue #1157 (I have run this test >100 times without any failures)


EshitaJoshi (approved):

GernotMaier:
@EshitaJoshi - thanks for the review!

PR #1149, opened 2024-09-23
Remove sim telarray native configuration files

orelgueta:
This PR removes our dependence on sim_telarray configuration files entirely. We only use simtools created configuration files from now on.

I had to skip the `testeff` tests because of this change since in the current version, `testeff` does not allow to provide a different configuration directory. A fix for that is already available. In addition, testeff requires a file which is currently unavailable in the DB (MR on gitlab is open). 

A couple of explicit additions were necessary because CORSIKA config does not write out configuration files. However, the atmospheric profile is necessary for all CORSIKA simulations, so now it is written out explicitly in the `corsika_config` step.

### DIFF
--- a/simtools/model/site_model.py
+++ b/simtools/model/site_model.py
@@ -2,6 +2,7 @@
 """Definition of site model."""
 
 import logging
+from pathlib import Path
 
 from simtools.model.model_parameter import ModelParameter
 
@@ -30,6 +31,7 @@ class SiteModel(ModelParameter):
         mongo_db_config,
         model_version,
         label=None,
+        array_model=None,
     ):
         """Initialize SiteModel."""
         self._logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


simtools/model/site_model.py
```diff
@@ -30,6 +31,7 @@ def __init__(
         mongo_db_config,
         model_version,
         label=None,
+        array_model=None,
```
GernotMaier, line 34: Missing entry in docstring.

GernotMaier (approved):
Thanks for this. Tiny comment on a missing entry in a docstring. Approved.

simtools/model/site_model.py
orelgueta: Thanks!

orelgueta:
Thanks for the review. Was about to merge but after fixing the docstring I see errors in all tests like

```
Error: Unable to resolve action `sonarsource/sonarqube-scan-action@v3`, unable to find version `v3`
```

I assume they are temporary so I will try rerunning the tests a bit later and if they all work (except the linter one) I will merge.

Issue #1153, opened 2024-09-24
Include `atm_trans_2200_1_3_0_0_0.dat` in camera efficiency simulator


orelgueta:
`testeff` uses `atm_trans_2200_1_3_0_0_0.dat` to perform a correction to the NSB attenuation. See e-mail excerpt from Konrad below.

> Concerning the NSB correction from the altitude of the ORM (where
> the B&E spectrum was taken) to the altitude of the Cherenkov telescope
> site (initially HEGRA, when this was written, now CTAO-N), that is a pretty
> small correction, guessing how extinction applies to diffuse light.
> The correct way would be to use a radiation transport code, apply the
> proper absorption and scattering cross sections, and the scattering phase
> angle distribution applying, and allow for multiple scattering.
> Both for La Palma ORM to CTAO-N and for Paranal ESO to CTAO-S, that is
> order of a percent correction and can hardly go wrong by more than a few
> tenths of a percent. Keep in mind that the solar UV/X-ray flux and solar
> wind
> can change the NSB by order 10 percent, even without aurorae.
> 
> If you have the proper NSB spectrum applying to the IACTs, you should
> disable that correction with the '-nc' (or '--no-correction') option.
> Given that the NSB pixel rates are usually not calculated from the direct
> NSB flux but from the one normalized to the CTA Reference NSB Level,
> the impact of this operation is pretty much zero. (And once we have actual
> pixel p.e. rates from actual telescopes, these should be used and not the
> ones based on a theoretical reference level.)

Currently I turned off this correction by setting `-nc` in `simulator_camera_efficiency.py`. Once we decide how to include `atm_trans_2200_1_3_0_0_0.dat` in the DB and in the code, this option should be reverted. This is until we get a proper NSB spectrum in the future.

orelgueta:
Fixed in #1154

PR #1154, opened 2024-09-24
Fix testeff after removing native sim_telarray cfg

orelgueta:
testeff did not work after removing the native sim_telarray configuration files. Two reasons for that are
- The executable does not allow to provide a directory with include files like we do for sim_telarray, so we couldn't point it to our model files.
- Testeff uses the default B&E spectrum for the calculation, unless a different spectrum is provided. Since we do not have a different spectrum at the moment, we need to provide the original B&E one ourselves. 

The first issue was fixed by Konrad providing a new testeff source code. I tested it with simtools by compiling it in the container after the build_all step. We need to update our sim_telarray tarball to one that includes this fix (not yet available). In the meantime, the integration tests check if the old testeff source code is used and if so, those tests are skipped.

The second issue was solved by adding the B&E spectrum to the DB and introducing a new parameter `nsb_reference_spectrum`. With time this will not be needed, but I suspect it will take a few years for that to be the case.

One more pending issue is that testeff requires also the atmospheric profile `atm_trans_2200_1_3_0_0_0.dat` to correct for the attenuation as in the original B&E. We avoid that for now by telling it not to apply this correction. However, this file was added to the DB and in the future we should reimplement this correction. I will open a separate issue.


orelgueta:
@GernotMaier, two issues/points with the tests here:
- The linting tests fail because of a non-ascii character which was added on purpose to one of the tests. Should I modify the test to avoid this issue or change the test (it's not really that important)?
- Once again I am running into the issue with the DB not updated because the MR in the model parameter repo (!63) is waiting for this PR for the schema and upload to the DB. Should I upload manually again? Open a separate PR just with the schema file?

GernotMaier:
I see that you've fixed the failing linting test. Good (although I don't understand why it wasn't an issue before).

For the interaction of MR / DB, see [this issue](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/issues/29)

orelgueta:
> I see that you've fixed the failing linting test. Good (although I don't understand why it wasn't an issue before).

I only implemented an example of a possible fix. You should tell me if you agree to it because it adds an exception for one file not being tested.

It wasn't an issue before because I added this character (ñ) now on purpose in order to test reading a latin encoding.

> 
> For the interaction of MR / DB, see [this issue](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/issues/29)

But this does not entirely solve the issue of this PR, right? 
Right now this PR is stuck because the DB does not contain a parameter it needs. In order to get that parameter in the DB, the CI on GitLab needs to update the DB. However, it won't do that until the schema validation passes. The missing schema file will not be in any of the docker images we can point to until this PR is merged. So, we are sort of stuck.

The only way to avoid this issue, as I see it, is to make sure that adding schema files is done in a separate PR first. Merge that and only then run the tests of the PR with actual code changes.

What do you suggest we do with this PR for now?

GernotMaier:
1. I've updated the perl command to exclude simply lines containing the string "latin". I think this is slightly better than excluding the full file
2. Yes, I think this needs to be the workflow: first update schemas and merge them. But we need to think what to do here, this is a bit painful.

orelgueta:
> 1. I've updated the perl command to exclude simply lines containing the string "latin". I think this is slightly better than excluding the full file

Nice, that's a better solution indeed.

> 2. Yes, I think this needs to be the workflow: first update schemas and merge them. But we need to think what to do here, this is a bit painful.

I can follow that workflow now as well. I can open a separate PR just for the schema and this PR will wait for the new DB until we merge it. Not a big deal I guess.

orelgueta:
This PR is waiting for the next release of the DB that contains a new parameter. For that we need to either switch to a tagged version of the docker image or wait for tomorrow for it to pull the latest.

GernotMaier:
MR for explicit tagging of image used in model_parameter repository: https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/merge_requests/66

orelgueta:
Yes, I need to release a new version of the model parameters, but I am taking advantage and also introducing the NSB correction to B&E first so we have both corrections in the new release.

orelgueta:
This is ready for review. The failed coverage test is not solvable, as explained in the test function, so I think we can safely ignore it.

simtools/schemas/model_parameters/correct_nsb_spectrum_to_telescope_altitude.schema.yml
```diff
@@ -16,7 +16,7 @@ data:
     unit: dimensionless
     default: "atm_trans_2200_1_3_0_0_0.dat"
 instrument:
-  class: Telescope
+  class: Camera
```
GernotMaier, line 19: Suggest to leave it as `Telescopes`. Requires merging of #1173

tests/integration_tests/test_applications_from_config.py
GernotMaier, line 1: Splitting this into functions is good! I suggest to open an issue that we move those functions into the simtools.testing module (which then allows to write unit tests for them).

(don't do this as part of this PR; this is good to be merged)

GernotMaier (approved):
Approve - with the suggestion below (replace instrument.class to Telescope after reviewing PR #1154

simtools/schemas/model_parameters/correct_nsb_spectrum_to_telescope_altitude.schema.yml
orelgueta: Done. Will wait to see that tests pass now.

tests/integration_tests/test_applications_from_config.py
orelgueta: Done

orelgueta:
Fixes #1153

Issue #1152, opened 2024-09-24
Update sim_telarray tarball


orelgueta:
The `testeff` software was updated. Once the new version is included in the tarball, we should update the tarball used to build the containers. I assign it to myself to keep track of the tarball availability, but @GernotMaier will actually update the tarball on his DESY space.

orelgueta:
It would be nice to get it from gitlab, yes. Maybe we take a quick look and see if it's easy. If it isn't we push it for the next maintenance period.

GernotMaier:
Looking at it:

- the gitlab is only sim_telarray, not the package for corsika/sim_telarray. I don't know how easy it would be to change our current workflow which assumes that both are in the same package
- gitlab allows to download the package as [tar file](https://gitlab.cta-observatory.org/Konrad.Bernloehr/sim_telarray/-/archive/master/sim_telarray-master.tar.gz), so this is not an issue. KB also updates the gitlab (but we would need a token to access the repository, as it is private)

orelgueta:
> * the gitlab is only sim_telarray, not the package for corsika/sim_telarray. I don't know how easy it would be to change our current workflow which assumes that both are in the same package

It's probably not too bad because the tarball we will get from sim_telarray will just replace the sim_telarray tarball of the usual corsika/sim_telarray tarball that we download from Konrad's website. It means we won't keep up with changes to e.g., eventio, but that we can do manually since those are rare.

So the workflow would probably remain as it is, except we will also download the tarball from gitlab and replace the sim_telarray tar package with this one. We would also have to delete the cfg directory (which we also do now).

> * gitlab allows to download the package as [tar file](https://gitlab.cta-observatory.org/Konrad.Bernloehr/sim_telarray/-/archive/master/sim_telarray-master.tar.gz), so this is not an issue. KB also updates the gitlab (but we would need a token to access the repository, as it is private)

The fact is private is why I couldn't find it before and look into it myself. I guess it is private because of the telescope models, but I am not sure it is relevant anymore since we published them. Maybe it's worth asking Konrad about making it public.

orelgueta:
A new tarball is available that includes the required change to testeff.c:

https://www.mpi-hd.mpg.de/hfm/CTA/MC/Software/Testing/corsika7.7_simtelarray.tar.gz

PR #1158, opened 2024-09-25
Update version number for corsika/sim_telarray package (20240913)

GernotMaier:
This PR required the following steps:

- download most recent CORSIKA/sim-telarray package from KB's testing website (date 2024/09/13)
- upload the package to sync and share
- define a new github secret for the actions CLOUD_SIMTEL_230913
- update the workflow for building the docker for CORSIKA/sim-telarray

### DIFF
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         production: ['prod6-sc', 'prod6-baseline', 'prod5', 'prod5-sc']
-        version: ['240923']
+        version: ['240913']
         hadronic_model: ['qgs2']
 
     steps:
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         production: ['prod6-sc', 'prod6-baseline', 'prod5', 'prod5-sc']
-        version: ['240318']
+        version: ['240923']
         hadronic_model: ['qgs2']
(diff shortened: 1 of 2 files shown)


GernotMaier:
The relevant action for the docker build of the new corsika/simtel_array version is [here](https://github.com/gammasim/simtools/actions/runs/11027285485/job/30625308828?pr=1158)

orelgueta:
Where do you see a package from the 23rd of September? The last one I see on the website is from the 13th of September, which means it wouldn't include the latest change to testeff. I might be looking in the wrong place though.

GernotMaier:
13th! Misread the numbers. I correct it to make sure we are not confused by it in the future.

orelgueta:
No, but then I would wait with this update until Konrad produces a new tarball. The trigger for this change is the updated testeff, all other changes are not really important.

GernotMaier:
Seems the 'patch' command is not part anymore of almalinux, as I get:

```
./build_all: line 1023: patch: command not found
```

I am adding it now and try again.

GernotMaier:
@orelgueta - I suggest a quick review and merge, even if it doesn't solve your testeff problem. The 'patch' issue was new and is solved here.

Successful building: https://github.com/gammasim/simtools/actions/runs/11031927431

orelgueta (approved):
From my point of view, the changes are fine, but the illuminator tests failed for some unrelated reason. I approve anyway. 

The only issue is that this PR will close #1152, which it shouldn't. So please reopen the issue once you merge.

GernotMaier:
I think we need to fix the illuminator issue first, e.g., in https://github.com/gammasim/simtools/actions/runs/11031923016/job/30639695323?pr=1158

It is looking for  atm_trans_2156_1_3_2_0_0_0.1_0.1.dat , is this related to the removal of the config files?

PR #1159, opened 2024-09-25
Add nsb reference spectrum schema

orelgueta:
See discussion in #1154. 

We first add the schema file so that new docker images are built with it. Then the MR on gitlab would be merged and a new release would be released. This would trigger an update to the DB that would allow #1154 to be merged.


GernotMaier (changes requested):
Shouldn't this file be in `simtools/schemas/model_parameters/`?

I was wondering why the `CI-schema-validation.yml` is not running, as it should be triggered by changes to this directory.

orelgueta:
Yes, sorry, I copied the file from the other branch to the wrong directory.

simtools/schemas/model_parameters/nsb_reference_spectrum.schema.yml
```diff
@@ -0,0 +1,34 @@
...
+---
+title: Schema for nsb_reference_spectrum model parameter
+version: 0.1.0
+meta_schema: simpipe-schema
+meta_schema_url: https://raw.githubusercontent.com/gammasim/simtools/main/simtools/schemas/model_parameter_and_data_schema.metaschema.yml
+meta_schema_version: 0.1.0
+developer_note: To be replaced by a data table
+name: nsb_reference_spectrum
+description: |-
+  Nightsky spectrum from Benn & Ellison (astro-ph/9909153 actually),
```
GernotMaier, line 11: I would remove the 'actually', as it doesn't really have meaning (without a Benn & Elisson reference), but I don't want to be pedantic!

GernotMaier (approved):

simtools/schemas/model_parameters/nsb_reference_spectrum.schema.yml
orelgueta: This was copy pasted directly from the spectrum file from KB, but I removed it, no problem.

orelgueta:
Tests fail here for version 5.0.0 because the funnel efficiency parameters were missing from Prod5. MR [64](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/merge_requests/64) should resolve that, once it is merged and a new release is created. Once the DB is updated, the tests here should pass, in which case we can merge this and then rerun the pipeline of the previous MR on the model parameters repo. Once that passes we can merge that, create another release, wait for the DB to update and then finally merge all other pending PRs. 

I assume we will find more issues by that point, but at least these ones will be resolved.

PR #1166, opened 2024-09-26
Remove option for uploading of metadata to DB

GernotMaier:
The metadata DB is depreciated and not used anymore.


GernotMaier:
I've left the `--type` option, although it has only one choice (`model_parameters`). I expect that we will need to expand this soon when adding tables to define productions.

orelgueta (approved):

PR #1171, opened 2024-09-27
Add schema for correction of NSB spectrum to the telescope altitude

orelgueta:
I am not sure about three things, 
- the name of the parameter, should it be more literal like `atmospheric_transmission_2200m`?
- I set the instrument to Camera because in the end this goes to the NSB pixel rate, however it is more related to the site of course. On the other hand, I don't know if I can use site because in sim_telarray_configuration we don't have site and I don't know if it would be read out and written out.
- I set the simulation software to simtools because `testeff` is not one of the options and I think that if I set it to sim_telarray it might not be read out.


GernotMaier (approved):

GernotMaier:
All your points are good, don't have any better suggestions. This should be clear enough.

orelgueta:
OK, thanks!

PR #1181, opened 2024-10-02
Improved array elements treatment

GernotMaier:
Improve pulling the list of available array elements (i.e., telescopes and calibration devices) from the DB. Add functionality to use new model parameter `design_model`.

### Background

The model for each array element is the combination of two models:

- a design or base model which covers the parameters common to all telescopes of a certain type (e.g., instrument `LSTS-design` for all southern telescopes)
- the telescope specific parameters (e.g., instrument field `LSTS-03`)

Until now it was assumed that there is a simple corresponding of design and specific model, meaning the design model was derived from the telescope type (meaning `LSTS-03` has the 'hardwired' design model 'LSTS' + 'design'.

The introduction of the design model parameter (see #1172 and [MR 86](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/merge_requests/68), which allows a more flexible handling of this relation ship (e.g., we can now have southern MSTs with the Northern camera model).

Note that this introduces new assumptions, which all simplifies the code:

- every array element model includes a 'design_model' parameter
- every array element has a design model (which is not the case yet; LSTN-design does not exist, see [this MR](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/merge_requests/69)).

Final note is regarding the changes in array_layout:  this module allows to add telescopes (array elements) which are not in the DB. The use case might be the testing of new telescope positions with the goal of converting telescope coordinates from e.g., UTM to CORSIKA coordinates. In this particular case, the assumption is that the corresponding design model for e.g. 'MSTS-483' is 'MSTS-design' (this is needed to retrieve e.g. the corsika_axis_height parameters).

### Code changes

- cleanup and simplification of functions pulling the list of array model elements
- moved function from db/db_handler to a new module called db/db_array_elements
- use a `cache` function to avoid multiple identical database queries
- added a function called db_handler.get_collection (as this is used in multiple place in db_handler and db_array_elements)

---

Everything below has been generated by Github copilot. Feel free to ignore it (maybe it is useful)

### Changes

This pull request introduces new functionality for retrieving array elements from the database and refactors existing database handling methods to enhance code modularity and readability. The most important changes include the addition of new functions for array element retrieval, the import of these functions into the database handler, and the refactoring of several methods to use the new `get_collection` method.

### New Functionality for Array Element Retrieval:
* Added `get_array_elements`, `get_array_element_list_for_db_query`, and `get_array_elements_of_type` functions to `simtools/db/db_array_elements.py` for retrieving array elements from the database.

### Refactoring for Code Modularity:
* Imported `db_array_elements` in `simtools/db/db_handler.py` to utilize the new array element functions.
* Replaced direct database client calls with the new `get_collection` method in various methods within `simtools/db/db_handler.py`, such as `get_model_parameters`, `read_mongo_db`, and `copy_array_element`. [[1]](diffhunk://#diff-76a4f9979ad6687c7a3fd974fd63e2f6e8f3a4b5d40198a567cb7482b169f0edL62-R62) [[2]](diffhunk://#diff-76a4f9979ad6687c7a3fd974fd63e2f6e8f3a4b5d40198a567cb7482b169f0edL190-R191) [[3]](diffhunk://#diff-76a4f9979ad6687c7a3fd974fd63e2f6e8f3a4b5d40198a567cb7482b169f0edL333-R353) [[4]](diffhunk://#diff-76a4f9979ad6687c7a3fd974fd63e2f6e8f3a4b5d40198a567cb7482b169f0edL728-R748)
* Renamed and refactored `_get_db_name` to `get_db_name` in `simtools/db/db_handler.py` for consistency and clarity.

These changes collectively improve the maintainability and extensibility of the codebase by centralizing database access logic and introducing new utilities for handling array elements.


GernotMaier:
@EshitaJoshi - in case you notice that the `@cache` function is not implemented in an optimal way: I have noticed this and this is corrected in a follow up PR (#1183 ). So don't worry about this here.

simtools/db/db_array_elements.py
```diff
@@ -0,0 +1,122 @@
...
+        "instrument", ASCENDING
+    )
+
+    _all_available_array_elements = {}
+    for doc in results:
+        try:
+            if doc["parameter"] == "design_model":
+                _all_available_array_elements[doc["instrument"]] = doc["value"]
+        except KeyError as exc:
+            raise KeyError("Incomplete array element entry in the database.") from exc
```
EshitaJoshi, line 51: Would it be worth logging or printing here which doc in results has the incomplete entry?

simtools/db/db_array_elements.py
```diff
@@ -0,0 +1,122 @@
...
+        List of array element model names as used in the DB.
+
+    """
+    _available_array_elements = get_array_elements(db, model_version, collection)
+    try:
+        return [array_element_name, _available_array_elements[array_element_name]]
+    except KeyError:
+        pass
+
+    if array_element_name in _available_array_elements.values():
```
EshitaJoshi, line 89: When would this if statement return true? I don't understand why it would since the get_array_elements function returns a dictionary with the instrument name as key and the design model as the value. Do we have situations where for eg. the MSTN-01 has the design model of MSTN-01 instead of MSTN-design? If so, what's the point of having separate design models? Even in that case MSTN-01 would exist as a key in the dictionary too.

simtools/db/db_array_elements.py
GernotMaier: Good point, added it.

simtools/db/db_array_elements.py
GernotMaier: This is e.g. the case for `array_element_name="MSTS-design"` - if you want to get the design model then the first statement with `return [array_element_name, _available_array_elements[array_element_name]]` will raise a KeyError.

The corresponding test is [here](https://github.com/gammasim/simtools/blob/9ca0b173df788c142f5532ff6921a98bf93faccb/tests/unit_tests/db/test_db_array_elements.py#L59).

EshitaJoshi (approved):
All looks good, thanks!

PR #1182, opened 2024-10-02
Sort explicitely DB queries to retrieve newest entry

GernotMaier:
Ensure that we retrieve always the most recent entry from the database by adding an explicit sorting to the pymongo command.

This resolves an issue occasionally occurring when running the unit test `test_db_handler.py::test_update_parameter_field_db`, where after updating a model parameter the read_mongo_db command reads the older and not the updated value.

Note that we do not have duplicated entries in our operational database. I am not entirely sure why this problem didn't occur earlier.

And because I've learned something about the mongoDB IDs:

```
from bson import ObjectId
id1=ObjectId("66fce0a6c39f413ae0412df1")
id2=ObjectId("66fce0a6c39f413ae0412e67");
print("ID1:", id1.generation_time)
print("ID2:", id2.generation_time)
print("ID1 < ID2: ", id1 < id2)
```

gives you the same time stamp (first four bytes)

```
ID1: 2024-10-02 05:56:54+00:00
ID2: 2024-10-02 05:56:54+00:00
ID1 < ID2:  True
```

Ordering is giving by the last three bytes (which are a counter).

### DIFF
--- a/simtools/db/db_handler.py
+++ b/simtools/db/db_handler.py
@@ -8,7 +8,7 @@ from threading import Lock
 import gridfs
 from bson.objectid import ObjectId
 from packaging.version import Version
-from pymongo import MongoClient
+from pymongo import ASCENDING, MongoClient
 from pymongo.errors import BulkWriteError
 
 from simtools.db import db_from_repo_handler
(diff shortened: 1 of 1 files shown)


orelgueta (approved):
We actually used to have something similar to get the latest tag, but it's not in the code anymore. Either way, this looks fine and affects only tests and operational code.

PR #1192, opened 2024-10-09
New array trigger definition

GernotMaier:
Implementation of new array trigger definition, see also discussions in model_parameter repository ([issue](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters/-/issues/28) and [merge request]()).

- change from file-type definition to explicit listing per telescope and trigger (stereo/mono) type in array_trigggers parameter
- updated json schema for the array_triggers dictionary
- added format checker for json_schema that trigger names are consistent and array element types are correct (for this: added a function in names.py to verify the array element type)

**Preliminary: this will make the integration tests fail until the code to write the simtel trigger configuration files is implemented**


orelgueta (approved):
You didn't officially ask for a review, but I came here from GitLab and it all looks OK so I approve.

PR #1195, opened 2024-10-10
Add CI to update developer software on a schedule

GernotMaier:
We run `pre-commit autoupdate` by hand in irregular intervals - mostly when we realize that there are issues or warnings are printed (right now there are warnings printed).

Add a CI which runs every Monday with pre-commit autoupdate. If necessary, the CI generates a pull request.

Also updated the current pre-commit.

Note that for testing, I didn't update the black version. As soon as this PR has been merged, we can test it to make sure that it picks up the missing update.


tobiaskleiner (approved):
Thanks @GernotMaier, good addition.

PR #1199, opened 2024-10-11
Refactor validate_data_dict and improved testing

### LINKED ISSUE #1193: Refactor validate_data._validate_data_dict
This function is too complex and needs to be simplified / split in several functions.

see https://github.com/gammasim/simtools/blob/1c3acd9c1e7498214527db7d3045aa2447804e5d/simtools/data_model/validate_data.py#L114

Also aim for 100% coverage in this case.

GernotMaier:
This is refactor PR with no change in functionality.

 `data_model.validate_data._validate_data_dict` is too complex and we split here the function into several. Also address the issue of a missing test for the call of `_validate_data_dict_using_json_schema`.

Closes #1193

### DIFF
--- a/simtools/data_model/validate_data.py
+++ b/simtools/data_model/validate_data.py
@@ -183,10 +183,15 @@ class DataValidator:
             if isinstance(self.data_dict["unit"], list | np.ndarray)
             else [self.data_dict["unit"]]
         )
-        if isinstance(value_as_list, np.ndarray):
+        try:
             value_as_list = value_as_list.tolist()
-        if isinstance(unit_as_list, np.ndarray):
+        except AttributeError:
+            pass
+        try:
             unit_as_list = unit_as_list.tolist()
+        except AttributeError:
+            pass
+
         unit_as_list = [None if unit == "null" else unit for unit in unit_as_list]
 
         return value_as_list, unit_as_list
--- a/simtools/data_model/validate_data.py
+++ b/simtools/data_model/validate_data.py
@@ -130,7 +130,49 @@ class DataValidator:
             raise KeyError("Data dict does not contain a 'name' or 'parameter' key.")
         self._data_description = self._read_validation_schema(self.schema_file_name, _name)
 
-        # validation assumes lists for values and units - convert to list if required
+        value_as_list, unit_as_list = self._get_value_and_units_as_lists()
+
+        for index, (value, unit) in enumerate(zip(value_as_list, unit_as_list)):
+            value_as_list[index], unit_as_list[index] = self._validate_value_and_unit(
+                value, unit, index
+            )
+
+        if len(value_as_list) == 1:
+            self.data_dict["value"], self.data_dict["unit"] = value_as_list[0], unit_as_list[0]
+
+        self._check_version_string(self.data_dict.get("version"))
+
+    def _validate_value_and_unit(self, value, unit, index):
+        """
+        Validate value, unit, and perform type checking and conversions.
+
+        Take into account different data types and allow to use json_schema for testing.
+        """
+        if self._get_data_description(index).get("type", None) == "dict":
+            self._validate_data_dict_using_json_schema(
+                self.data_dict["value"], self._get_data_description(index).get("json_schema")
+            )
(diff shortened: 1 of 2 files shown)


simtools/data_model/validate_data.py
```diff
@@ -141,26 +183,13 @@ def _validate_data_dict(self):
             if isinstance(self.data_dict["unit"], list | np.ndarray)
             else [self.data_dict["unit"]]
         )
+        if isinstance(value_as_list, np.ndarray):
+            value_as_list = value_as_list.tolist()
+        if isinstance(unit_as_list, np.ndarray):
+            unit_as_list = unit_as_list.tolist()
```
orelgueta, line 189: The following would be more pythonic:

```
try:
    value_as_list = value_as_list.tolist()
except AttributeError:
    pass
```

tests/unit_tests/data_model/test_validate_data.py
```diff
@@ -771,3 +771,56 @@ def test_validate_data_dict_using_json_schema(caplog):
...
+    data_validator.data_dict = {"value": [100, 200], "unit": ["m", "null"]}
+    values, units = data_validator._get_value_and_units_as_lists()
+    assert values == [100, 200]
+    assert units == ["m", None]
+
+    # Test with mixed types
+    data_validator.data_dict = {"value": [100, 200], "unit": np.array(["m", "cm"])}
+    values, units = data_validator._get_value_and_units_as_lists()
+    assert values == [100, 200]
+    assert units == ["m", "cm"]
```
orelgueta, line 807: I think this is already tested above in the second test (I assume copilot did this). You can change the first one to be all with the same unit or just delete this.

tests/unit_tests/data_model/test_validate_data.py
```diff
@@ -771,3 +771,56 @@ def test_validate_data_dict_using_json_schema(caplog):
...
+
+
+def test_validate_value_and_unit_for_dict(reference_columns):
+    data_validator = validate_data.DataValidator()
+    data_validator._data_description = reference_columns
+
+    # Test case for dict type
+    data_validator.data_dict = {"value": {"key": "value"}, "unit": None, "type": "dict"}
```
orelgueta, line 815: Shouldn't the unit here be `"null"`? I don't remember which of the options we decided to stick with, but either way it is inconsistent with the example in the tests above.

orelgueta (commented):
A couple of minor comments

simtools/data_model/validate_data.py
GernotMaier: But this would replace two lines with four lines? (for both value_as_list and unit_as_list).

Would that really be more pythonic? (I don't care, but I don't like more lines)

tests/unit_tests/data_model/test_validate_data.py
GernotMaier: Removed it. Indeed a copilot invention.

tests/unit_tests/data_model/test_validate_data.py
GernotMaier: You are correct, we are using everywhere "null", so I've added a test for "null".

I want to keep the test for `None`, as the "null" is used solely because of the usage of string-type arrays (as in sim-telarray). I hope that on a medium time scale, we can move away from these strings.

GernotMaier:
@orelgueta - thanks for the review!

Let me know if it is ok to keep the if-statements. I really don't care, but have a slight preference to less lines of code.

simtools/data_model/validate_data.py
orelgueta: The "pythonic way" is less about the number of lines but on number of operations or the old saying "easier to ask forgiveness than permission". So I would vote for try..except. I am approving anyway though.

orelgueta (approved):
Approving, but stating my preference for a change.

simtools/data_model/validate_data.py
GernotMaier: ok...added the try and except although I am not convinced at all. Thanks!

PR #1201, opened 2024-10-14
Add format checkers for astropy units of time and length

GernotMaier:
This pull request introduces new format checkers for astropy units of time and length, updates the schema to use these new format checkers, and adds corresponding unit tests.

Closes #1200 

### New Format Checkers:
* Added `check_astropy_unit_of_time` to validate astropy units of time. (`simtools/data_model/format_checkers.py`)
* Added `check_astropy_unit_of_length` to validate astropy units of length. (`simtools/data_model/format_checkers.py`)

### Schema Updates:
* Updated the schema to use `astropy_unit_of_time` for time units. (`simtools/schemas/model_parameters/array_triggers.schema.yml`)
* Updated the schema to use `astropy_unit_of_length` for length units. (`simtools/schemas/model_parameters/array_triggers.schema.yml`)

### Unit Tests:
* Added tests for `check_astropy_unit_of_time` and `check_astropy_unit_of_length` to ensure proper validation and error handling. (`tests/unit_tests/data_model/test_format_checkers.py`)
* Imported `re` and `UnitConversionError` to support the new tests. (`tests/unit_tests/data_model/test_format_checkers.py`)

PR #1202, opened 2024-10-14
Remove unneeded if statement

GernotMaier:
Removes an if statement for array_triggers.width value and unit which is not needed. The two checks are done already in the data validation using jsonschema and are not needed to be repeated.

### DIFF
--- a/simtools/simtel/simtel_config_writer.py
+++ b/simtools/simtel/simtel_config_writer.py
@@ -365,11 +365,8 @@ class SimtelConfigWriter:
             trigger_dict = self._get_array_triggers_for_telescope_type(array_triggers, tel_type)
             trigger_lines[tel_type] = f"Trigger {trigger_dict['multiplicity']['value']} of "
             trigger_lines[tel_type] += ", ".join(map(str, tel_list))
-            if all(trigger_dict["width"][key] is not None for key in ["value", "unit"]):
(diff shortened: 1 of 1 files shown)


simtools/simtel/simtel_config_writer.py
```diff
@@ -365,11 +365,8 @@ def _write_array_triggers_file(self, array_triggers, model_path, telescope_model
             trigger_dict = self._get_array_triggers_for_telescope_type(array_triggers, tel_type)
             trigger_lines[tel_type] = f"Trigger {trigger_dict['multiplicity']['value']} of "
             trigger_lines[tel_type] += ", ".join(map(str, tel_list))
-            if all(trigger_dict["width"][key] is not None for key in ["value", "unit"]):
-                width = trigger_dict["width"]["value"] * u.Unit(trigger_dict["width"]["unit"]).to(
-                    "ns"
-                )
-                trigger_lines[tel_type] += f" width {width}"
+            width = trigger_dict["width"]["value"] * u.Unit(trigger_dict["width"]["unit"]).to("ns")
```
orelgueta, line 368: A unit will also be provided always, right? If it is empty you would get `dimensionless` instead I assume.

orelgueta (approved):
A small question, approving already

simtools/simtel/simtel_config_writer.py
GernotMaier: The json schema requires an astropy_unit (and with PR #1201 even an astropy unit of time)

PR #1207, opened 2024-10-14
Pre commit autoupdate - add simtools installation

GernotMaier:
Pre-commit autoupdate (see PR #1195) run a `pre-commit --all-files` after updating the pre-commit software.

This is good (as we can see if simtools passes any updated pre-commit tools), but requires an installation of the simtools environment. This PR adds a simple pip install to do this.


tobiaskleiner (approved):
Thanks for the update @GernotMaier! All good.

PR #1208, opened 2024-10-14
Astropy units bugfix

EshitaJoshi:
This PR fixes a bug in the function `.get_parameter_value_with_unit()` where the units were strings (eg. "['m', 's', 'null']") instead of a list. Null entries within this list of units were not being returned as dimensionless, but were instead raising the following error: `ValueError: 'cm, deg, null, null' did not parse as unit: At col 9, null is not a valid unit.  If this is meant to be a custom unit, define it with 'u.def_unit'. To have it recognized inside a file reader or other code, enable it with 'u.add_enabled_units'. For details, see https://docs.astropy.org/en/latest/units/combining_and_defining.html`

### Changes:

- Units are now being converted from a string to a comma-separated list. This was not done using `simtools.utils.general.convert_string_to_list()` as some units have spaces within them, which this function would interpret as different entries in a list.
- Null units in lists are treated as dimensionless.
- A list of astropy quantities is returned if values are a list.
- `AttributeError` added for the case where units are not provided (`NoneType`), in which case the function returns the value without a unit.

### Unit-testing checks for:

- Lists of values where some values have null units (eg. [cm, deg, null, null])
- Lists of values with a single unit
- Units with spaces in them
- Values without units are not returned as astropy quantities
- Values with units are returned as astropy quantities


tests/unit_tests/model/test_model_parameter.py
```diff
@@ -87,6 +87,20 @@ def test_get_parameter_value(telescope_model_lst):
 def test_get_parameter_value_with_unit(telescope_model_lst):
     tel_model = telescope_model_lst
 
+    # check handling of list of values and units including null units
+    t_1 = tel_model.get_parameter_value_with_unit("focus_offset")
+    assert isinstance(t_1, list)
+    assert isinstance(t_1[0], u.Quantity)  # list of quantities returned
```
GernotMaier, line 93: Do you want to test in this case as well that unit  'null' in `t_1[2]` is converted correctly?

GernotMaier (approved):
Thanks for this!

A tiny suggestion of adding one more test, but otherwise ready to be merged.

tests/unit_tests/model/test_model_parameter.py
EshitaJoshi: Good idea, done.

PR #1221, opened 2024-10-30
Introduce selective log levels via fixture

tobiaskleiner:
This PR adds a fixture to set the log level context. In some cases the context was added where it was previously missing. In some cases log levels where adjusted to better reflect the log message.

This adresses the problem that previously we set `logger.setLevel(logging.DEBUG) `for almost all unit tests.

### DIFF
--- a/simtools/ray_tracing.py
+++ b/simtools/ray_tracing.py
@@ -75,7 +75,7 @@ class RayTracing:
     ):
         """Initialize RayTracing class."""
         self._logger = logging.getLogger(__name__)
-        self._logger.debug("Initializing RayTracing class")
+        self._logger.info("Initializing RayTracing class")
 
         self.simtel_path = Path(simtel_path)
         self._io_handler = io_handler.IOHandler()
@@ -155,7 +155,7 @@ class RayTracing:
         tuple
             Tuple containing source distance and mirror numbers.
         """
-        self._logger.debug(
+        self._logger.info(
             "Single mirror mode is activated - "
             "source distance is being recalculated to 2 * focal length "
             " (this is not correct for dual-mirror telescopes)."
--- a/simtools/ray_tracing.py
+++ b/simtools/ray_tracing.py
@@ -75,7 +75,7 @@ class RayTracing:
     ):
         """Initialize RayTracing class."""
         self._logger = logging.getLogger(__name__)
-        self._logger.info("Initializing RayTracing class")
+        self._logger.debug("Initializing RayTracing class")
 
         self.simtel_path = Path(simtel_path)
         self._io_handler = io_handler.IOHandler()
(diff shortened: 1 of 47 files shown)


simtools/ray_tracing.py
```diff
@@ -75,7 +75,7 @@ def __init__(
     ):
         """Initialize RayTracing class."""
         self._logger = logging.getLogger(__name__)
-        self._logger.debug("Initializing RayTracing class")
+        self._logger.info("Initializing RayTracing class")
```
GernotMaier, line 78: I would suggest to keep this on the debug level. We don't need to know each time with initialize RayTracing that this is happening. This is debug information.

simtools/ray_tracing.py
```diff
@@ -155,7 +155,7 @@ def _initialize_single_mirror_mode(self, mirror_numbers):
         tuple
             Tuple containing source distance and mirror numbers.
         """
-        self._logger.debug(
+        self._logger.info(
```
GernotMaier, line 158: Also think this should be on the debug level, as this is a comment to be used for developers (we have an open issue on it).

simtools/utils/general.py
```diff
@@ -165,7 +165,7 @@ def collect_data_from_file_or_dict(file_name, in_dict, allow_empty=False):
         return dict(in_dict)
 
     if allow_empty:
-        _logger.debug("Input has not been provided (neither by file, nor by dict)")
+        _logger.warning("Input has not been provided (neither by file, nor by dict)")
```
GernotMaier, line 168: Keep this at debug: we explicitly `allow_empty`, so there is no need to warn that it actually happens.

simtools/utils/general.py
```diff
@@ -816,7 +816,7 @@ def read_file_encoded_in_utf_or_latin(file_name):
         with open(file_name, encoding="utf-8") as file:
             lines = file.readlines()
     except UnicodeDecodeError:
-        logging.debug("Unable to decode file using UTF-8. Trying Latin-1.")
+        logging.error("Unable to decode file using UTF-8. Trying Latin-1.")
```
GernotMaier, line 819: Keep this as debug as it is not an `error`. We continue trying to read the file taking the new circumstances into accounts. Errors are raised laterl

GernotMaier (changes requested):
Thanks for this! Clearly we don't want to have log-level settings interfering between tests.

Couple of comments below:

- some changes of log levels are not correct
- (discussed in person) that we want to use consistently `with caplog.at_level(logging.WARNING):` blocks.

simtools/ray_tracing.py
tobiaskleiner: Okey, set them now back to debug, also the following three you commented on.

tobiaskleiner:
Thanks @GernotMaier for the review. I have now reverted back to with statements and added a few more where they were previously missing. Also now there should be no more general statements to set log levels globally. Let me know if everything is alright now.

PR #1237, opened 2024-11-12
removing hardcoding of sim_telarray path

EshitaJoshi:
The integration tests were using a hardcoded sim_telarray path instead of the one set in the .env file.
This PR is a minor change to remove this hardcoding.


GernotMaier (approved):
All good, thanks!

PR #1238, opened 2024-11-14
Move functions from test_applications_from_config.py to simtools.testing

### LINKED ISSUE #1174: Move functions from `test_applications_from_config.py` to `simtools.testing`
The main function in `test_applications_from_config.py` was split into smaller functions. We should move those to `simtools.testing` and write unit tests for them.

GernotMaier:
The integration test application `test_applications_from_config.py` contains rather complicated functions to prepare, run, and validate the integration tests.

This PR moves:

- all methods to corresponding modules in the `simtools.testing` area
- new modules are `testing.configuration` and `testing.helpers`
- renamed the module `testing.compare_output` to `testing.validate_output`
- added unit tests for all modules

Side note: needed to move the pytest package from the test section to the general in pyproject toml, as some of modules in simtools.testing are importing pytest.

I suggest not to do a line-by-line code review, but review the general refactoring and naming. No new functionality has been added, but code moved around. Also not sure if we need a line-by-line review of units tests.

Closes #1174


simtools/testing/configuration.py
```diff
@@ -0,0 +1,225 @@
...
+    model_version_config = config["CONFIGURATION"]["MODEL_VERSION"]
+    if model_version_requested != model_version_config:
+        pytest.skip(
+            f"Model version requested {model_version_requested} not supported for this test"
+        )
+
+
+def _prepare_configuration(config, output_path, model_version=None):
+    """
+    Prepare test configuration.
```
tobiaskleiner, line 133: Naming and desciption is very similar to above function `configuration`. Suggest to rename it and also to clarify what `configuration string` means in the docstring.

simtools/testing/configuration.py
```diff
@@ -0,0 +1,225 @@
...
+                {
+                    "APPLICATION": app,
+                    "TEST_NAME": "auto-version",
+                    "CONFIGURATION": {"VERSION": True},
+                },
+                {"APPLICATION": app, "TEST_NAME": "auto-no_config"},
+            ]
+        )
+
+    if get_test_names:
```
tobiaskleiner, line 54: Suggest to always return both, to avoid calling it twice in `test_application_from_config`.

tobiaskleiner (commented):
@GernotMaier thanks for the PR! The changes look good overall, I have added few minor comments.

simtools/testing/configuration.py
GernotMaier: Renamed it to `_prepare_test_options`. Improved also the docstring.

simtools/testing/configuration.py
GernotMaier: Good point, I've changed the logic accordingly.

GernotMaier:
@tobiaskleiner - I think I've addressed your suggestions. Let me know if there is anything else I can do.

Thanks for the review!

PR #1241, opened 2024-11-18
Type hints and input validation

tobiaskleiner:


simtools/model/model_utils.py
```diff
@@ -37,19 +42,19 @@ def compute_telescope_transmission(pars, off_axis):
...
     ----------
     telescope_model_name: str
-        Telescope model name (ex. LSTN-01).
+        Name of the telescope model.
 
     Returns
     -------
     bool
-        True if the telescope is a two mirror one.
+        True if it is a two-mirror telescope, False otherwise.
```
GernotMaier, line 57: As booleans have only two states, it is probably enough to describe the Trues state (no need of `False otherwise`)

GernotMaier:
Following https://github.com/gammasim/simtools/blob/main/simtools/schemas/model_parameters/telescope_transmission.schema.yml , we should allow for 6 parameters.

tobiaskleiner:
> Following https://github.com/gammasim/simtools/blob/main/simtools/schemas/model_parameters/telescope_transmission.schema.yml , we should allow for 6 parameters.

from the schema:

```
The $p_5$ parameter may not be available/supported and is effectively
  zero/unused if not.
```

So I should raise an error if len(pars) is other than 5 or 6?

PR #1250, opened 2024-11-25
Simplify yaml/json file reading

GernotMaier:
This is a small maintenance PR which simplifies the often used function `general.collect_data_from_file_or_dict` to `general_collect_data_from_file`.

Reason for changes:

- simplification
- the second argument of `in_dict` in general.collect_data_from_file_or_dict` is counterintuitive, as all the function does is to return the original dict (so why call it at all?)
- the second argument was used once only in the whole code base and not using it actually simplifies the code


tobiaskleiner (approved):
Hi @GernotMaier, thanks! Changes look good and this makes the reading more straight-forward.

GernotMaier:
@tobiaskleiner - thank you for the quick review!

PR #1253, opened 2024-11-26
type hints for models

tobiaskleiner:
This is a first of probably several PRs that add type hinting to the codebase, this should make developping and debuging easier.


simtools/model/array_model.py
```diff
@@ -108,7 +108,7 @@ def _initialize(self, site, array_elements_config):
             array_elements = self._get_array_elements_from_list(
                 site_model.get_array_elements_for_layout(self.layout_name)
             )
-        if array_elements == {}:
+        if not array_elements:
```
GernotMaier, line 111: Suggest to check as well for the empty dict - we need to have array elements to continue.

simtools/model/camera.py
```diff
@@ -392,34 +400,30 @@ def _find_neighbors(x_pos, y_pos, radius):
...
-        neighbors: numpy.array_like
-            Array of neighbor indices in a list for each e.g., pixel.
+        list of lists
+            Array of neighbor indices in a list for each pixel
         """
-        points = np.array([x_pos, y_pos]).T
-        indices = np.arange(len(x_pos))
-        kdtree = KDTree(points)
-        neighbors = [kdtree.query_ball_point(p, r=radius) for p in points]
+        tree = KDTree(np.column_stack([x_pos, y_pos]))
```
GernotMaier, line 406: As these lines have been changed, would it be possible to add unit test coverage? This would make the change much clearer.

simtools/model/model_utils.py
```diff
@@ -37,19 +37,19 @@ def compute_telescope_transmission(pars, off_axis):
...
     ----------
     telescope_model_name: str
-        Telescope model name (ex. LSTN-01).
+        Name of the telescope model.
 
     Returns
     -------
     bool
-        True if the telescope is a two mirror one.
+        True if it is a two-mirror telescope, False otherwise.
```
GernotMaier, line 52: Mentioned that before 'False otherwise' is not needed for the description of a boolean.

GernotMaier (approved):
Thanks a lot for that - this improves readability significantly.

I am approving, but there are two minor comments which would be great if they are addressed.

simtools/model/array_model.py
tobiaskleiner: `not array_elements` evaluates to True for any empty dict, list, str etc. This is more general than just a check for empty dict.

GernotMaier: Yes - you are right!

PR #1254, opened 2024-11-26
Add uc 08 part 2 computing resources

tobiaskleiner:
This is the second split part of the UC8 PR containting the computing and storage estimate application.


simtools/applications/production_calculate_resource_estimates.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/python3
+
+r"""
+Estimate the required computing resources for a simulation based on command-line arguments.
```
GernotMaier, line 4: Remove 'based on command-line arguments'. Not sure what you mean with that? All simtools are configured through the command line.

simtools/applications/production_calculate_resource_estimates.py
```diff
@@ -0,0 +1,159 @@
...
+        help="Name of the file to save the resource estimates.",
+        type=str,
+        required=False,
+        default="estimated_resources.json",
+    )
+
+    return config.initialize(db_config=False)
+
+
+def load_existing_data(file_path: str) -> list:
```
GernotMaier, line 95: please use utils.general.collect_data_from_file to read from a yaml file.

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+    Attributes
+    ----------
+    grid_point : dict
+        Dictionary containing parameters such as energy (TeV), azimuth (deg),
+        zenith (deg), and night sky background.
+    simulation_params : dict
+        Dictionary containing simulation parameters, including the number of events and site.
+    existing_data : list of dict, optional
+        List of dictionaries with historical data of compute and storage
+          resources for existing simulations.
```
GernotMaier, line 26: `lookup_file` missing in attribute description.

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+        ----------
+        lookup_file : str
+            Path to the YAML file containing the lookup table.
+
+        Returns
+        -------
+        dict
+            Dictionary containing the lookup table.
+        """
+        with open(lookup_file, encoding="utf-8") as file:
```
GernotMaier, line 69: Please use collect_data_from_file from the general module.

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+            lookup = yaml.safe_load(file)
+
+        # return as quantities
+        for _, data in lookup.items():
```
GernotMaier, line 73: `for data in lookup.value()`

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+            for _, resources in data["Zenith"].items():
```
GernotMaier, line 74: Same here, use `values()`

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+                resources["compute_per_event"] = u.Quantity(
```
GernotMaier, line 75: Avoid duplication:

```
        for key in ["compute_per_event", "storage_per_event"]:
            resources[key] = u.Quantity(
                resources[key]["value"], resources[key]["unit"]
            )
```

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Returns
+        -------
+        dict
+            A dictionary with guessed estimates for compute and storage resources, with units.
+        """
+        zenith = self.grid_point["zenith"]
+        zeniths = sorted(self.lookup_table[self.site]["Zenith"].keys())
+
+        if zenith <= zeniths[0]:
```
GernotMaier, line 160: Are we sure that we want to allow `<` or `>` in this context? This might mean that resource estimates are off from the actual needs if the zenith angle is very different from those available in the table. I would suggest not to allow that and issue an exception?

simtools/production_configuration/derive_computing_resources.py
```diff
@@ -0,0 +1,207 @@
...
+                "compute_per_event"
+            ]
+            storage_per_event = self.lookup_table[self.site]["Zenith"][zeniths[-1]][
+                "storage_per_event"
+            ]
+        else:
+            lower_bound = max(e for e in zeniths if e <= zenith)
+            upper_bound = min(e for e in zeniths if e >= zenith)
+            lower_values = self.lookup_table[self.site]["Zenith"][lower_bound]
+            upper_values = self.lookup_table[self.site]["Zenith"][upper_bound]
```
GernotMaier, line 178: Please fix the missing coverage.

tests/integration_tests/config/production_calculate_resource_estimates.yml
GernotMaier, line 1: Can you add also an integration test for using the `--existing_data` command line parameters?

Wondering also if we should test the results? But maybe not needed, as the resource computation is tested in the unit tests.

GernotMaier (changes requested):
Thanks a lot! So much easier to review when the scope is limited to one application.

Below some minor comments on details. My main issue is that this approach relies that lookup table of previous estimates are done with exactly the same MC parameters as the planned productions. This includes especially MC energy range and spectral index, but also e.g. number of telescopes, NSB, etc. There is also a dependency on the primary.

Maybe best to discuss this in person?

tobiaskleiner:
> Thanks a lot! So much easier to review when the scope is limited to one application.
> 
> Below some minor comments on details. My main issue is that this approach relies that lookup table of previous estimates are done with exactly the same MC parameters as the planned productions. This includes especially MC energy range and spectral index, but also e.g. number of telescopes, NSB, etc. There is also a dependency on the primary.
> 
> Maybe best to discuss this in person?

@GernotMaier , thanks for the review. The implementation allows for interpolation of the storage and computing requirements, which means that we would only require a wide enough coverage of the parameter space in zenith, az, NSB etc. We also have a first estimate of a comparison for the compute and storage requirements between the alpha and the hyperarray that we planned to use for interpolation for the telescope numbers. Repeating the test simulations (with additionally maybe smaller array setups) would require that the importance sampling is ready. Let's talk in person about this and I wait until then with your other comments.

GernotMaier:
Closing the merge requests as they are superseded by other development.

PR #1258, opened 2024-11-26
Update README.md

tobiaskleiner:
Add Eshita to README


GernotMaier (approved):
Very good.

But need to check that we don't duplicated too much. I think we have already the CITATION.cff file and could this list remove (this would be my suggestion)

PR #1267, opened 2024-12-03
Simple plotting application for tabular data

GernotMaier:
Add an application to plot tabular data from file.

- allow to read a file from the file system or from the model parameter database
- introduce a new module called `io_operations:legacy_data_handler` to allow reading of the many legacy tables provided in different data formats. This module should hide the complexity of the reading of these files and return in all cases astropy tables. This approach ensures that the knowledge of the legacy data formats is needed in one place only and we can write all other code in the expectation to get tables. This is a start and implemented only for the plotting of the single pe amplitudes. It is expected that many further adaptions are needed for the different data we want to plot.
- introduce a new module `simtel:simtel_table_reader` to read all tables available through the model database. This might serve in future as base for moving the data structure towards astropy tables.
- allow to configure plots through a configuration files. This is required, as plotting always require many options which would overwhelm the command line.

Added several examples for plotting configuration into tests/resources/plot_table_data/. I suggest to look into some of them and run

```
python src/simtools/applications/plot_tabular_data.py --output_file my_plot --plot_config tests/resources/plot_table_data/plot_atmospheric_transmission_north_south.yml
```

Note some of the options:

- `NORMALIZE_Y` to normalize the data
- 'SELECT_VALUES` to select from a n-dimensional tables certain values for plotting (see plot_atmospheric_transmission_south.yml for application).

This is not complete, but the PR is getting big, some missing points are therefore listed in issue #1284.

Finally: attached a list of file of all "Files" into the model parameter DB. All with "+" in the first column are covered by this PR, all other files are actually not model parameters.

[PLOT_FILES.txt](https://github.com/user-attachments/files/18166579/PLOT_FILES.txt)


EshitaJoshi (approved):
Didn't see any problems with the code and the plots look really nice. Thanks for the PR!

GernotMaier:
Thanks @EshitaJoshi  for the review of this big PR!

PR #1276, opened 2024-12-04
Minor changes to model parameter report generating

GernotMaier:
(this is a intermediate PR to be merged into the `model_reports` branches; full review should be applied before `model_reports` is merged to main)

This PR includes changes to align PR #1185 with the general simtools code base:

- use of db functions to read files form db
- use pathlib to make directories (if possible, we don't use subprocess)

Fixed a bug resulting in the addition of only list- or file type model parameters. Single values were excluded (e.g., `num_gains`).

Added also a new function to names to get the list of 'classes' for the instrument type (e.g., for `telescope`, this returns `("Structure", "Camera", "Telescope"))`. This allows to define this list once throughout the code (and avoid mistakes in future when these lists are changing.

### DIFF
--- a/src/simtools/applications/docs_produce_reports.py
+++ b/src/simtools/applications/docs_produce_reports.py
@@ -9,7 +9,6 @@ The reports are then uploaded as GitLab Pages using GitLab's CI workflow.
 """
 
 import logging
-import subprocess
 from itertools import groupby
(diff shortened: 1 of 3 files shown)


EshitaJoshi (approved):
Thank you for the PR! It's much cleaner and reads better now.

PR #1277, opened 2024-12-05
CTAO CR spectra package now available from conda

GernotMaier:
Packages in `environment.yml` should preferably be installed through conda. Add `ctao-dpps-cosmic-ray-spectra`, which is since yesterday available through conda.


EshitaJoshi (approved):
Thanks for the PR!

PR #1281, opened 2024-12-13
Improve db configuration checks

GernotMaier:
This is a minor update of inconsistent behavior when running simtools with incomplete database configuration. Essentially the code didn't notice at the configuration step that something was wrong, but much later when trying to call the DB.

Two changes:

- added a jsonschema based validation of the DB configuration dictionary in db_handler. This handles also the two optional parameter correctly
- fix a small bug in the setting of the default db_configuration in the configurator: if any parameter was missing (even an optional one), an error was thrown.

### DIFF
--- a/src/simtools/db/db_handler.py
+++ b/src/simtools/db/db_handler.py
@@ -35,6 +35,7 @@ jsonschema_db_dict = {
             "type": "integer",
             "minimum": 1,
             "maximum": 65535,
+            "default": 27017,
             "description": "Port to use",
         },
         "db_api_user": {"type": "string", "description": "API username"},
--- a/src/simtools/db/db_handler.py
+++ b/src/simtools/db/db_handler.py
@@ -26,7 +26,7 @@ logging.getLogger("pymongo").setLevel(logging.WARNING)
 
 
 jsonschema_db_dict = {
-    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$schema": "https://json-schema.org/draft/2020-12/schema#",
     "type": "object",
     "description": "MongoDB configuration",
     "properties": {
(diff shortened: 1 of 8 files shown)


src/simtools/db/db_handler.py
```diff
@@ -24,20 +25,47 @@
 # The above comment is because pylint does not know that DatabaseHandler.db_client is subscriptable
 
 
+jsonschema_db_dict = {
+    "$schema": "https://json-schema.org/draft-07/schema#",
```
EshitaJoshi, line 29: I seem to get a warning from the `jsonschema` library `DeprecationWarning: The metaschema specified by $schema was not found. Using the latest draft to validate, but this will raise an error in the future.` Do you get this too?

GernotMaier: Thanks a lot for that - I didn't see the warning when I worked on the code (I can only see it when I run pytest with the "-s" option.

I've updated the jsonschema version throughout the code (in some schema files unrelated to the DB).

Fixed also one issue noticed now when running tests without a local '.env' file (which we anyway shouldn't; but let's keep the code robust)

src/simtools/db/db_handler.py
```diff
@@ -24,20 +25,47 @@
 # The above comment is because pylint does not know that DatabaseHandler.db_client is subscriptable
 
 
+jsonschema_db_dict = {
+    "$schema": "https://json-schema.org/draft/2020-12/schema#",
+    "type": "object",
+    "description": "MongoDB configuration",
+    "properties": {
+        "db_server": {"type": "string", "description": "DB server address"},
+        "db_api_port": {
```
orelgueta, line 34: Should we add a default here as well? We usually only use 27017.

orelgueta (commented):
Skimmed through it, added one small comment, not blocking.

src/simtools/db/db_handler.py
GernotMaier: Done.

GernotMaier:
@EshitaJoshi  - could you have another look? I've fixed the schema url and added the suggestion by Orel. Note that I have to look into the other schema in src/simtools/schema for up-to-date schema urls. I do that in a separate PR.

EshitaJoshi (approved):
Looks good, thanks for the PR!

PR #1283, opened 2024-12-16
Replace os.system by subprocess.run and Pathlib commands.

GernotMaier:
Two maintenance issue of minor importance:

- replaces os.system (declared as deprecated, see [here](https://docs.python.org/3/library/subprocess.html)) by subprocess.run, which allows also better control and less reliance on shell commands. subprocess also allows a cleaner handling of stdout and stderr
- replaces file system access to Pathlib commands (as we usually do).

Additionally

- switched on ruff checker PTH (flake8-pathlib), which should helps us to keep this consistent in future

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
         args: ["--line-length=100"]
   - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff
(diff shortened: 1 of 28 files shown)


EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thank you @EshitaJoshi !

PR #1285, opened 2024-12-18
Move test_commandline_tools as test after simtools-prod container generation.

### LINKED ISSUE #1229: test_commandline_tools in CI-Integration tests fails when adding new applications
This is an issue related to failing CI-integration tests when adding new applications.

The failure is due to the usage for the simtools-prod container, which has not been updated yet during a pull request and therefore does not include the new application.

Not sure what the best solutions are:

- run this test as an additional step in the docker container generation CI ([build-docker-images.yml](.github/workflows/build-docker-images.yml)
- run this test only for merge-to-main (after the CI docker generation)
- drop the test? I am not sure actually what we are testing? We ensure that the applications defined in pyproject.toml exist? That should always be the case, otherwise `pip install` would not work (which we test in `testbuilding` in the same workflow).

Any input?

GernotMaier:
Move the `test_commandline_tools` from the integration test workflow to the `build-docker-images` workflow. The tests are now run after a new docker image is built and push. This ensures that the test finds all applications, even those added in a pull request.

Closes #1229


GernotMaier:
This works successfully, see https://github.com/gammasim/simtools/actions/runs/12390525072

GernotMaier:
Stupid me...for pull requests we don't push docker images and therefore it doesn't work (see https://github.com/gammasim/simtools/actions/runs/12395361896/job/34601535791). I tested it triggering the workflows by hand (and then we push). Will have to figure out something differently.

GernotMaier:
@EshitaJoshi - this is now ready for review. The solution is quite simple in the end.

For the successful tests, see e.g. https://github.com/gammasim/simtools/actions/runs/12399747739

EshitaJoshi (approved):
Thank you for the PR!

PR #1289, opened 2024-12-18
Handle return codes correctly for simulation productions

### LINKED ISSUE #1146: Skip `pack_for_grid_register` if the simulation fails
In the application `simulate_prod`, if the simulation fails, the `pack_for_grid_register` step still runs and copies files to the local directory. That's not the expected behaviour. It's best to skip this step if `simulator.simulate()` fails. It would also avoid printing a "success" message which is currently printed regardless of the outcome.

GernotMaier:
The `simulate_prod` application calls CORSIKA / sim_telarray but had no handling of the return codes when calling those processes. This means failed simulation runs could still be labeled as successful.

This PR collect the return value from `subprocess.run` and evaluates it. For any return values unequal zero, the code raises now a `JobExecutionError`.

This fixes also the issue "Skip pack_for_grid_register if the simulation fails" #1146.

Closes #1146

### DIFF
--- a/docs/changes/1289.bugfix
+++ b/docs/changes/1289.bugfix
@@ -1,0 +1,1 @@
+Add correct handling of return codes from simulation software to prevent reports of success even though a job failed (`simulate_prod`)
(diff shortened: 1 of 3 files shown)


tobiaskleiner (approved):
@GernotMaier thanks - all good.

GernotMaier:
@tobiaskleiner - thanks again!

PR #1291, opened 2024-12-19
Set correct activity:end in metadata

### LINKED ISSUE #528: Set correct activity:end time
We currently set in the metadata `activity_dict['end'] = activity_dict['start']`, which is not correct (especially for tools which takes a bit longer. Fix this and set the correct execution time.

_Originally posted by @GernotMaier in https://github.com/gammasim/simtools/pull/490#discussion_r1289684186_

GernotMaier:
The metadata collector sets currently activity:end to activity:start time.

Fix this by introducing a new function MetadataCollector::get_top_level_metadata() which updates the activity:end time. This function is usually called when passing the metadata dictionary on to the data writer.

Closes #528

### DIFF
--- a/docs/changes/1291.bugfix
+++ b/docs/changes/1291.bugfix
@@ -1,0 +1,1 @@
+Fix setting of activity:end time in the metadata collector.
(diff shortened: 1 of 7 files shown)


tobiaskleiner (approved):

GernotMaier:
@tobiaskleiner - thank you!

PR #1308, opened 2025-01-14
Simplify `db_handler.add_new_parameter` and add parameter dict validation before writing

GernotMaier:
This PR changes / adds two items:

- simplification of the API of `db_handler.add_new_parameter`: the current/old implementation had an imprint of the model parameter dictionary definition in its API. We now pass on directly the diff.
- generalized the validation of the parameter dictionary before writing to the database: use now the corresponding schema files and the data validation routines


tests/unit_tests/db/test_db_handler.py
```diff
@@ -212,64 +212,76 @@ def test_copy_array_element_db(db, random_id, io_handler, model_version):
...
+        "file": False,
+    }
+
+    par_dict_int = copy.deepcopy(tmp_par_dict)
+    par_dict_int["parameter"] = "num_gains"
+    par_dict_int["value"] = 3
+    par_dict_int["type"] = "int64"
+    with pytest.raises(
+        ValueError,
+        match=re.escape("Value for column '0' out of range. ([3, 3], allowed_range: [1, 2])"),
```
tobiaskleiner, line 233: Why do we get here a list [3, 3]?

GernotMaier: The validation code converts everything into astropy table columns (even single numbers) to simplify the code. This allows to use the column range checks. So `[3,3]` is `{col_min}, {col_max}`. I think this should be ok.

tobiaskleiner (approved):

GernotMaier:
Thanks @tobiaskleiner

PR #1309, opened 2025-01-16
Improvements to handle different versions of schemas.

### LINKED ISSUE #1197: How to handle change in model parameter schema versions
The model parameter schema might change with time, this is why they have a version field.

We are missing a mechanism to handle different versions.

Current example use case:

- the model parameter definition of array_triggers is changing significantly
- we might not want to change all model versions to the new schema

Right now, we have only two versions and this is not a lot of work. For the future, this will be an issue.

As a starting point, I suggest the following:

- add a `schema_version` field to all model parameters
- use the possibility to save several documents in one yaml file to keep all schema versions of a model parameter
- add functionality to simtools to read the correct document for a given version in the model parameters

This would solve at least the issue of keeping records of model parameters and introduce links between schemas and parameters. This will not solve the issue of changes required to "use" the model parameters in simtools (e.g., above changes to the trigger definition requires changes in reading and writing the simtel configuration files).

GernotMaier:
Schemas for input/output change with time with fields being added, renamed, or removed.

This PR adds functionality to handle different versions of schema files:

- allow for several schema definitions in a single yaml file using yaml documents (see example in this PR in src/simtools/schemas/model_parameter.metaschema.yml).
- select correct schemas by using the 'version' field.

Required a small improvement in general.collect_data_from_file to handle multi-document yaml files.

Removes also an obsolete (and duplicated) test file: tests/resources/array_element_position_ground.json

Closes #1197


src/simtools/schemas/model_parameter.metaschema.yml
```diff
@@ -2,12 +2,99 @@
...
+  SimtoolsModelParameters:
+    description: ""
+    type: object
+    properties:
+      _id:
+        type: string
+        description: "DB unique identifier"
+      entry_date:
+        type: string
+        description: "Value entry data"
```
tobiaskleiner, line 23: Value entry date

tests/unit_tests/utils/test_general.py
```diff
@@ -41,6 +41,12 @@ def test_collect_dict_data(io_handler, caplog) -> None:
     assert isinstance(_dict, dict)
     assert len(_dict) > 0
 
+    # file with several documents
```
tobiaskleiner, line 44: Maybe you can add a test here to check that `yaml.composer.ComposerError` is thrown?

tests/unit_tests/data_model/test_metadata_model.py
```diff
@@ -29,6 +29,19 @@ def test_load_schema():
...
+        metadata_model._load_schema(schema_file)
+
+    _schema_1, _ = metadata_model._load_schema(schema_file, "0.1.0")
+    assert _schema_1["version"] == "0.1.0"
+    _schema_2, _ = metadata_model._load_schema(schema_file, "0.2.0")
+    assert _schema_2["version"] == "0.2.0"
+
+    with pytest.raises(ValueError, match=r"^Schema version 0.2 not found in"):
+        metadata_model._load_schema(schema_file, "0.2")
+
```
tobiaskleiner, line 44: Would be good to also check for the warning from `_logger.warning(f"Schema version {schema_version} does not match {schema.get('version')}")`

src/simtools/schemas/model_parameter.metaschema.yml
GernotMaier: Fixed.

src/simtools/schemas/model_parameter.metaschema.yml
```diff
@@ -2,12 +2,99 @@
...
+      value:
+        anyOf:
+          - type: boolean
+          - type: number
+          - type: string
+        description: "Value of the parameter."
+      parameter_version:
+        anyOf:
+          - type: string
+          - type: "null"
```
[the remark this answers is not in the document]
GernotMaier: You are right, parameter_version and schema_version should always be filled. I removed the 'null' statements. `unique_id` is not filled at this point, so we should keep allowing 'null'.

tests/unit_tests/utils/test_general.py
GernotMaier: We don't raise ` yaml.composer.ComposerError`. This error is thrown by yaml if you try to open a multi-document yaml file with `yaml.load`. In this case, we want to load all documents with `yaml.safe_load_all`. So the test using a multi-document yaml is covering this.

tests/unit_tests/data_model/test_metadata_model.py
GernotMaier: Good point, added a test.

GernotMaier:
@tobiaskleiner - thanks for the review. I think I've addressed your points. please let me you know if they are ok.

tobiaskleiner:
> @tobiaskleiner - thanks for the review. I think I've addressed your points. please let me you know if they are ok.

@GernotMaier Yes all good now, thanks!

PR #1310, opened 2025-01-17
Simulation model refactoring with refactoring of db_handler.

GernotMaier:
This is a significant update to allow simtools with the improved and simplified structure of the simulation model.

- model parameters are stored in a new database called `CTAO-Simulation-ModelParameters-LATEST`, which is a copy of the new repository [simulation-models](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/tree/main) (check branch v1.0.0-parameters)
- introduction of difference between `parameter_version` and `model_version` (required a lot of small changes; especially when initializing the configurator, 'model_version' needs now to be specified)
- introduction of production tables, which list for a given `model_version` all relevant parameters and their versions
- removes reading / access to 'derived' DB parameters (these parameters are now in the model parameter DB).

Many code changes, but also significant simplifications.

Open points:

- [x] [database_scripts/upload_from_model_repository_to_db.sh](https://github.com/gammasim/simtools/pull/1310/files#diff-75c53c06a2b9462c72fc88921c1967d0dafbd114d1f37e8220c4e10bb4c93e7e): branch used for upload of parameters from the SimulationModel repository is temporary set to `v1.0.0-parameters`.
- [x] 3 integration tests fail due to #1315


tests/unit_tests/db/test_db_handler.py
```diff
@@ -2,16 +2,33 @@
...
+    """Reset db_client before each test."""
+    # If using the class-level db_client:
+    db_handler.DatabaseHandler.db_client = None
+    db_handler.production_table_cached = {}
+    db_handler.model_parameters_cached = {}
+    yield  # allows the test to run
+    # After the test, reset any side-effects (if necessary):
+    db_handler.DatabaseHandler.db_client = None
+    db_handler.production_table_cached.clear()
+    db_handler.model_parameters_cached.clear()
```
orelgueta, line 29: If you do this at the beginning of every test, why do it also at the end of each? Also, is there a reason here you use `.clear()` and above you set an empty dictionary?

tests/unit_tests/db/test_db_handler.py
```diff
@@ -2,16 +2,33 @@
...
+from bson.objectid import ObjectId
 
 from simtools.db import db_handler
+from simtools.utils import names
 
 logger = logging.getLogger()
 
 
+@pytest.fixture(autouse=True)
+def reset_db_client():
```
orelgueta, line 19: How do you make sure that this is called after the `_db_cleanup` fixture? If it's called before, then the sanboxes will not be deleted (or perhaps an error would occur).

tests/unit_tests/db/test_db_handler.py
```diff
@@ -33,11 +50,46 @@ def _db_cleanup(db, random_id):
...
+    return "test_db"
+
+
+@pytest.fixture
+def test_file():
+    return "test_file.dat"
+
+
+@pytest.fixture
+def test_file2():
```
orelgueta, line 64: Surprised the linters didn't complain about camel case (`test_file_2`)

tests/unit_tests/db/test_db_handler.py
```diff
@@ -76,360 +163,239 @@ def test_find_latest_simulation_model_db(db, db_no_config_file, mocker):
...
-    assert db_copy.mongo_db_config["db_simulation_model"] == "CTAO-Simulation-Model-v0-3-19"
+    assert (
+        db_copy.mongo_db_config["db_simulation_model"] == "CTAO-Simulation-ModelParameters-v0-3-19"
+    )
+
 
+def test_get_model_parameter(db, mocker):
+    """Test get_model_parameter method."""
+    mock_read_mongo_db = mocker.patch.object(
+        db, "_read_mongo_db", return_value={"parameter": "value"}
```
orelgueta, line 187: Does it really return a dictionary with this structure of just `{"parameter": "value"}`? If not, I suggest to make this a bit more realistic. It doesn't matter for the test since it is just mocking it, but it helps in understanding the code by looking at the test.
(There are more cases below.)

Edit: After seeing the amount of cases with random values below, I won't insist. This one feels different cause it's about structure rather than values, but OK.

tests/unit_tests/db/test_db_handler.py
```diff
@@ -76,360 +163,239 @@ def test_find_latest_simulation_model_db(db, db_no_config_file, mocker):
...
+    mock_read_mongo_db.assert_called_once_with(query=query, collection_name=collection)
+    assert result == {"parameter": "value"}
+
+
+def test_get_model_parameters(db, mocker):
+    """Test get_model_parameters method."""
+    mock_get_production_table = mocker.patch.object(
+        db,
+        "_read_production_table_from_mongo_db",
+        return_value={"parameters": {"LSTN-01": {"param1": "v1"}}},
```
orelgueta, line 268: Realistic values?

Edit: There are too many examples below, impossible to come up with realistic values for all, so you can forget it.

tests/unit_tests/db/test_db_handler.py
```diff
@@ -446,7 +412,757 @@ def test_get_collections(db, db_config, fs_files):
...
+            "_id": mock_find.return_value[1]["_id"],
+            "value": "value2",
+            "entry_date": mock_find.return_value[1]["_id"].generation_time,
+        },
+    }
+
+    mocker.patch.object(db.get_collection.return_value, "find", return_value=[])
+    with pytest.raises(
+        ValueError,
+        match=r"The following query for test_collection returned zero results: {'parameter_version': '1.0.0'}",
```
orelgueta, line 628: Line is too long. How did it pass the pre-commit hooks?

tests/unit_tests/db/test_db_handler.py
```diff
@@ -446,7 +412,757 @@ def test_get_collections(db, db_config, fs_files):
...
+        "parameters": param,
+        "design_model": {},
+        "entry_date": mock_find_one.return_value["_id"].generation_time,
+    }
+
+    # test with no results
+    mocker.patch.object(db.get_collection.return_value, "find_one", return_value=None)
+    with pytest.raises(
+        ValueError,
+        match=r"The following query returned zero results: {'model_version': '1.0.0', 'collection': 'telescopes'}",
```
orelgueta, line 686: Line too long

tests/unit_tests/db/test_db_handler.py
```diff
@@ -446,7 +412,757 @@ def test_get_collections(db, db_config, fs_files):
...
+    )
+    mock_reset_parameter_cache = mocker.patch.object(db, "_reset_parameter_cache")
+
+    par_dict = {"parameter": "param1", "value": "value1", "file": True}
+    collection_name = "telescopes"
+    file_prefix = None
+
+    with pytest.raises(
+        FileNotFoundError,
+        match=r"The location of the file to upload, corresponding to the param1 parameter, must be provided.",
```
orelgueta, line 909: Too long

tests/unit_tests/model/test_model_parameter.py
```diff
@@ -162,11 +162,11 @@ def test_load_parameters_from_db(telescope_model_lst, mocker):
     telescope_copy = copy.deepcopy(telescope_model_lst)
     mock_db = mocker.patch.object(DatabaseHandler, "get_model_parameters")
     telescope_copy._load_parameters_from_db()
-    assert mock_db.call_count == 2
+    assert mock_db.call_count == 4
```
orelgueta, line 165: Is this a bad sign that we went up to 4?

orelgueta (commented):
Very nice! 
Good to see that the db_handler tests are now mostly (all?) with mocking. 

My comments are small and random. However, I don't approve yet in case you want to make changes based on them.

src/simtools/db/db_handler.py
```diff
@@ -180,66 +172,103 @@ def _find_latest_simulation_model_db(self):
...
-            DatabaseHandler.model_parameters_cached[_array_elements_cache_key] = pars
+                parameter_version_table = production_table["parameters"][array_element]
+            except KeyError:  # allow missing array elements (parameter dict is checked later)
+                continue
+            DatabaseHandler.model_parameters_cached[cache_key] = self._read_mongo_db(
+                query=self._get_query_from_parameter_version_table(
+                    parameter_version_table, array_element, site
+                ),
+                collection_name=collection,
+            )
```
orelgueta, line 270: (Sorry, I missed this file when I did the first pass of the review.)

Do I understand correctly that in this line you read from MongoDB and execute the query once per array element? Would it be more efficient to build a larger query and return all array elements with one call?

src/simtools/db/db_handler.py
```diff
@@ -263,99 +292,106 @@ def get_collection(self, db_name, collection_name):
...
+        if db_name not in self.list_of_collections:
+            self.list_of_collections[db_name] = DatabaseHandler.db_client[
+                db_name
+            ].list_collection_names()
+        collections = self.list_of_collections[db_name]
+        if model_collections_only:
+            return [
+                collection
+                for collection in collections
+                if not collection.startswith("fs.") and collection != "metadata"
```
orelgueta, line 322: Will we still have a metadata collection after the restructuring?

src/simtools/db/db_handler.py
```diff
@@ -263,99 +292,106 @@ def get_collection(self, db_name, collection_name):
...
-        array_element_name,
-        model_version,
-        run_location,
-        collection_name,
-        write_files=True,
-        only_applicable=False,
+        instance_ids = {}
+        for file_name in file_names:
+            if Path(dest).joinpath(file_name).exists():
+                instance_ids[file_name] = "file exits"
```
orelgueta, line 361: exists

src/simtools/db/db_handler.py
```diff
@@ -364,154 +400,87 @@ def read_mongo_db(
...
+        list
+            Sorted list of all array element names found in collection
+        """
+        production_table = self._read_production_table_from_mongo_db(collection, model_version)
+        all_array_elements = production_table["parameters"]
+        return sorted(
+            [
+                entry
+                for entry in all_array_elements
+                if entry.startswith(array_element_type) and "-design" not in entry
```
orelgueta, line 482: Might be useful to mention in the docstring the fact this skips the design (I wouldn't necessarily expect that).

src/simtools/db/db_handler.py
```diff
@@ -541,63 +510,24 @@ def get_simulation_configuration_parameters(
             if simulation_software is not valid.
         """
         if simulation_software == "corsika":
-            return self.get_corsika_configuration_parameters(model_version)
+            return self.get_model_parameters(
+                None,
+                None,
+                model_version,
+                collection="configuration_corsika",
+            )
```
orelgueta, line 518: This doesn't depend on the site (or more accurately, the atmospheric profile dependence is in a different table), so setting the site/array_element to `None` makes sense. However, it reminded me that some of the parameters here depend on the primary (muons). How do we deal with it? We can postpone this issue for later of course, but if you agree that we should deal with it, I will open an issue so we don't forget.

orelgueta (commented):
Sorry, missed the db_handler file in the original pass. A couple of extra comments on that file.

database_scripts/upload_from_model_repository_to_db.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
 # Upload model parameter from repository to a local or remote mongoDB.
 #
+# Execute this scripts from the ./database_scripts directory.
 # Cover 'source .env': the script ensure that this file exists:
 # shellcheck disable=SC1091
 
-DB_SIMULATION_MODEL_URL="https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/model_parameters.git"
-DB_SIMULATION_MODEL_BRANCH="main"
+DB_SIMULATION_MODEL_URL="https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models"
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/applications/convert_all_model_parameters_from_simtel.py
```diff
@@ -6,6 +6,8 @@
     ready to be submitted to the model database. Prints out parameters which are not found
     in simtel configuration file and parameters which are not found in simtools schema files.
 
+    Note that all parameter are assigned the same parameter version.
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/applications/convert_all_model_parameters_from_simtel.py
```diff
@@ -30,7 +32,7 @@
           --simtel_cfg_file all_telescope_config_la_palma.cfg\\
           --simtel_telescope_name CT1\\
           --telescope LSTN-01\\
-          --model_version "2024-03-06"
+          --parameter_version "0.0.1"
```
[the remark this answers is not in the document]
GernotMaier: It is an example..but you are right, let's set it to `1.0.0`. Fixed.

src/simtools/applications/db_get_parameter_from_db.py
```diff
@@ -26,26 +26,21 @@
...
-         'File': True,
-         'Type': 'str',
-         'Value': 'mirror_CTA-N-LST1_v2019-03-31.dat',
-         'Version': '5.0.0',
-         '_id': ObjectId('608834f257df2db2531b8e78'),
-         'entry_date': datetime.datetime(2021, 4, 27, 15, 59, 46, tzinfo=<bson.tz_util.FixedOffset \
-          object at 0x7f601dd51d80>)}
+        simtools-db-get-parameter-from-db --parameter mirror_list \\
+                --site North --telescope LSTN-01 \\
+                --parameter_version 5.0.0
```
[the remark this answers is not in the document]
GernotMaier: Yes, Fixed.

src/simtools/applications/db_get_parameter_from_db.py
```diff
@@ -94,37 +90,54 @@ def main():  # noqa: D103
...
+    # get parameter using 'model_version'
+    elif args_dict["model_version"] is not None:
+        if args_dict["telescope"]:
+            pars = db.get_model_parameters(
+                site=args_dict["site"],
+                array_element_name=args_dict["telescope"],
+                model_version=args_dict["model_version"],
+                collection=(
+                    "configuration_sim_telarray"
+                    if args_dict.get("db_collection") == "configuration_sim_telarray"
```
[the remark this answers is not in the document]
GernotMaier: This line can be simplified to `collection=args_dict.get("db_collection", "telescopes"),`: if no collection is given on the command line, we assume the user wants the telescope collection. We otherwise expect that the user to provide the collection.

Surprisingly, we don't have a way to get the collection for a given parameter . Although it is not possible for all parameters to get a unique answer (e.g., `array_element_position_ground` could be `telescopes` or `calibration_devices`). Although we have `names.get_collection_name_from_array_element_name()`, which would resolve this. Anyway, the best solution would be the collection(s) to the schema files. Should we do this?

tests/unit_tests/db/test_db_handler.py
GernotMaier: I know what you mean, not ideal. But would change the tests, so dummy values should be fine.

tests/unit_tests/db/test_db_handler.py
GernotMaier: pylint does not run for `tests` (see `.pre-commit-config.yaml`). 

Fixed.

tests/unit_tests/db/test_db_handler.py
GernotMaier: Fixed.

tests/unit_tests/db/test_db_handler.py
GernotMaier: Fixed.

src/simtools/db/db_handler.py
GernotMaier: Yes, we call query the DB per array element - this is what we have always done. I don't know what is more efficient (I could imagine that simpler queries are faster). What I want to implement is to get a defined set of parameters for all array elements with one query (e.g., get all array element positions). This is beyond this PR - but we need to test DB query performance again and if necessary improve it.

src/simtools/db/db_handler.py
GernotMaier: We don't have a metadata collection since PR #1166. Removed above reference to "metadata"

src/simtools/db/db_handler.py
GernotMaier: Fixed.

src/simtools/db/db_handler.py
GernotMaier: Please open an issue.

tests/unit_tests/model/test_model_parameter.py
GernotMaier: No, this is expected - we call `get_model_parameters` twice in `_load_parameters_from_db()` (for `telescope` and `sites` parameter)

GernotMaier:
Thanks for the review! Let me know if this is all good.

If approved, I wait with merging until #1322 has been completed and merged plus the model parameters in the Simulation-Models DB/repository has been confirmed to be correct.

src/simtools/applications/db_get_parameter_from_db.py
```diff
@@ -94,37 +90,54 @@ def main():  # noqa: D103
...
+                collection=(
+                    args_dict["db_collection"] if args_dict.get("db_collection") else "sites"
+                ),
+                array_element_name=None,
+            )
     else:
-        pars = db.get_site_parameters(args_dict["site"], args_dict["model_version"])
-    if args_dict["parameter"] not in pars:
+        raise ValueError("Either 'parameter_version' or 'model_version' must be provided.")
+    param = args_dict["parameter"]
```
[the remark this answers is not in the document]
GernotMaier: Which KeyError do you mean? Both args_dict entries (for parameter and model-version) return `None` if not set, so this is not a KeyError.

src/simtools/db/db_model_upload.py
```diff
@@ -0,0 +1,139 @@
...
+                logger.info(f"No production table for {collection} in model version {model.name}")
+                continue
+            logger.info(f"Adding production table for {collection} to the database")
+            db.add_production_table(
+                db_name=args_dict["db_name"],
+                production_table=data,
+            )
+
+
+def _read_production_table(model_dict, file, model_name):
```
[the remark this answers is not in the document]
GernotMaier: It needs to update existing ones. Remember that in the repository, we save the production tables per telescope, while in the database we have one big one for all in the `telescopes` collection.

src/simtools/layout/array_layout.py
```diff
@@ -386,7 +386,9 @@ def _read_table_from_json_file(self, file_name):
         with Path(file_name).open("r", encoding="utf-8") as file:
             data = json.load(file)
 
-        position = gen.convert_string_to_list(data["value"])
+        position = data["value"]
+        if isinstance(position, str):
+            position = gen.convert_string_to_list(position)
```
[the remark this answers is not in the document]
GernotMaier: Yes, and I won't add a test. `convert_string_to_least` will go anyway, see issue #1317

src/simtools/model/array_model.py
```diff
@@ -314,16 +320,17 @@ def _get_telescope_position_parameter(
             Dict with telescope position parameters.
         """
         return {
+            "schema_version": "0.1.0",
```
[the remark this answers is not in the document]
GernotMaier: Not good and should be changed. I've started go generalize this by setting the default to the most recent version in the `model_parameter.metaschema.yml` (see branch `write-prod5-model-parameters` and #1327). Let's leave it hear and I correct it soon.

src/simtools/schemas/production_tables.schema.yml
```diff
@@ -0,0 +1,41 @@
...
+        additionalProperties:
+          type: object
+          description: Model parameter.
+          additionalProperties:
+            type: string
+            description: Parameter version (semantical versioning).
+            pattern: '^\d+\.\d+\.\d+$'
+        propertyNames:
+          description: Allowed parameter name patterns.
+          pattern: '^([A-Za-z](ST|LL)[N,S,x]-\d{2,3}|[A-Za-z](ST|LL)[N,S,x]-design|OBS-(North|South)|Dummy-Telescope)$'
```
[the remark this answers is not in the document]
GernotMaier: Yes, this was a conscious decision as we have no SCT configuration yet.

(but respect for checking the pattern in such a detail)

tests/unit_tests/corsika/test_corsika_config.py
```diff
@@ -29,9 +29,12 @@ def corsika_configuration_parameters(gcm2):
     return {
         "corsika_iact_max_bunches": {"value": 1000000, "unit": None},
         "corsika_cherenkov_photon_bunch_size": {"value": 5.0, "unit": None},
-        "corsika_cherenkov_photon_wavelength_range": {"value": "240. 700.", "unit": "nm"},
+        "corsika_cherenkov_photon_wavelength_range": {"value": [240.0, 700.0], "unit": "nm"},
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

tests/integration_tests/config/db_get_parameter_from_db_telescope_parameter_version.yml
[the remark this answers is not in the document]
GernotMaier: Yes. Added.

tests/resources/model_parameters/array_element_position_ground.json
```diff
@@ -1,11 +1,16 @@
...
-    "value": "-4.79 -499.24 39.75",
+    "instrument": "MSTN-09",
+    "site": "North",
+    "parameter_version": "2.0.0",
+    "unique_id": null,
+    "value": [
+        221.68,
+        -355.66,
+        49.0
+    ],
```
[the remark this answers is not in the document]
GernotMaier: Yes, but this will change very very soon.

tests/unit_tests/db/test_db_handler.py
GernotMaier: This was on advice from ChatGTP:

The main difference between assigning an empty dictionary (`{}`) and calling the `clear()` method on an existing dictionary is how they operate on references and what happens to the object they are applied to.

1. **Assigning an empty dictionary (`{}`):**
   - A new dictionary object is created.
   - The reference to the old dictionary object is replaced with the reference to the new dictionary.
   - If other parts of the code hold references to the original dictionary, those references will still point to the old dictionary, which remains unaffected by this operation.

2. **Using `clear()`:**
   - The `clear()` method removes all the items from the existing dictionary, leaving it empty.
   - It does not create a new dictionary but modifies the existing one in-place.
   - Any other references to the same dictionary will observe the clearing, as the dictionary object itself is still the same.

### Implications in the Given Code:
```python
db_handler.production_table_cached = {}   # Replaces the existing dictionary with a new empty one.
db_handler.production_table_cached.clear()  # Empties the existing dictionary in place.
```

- In `reset_db_client`, the decision to use `clear()` after the test ensures any references to the dictionaries (from elsewhere in the application or other threads) will remain intact but emptied, preventing potential bugs where stale data from earlier tests is observed in shared contexts.
- If you assign `{}`, other parts of the application that rely on the original dictionary object will not see this update because they hold references to the old, now discarded object.

### Conclusion:
- Use `clear()` when the dictionary's object identity (shared references) must be preserved.
- Use `{}` when you want to replace the dictionary entirely with a new object.
-

tests/unit_tests/db/test_db_handler.py
GernotMaier: Very good point. I have been wondering where all the sandboxes come from in the DB (just removed 128 sandboxes). I've combined both functions now into one to make sure that DBs are removed in the correct sequence. Good catch!

tests/unit_tests/db/test_db_handler.py
GernotMaier: We have relaxed linting rules for the test functions.

tests/unit_tests/db/test_db_handler.py
GernotMaier: But this is more than sufficient for the tests we are running here. The full dict is really not necessary. Let me know if you disagree.

src/simtools/applications/db_get_parameter_from_db.py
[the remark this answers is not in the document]
orelgueta: No, it's probably not worth it. What could be added (in a separate PR), is that the db_handler would print out the names of the available collections in case the collection was not found. Otherwise we can leave it as is (or change to your simplified suggestion of `collection=args_dict.get("db_collection", "telescopes")`)

src/simtools/applications/db_get_parameter_from_db.py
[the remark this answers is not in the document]
orelgueta: Sorry, I didn't expect github to show 4 lines. 
What I meant is that the assignment `param = args_dict["parameter"]` is not really needed, you can just use `args_dict["parameter"]` directly everywhere. However, if you did add this assignment, you might as well change line 130:

`raise KeyError(f"The requested parameter, {args_dict['parameter']}, does not exist.")`

to use `param` in it.

src/simtools/db/db_model_upload.py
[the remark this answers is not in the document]
orelgueta: Would doing something like

```
model_dict = {}
for model in filter(Path.is_dir, input_path.iterdir()):
        logger.info(f"Reading production tables for model version {model.name}")
        for file in sorted(model.rglob("*json")):
            model_dict.update(_read_production_table(model_dict, file, model.name))
```

and returning a dict work? Again, not important, just if you also agree that modifying an input variable is "unsafe".

src/simtools/layout/array_layout.py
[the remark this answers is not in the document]
orelgueta: OK!

src/simtools/model/array_model.py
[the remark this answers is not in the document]
orelgueta: OK!

tests/unit_tests/db/test_db_handler.py
orelgueta: Excellent

tests/unit_tests/db/test_db_handler.py
orelgueta: As you can see, I am not as relaxed as the linters...

tests/unit_tests/db/test_db_handler.py
orelgueta: As I wrote in the edit, it would become too cumbersome to change everywhere as there are many cases. The only reason I brought it up is because sometimes I read the test in order to understand the code/input/output. In this case it doesn't help to do it. It is not super important though I figured it out in the end, so you can leave it as is.

src/simtools/db/db_handler.py
orelgueta: Discussed in person, all good.

src/simtools/db/db_handler.py
orelgueta: Done (#1328).

orelgueta:
Most comments were addressed. Minor answers remain unresolved, but none are critical. Will not insist on any of them, will approve once an OK is given that all can be ignored/addressed.

src/simtools/applications/db_get_parameter_from_db.py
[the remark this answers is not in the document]
GernotMaier: We don't have this information anywhere stored in a nice way. I suggest #1329.

src/simtools/applications/db_get_parameter_from_db.py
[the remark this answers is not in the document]
GernotMaier: OK. Replaced param (although I am not convinced that it improves readability).

src/simtools/db/db_model_upload.py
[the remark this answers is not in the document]
GernotMaier: I honestly don't see an issue here. But don't ask me, I am a C++ programmer and allowed to modify anything unless there is a `const` keyword.

tests/unit_tests/db/test_db_handler.py
GernotMaier: OK. Changed to `t_e_s_t_f_i_l_e_2.dat`

tests/unit_tests/db/test_db_handler.py
GernotMaier: Looked again into it, and I am a bit puzzled why I have added them at all. The correct syntax would be `db_handler.DatabaseHandler.production_table_cached` (as they are class level variables). But tests run through without those lines (just checked also with ` --count 25 --random-order` and it works). Hmm.

src/simtools/applications/db_get_parameter_from_db.py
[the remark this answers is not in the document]
orelgueta: OK!

src/simtools/db/db_model_upload.py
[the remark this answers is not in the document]
orelgueta: OK, no problem then.

tests/unit_tests/db/test_db_handler.py
orelgueta: Makes sense that it works without them because you do it at the beginning of every test anyway. Either way, if these lines are removed and there are no issues, then this point is resolved.

orelgueta (approved):
Thanks!

GernotMaier:
Great, thanks for all the input @orelgueta .

As mentioned before, this branch will not be merged but serve as a base-development branch for the different issues to be solved before we move to the new database. So a couple of merge requests into this branch will follow soon....

PR #1312, opened 2025-01-22
Preparation of data validation routines for db restructering

GernotMaier:
Improvements to the data validation routines:

- generalize validation of all json files in a directory. The command line was called `--model_parameters_directory` and is now called `--file_directory`, as the validation application can be applied to files other than model parameter files
- the validation of all files in a directory using `--file_directory` can now be applied to schema validation (validating that all keys are there; fields are the correct type) and the more extensive validation using the schemas in `model_parameters` (which determine for each parameter the correctness of entries)
- allow for array-type `value` entries either the sim_tel-like string ("1 2 3 4") or an actual list ([1,2,3,4]); at the necessary conversion (I think on the long-term we want to get rid of the sim_tel-like strings)
- `value`-entries can be `None` now.
- added a static function `validate_model_parameter(parameter_dict)` for simple and fast validation of a model parameter dictionary

This is an updated required for the DB restructuring #1310 .

### DIFF
--- a/src/simtools/applications/validate_file_using_schema.py
+++ b/src/simtools/applications/validate_file_using_schema.py
@@ -41,6 +41,7 @@ import jsonschema
 
 import simtools.utils.general as gen
 from simtools.configuration import configurator
+from simtools.constants import MODEL_PARAMETER_SCHEMA_PATH
 from simtools.data_model import metadata_collector, metadata_model, validate_data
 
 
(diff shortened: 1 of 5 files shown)


tobiaskleiner (approved):
All good, thanks @GernotMaier .

GernotMaier:
Thanks - @tobiaskleiner

PR #1314, opened 2025-01-22
Fix site in integration test

GernotMaier:
Array elements for this tests are all for the southern site; site element should therefore be set to South.


EshitaJoshi (approved):
Thanks for the PR!

PR #1320, opened 2025-01-22
Fix label setting in validate_camera_efficiency

### LINKED ISSUE #1318: Label ignored in validate_camera_efficiency
The label configuration setting in `validate_camera_efficiency` is ignored and instead the label is sort of hardcoded to be `validate_camera_efficiency` (see code below).

https://github.com/gammasim/simtools/blob/0b86eeee4837bdac3567eedd9bd014a55dc11eab/src/simtools/applications/validate_camera_efficiency.py#L100

Not sure who wrote this code (cause we lost our history...), so I tag @GernotMaier to ask if you know of a good reason for it.

GernotMaier:
Allows to set label through the command line. Default value is simply `validate_camera_efficiency`.

Closes #1318

### DIFF
--- a/src/simtools/applications/validate_camera_efficiency.py
+++ b/src/simtools/applications/validate_camera_efficiency.py
@@ -106,7 +106,7 @@ def main():  # noqa: D103
     ce = CameraEfficiency(
         db_config=_db_config,
         simtel_path=args_dict["simtel_path"],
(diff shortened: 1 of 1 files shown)


EshitaJoshi (approved):
Thanks for the PR!

PR #1322, opened 2025-01-23
Allow for validation of simtools-generated configuration files for sim_telarray with reference files.

GernotMaier:
Add functionality to compare simtools-generated configuration files for sim_telarray with reference file:

- line-by-line string comparison (this might lead to issue with comparison of floats; for now it seems to work)
- reference files are available in [tests/resources/sim_telarray_configurations/](tests/resources/sim_telarray_configurations/)
- implemented for the integration tests `simulate_prod_gamma_40_deg_North_check_output.yml` and `simulate_prod_gamma_40_deg_South_check_output.yml` (see the keyword `TEST_SIMTEL_CFG_FILES`)
- changed `simulate_prod_gamma_40_deg_South_check_output.yml`  to beta layout to allow to test also `LSTS`

Some minor fine tuning in the file comparison is required as lines like `label` and 'config_version' (as they change for different integration tests).

Note: there is no need to review the reference configuration files in `tests/resources/sim_telarray_configurations`.

**Important!!** Model parameters in the reference files are not entirely correct but are 'as exported' from the current version of the model parameter database. #1326 provides corrections so that the model parameters will be correct.


GernotMaier:
> Code looks good. I have only skimmed at the config files and not reviewed them, as you said. Two questions - why are we not including units information for the parameters? 

These are configuration files used by sim_telarray. There are 'expected' units for sim_telarray which we have written up in the schema files. All value sin sim_telarray needs to be given in these expected units.

> And, how are these config files going to be updated when the model parameters are updated in the future? Will that have to be done manually?

We don't expect any updates to these reference files (until we see bugs), as we don't expect any future model versions to be distributed with sim_telarray (they will be distributed using simtools and the Simulation-Models database). So the tests included in this PR are ensuring that our base models (prod5/prod6) are correct (plus all the machinery to write these base models in simtools)

EshitaJoshi (approved):
Code is good, thanks for the PR!

GernotMaier:
@EshitaJoshi - thank you!

PR #1324, opened 2025-01-24
Clear default sim_telarray cfg directories from all calls

### LINKED ISSUE #1132: Option to remove unused preprocessor commands in sim_telarray
sim_telarray pfp commands include several unused directories by default such as:
`-I. -I/workdir/sim_telarray/sim_telarray/cfg -I/workdir/sim_telarray/sim_telarray/cfg/common -I/workdir/sim_telarray/sim_telarray/cfg/hess -I/workdir/sim_telarray/sim_telarray/cfg/hess2 -I/workdir/sim_telarray/sim_telarray/cfg/hess3 -I/workdir/sim_telarray/sim_telarray/cfg/hess5000 -I/workdir/sim_telarray/sim_telarray/cfg/CTA -I/workdir/sim_telarray/sim_telarray/cfg/MAGIC`

Most/all should be removable to avoid accidential configuration parameters.

orelgueta:
A general function is added to do this in order to avoid a bit of code duplication with a hardcoded command. It might also be needed to expand in the future, so it is easier this way.

Closes #1132

### DIFF
--- a/tests/unit_tests/utils/test_general.py
+++ b/tests/unit_tests/utils/test_general.py
@@ -714,3 +714,29 @@ def test_convert_keys_in_dict_to_lowercase():
     input_data = {"Key1": 123, "Key2": [1, 2, 3], "Key3": {"NestedKey1": "value1"}}
     expected_output = {"key1": 123, "key2": [1, 2, 3], "key3": {"nestedkey1": "value1"}}
     assert gen.convert_keys_in_dict_to_lowercase(input_data) == expected_output
+
+
+def test_clear_default_sim_telarray_cfg_directories():
+    """
+    Test the clear_default_sim_telarray_cfg_directories function.
+    """
+
+    # Test with a simple command.
+    command = "run_simulation"
+    expected_output = "SIM_TELARRAY_CONFIG_PATH='' run_simulation"
+    assert gen.clear_default_sim_telarray_cfg_directories(command) == expected_output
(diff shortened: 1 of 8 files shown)


tests/unit_tests/utils/test_general.py
```diff
@@ -714,3 +714,29 @@ def test_convert_keys_in_dict_to_lowercase():
     input_data = {"Key1": 123, "Key2": [1, 2, 3], "Key3": {"NestedKey1": "value1"}}
     expected_output = {"key1": 123, "key2": [1, 2, 3], "key3": {"nestedkey1": "value1"}}
     assert gen.convert_keys_in_dict_to_lowercase(input_data) == expected_output
+
+
+def test_clear_default_sim_telarray_cfg_directories():
+    """
+    Test the clear_default_sim_telarray_cfg_directories function.
```
GernotMaier, line 721: I am not convinced of the necessity of this docstring...it repeats the name of the test function. Anyway, doesn't hurt.

GernotMaier (approved):
Good, thanks!

tests/unit_tests/utils/test_general.py
orelgueta: Clearly I did not explicitly write the docstring and I agree it is not necessary. But since it doesn't hurt, I didn't bother deleting it.

PR #1330, opened 2025-01-28
Schema validation bug fixes.

GernotMaier:
**Important: this PR merges into db-simulation-model-refactoring**

This PR addresses the following points for `simtools-validate-file-using-schema`

- raise 'FileNotFoundError" when no files are found using the command line option `--file_directory` (which tells the code to validate all json files in this directory)
- allow for 'None' for entries 'site' and 'instrument' (e.g., most `configuration_corsika` parameters are site and instrument independent)
- fix missing name in test file for `mirror_list`

### DIFF
--- a/src/simtools/applications/validate_file_using_schema.py
+++ b/src/simtools/applications/validate_file_using_schema.py
@@ -116,7 +116,7 @@ def _get_schema_file_name(args_dict, data_dict=None):
     return schema_file
 
 
-def _get_file_list(file_directory=None, file_name=None):
+def _get_json_file_list(file_directory=None, file_name=None):
     """Return list of json files in a directory."""
     file_list = []
     if file_directory is not None:
@@ -137,7 +137,9 @@ def validate_schema(args_dict, logger):
     the metadata section of the data dictionary.
 
     """
-    for file_name in _get_file_list(args_dict.get("file_directory"), args_dict.get("file_name")):
+    for file_name in _get_json_file_list(
+        args_dict.get("file_directory"), args_dict.get("file_name")
+    ):
         try:
             data = gen.collect_data_from_file(file_name=file_name)
         except FileNotFoundError as exc:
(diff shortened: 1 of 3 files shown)


GernotMaier:
@tobiaskleiner - note that the three integration tests are expected to fail for now (this is similar to those in PR #1310). This will be fixed soon (one of the many things) - but is unrelated to the topic of this PR.

src/simtools/applications/validate_file_using_schema.py
```diff
@@ -116,6 +116,19 @@ def _get_schema_file_name(args_dict, data_dict=None):
     return schema_file
 
 
+def _get_file_list(file_directory=None, file_name=None):
```
tobiaskleiner, line 119: Maybe consider renaming the function to get_json_file_list

src/simtools/schemas/model_parameter.metaschema.yml
```diff
@@ -25,14 +25,19 @@ definitions:
...
       site:
-        type: string
+        type:
+          - string
+          - "null"
         description: "Associated CTAO site."
         enum:
           - North
           - South
+          - null
```
tobiaskleiner, line 40: If we accept null here: Could this result in some conflicts where we missed to add the site?

src/simtools/applications/validate_file_using_schema.py
GernotMaier: Good suggestion. Done.

src/simtools/schemas/model_parameter.metaschema.yml
GernotMaier: Site always needs to be presents, it is part of the list of required parameters (further down in the file):

```
   required:
      - file
      - instrument
      - parameter_version
      - schema_version
      - site
      - type
      - unit
      - value
```

tobiaskleiner: Ok missed this one. All good then.

tobiaskleiner (approved):

GernotMaier:
Thanks @tobiaskleiner

PR #1334, opened 2025-01-29
Set trigger current limit to 20 uA

GernotMaier:
see similar changes in the Simulation-Models repository.


orelgueta (approved):

GernotMaier:
Thanks @orelgueta

PR #1347, opened 2025-02-03
Remove duplication for schema validation plus some simplification

GernotMaier:
The construct of calling 

`jsonschema.validate(data, schema=json_schema, format_checker=format_checkers.format_checker)`

was duplicated - and would be further replicated for similar usages. This is especially not ideal, as we use a custom format checker to allow e.g., the check of astropy units. 

There is almost no new code in this PR, more consolidation / removal of duplication:

- move `validate_schema` and `load_schema` from `simtools.data_model.metadata_model` to `simtools.data_model.schema` (and rename it to `validate_dict_using_schema`)
- remove a duplicated function in `simtools.data_model.validate_data`
- moved and updated tests.
- removed some unused code (e.g., we don't have any data dicts including `meta_schema_url` (we had a long time ago))


EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thank you for the quick review!

PR #1349, opened 2025-02-04
Add additional queries to db handler to retrieve model versions etc.

### LINKED ISSUE #1344: Additional DB queries needed for model parameter reporting
We've identified the following queries required for model parameter reports, which should be added to the db_handler module:

- return list of all available model versions in the DB
- return list of array elements / site for a given model version
- return all model parameters (all parameter versions) for a given array element

GernotMaier:
Additional database queries are required to generate the model parameter reports, see issue #1344 👍 

- return list of all available model versions in the DB: new `db_handler.get_model_versions()`
- return list of array elements / site for a given model version: new `db_handler.get_array_elements()` (default collection is `telescopes`, but can also be e.g. `calibration_devices`
- return all model parameters (all parameter versions) for a given array element: updated `db_handler.get_model_parameters()` and allow `model_version` to be `None` (or a list of model versions). If `None`, the parameters for all available model versions is returned.

Closes #1344

### DIFF
--- a/src/simtools/db/db_handler.py
+++ b/src/simtools/db/db_handler.py
@@ -238,7 +238,9 @@ class DatabaseHandler:
         -------
         dict containing the parameters
         """
-        model_versions = [model_version] or self.get_model_versions(collection)
+        model_versions = (
+            self.get_model_versions(collection) if model_version is None else [model_version]
+        )
 
         pars = {}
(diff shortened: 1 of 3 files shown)


EshitaJoshi (approved):
Thanks for the PR - this should simplify things quite a lot.

PR #1354, opened 2025-02-06
Add bin edges to hdf5

tobiaskleiner:


GernotMaier (changes requested):
Unittests pass, but generate a lot of warnings, see e.g.:

https://github.com/gammasim/simtools/actions/runs/13289919308/job/37107837607

Warnings appear in test_corsika_histograms.py which is using the hdf5_handler.

GernotMaier (approved):
Looks good now, thanks!

PR #1355, opened 2025-02-07
Build optimized corsika binaries and allow different corsika versions

GernotMaier:
This is a bit of a demonstrator on how to build SimPipe images with CORSIKA, sim_telarray, and simtools included and allow for different versions of all software packages plus different compile options. Expect that this setup in a somewhat modified form will by part of the DPPS-SimPipe sub-project build.

The build matrix includes:

- versions of the tar packages of sim_telarray / CORSIKA (currently named using the date plus the versions of the included CORSIKA/bernlohr package versions). Note that the plan is in future to separate CORSIKA and sim_telarray (not included in this PR). As a temporary solution, these tar packages are downloaded from the cloud storage
- allow different compile options (e.g. prod6, prod6-SC) and different hadronic interaction models  (at this point only the highE model is available; straight forward to extend to the lowE models)
- allow to add extra_defines (e.g., for prod6-sc)
- tried to synchronize the docker files to minimize differences (simplifies maintainenance)

Important addition of this PR is transfer of the build scripts for optimized versions of CORSIKA:

- build script contents copied from [here](https://gitlab.cta-observatory.org/cta-computing/dpps/workload/production-configurations/-/tree/master/utils/build_opt?ref_type=heads); also the tar package with the patches is copied from there and upload to the cloud storage
- docker/Dockerfile-prod-opt includes the options to build the optimized CORSIKA code **and** install simtools into the same container (this deviates from the previous separation of corsika/simtel_array/simtools images; adding simtools is so straightforward that it is simpler to add it here then to have a separate step)

Images are build in the github action .github/workflows/build-docker-corsika-simtelarray-image.yml

Note:

- as this is a demonstrator, the full build matrix is not set (e.g.., all optimisation options of `avx_instruction: ['avx2', 'avx512f', 'sse4', 'avx']`)
- I could not find a nice way to not repeat the build matrix for the different jobs. This is not nice
- this workflow continuous to generate a non-optimized version (as before) to be used for the simtools-dev and simtools-prod containers

Tests of the generated images are still ongoing (need to decide how we want to test the generate images; I think now all avx instruction set is supported by the github provided runners).

An example workflow for two different simtel package versions can be found here: https://github.com/gammasim/simtools/actions/runs/13198633221

Open question (for follow up pull requests):

- somewhere we need to configure the allowed versions of CORSIKA (I think the optimisation requires v77500 or higher)
- the application of CORSIKA-version dependent patches is not nice. But I guess there will be not so many more follow up patches and this is fine?


GernotMaier:
@orelgueta - this is mostly interesting for you.

Spend a lot of time testing the optimized docker images on the DESY cluster using apptainers and the HT Condor tool.

- newer corsika / sim-telarray packages (e.g. with 77550) is not running CORSIKA successfully due to the changes Ive indicated in the mail to Luisa and Alice:

```
 floating point exceptions in the atmospheric interpolation routines:

 FITTED ATMOSPHERIC PARAMETERS:
 HLAY =   0.0000000000000000        400000.00000000000        1000000.0000000000        2500000.0000000000        11600000.000000000
 AATM =  -164.33804701596569       -133.71028031630613       -1.2869957446645230       -3.2473716188974322E-005   3.0954848557868938E-004
 BATM =   1209.0680470159657        1178.3361519842442        1372.0759298924836        992.54573295922296        1.0000000000000000
 CATM =   994186.38000000000        963373.44359814248        636143.04000000004        683985.15099011839        38766140230.233871
 START : PROBLEM WITH ATMOSPHERIC PARAMETERS
 INCONSISTENT LAYER =  2 DIFFERENCE: HEIGHT(THICKL)-HLAY = -0.1806496002E-02
Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG
STOP 1
```

Looking at corsika-77550.patch, it seems that Konrad made some adjustments to the interpolation of the atmospheres.

- trying one version before: CORSIKA 77500 fails earlier with:

```
 FILOPN: INCORRECT DATA SET NAME
 /dev/null
 PLEASE READ THE USERS GUIDE
 SEE KEYWORD: DIRECT
 IF YOU DO NOT WANT PARTICLE OUTPUT FILE
 PLEASE USE: PAROUT F F
Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG
STOP 1
```

Comparing input files, I've realized that the output files (DIRECT/PAROUT) are written out differently for the two version:

Screenshot 2025-02-10 at 13 57 21

I have been trying to see if there is an easy back-fix in corsika_inputs.cc, but without success. Any good ideas?

GernotMaier:
Above issue has been solved by changing in simtools.corsika.corsika_config the function _corsika_configuration_debugging_parameters to:

```
          "DIRECT": ["./"],
            "PAROUT": ["F", "F"],
```

Was:

```
"DIRECT": ["/dev/null"],
```

GernotMaier:
> One comment which is sort of related, for the prod6-sc option, please add the option to increase the number of allowed telescopes. E.g., the command for compilation should be
> 
> `env EXTRA_DEFINES="-DMAXIMUM_TELESCOPES=128" ./build_all prod6-sc qgs2 gsl`
> 
> The reason for it is that with 33 SCTs and 8 NectarCams, we pass the allowed limit in `build_all` for the `prod6-sc` option. It's a bit dangerous as well because no real error message is give, sim_telarray just ignores the last N telescopes.

Is this only for `prod6-sc` or for any other build option? This is our current list: `production: ['prod6-sc', 'prod6-baseline', 'prod5', 'prod5-sc']` (and I guess this should not expand significantly in future)

orelgueta:
It is an issue only for `prod6-sc` (for now). And yes, I don't expect it will expand in the future.

.github/workflows/build-docker-corsika-simtelarray-image.yml
```diff
@@ -60,17 +100,107 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: download-corsika
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        version:
+          - { simtel_version: '240205', corsika_version: '77500', bernlohr_version: '1.67' }
+#          - { simtel_version: '240927', corsika_version: '77550', bernlohr_version: '1.68' }
```
orelgueta, line 134: I assume you use the older one here and the newer one above just to test both and it's not a mistake, right?

orelgueta:
> This is a bit of a demonstrator on how to build SimPipe images with CORSIKA, sim_telarray, and simtools included and allow for different versions of all software packages plus different compile options. Expect that this setup in a somewhat modified form will by part of the DPPS-SimPipe sub-project build.

What does this mean for this specific PR?
The code looks good and I don't mind approving it already (with a couple of questions below), but will you merge it with commented out parts and not a full build matrix?


 > The build matrix includes:
> 
> * versions of the tar packages of sim_telarray / CORSIKA (currently named using the date plus the versions of the included CORSIKA/bernlohr package versions). Note that the plan is in future to separate CORSIKA and sim_telarray (not included in this PR). As a temporary solution, these tar packages are downloaded from the cloud storage

Why do we need to support different versions of CORSIKA?
I understand why in theory it would be good to have this option, but in practice, we don't expect many (any?) CORSIKA versions in the future (till C8) and if we decide to support just Prod5/Prod6 of the past, the latest CORSIKA will cover both.

> * docker/Dockerfile-prod-opt includes the options to build the optimized CORSIKA code **and** install simtools into the same container (this deviates from the previous separation of corsika/simtel_array/simtools images; adding simtools is so straightforward that it is simpler to add it here then to have a separate step)

It is straightforward to add simtools, I agree, but we decided to separate in the past because of the complexity of building CORSIKA/sim_telarray, not simtools. That is, we built simtools often, and we didn't want to wait for sim_telarray each time, considering it rarely changes, so we separated them. I don't think we should change this strategy. 
It would make sense if you plan to build the optimised versions only on release though. Is that your plan?



> * as this is a demonstrator, the full build matrix is not set (e.g.., all optimisation options of `avx_instruction: ['avx2', 'avx512f', 'sse4', 'avx']`)
> * I could not find a nice way to not repeat the build matrix for the different jobs. This is not nice

I also couldn't think of a way to do it, but I don't think it's too bad.

> * this workflow continuous to generate a non-optimized version (as before) to be used for the simtools-dev and simtools-prod containers

For the deployment with WMS, we would need this non-optimised version to be labeled as well like "package_noOpt". It allows DIRAC to choose this version in case the node CPU does not support any of the avx instructions available. It's enough to just add a label to the non-optimised version.


> * somewhere we need to configure the allowed versions of CORSIKA (I think the optimisation requires v77500 or higher)

See comment above it above.

> * the application of CORSIKA-version dependent patches is not nice. But I guess there will be not so many more follow up patches and this is fine?

I agree it is not an issue and if we stick to just the latest version then you don't have to worry about different patch versions either.

.github/workflows/build-docker-corsika-simtelarray-image.yml
GernotMaier: I have commented out those options which we are not using to keep the number of github workflows running at a necessary minimum. I however would like to keep the options in the workflow file to indicate how to expand the build matrix. 

We have used until now always 77500 with 1.67 (and not the newer package).

GernotMaier:
> > This is a bit of a demonstrator on how to build SimPipe images with CORSIKA, sim_telarray, and simtools included and allow for different versions of all software packages plus different compile options. Expect that this setup in a somewhat modified form will by part of the DPPS-SimPipe sub-project build.
> 
> What does this mean for this specific PR? The code looks good and I don't mind approving it already (with a couple of questions below), but will you merge it with commented out parts and not a full build matrix?

This means:

- for now this should go into simtools - but we are aware that building packages will happen in future as part of the DPPS SImPipe project (using Helm).
- for this reason I also minimized the build matrix to one CORSIKA version and one optimization, which is enough for testing and demonstration.

If ok with you, my suggestion would be to merge this now into main, but be prepared to remove it as soon as it is replaced by functionality elsewhere.

> 
> > The build matrix includes:
> > 
> > * versions of the tar packages of sim_telarray / CORSIKA (currently named using the date plus the versions of the included CORSIKA/bernlohr package versions). Note that the plan is in future to separate CORSIKA and sim_telarray (not included in this PR). As a temporary solution, these tar packages are downloaded from the cloud storage
> 
> Why do we need to support different versions of CORSIKA? I understand why in theory it would be good to have this option, but in practice, we don't expect many (any?) CORSIKA versions in the future (till C8) and if we decide to support just Prod5/Prod6 of the past, the latest CORSIKA will cover both.

We need to support for any external/internal software components multiple CORSIKA version. I am not sure I understand - there will be update to CORSIKA (7 and 8) and we need to be able to produce simulations with any "accepted" version. In the current example, what is the impact of moving to 775550? To study that we need production images for both new and old version.


> 
> > * docker/Dockerfile-prod-opt includes the options to build the optimized CORSIKA code **and** install simtools into the same container (this deviates from the previous separation of corsika/simtel_array/simtools images; adding simtools is so straightforward that it is simpler to add it here then to have a separate step)
> 
> It is straightforward to add simtools, I agree, but we decided to separate in the past because of the complexity of building CORSIKA/sim_telarray, not simtools. That is, we built simtools often, and we didn't want to wait for sim_telarray each time, considering it rarely changes, so we separated them. I don't think we should change this strategy. It would make sense if you plan to build the optimised versions only on release though. Is that your plan?

Exactly, if you look at the build options they are:

```
on:
  workflow_dispatch:
  pull_request:
    paths:
      - './docker/Dockerfile-simtelarray'
      - './docker/Dockerfile-prod-opt'
  release:
    types: [published]
  schedule:
    - cron: '0 0 * * 0'
```

This workflow runs only on changes to the docker files, on a published release, plus once a week for testing (we could remove this if required; but the scheduled workflow does not push any new images to the package registry). So I expect this to run not very often (we release maybe once a month?) - the overhead of building simtools seems to be minimal.

Note again the we will be moving on building in DPPS SimPipe.

> 
> > * as this is a demonstrator, the full build matrix is not set (e.g.., all optimisation options of `avx_instruction: ['avx2', 'avx512f', 'sse4', 'avx']`)
> > * I could not find a nice way to not repeat the build matrix for the different jobs. This is not nice
> 
> I also couldn't think of a way to do it, but I don't think it's too bad.
> 
> > * this workflow continuous to generate a non-optimized version (as before) to be used for the simtools-dev and simtools-prod containers
> 
> For the deployment with WMS, we would need this non-optimised version to be labeled as well like "package_noOpt". It allows DIRAC to choose this version in case the node CPU does not support any of the avx instructions available. It's enough to just add a label to the non-optimised version.

So there is no flexibility in DIRAC on how we label the different versions? This is actually an interface we need to describe somewhere.

> 
> > * somewhere we need to configure the allowed versions of CORSIKA (I think the optimisation requires v77500 or higher)
> 
> See comment above it above.
> 
> > * the application of CORSIKA-version dependent patches is not nice. But I guess there will be not so many more follow up patches and this is fine?
> 
> I agree it is not an issue and if we stick to just the latest version then you don't have to worry about different patch versions either.

We should discuss the versioning in future, but I am pretty much convinced that we need to be able to handle different CORSIKA versions.

Thanks for reviewing this PR!

.github/workflows/build-docker-corsika-simtelarray-image.yml
orelgueta: Right, but above you commented out 1.67 and left 1.68. That's why I asked.

GernotMaier: Yes! This is from testing. I've changes this now and all are at 1.67.

GernotMaier:
In case we need the github builds, let's get ready for that in a follow up issue / pull request (I having trouble with those long discussions in PRs). 

What is left to do?

- change non-optimized image to expected naming for dirac (`_noOpt` - does that need to be at the end of the name? Node that the package naming starts to be a bit complex (e.g., [simtools-corsika-sim-telarray-240927-corsika-77550-bernlohr-1.68-prod6-sc-qgs2](https://github.com/orgs/gammasim/packages/container/package/simtools-corsika-sim-telarray-240927-corsika-77550-bernlohr-1.68-prod6-sc-qgs2))
- as there is (to my knowledge) no easy way to get the build parameters and versions from inside the container, I suggest to add a simple yaml file in `/workdir/sim_telarray` containing the build arguments (e.g., this would solve also the issue that corsika does not have a `--version` flag).

Anything else?

.github/workflows/build-docker-corsika-simtelarray-image.yml
orelgueta: OK!

orelgueta:
> In case we need the github builds, let's get ready for that in a follow up issue / pull request (I having trouble with those long discussions in PRs).

Yes, it can definitely be in a future PR. It looks like it would be just uncommenting the other avx instructions, so a trivial one.


> 
> What is left to do?
> 
> * change non-optimized image to expected naming for dirac (`_noOpt` - does that need to be at the end of the name? Node that the package naming starts to be a bit complex (e.g., [simtools-corsika-sim-telarray-240927-corsika-77550-bernlohr-1.68-prod6-sc-qgs2](https://github.com/orgs/gammasim/packages/container/package/simtools-corsika-sim-telarray-240927-corsika-77550-bernlohr-1.68-prod6-sc-qgs2))

If I understood the code correctly, the avx instructions are always added to the end. So the two examples for `avx2` and `noOpt` would be:

- `simtools-corsika-sim-telarray-240927-corsika-77550-bernlohr-1.68-prod6-sc-qgs2-avx2`
- `simtools-corsika-sim-telarray-240927-corsika-77550-bernlohr-1.68-prod6-sc-qgs2-noOpt`

We might change the latter at some point to fit with the requirements of WMS, but having this now would make it trivial in the future.

> * as there is (to my knowledge) no easy way to get the build parameters and versions from inside the container, I suggest to add a simple yaml file in `/workdir/sim_telarray` containing the build arguments (e.g., this would solve also the issue that corsika does not have a `--version` flag).

Technically you can get the CORSIKA version by running the following on a CORSIKA produced file

`${SIM_TELARRAY_PATH}/bin/extract_corsika_tel --header-only --only-telescopes 1-999`

and extracting the env variables you are interested in. Other build flags can be extracted by running sim_telarray and extracting info from the log file. 

However, these options are definitely not easy, so I fully support adding a yaml file with this info.


> 
> Anything else?

No, nothing else and even those changes above can be done in a separate PR if you wish (in which case I will approve this one already now).

GernotMaier:
Good. I've started a new issue, see #1381

PR #1356, opened 2025-02-07
UC8-tool1-get-limits-from-hdf5

tobiaskleiner:
This PR adds a tool to derive the limits for energy, radial distance, and viewcone using the LimitCalculator.

Required are hdf5 files that are generated with `simtools-generate-simtel-array-histograms` using the --hdf5 flag. 

The limits\thresholds are then derived from these hdf5 files for a user specified event loss fraction.


pyproject.toml
GernotMaier, line 1: Please add h2py also to the environment.yml file. We unfortunately have to define dependencies in both files (would be nice if conda could read pyproject.toml)

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/python3
+
+r"""
+Derives the limits for energy, radial distance, and viewcone using the LimitCalculator.
```
GernotMaier, line 4: I suggest to modify to

- "Derives limits for ... to be used in CORSIKA simulations"

Otherwise unclear for what it is used.

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+
+The limits are derived based on the event loss fraction specified by the user.
+
+Command line arguments
+----------------------
+hdf5_file (str, required)
```
GernotMaier, line 10: Suggest to name the file not after the file format but after the content (e.g. `event_data_file`). You otherwise have to change many places in the code when deciding to use another file format.

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+    Path to the HDF5 file containing the event data.
+loss_fraction (float, required)
+    Fraction of events to be lost.
+
+
+Example
+-------
+Derive limits for a given HDF5 file with a specified loss fraction.
```
GernotMaier, line 18: Again, not important at this point that it is hdf5 file.

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+
+.. code-block:: console
+
+    simtools-derive-limits\\
+        --hdf5_file path/to/hdf5_file.hdf5 \\
```
GernotMaier, line 23: Choose a better, not `hd5_file.hdf5`.

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+        --loss_fraction 0.1
```
GernotMaier, line 24: This is probably a rather unrealistic loss fraction or? Do you want to give a realistic value in the example?

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+    logger = logging.getLogger()
+    logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
+
+    hdf5_file_path = args_dict["hdf5_file"]
+    loss_fraction = args_dict["loss_fraction"]
+
+    _logger.info(f"Loading HDF5 file: {hdf5_file_path}")
+    tables = read_hdf5(hdf5_file_path)
+
+    _logger.info("Initializing LimitCalculator")
```
GernotMaier, line 82: Is this really so important that it needs to be logged?

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+    calculator = LimitCalculator(tables)
+
+    _logger.info("Computing lower energy limit")
+    lower_energy_limit = calculator.compute_lower_energy_limit(loss_fraction)
+    _logger.info(f"Lower energy threshold: {lower_energy_limit} TeV")
+
+    _logger.info("Computing upper radial distance")
```
GernotMaier, line 89: Same here - not sure that info is needed. You print out the result, so it is clear that the scatter radius has been calculated.

src/simtools/applications/derive_limits.py
```diff
@@ -0,0 +1,99 @@
...
+    upper_radial_distance = calculator.compute_upper_radial_distance(loss_fraction)
+    _logger.info(f"Upper radius threshold: {upper_radial_distance} m")
+
+    _logger.info("Computing viewcone")
+    viewcone = calculator.compute_viewcone(loss_fraction)
+    _logger.info(f"Viewcone threshold: {viewcone} degrees")
```
GernotMaier, line 95: Is it a threshold, or the radius (diameter?)

src/simtools/production_configuration/limits_calculation.py
```diff
@@ -0,0 +1,161 @@
+"""Calculate the thresholds for energy, radial distance, and viewcone."""
+
+import numpy as np
+
+
+class LimitCalculator:
+    """
+    Class to compute thresholds/limits for energy, radial distance, and viewcone.
```
GernotMaier, line 8: "Compute thresholds/limits..."

(clear that it is a class, see two lines higher...)

GernotMaier: Maybe also add to the help here that what kind of histograms are read and they have been generated with sim_telarray.

src/simtools/production_configuration/limits_calculation.py
```diff
@@ -0,0 +1,161 @@
...
+            ):
+                self.angle_to_observing_position__triggered_showers_ = table
+            elif "Title" in table.meta and table.meta["Title"] == "event_weight__ra3d__log10_e__":
+                self.event_weight__ra3d__log10_e__ = table
+
+    def compute_threshold(
+        self, event_weight_array, bin_edges, loss_fraction, axis=0, limit_type="lower"
+    ):
+        """
+        Compute the threshold based on the loss fraction.
```
GernotMaier, line 41: is "threshold" correct, as you used also 'upper' limit? Maybe "compute_limits"?

src/simtools/production_configuration/limits_calculation.py
```diff
@@ -0,0 +1,161 @@
...
+        total_events = np.sum(projection)
+        threshold = (1 - loss_fraction) * total_events
+        bin_index = np.searchsorted(cumulative_sum, threshold)
+        if limit_type == "upper":
+            bin_edge_value = bin_edges[bin_index]
+        elif limit_type == "lower":
+            bin_edge_value = bin_edges[-bin_index]
+        return bin_index, bin_edge_value
+
+    def compute_lower_energy_limit(self, loss_fraction):
```
GernotMaier, line 81: At what point do we know that this is an energy in GeV (or TeV?)?

src/simtools/production_configuration/limits_calculation.py
```diff
@@ -0,0 +1,161 @@
...
+                for name in self.event_weight__ra3d__log10_e__.dtype.names
+            ]
+        )
+        bin_edges = self.event_weight__ra3d__log10_e__.meta["x_bin_edges"]
+        _, upper_bin_edge_value = self.compute_threshold(
+            event_weight_array, bin_edges, loss_fraction, axis=1, limit_type="upper"
+        )
+        return upper_bin_edge_value
+
+    def compute_viewcone(self, loss_fraction):
```
GernotMaier, line 133: Assume this is in 'deg'? Do we want to use units?

GernotMaier (changes requested):
Minor updates requested. Mostly wording and units. Thanks!

src/simtools/production_configuration/limits_calculation.py
tobiaskleiner: Not sure its needed here: There are no units in the tables, we don't write them out currently and there are no calculations involved, but only extracting the bin values (degree, TeV). I have added some comments about the units and to the print outs - let me know if you think this is sufficient.

GernotMaier (approved):
Thank you - all good!

PR #1362, opened 2025-02-11
Introduce MSTx-NectarCam and MSTx-FlashCam 

GernotMaier:
This PR introduces functionality to handle design models located on several sites. This mostly required changes in the validation modules:

- allow model parameter key 'site' to be a list of sites (e.g., ['North', 'South'] instead of only one of them); requires introduction of a new version of the [model_parameter meta schema](src/simtools/schemas/model_parameter.metaschema.yml) (new version is 0.3.0)
- replace `MSTN-design` by `MSTx-NectarCam` and `MSTS-design` by `MSTx-FlashCam`. This requires to introduce in the list of array elements the following new entry:
```
  MSTx:
    collection: "telescopes"
    observatory: "CTAO"
    site: ["North", "South"]
    design_types: ["design", "test", "FlashCam", "NectarCam"]
    description: "Mid-sized telescope"
```
- added a new function to get the check the allowed design types: `names.array_element_design_types`
- note that the test for the design type is now case sensitive (as the mongoDB query is case sensitive)
- improve inconsistent testing of parameter versions and file names for parameter (testing for typically errors found during setup of the DB); this required then as a follow up to rename the model parameter json files in tests/resources to be consistent with our standards
- corresponding [issue](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/issues/2) and [merge request](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/merge_requests/13)

**Requires a simulation model database >0.3.0**


src/simtools/utils/names.py
```diff
@@ -284,6 +317,26 @@ def get_array_element_type_from_name(name):
...
+    ----------
+    name: str
+       Array element name
+
+    Returns
+    -------
+    str
+        Design model name.
+    """
+    return f"{get_array_element_type_from_name(name)}-design"
```
EshitaJoshi, line 337: I'm not sure I understand the usefulness of this function. Earlier the code was just adding a '-design' at the end of the string when it needed to, so what does this function add? I'm not sure it adds clarity either because it hides away what the function is doing so it looks like it is retrieving the correct design model for the telescope name given.
Maybe it would make sense to keep this function if the code checks for `MSTN` and `MSTS` array element types and returns `MSTx-NectarCam` and `MSTx-FlashCam` accordingly?

tests/integration_tests/config/validate_camera_efficiency_mstn.yml
```diff
@@ -3,13 +3,13 @@ CTA_SIMPIPE:
   TEST_NAME: MSTN
   CONFIGURATION:
     SITE: North
-    TELESCOPE: MSTN-design
+    TELESCOPE: MSTx-NectarCam
```
EshitaJoshi, line 6: Would it make sense to rename this to `validate_camera_efficiency_mstx_nectarcam`?

tests/integration_tests/config/validate_camera_efficiency_msts.yml
```diff
@@ -3,14 +3,14 @@ CTA_SIMPIPE:
   TEST_NAME: MSTS
   CONFIGURATION:
     SITE: South
-    TELESCOPE: MSTS-design
+    TELESCOPE: MSTx-FlashCam
```
EshitaJoshi, line 6: Would it make sense to rename this to `validate_camera_efficiency_mstx_flashcam`? Specially since some southern MSTs also have NectarCam, would it make sense to have another integration test with telescope `MSTx-NectarCam` and site `South`?

tests/integration_tests/config/validate_camera_fov_run.yml
```diff
@@ -3,9 +3,9 @@ CTA_SIMPIPE:
   TEST_NAME: fov-run
   CONFIGURATION:
     SITE: North
-    TELESCOPE: MSTN-design
+    TELESCOPE: MSTx-NectarCam
```
EshitaJoshi, line 6: Should we also be validating the fov for FlashCam?

EshitaJoshi (commented):
Thanks for the PR! I have a few minor comments.

tests/integration_tests/config/validate_camera_fov_run.yml
GernotMaier: Not really necessary, we are validating the 'validate_camera_fov' application and not the the FOV of all telescope types. But I see that this is not consistent with the earlier validate_camera_efficiency tests.

tests/integration_tests/config/validate_camera_efficiency_mstn.yml
GernotMaier: Good point. Changed it.

src/simtools/utils/names.py
GernotMaier: Good point. Changed it.

GernotMaier: Sorry - Ignore that comment - that was for renaming of the validate_camera_efficiency_mstn/s below.

GernotMaier: To your question: this function is mostly there for backwards compatibility, but I see the point. I don't want to hardcode NectarCam and FlashCam, this is why we do the following:

- change `get_design_model_from_name` to `guess_design_model_from_name` to indicate a certain uncertainty
- opened an issue that this should be improved fixed: #1391

tests/integration_tests/config/validate_camera_efficiency_msts.yml
GernotMaier: Changed the file names:

```
tests/integration_tests/config/validate_camera_efficiency_mstx_flashcam_south.yml
tests/integration_tests/config/validate_camera_efficiency_mstx_nectarcam_north.yml
```

EshitaJoshi (approved):
Thanks for the PR!

PR #1363, opened 2025-02-12
Missing sphinx documentation

GernotMaier:
Any new module requires a corresponding markdown file in `docs/source/api-reference` and an entry in `docs/source/api-reference/index.md` (I realize that https://gammasim.github.io/simtools/developer-guide/documentation.html requires updates). We have a test that this is the case and e.g. [this one](https://github.com/gammasim/simtools/actions/runs/13284864252) is failing due to undocumented modules.

This PR adds missing documentation for some modules module (especially for production_configuration). I don't understand why tests didn't fail before (!?!) - I've checked the pull request and the tests where successful at this point. However, there seems to be a bug in the bash-style shell commands for the test; this is fixed in this PR.
 
These are the changes for this PR:

- introduction production configuration markdown file
- fix errors in docstrings to get sphinx compile without errors and warnings (note! This required to remove some type declaration from some of the functions in production_configuration modules. I've spent some time on trying to solve this, but sphinx does not accept these)
- removed duplication in module and class documentation
- added entries for legacy_data, plot_table, and simtel_table_reader

### DIFF
--- a/.github/workflows/CI-docs.yml
+++ b/.github/workflows/CI-docs.yml
@@ -52,15 +52,16 @@ jobs:
       - name: Check for complete API documentation
         run: |-
           MODULES=$(find src/simtools -type f -name "*.py" \
-            ! -path "src/simtools/applications/*" \
-            ! -name "__init__.py" ! -name "version.py" \
-            ! -path "src/simtools/_*" -prune)
-          FULLY_DOCUMENTED="TRUE"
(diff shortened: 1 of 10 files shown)


tobiaskleiner (approved):
Thanks @GernotMaier and strange this error didn't came up before.

GernotMaier:
Thanks @tobiaskleiner

PR #1365, opened 2025-02-13
Add fixed database version for testing

### LINKED ISSUE #1364: Remove `LATEST` from database name configuration
The name of the DB including its version is usually configured using:

```
SIMTOOLS_DB_SIMULATION_MODEL='CTAO-Simulation-ModelParameters-LATEST'
```
(e.g. in the .env files; or as environmental variables in the github workflow)

This is very efficient when database updates do not involve updates to the database structure. However:

- not all simtools version work with all database versions. We loose this 'knowledge' when using `LATEST`.
- database structure changes are hard to integrate into simtools, as PR tends to fail (as wrong versions of the DB are used)

 I therefore suggest as an intermediate solution to:

- explicit name the most recent and compatible database version for each release in the `.env_template` file
- change the github workflows to read out those values from the `.env_template` file

We need to find a long-term solution on how to configure simtools, as certain version of simtools will probably work only with a range of CORSIKA / sim_telarray versions plus a range of database versions.

GernotMaier:
Fix database version to be used in unit and integration testing.

Add this to `.en_update` instead of "LATEST" and add steps to read this value in the unit and integration tests.

This is a temporary solution until we implemented a more consistent way of setting versions for databases and software. 

Closes #1364

### DIFF
--- a/.env_template
+++ b/.env_template
@@ -6,7 +6,7 @@ SIMTOOLS_DB_SERVER='cta-simpipe-protodb.zeuthen.desy.de' # MongoDB server
 SIMTOOLS_DB_API_USER=YOUR_USERNAME # username for MongoDB: ask the responsible person
 SIMTOOLS_DB_API_PW=YOUR_PASSWORD # Password for MongoDB: ask the responsible person
 SIMTOOLS_DB_API_AUTHENTICATION_DATABASE='admin'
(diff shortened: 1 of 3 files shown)


tobiaskleiner (approved):

GernotMaier:
Thank you @tobiaskleiner

PR #1366, opened 2025-02-13
Validate meta_schema_url

### LINKED ISSUE #1269: Hardwired schema urls for model parameters
All schema files include a field called meta_schema_url, which is the url of the metaschema in the github repository.

The url points to the main branch, which is not ideal as any changes to directory or file names means that older versions don't work anymore. Discuss if there is a better solution.

GernotMaier:
Some of the metaschema urls have been incorrect in the past (especially after changes in names or directory structure). Add here a simple check to validate that the ULR of the metaschema ulr exists.

Closes #1269

### DIFF
--- a/src/simtools/data_model/schema.py
+++ b/src/simtools/data_model/schema.py
@@ -120,9 +120,8 @@ def validate_dict_using_schema(data, schema_file=None, json_schema=None):
     except jsonschema.exceptions.ValidationError as exc:
         _logger.error(f"Validation failed using schema: {json_schema}")
         raise exc
-    if data.get("meta_schema_url"):
-        if not gen.url_exists(data["meta_schema_url"]):
-            raise FileNotFoundError(f"Schema file not found: {data['meta_schema_url']}")
+    if data.get("meta_schema_url") and not gen.url_exists(data["meta_schema_url"]):
+        raise FileNotFoundError(f"Schema file not found: {data['meta_schema_url']}")
 
     _logger.debug(f"Successful validation of data using schema ({json_schema.get('name')})")
 
--- a/src/simtools/data_model/schema.py
+++ b/src/simtools/data_model/schema.py
@@ -120,6 +120,10 @@ def validate_dict_using_schema(data, schema_file=None, json_schema=None):
     except jsonschema.exceptions.ValidationError as exc:
         _logger.error(f"Validation failed using schema: {json_schema}")
         raise exc
+    if data.get("meta_schema_url"):
+        if not gen.url_exists(data["meta_schema_url"]):
+            raise FileNotFoundError(f"Schema file not found: {data['meta_schema_url']}")
+
     _logger.debug(f"Successful validation of data using schema ({json_schema.get('name')})")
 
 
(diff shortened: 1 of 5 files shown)


src/simtools/data_model/schema.py
```diff
@@ -120,6 +120,9 @@ def validate_dict_using_schema(data, schema_file=None, json_schema=None):
     except jsonschema.exceptions.ValidationError as exc:
         _logger.error(f"Validation failed using schema: {json_schema}")
         raise exc
+    if data.get("meta_schema_url") and not gen.url_exists(data["meta_schema_url"]):
+        raise FileNotFoundError(f"Schema file not found: {data['meta_schema_url']}")
+
```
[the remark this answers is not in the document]
GernotMaier: Agree.

src/simtools/utils/general.py
```diff
@@ -82,6 +82,29 @@ def is_url(url):
...
+        URL to be checked.
+
+    Returns
+    -------
+    bool
+        True if URL exists.
+    """
+    try:
+        with urllib.request.urlopen(url, timeout=5) as response:
+            return response.status == 200
```
EshitaJoshi, line 101: Do we want to make the acceptable response more general? As in, catch 2xx - 3xx statuses in the event of redirection?

src/simtools/utils/general.py
```diff
@@ -82,6 +82,29 @@ def is_url(url):
...
+    except urllib.error.HTTPError as e:
```
EshitaJoshi, line 102: HTTPError catches bad requests only, should we also add a TimeoutError?

src/simtools/utils/general.py
```diff
@@ -82,6 +82,29 @@ def is_url(url):
...
+        return e.status == 200
```
EshitaJoshi, line 103: Would this not always return False? I would expect the error status to be something like 404, not 200.

src/simtools/utils/general.py
```diff
@@ -82,6 +82,29 @@ def is_url(url):
...
+    except Exception:  # pylint: disable=broad-except
```
EshitaJoshi, line 104: Why the general exception? Would URLError be better here?

EshitaJoshi (commented):
Minor comments, mainly about exception handling. Would it make sense to accept 2xx (success) and 3xx (redirect) statuses and set an exception for 4xx (client error) and 5xx (server error) statuses and return False directly?

src/simtools/utils/general.py
GernotMaier: Not sure if we want to accept redirections. This is used for now only for schema validation, and I would prefer to be correct and don't accept redirection. Can you think of other return values to accept? I had a look and I don't see any which seem to be appropriate.

src/simtools/utils/general.py
GernotMaier: Good points - I've updated now the exception handling to catch the more general URLError and the one relevant when None is passed on.

GernotMaier:
@EshitaJoshi - thanks for the comments. Addressed and improved the error handling. Let me know if this is ok.

EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thank you @EshitaJoshi

PR #1368, opened 2025-02-14
Make sure that Changelog entries are not forgotten.

### LINKED ISSUE #1367: Changelogs
Seems like we are constantly forgetting to add the changelog entries into docs/changes.

Suggest to add a CI to test for this - more or less copy the ctapip CI : https://github.com/cta-observatory/ctapipe/blob/main/.github/workflows/changelog.yml

GernotMaier:
We forget too often the changelog entries when opening pull requests.

Here we copy and adapt the corresponding [ctapipe workflow](https://github.com/cta-observatory/ctapipe/blob/main/.github/workflows/changelog.yml) to test for updates of the `docs/changes` directory.

In case no changelog entry is required, please label the PR with `no-changelog-needed` (see label on the right)

Closes #1367


tobiaskleiner (approved):
All good, thanks.

PR #1374, opened 2025-02-18
Minor typo and pylint corrections

GernotMaier:
Typo corrections, no functional changes.

### DIFF
--- a/src/simtools/applications/generate_corsika_histograms.py
+++ b/src/simtools/applications/generate_corsika_histograms.py
@@ -152,7 +152,7 @@ r"""
      .. code-block:: console
 
         simtools-generate-corsika-histograms --iact_file /workdir/external/simtools/\\
-        testsresources/tel_output_10GeV-2-gamma-20deg-CTAO-South.corsikaio \\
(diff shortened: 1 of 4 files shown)


GernotMaier:
@tobiaskleiner - we discussed one wrong docstring in the call today - I don't remember, can you help me so that we correct it as well here?

tobiaskleiner:
> @tobiaskleiner - we discussed one wrong docstring in the call today - I don't remember, can you help me so that we correct it as well here?

it was `generate_simtel_array_histograms.py`?

GernotMaier:
But what is the problem:

```
   Write sim_telarray histograms into pdf and hdf5 files.

    It accepts multiple lists of histograms files, a single list or a histogram file.
    Each histogram is plotted in a page of the pdf file if the --pdf option is activated.
```

The last sentence is a bit convoluted, but probably fine.

tobiaskleiner (approved):

tobiaskleiner:
Looks good, thanks @GernotMaier. Regarding the formulation I also can't find anymore what was our initial problem. Maybe we were interpreting something false.

PR #1380, opened 2025-02-19
Add reading of database collections from model parameter schemas

### LINKED ISSUE #1329: Add collection to model parameter schemas
Add a field listing to associated collections to the model parameter schemas. This will allow to remove the various listing / setting of database collections throughout the code.

Note that a model parameter can be part of several collections (e.g., the position on the ground for telescopes and calibration_devices).

GernotMaier:
Add functionality to get the database collection name for a given parameter or array element name.

This simplifies the code in getting e.g. a single model parameter from the database - until one needed to know the collection name.

The implementation was done by trying to have a minimum amount of changes to existing code:

- instead of adding a `collection` field to the model parameter schemas in [src/simtools/schemas/model_parameters/](src/simtools/schemas/model_parameters/), the `instrument:class` field is used
- therefore updated the `instrument:class` field for a small number of schemas (mostly for configuration_sim_telarray and configuration_corsika collections)
- did **not** update the schema version in `src/simtools/schemas/model_parameter_and_data_schema.metaschema.yml`, despite changes: the changes don't break backward compatibility and therefore no version update is required
- otherwise mostly changes in `simtools.utils.names` to add new function to get the collection names.

Closes #1329


src/simtools/utils/names.py
```diff
@@ -305,43 +330,67 @@ def get_site_from_array_element_name(name):
...
+    ------
+    KeyError
+        If the parameter name is not found in the list of model parameters
+    """
+    _parameter_names = all_model_parameters()
+    try:
+        class_key = _parameter_names[parameter_name].get("instrument", {}).get("class")
+    except KeyError as exc:
+        raise KeyError(f"Parameter {parameter_name} without schema definition") from exc
+    return class_key_to_db_collection(class_key)
```
EshitaJoshi, line 391: This returns just one collection per parameter, right? I thought that a parameter could be used in more than one collection - in that case, what is returned? Or is that not the case anymore?

GernotMaier: Each parameter is in one collection. Remember collections correspond to the mongoDB collections, meaning if a parameter would be in multiple collections, we would duplicate data.

EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thank for the review, @EshitaJoshi

PR #1389, opened 2025-02-21
Generalize integration tests configuration to application workflows

GernotMaier:
Generalize the configuration files used for integration tests (see `tests/integration_tests/config/`) to describe more general application workflows:

- allow now several applications to be configured and executed (instead of running only a single simtools application)
- added a `LOG_FILE` entry to allow to specify stdout/stderr ouput
- renamed `integration_tests_config.metaschema.yml` to `application_workflow.metaschema.yml` to improve the naming
- added `SCHEMA_VERSION` and `SCHEMA_NAME` as allowed fields plus as fields in the integration tests configurations
- added a new json-schema document to application_workflow.metaschema.yml` with schema_version `0.3.0` (why do we no schema version `0.1.0`.

For the review: all files in `tests/integration_tests/config/` have been modified using a script. Reviewing 1-2 of them should be enough.


src/simtools/schemas/application_workflow.metaschema.yml
```diff
@@ -0,0 +1,244 @@
...
+              description: "Condition to apply to the cut column."
+              type: string
+    title: ApplicationWorkflow
+...
+---
+$schema: http://json-schema.org/draft-06/schema#
+$ref: '#/definitions/SimtoolsApplicationWorkflowConfiguration'
+title: SimPipe application workflow configuration metaschema
+description: YAML representation of application workflow configuration metaschema
+version: 0.2.0
```
tobiaskleiner, line 138: Where do we use 0.2.0, I can't see it?

GernotMaier: You are right, we don't use it anywhere anymore. So I probably should remove the 0.2.0 definition?

tobiaskleiner: Yes, I think it can be removed.

tobiaskleiner (approved):
Looks good, thanks @GernotMaier. Please remove the 0.2.0 definitions before merging.

GernotMaier:
Thanks @tobiaskleiner  - removed it. Waiting now for the tests before merging.

PR #1390, opened 2025-02-21
Increase coverage and improve function naming in names.py

GernotMaier:
The multiple changes in the past 2 weeks "lowered" the coverage of `simtools.utils.names`. As this is a fundamentally important base module of simtools, good coverage is absolutely essentially.

Improve with the coverage with this PR. Improve also naming of the function return instrument classes for a given collection.


tobiaskleiner (approved):
Thanks @GernotMaier, looks good.

PR #1392, opened 2025-02-21
Name test module for docs_read_parameters according to module name

GernotMaier:
Unit test files should follow the `test_<module name>` convention.


EshitaJoshi (approved):
Thanks for the PR!

PR #1397, opened 2025-02-25
Fix bug in config reading from APPLICATIONS dict

GernotMaier:
PR #1389 introduced an additional level in the definition of application config files (`APPLICATIONS`). There is a bug in the implementation which is fixed with this PR: the possibility to run any simtools using a config file instead specifying all command line options was broken (meaning e.g., `simtools-derive-photon-electron-spectrum --config my_config.yml`). 

Code was also simplified - no KEYERROR is now issued and therefore the unit test can be shortened.

Note that this expect that there is only one application config defined in the configuration files (there is another PR open #1379 , which introduces the possibility to run several applications). An example of a configuration file is:

```
CTA_SIMPIPE:
  APPLICATIONS:
  - APPLICATION: simtools-derive-photon-electron-spectrum
    TEST_NAME: LST_legacy_format
    CONFIGURATION:
      INPUT_SPECTRUM: ./tests/resources/SinglePhe_spectrum_totalfit_19pixel-average_20200601.csv
      AFTERPULSE_SPECTRUM: ./tests/resources/LST_afterpulses.dat
      STEP_SIZE: 0.02
      USE_NORM_SPE: true
      OUTPUT_PATH: simtools-output
      OUTPUT_FILE: test_single_pe_lst.ecsv
      USE_PLAIN_OUTPUT_PATH: true
      LOG_LEVEL: DEBUG
```

### DIFF
--- a/src/simtools/configuration/configurator.py
+++ b/src/simtools/configuration/configurator.py
@@ -287,16 +287,14 @@ class Configurator:
             )
             # yaml parser adds \n in multiline strings, remove them
             _config_dict = gen.remove_substring_recursively_from_dict(_config_dict, substring="\n")
-            if "CTA_SIMPIPE" in _config_dict:
-                try:
-                    self._fill_from_config_dict(
-                        input_dict=gen.change_dict_keys_case(
-                            _config_dict["CTA_SIMPIPE"]["CONFIGURATION"],
-                        ),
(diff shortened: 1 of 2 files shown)


GernotMaier:
Thank you @tobiaskleiner

Issue #1055, opened 2024-07-11
simtool configuration requires a significant review and update


GernotMaier:
All simtools print the tool configuration when giving the `--help` parameter. It is evident that 

1. a lot of the printed parameters are except for edge cases not relevant for most users
2. some of the printed / configurable parameters are not relevant at all / should not be changed

This requires a review and improvement of the configuration classes to allow for a more fine-grained selection of the command line configuration.

We might consider as well a two-step help:

- `--help` prints the most important parameters
- `--full_help` (need a better name) prints all

GernotMaier:
Adding to this issue the inconsistent handling of command line parameters with units:

1. `-zenith 20` (implicitly expecting that this is in degrees)
2. `-zenith "20 deg" (enforcing units; read in as string and convert to astropy quantity)
3. `-zenith "20. * u.deg" (adding a `type=parse_quantity,` to argparser and read it in directly as astropy quantity)

We should handle this consistently among all tools.

GernotMaier:
Some application should also work without any options given (e.g., `simtools-print-version`). At this point, without any option the help messages are printed.

GernotMaier:
Check also the `output_file_from_default` setting and that we really want to set always an output_file (some `ecsv`) file, even if not given on the command line.

PR #1398, opened 2025-02-25
Improvements for single pe setting workflow

GernotMaier:
This PR include several improvements for the single pe setting workflow and is related to [this merge request](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-model-parameter-setting/-/merge_requests/4):

- simplifications in running `simtools-run-applications`
- addition of metadata when running `imtools-plot-tabular-data`
- improved collection of metadata plus some additional fields filled

## simtools-run-applications

Simplified significantly the configuration part. Note, this requires that the tool is used in a setting similar to the [Simulation Model Parameter Setting](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-model-parameter-setting) repository: configuration file is given for a file path `.../input/ISODATE-.../config.yml` and output will be written to `.../output/ISODATE-.../`. This assumption should be fine, as we developed this application exactly for this purpose.

## simtools-plot-tabular-data

- add writing of metadata using the new `MetadataCollector:dump()` function (see below)

## Metadata writing and schemas

- allow to read several metadata files (allow on the command line a list of metadata files (with or without wildcard expressions)); this required several changes very instead of reading / modify the metadata, we have to loop over all metadata files in the list and run the operation
   - for this: added to `simtools.utils.general` a function called `generate_list_of_files(file_names)`, which returns e.g., the list of files matching `["*.yml", "*.json"]`.
- remove spurious "\n" from descriptions when writing metadata dictionaries to yaml. 
- addition of 'AUTHOR' to metadata field 'CONTEXT:DOCUMENT' (see also issue #1399)
- improve to collection of metadata for cases where not all fields are filled from the input (e.g., of no `PRODUCT.DATA.FORMAT` is given, the format is derived from the output file suffix.
- several changes to allow writing of metadata from `simtools-derive-photon-electron-spectrum` (removed some required fields from schema; improved metadata collector); note again #1399)
- move write function for metadata from `simtools.data_model.model_data_writer` to `simtools.data_model.metadata_collector` (meaning the `owner` of the metadata is now writing it)
  - in this context add a static function `MetadataCollector:dump()` to easily write the metadata describing an activity (actictivity="execution of a simtools")
- writing of metadata requires that the instrument is given for which the associated data is valid. Added therefore on the command to the simtools used for the single-pe setting worfklow the possibility to configure this on the command line
- metadata can now written to a file which name contains the name of the tool used (e.g., `spe_LST_2022-04-27_AP2.0e-4.plot_tabular_data.meta.yml`)


src/simtools/applications/plot_tabular_data.py
```diff
@@ -56,7 +57,7 @@ def _parse(label, description, usage):
         type=str,
         required=True,
     )
-    return config.initialize(db_config=True)
+    return config.initialize(db_config=True, simulation_model=["telescope"])
```
orelgueta, line 60: If I dig deep I might find an answer to this question, but do we really need the telescope models for every call of this application? The application name is quite general and just from it I would be a bit surprised of this behaviour.

(Technically that is also true for the DB connection, but never mind.)

src/simtools/applications/run_application.py
```diff
@@ -111,13 +111,10 @@ def read_application_configuration(configuration_file, logger):
     application_config = gen.collect_data_from_file(configuration_file).get("CTA_SIMPIPE")
     place_holder = "__SETTING_WORKFLOW__"
     workflow_dir, setting_workflow = get_subdirectory_name(configuration_file)
-    output_path = str(workflow_dir).replace("input", "output") + setting_workflow
+    output_path = str(workflow_dir).replace("input", "output") + "/" + setting_workflow
```
orelgueta, line 114: I assume workflow_dir is already a `Path`. In which case, it would be nicer to do

`output_path = workflow_dir.with_name(workflow_dir.name.replace("input", "output")) / setting_workflow`

This gets rid of string manipulations of paths (and would return a `Path` which would mean that in the following line you can remove the `Path` definition, but it doesn't hurt).

src/simtools/data_model/metadata_collector.py
```diff
@@ -256,31 +335,32 @@ def _read_input_metadata_from_file(self, metadata_file_name=None):
...
-        elif Path(metadata_file_name).suffix == ".ecsv":
-            _input_metadata = self._read_input_metadata_from_ecsv(metadata_file_name)
-        else:
-            self._logger.error("Unknown metadata file format: %s", metadata_file_name)
-            raise gen.InvalidConfigDataError
+            return None
+
+        metadata = []
+        for metadata_file in metadata_files:
+            self._logger.debug("Reading meta data from %s", metadata_file)
```
orelgueta, line 352: That's the correct way to do string formatting in logging, but I thought we were consistent in not using it anywhere and choose f-string instead. Not a big deal either way.

src/simtools/utils/general.py
```diff
@@ -455,6 +456,34 @@ def find_file(name, loc):
     raise FileNotFoundError(msg)
 
 
+def generate_list_of_files(file_names):
```
orelgueta, line 459: Is "generate" the correct name? This doesn't generate a list of files but instead finds the files which are included in the wildcards, no?

tests/unit_tests/utils/test_general.py
```diff
@@ -807,6 +807,22 @@ def test_get_list_of_files_from_command_line(tmp_test_directory) -> None:
...
+    with pytest.raises(ValueError, match=r"^No file list provided"):
+        gen.generate_list_of_files(None)
+
+    assert gen.generate_list_of_files("LICENSE") == [Path("LICENSE")]
+    yml_list = gen.generate_list_of_files("tests/resources/*.yml")
+    assert len(yml_list) > 0
+    yml_and_ecvs_list = gen.generate_list_of_files(
+        ["tests/resources/*.yml", "tests/resources/*.ecsv"]
+    )
+    assert len(yml_and_ecvs_list) > len(yml_list)
```
orelgueta, line 820: Change to `>=` in case we decide to abandon all ecsv files in the future? Might make this test less safe though.

orelgueta (commented):
Very minor comments, could approve already, but the name of the "generate" function is something I would like to resolve first.

src/simtools/applications/plot_tabular_data.py
GernotMaier: No, this is an open issue that the naming of our configuration sections should be updated: this adds the "--telescope" as a command line option. This should be changed to an 'Instrument' section (or something similar). On the todo list...also related to the other points in #1055. No telescope model is needed!

src/simtools/data_model/metadata_collector.py
GernotMaier: Fixed.

tests/unit_tests/utils/test_general.py
GernotMaier: I prefer to leave it, as the purpose of the test is to check if a list of wildcard pattern works.

src/simtools/utils/general.py
GernotMaier: It generates a list of files using the wildcards. To emphasize the wildcard issue, I've renamed it to resolve_file_patterns()

GernotMaier:
@orelgueta - please have another look. I think I've addressed all your problems. Thanks for the review!

src/simtools/applications/plot_tabular_data.py
orelgueta: OK, the question still stands as to why we need the `--telescope` command line option for every call to this application, but it's not a blocking point.

src/simtools/applications/run_application.py
orelgueta: Not about number of characters, but about not performing string operations on paths.

src/simtools/utils/general.py
orelgueta: Much better, thanks!

orelgueta (approved):

src/simtools/applications/plot_tabular_data.py
GernotMaier: It is optional, so you don't have to give it (look e.g., at tests/integration_tests/config/derive_photon_electron_spectrum_lst_ecsv.yml).

This is used only to get a minimum set of metadata.

PR #1400, opened 2025-02-26
Set user in docker images and add possibility to add user information from env/command line

GernotMaier:
Add two related items:

- default user in all docker containers is now `ctao` and not `root`
- add possibility to add user information, which is used for a more complete collection of the metadata information:

```
user:
  --user_name USER_NAME
                        user name (default: None)
  --user_organization USER_ORGANIZATION
                        user organization (default: None)
  --user_email USER_EMAIL
                        user email (default: None)
  --user_orcid USER_ORCID
                        user ORCID (default: None)
```

These values can also be specified in your `.env` file:

```
SIMTOOLS_USER_NAME='G.Maier'
SIMTOOLS_USER_ORCID='0000-0001-9868-4700'
```

### DIFF
--- a/.env_template
+++ b/.env_template
@@ -8,6 +8,10 @@ SIMTOOLS_DB_API_PW=YOUR_PASSWORD # Password for MongoDB: ask the responsible per
 SIMTOOLS_DB_API_AUTHENTICATION_DATABASE='admin'
 SIMTOOLS_DB_SIMULATION_MODEL='CTAO-Simulation-ModelParameters-v0-4-0'
 SIMTOOLS_SIMTEL_PATH='/workdir/sim_telarray'
+SIMTOOLS_USER_NAME='simtools-user'
+SIMTOOLS_USER_ORGANIZATION='CTAO'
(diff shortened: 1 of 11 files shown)


orelgueta (approved):
Maybe in a separate PR, but don't you want to add a check that if metadata is filled with the default username an error is raised?

GernotMaier:
Do you mean we should get completely of `getpass.getuser()` in case it is not set through any other option? This requires then that the user name is defined in the `.env` file and we have to modify the CI github workflows (as there the `.env` settings are set through environmental variable). 

Do we want to start with issuing a warning instead of an error?

orelgueta:
That was my thought, yes. However, I am not 100% how we enforce it when it runs in the CI (having a CI username isn't that useful). Perhaps in that case it is either a required filled when the workflow is triggered (if done manually) or it has to come from the input metadata, if it exists. 

Either way, I agree it is not something we can enforce at this stage with so many unknowns. A warning would be enough and even that I wouldn't insist on (it's approved, so you can merge already).

PR #1401, opened 2025-02-26
Remove CTAO user in dev container

GernotMaier:
Otherwise cannot run `pip install -e .`


tobiaskleiner (approved):
Thanks @GernotMaier, can be merged.

PR #1411, opened 2025-03-03
Fix date writing to ensure ISO date format

GernotMaier:
Noticed today with failing integration tests (schema validation issues) that `datetime.datetime.now().isoformat()` is not writing ISO time format strings (???).

Fixed this with the pull request below. To avoid repeating the required options, added a general function `simtools.gen.now_date_time_in_isoformat()`. Adjusted all files in `test/resources` to the correct format.

### DIFF
--- a/docs/changes/1411.bugfix.md
+++ b/docs/changes/1411.bugfix.md
@@ -1,0 +1,1 @@
+Fix date writing to ensure ISO-conform date format.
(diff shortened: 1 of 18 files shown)


maxnoe:
> that datetime.datetime.now().isoformat() is not writing ISO time format strings (???).

Of course it does, just in local time, not in UTC.

EshitaJoshi (approved):
Thanks for the PR!

PR #1417, opened 2025-03-05
Switch on ruff formatting

GernotMaier:
Was wondering why long lines are not fixed anymore with pre-commit. Ruff formatting needs to be switched on explicitly (remember that we have removed black to avoid using several tools recently; see PR #1407.

ruff and black are nearly identical, but not exactly  (see the [ruff FAQ](https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black)). This means running it over all files changed 60 files. Scanning through the changes, I do think it improved the readability of the code and we should apply them.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,7 @@ repos:
   hooks:
   - id: ruff
     args: ["--fix"]
+  - id: ruff-format
 # pylint
 - repo: local
   hooks:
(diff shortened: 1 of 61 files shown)


tobiaskleiner (approved):

tobiaskleiner (approved):
Thanks @GernotMaier , all good.

PR #1419, opened 2025-03-06
Refinements of setting workflows

GernotMaier:
Add a couple of improvements for setting workflows - mostly on the example of the single pe setting:

- allow in `simtools-run-application` to specify to run certain steps only and skipping some of them. This is useful for debugging. E.g. for a 3-step workflow, with `--steps 2 3`, only step 2 and 3 are executed
- improved docstrings in the single_pe application

### DIFF
--- a/src/simtools/applications/run_application.py
+++ b/src/simtools/applications/run_application.py
@@ -1,12 +1,30 @@
 #!/usr/bin/python3
 
 """
-Run simtools applications from configuration files for setting workflows.
+Run several simtools applications using a configuration file.
 
 Allows to run several simtools applications with a single configuration file, which includes
 both the name of the simtools application and the configuration for the application.
 
-Strong assumption on the directory structure for input and output files of applications.
+This application is used for model parameter setting workflows.
+Strong assumptions are applied on the directory structure for input and output files of
+applications.
+
+Example
+-------
+
+Run the application with the configuration file 'config_file_name':
+
+.. code-block:: console
+
+    simtools-run-application --configuration_file config_file_name
+
+Run the application with the configuration file 'config_file_name', but skipping all steps except
+step 2 and 3 (useful for debugging):
+
+.. code-block:: console
+
+    simtools-run-application --configuration_file config_file_name --steps 2 3
 
 """
 
@@ -49,6 +67,12 @@ def _parse(label, description, usage):
         required=True,
         default=None,
     )
+    config.parser.add_argument(
+        "--steps",
+        type=int,
+        nargs="+",
+        help="List of steps to be execution (e.g., '--steps 7 8 9'; ignore to run all)."
+    )
     return config.initialize(db_config=True)
 
 
@@ -81,7 +105,7 @@ def get_subdirectory_name(path):
         raise ValueError(f"Could not find subdirectory under 'input': {exc}") from exc
 
 
-def read_application_configuration(configuration_file, logger):
+def read_application_configuration(configuration_file, steps, logger):
     """
     Read application configuration from file and modify for setting workflows.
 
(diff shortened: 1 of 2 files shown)


src/simtools/applications/derive_photon_electron_spectrum.py
```diff
@@ -74,6 +74,20 @@ def _parse(label):
...
+    config.parser.add_argument(
+        "--scale_afterpulse_spectrum",
+        help="Scale afterpulse spectrum by the given factor",
+        type=float,
+        default=1.0,
+        required=False,
+    )
+    config.parser.add_argument(
+        "--threshold_afterpulse_spectrum",
+        help="Threshold for afterpulse ratio calculation",
```
orelgueta, line 86: Ratio calculation?

src/simtools/applications/run_application.py
```diff
@@ -49,6 +67,12 @@ def _parse(label, description, usage):
         required=True,
         default=None,
     )
+    config.parser.add_argument(
+        "--steps",
+        type=int,
+        nargs="+",
+        help="List of steps to be execution (e.g., '--steps 7 8 9'; ignore to run all).",
```
orelgueta, line 74: I suggest to write either "do not specify to run all" or remove that comment entirely. As is it is a bit confusing to me.

src/simtools/camera/single_photon_electron_spectrum.py
```diff
@@ -119,13 +119,14 @@ def _derive_spectrum_norm_spe(self):
             f"{self.args_dict['simtel_path']}/sim_telarray/bin/norm_spe",
             "-r",
             f"{self.args_dict['step_size']},{self.args_dict['max_amplitude']}",
-            tmp_input_file.name,
         ]
         if tmp_ap_file:
-            command.insert(1, "-a")
-            command.insert(2, f"{tmp_ap_file.name}")
+            command.extend(["-a", f"{tmp_ap_file.name}"])
+            command.extend(["-s", f"{self.args_dict.get('scale_afterpulse_spectrum', 1.0)}"])
```
orelgueta, line 125: The default is defined here and in the argument parsing itself. Obviously not a problem, but not needed here. (I don't think it means changing the default would mean changing in two places cause I assume the default value here will never be used.)

src/simtools/camera/single_photon_electron_spectrum.py
```diff
@@ -119,13 +119,14 @@ def _derive_spectrum_norm_spe(self):
...
+            command.extend(["-t", f"{self.args_dict.get('threshold_afterpulse_spectrum', 4.0)}"])
```
orelgueta, line 126: Same comment about default value

orelgueta (approved):
Very minor comments, approving already

src/simtools/applications/derive_photon_electron_spectrum.py
GernotMaier: (note this parameter is replace in #1419, so no need to discuss here. This is the threshold used for the ratio of afterpulse to prompt calculation.

src/simtools/applications/run_application.py
GernotMaier: Good suggestion, changed it.

src/simtools/camera/single_photon_electron_spectrum.py
GernotMaier: Good point, I removed it here and left it at the parsing stage (required an update in the unit tests)

src/simtools/camera/single_photon_electron_spectrum.py
GernotMaier: Fixed.

PR #1420, opened 2025-03-06
Improvements to production docker images: arm compatibility; run integration tests

GernotMaier:
Two improvements regarding the production images:

- remove `avx2` flag for non-optimized images to allow to run them on arm-based CPUs (still need to use `podman run --arch amd64 ...` to execute)
- add integration test to run on the production images with non-optimized CORSIKA; this was noted recently that we need to test these images. 

Note that this is in some sense a duplication of what has been / will be added in the SimPipe gitlab repository. However, for it is easier to develop these steps here.

Minor changes: 

- docker image tags are call small caps. Changed there the name for non-optimized images from `noOpt` to `no-opt`.
- increased the allowed line length in yaml files in the pre-commit section to 300 characters (from 250). Reason is that I simple cannot get the CI running without having the `image` file in one single line.

### DIFF
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -210,7 +210,9 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}-prod-opt
       - name: set image tag
-        run: echo "image_tag=${{ steps.meta.outputs.tags }}" >> $GITHUB_OUTPUT
+        run: |
+            TAG=$(echo "${{ steps.meta.outputs.tags }}" | grep 'latest' | head -n1)
+            echo "image_tag=$TAG" >> "$GITHUB_ENV"
         id: set-image-tag
     outputs:
       image_tag: ${{ steps.meta.outputs.tags }}
--- a/.github/workflows/build-docker-corsika-simtelarray-image.yml
+++ b/.github/workflows/build-docker-corsika-simtelarray-image.yml
@@ -189,7 +189,7 @@ jobs:
             type=semver,pattern={{major}}.{{minor}}.{{patch}}
             type=schedule,pattern={{date 'YYYYMMDD'}}
             type=raw,value={{date 'YYYYMMDD-HHmmss'}}
-          images: ${{ env.REGISTRY }}/gammasim/simtools-prod-opt-${{ env.LABELS }}
+          images: ${{ env.REGISTRY }}/gammasim/simtools-prod-${{ env.LABELS }}
           flavor: latest=true
 
       - name: Build and push optimized image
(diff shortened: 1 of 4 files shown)


.github/workflows/build-docker-corsika-simtelarray-image.yml
```diff
@@ -205,3 +210,52 @@ jobs:
...
+        version:
+          - {simtel_version: '240205', corsika_version: '77500', bernlohr_version: '1.67'}
+          # - {simtel_version: '240927', corsika_version: '77550', bernlohr_version: '1.68'}
+        hadronic_model: ['qgs2']
+        production:
+          - {build_opt: 'prod6-sc', extra_def: '-DMAXIMUM_TELESCOPES=128'}
+          - {build_opt: 'prod6-baseline', extra_def: ''}
+        avx_instruction: ['no_opt']
+    container:
+      image: ghcr.io/gammasim/simtools-prod-opt-${{ matrix.version.simtel_version }}-corsika-${{ matrix.version.corsika_version }}-bernlohr-${{ matrix.version.bernlohr_version }}-${{ matrix.production.build_opt }}-${{ matrix.hadronic_model }}-${{ matrix.avx_instruction }}
```
orelgueta, line 234: What's the `opt` in `prod-opt` here? 

Reading further, I understand it comes from Docker image name. Are they still separate and the `opt` is needed?

.github/workflows/build-docker-corsika-simtelarray-image.yml
```diff
@@ -205,3 +210,52 @@ jobs:
...
+        SIMTOOLS_DB_API_USER: ${{ env.SIMTOOLS_DB_API_USER }}
+        SIMTOOLS_DB_API_PW: ${{ env.SIMTOOLS_DB_API_PW }}
+        SIMTOOLS_DB_SERVER: ${{ env.SIMTOOLS_DB_SERVER }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set sim_telarray path
+        run: |
+          echo "PATH=\$PATH:/usr/bin:/usr/local/bin:$SIMTOOLS_SIMTEL_PATH" >> "$GITHUB_ENV"
```
orelgueta, line 247: Do you really need to explicitly set `/usr/bin:/usr/local/bin`? There's no harm in it of course, but if it's not needed I would remove it (who knows, maybe in a future OS it will change).

docker/Dockerfile-prod-opt
```diff
@@ -83,20 +84,20 @@ RUN echo "build_opt: $BUILD_OPT" > /workdir/sim_telarray/build_opts.yml && \
 FROM almalinux:9.5-minimal
 WORKDIR /workdir
 ARG BUILD_BRANCH=main
+ARG SIMTOOLS_REQUIREMENT="git+https://github.com/gammasim/simtools.git"
 COPY --from=build_image /workdir/sim_telarray/ /workdir/sim_telarray/
 
 RUN microdnf update -y && microdnf install -y \
     bc findutils gcc-c++ git gsl-devel gcc-fortran libgfortran \
-    procps python3.11-pip zstd && \
+    procps python3.12-devel zstd && \
```
orelgueta, line 92: Should the python version be an argument that updates when we updated the supported versions?

orelgueta (commented):
Very minor comments.

.github/workflows/build-docker-corsika-simtelarray-image.yml
GernotMaier: Removed the `opt`. Good point.

.github/workflows/build-docker-corsika-simtelarray-image.yml
GernotMaier: Yes, needed - running the workflow without it will give you:

```
OCI runtime exec failed: exec failed: unable to start container process: exec: "sh": executable file not found in $PATH: unknown
```

docker/Dockerfile-prod-opt
GernotMaier: Good point, added.

GernotMaier:
Thanks @orelgueta for the review. I tried to address them, let me know if there is anything else.

orelgueta (approved):
Apologies for the delay, all good now!

GernotMaier:
Thanks a lot @orelgueta  - and no worries, no other PR depends directly on this work.

Issue #1317, opened 2025-01-22
Store arrays in model parameters as arrays and not as strings to simplify code.


GernotMaier:
Arrays are stored in the model parameters typically as strings, similar to the way sim_telarray is doing it.

Example:

```
{
    "schema_version": "0.2.0",
    "parameter": "telescope_transmission",
    "instrument": "LSTN-design",
    "site": "North",
    "parameter_version": "1.0.0",
    "unique_id": null,
    "value": "0.969 0.0 0.0 0.0 0.0 0.0",
    "unit": "null, null, null, null, null, null",
    "type": "float64",
    "file": false
}
```

This requires multiple string-to-list and list-to-string conversion in the code (`general.convert_string_to_list` and `gen.convert_list_to_string`.). I suggest that we store arrays as arrays (meaning above would be stored as `[0.969, 0.0, 0.0, 0.0, 0.0, 0.0]`) and we do only once a conversion to strings: when writing the sim_telarray configuration files.

PR #1423, opened 2025-03-10
Bug fix in DB-upload related string to list conversion

GernotMaier:
This is a bug fix to another issue (seems to be a weekly thing) of the string to list conversion needed for the DB upload (really should address this, see #1317).

This address a bug where a two number array as `240 1000.` is interpreted as `240.` and unit `Unit(dimensionless with a scale of 700)`.

Added a fix assuming the we don't use dimensionless units with a scale. Added also a test.

(as mentioned, this will all go away when addressing #1317).


EshitaJoshi (approved):
Thanks for the PR!

PR #1424, opened 2025-03-10
Run unit and integration test on database setup in local test environment.

GernotMaier:
Add functionality to use a local model parameter database in the test environment for unit and integration tests:

- testing will not be done anymore using the production DB and will not affect any production usage
- significant acceleration of testing 
  - integration tests: 40 min to 18 min
  - unit tests: 14 min to 4 min
- added `mongodb` as allowed DB for direct connection 
- changed option to run tests in parallel from `-n auto` to `-n 4` (github runners provide 4 CPUs; but `-n auto` is using only 2 CPUs in parallel).

Note that this required to remove the unit testing on macOS, as the mongoDB service is not available there. I do think we are fine developing for linux only.


orelgueta (approved):
This is all fine as it is, however, wouldn't it be more natural (and quicker/safer) to copy the production DB to the local instance rather than uploading to it from gitlab? 

Otherwise, the reduction in time is thanks to the local DB or thanks to the `-n 4`?

GernotMaier:
> This is all fine as it is, however, wouldn't it be more natural (and quicker/safer) to copy the production DB to the local instance rather than uploading to it from gitlab?

But then we would run in danger that tests affect productions, which we definitely want to avoid.

What does 'safer' mean in this context?

> 
> Otherwise, the reduction in time is thanks to the local DB or thanks to the `-n 4`?

Most of the reduction is due to the local DB, but can't quantify it. 

Thanks for the review!

orelgueta:
> > This is all fine as it is, however, wouldn't it be more natural (and quicker/safer) to copy the production DB to the local instance rather than uploading to it from gitlab?
> 
> But then we would run in danger that tests affect productions, which we definitely want to avoid.

No, you would just export the latest production DB and upload it to the local one you spin. That shouldn't affect anything in production and would also have the benefit of running the test with the actual DB rather than the repository which might not be exactly the same as the DB (that's part of the "safer" in my comment). The other "safer" point is that our DB should have a higher availability fraction than our GitLab instance (no actual data on it, just from experience of GitLab failing every once in a while). Lastly, I assume exporting and importing DB entries is quicker than the repo reading and uploading.

GernotMaier:
Agree with all what you said.

However, uploading from gitlab to the DB is an essential part of our workflow and by definition the outcome must be equivalent. And unfortunately, this is the only place where it is 'tested' (I agree that this might also be tested differently).

orelgueta:
OK, then would you agree that I open an issue so that in the future we implement both? That is, replace the workflow implemented here with one that exports from the production DB and add a new separate test for `upload_from_repo`?

GernotMaier:
Yes, please open an issue - although I really think the main point is to ensure that gitlab/DB is identical. Then everything else does not matter.

Issue #1403, opened 2025-02-27
Additional plotting functionality needed


EshitaJoshi:
Need an application to plot pixel data (as was previously done [here](https://gitlab.cta-observatory.org/cta-consortium/aswg/simulations/simulation-model/simulation-model-description/-/blob/master/produceReport.py#L1258)) to be able to generate plots for parameters such as ```camera_config_file```.

Need to also be able to plot data files for parameters using ```parameter_version``` and not (only) ```model_version```.

PR #1426, opened 2025-03-11
Improve reading of single parameter and files from DB

GernotMaier:
Simplify and improve reading of a single parameter entry and exporting from model parameter files:

- allow to retrieve model parameter dict using `DBHandler.get_model_parameter` for a given `parameter_version` or `model_version`  (before: only for a `parameter_version`); tried to generalize the reading of parameter with the choice of both version types.
- allow to export using the same function tables files
- allow to export (if applicable) tables files in astropy ecsv files.

Adjusted the application `simtools-db-get-parameter` for these changes with additional options.

This approach simplified especially the plotting routes. Plotting of parameter tables is now possible using the `parameter_version`, see example in [tests/resources/plot_table_data/plot_single_pe_config_parameter.yml](tests/resources/plot_table_data/plot_single_pe_config_parameter.yml).

Improved testing and added missing unit tests.

Addresses partly #1403.

### DIFF
--- a/docs/changes/1426.feature.md
+++ b/docs/changes/1426.feature.md
@@ -1,0 +1,1 @@
+Improve reading of single parameter and files from database. Add plotting of tabular data using `parameter_version`.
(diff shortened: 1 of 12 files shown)


EshitaJoshi (approved):
Thanks for the PR!

PR #1409, opened 2025-02-28
Comparing parameters across versions

EshitaJoshi:
Using production tables now to compare model parameters across model versions.

Closes Issue #1386 .

PR #1432, opened 2025-03-12
Adding function to get parameters for all model versions

EshitaJoshi:
Function returns a dictionary with model version as key and another dictionary as value. This dictionary has parameter names as key and a dictionary of parameter data as value.


src/simtools/reporting/docs_read_parameters.py
```diff
@@ -104,6 +104,7 @@ def get_array_element_parameter_data(self, telescope_model, collection="telescop
             site=telescope_model.site,
             array_element_name=telescope_model.name,
             collection=collection,
+            model_version="6.0.0",
```
GernotMaier, line 107: Why do we have to fix "6.0.0" here? At least when called from line 174, the model_version would be available as a parameter. Not sure about the call from line 199 - is this function called for a fixed model_version?

GernotMaier (approved):
Looks good - just have a look at the hardwired model version numbers. I don't understand why this is needed and what the implications are.

src/simtools/reporting/docs_read_parameters.py
EshitaJoshi: This change is just so that the unit tests pass here. The docs_read_parameters goes through quite a bit of changes in PR #1409, so this change here is just temporary. I can't push my local changes to that PR until this is merged to main since the function doesn't yet exist in db_handler.

PR #1433, opened 2025-03-13
Change base python version to 3.12

GernotMaier:
..as we are mid-way through the lifetime of 3.11...update the "base version" to 3.12 (I wanted to change to 3.13, but there are no microdfn package available yet for 3.13).

Added the python version as build argument for the docker containers.

Actually the main reason for this PR is that I've noticed that this unit test fails for good reason - but only for 3.13: `tests/unit_tests/job_execution/test_job_manager.py::test_submit_local_real_failure`
There was a bug due to incorrect mocking given "TypeError: '>' not supported between instances of 'MagicMock' and 'int'" (interesting that this was not an issue for other python versions).

Surprised about the number of places we have set the python version.


.github/workflows/CI-unittests.yml
```diff
@@ -25,14 +25,18 @@ jobs:
           - os: ubuntu-latest
             python-version: "3.11"
```
tobiaskleiner, line 26: also 3.12 here?

GernotMaier: Not here - we are still testing as part of the matrix for 3.11 (plus 3.12, 3.13). I think there is no reason to increase the minimum python version required.

tobiaskleiner (approved):
@GernotMaier all good then, thanks!

GernotMaier:
Thanks @tobiaskleiner

PR #1439, opened 2025-03-18
Use full array pointings if available

tobiaskleiner:
Small PR to include the option to add the full array pointings if present in the mc header. Also adds a log message if this is the case. Currently we have simulations with fixed pointings, so this is not used.

### DIFF
--- a/src/simtools/production_configuration/extract_mc_event_data.py
+++ b/src/simtools/production_configuration/extract_mc_event_data.py
@@ -155,7 +155,9 @@ class MCEventExtractor:
             array_azimuth = None
             for eventio_object in f:
                 if isinstance(eventio_object, MCRunHeader):
-                    self._process_mc_run_header(eventio_object, data_lists)
+                    array_altitude, array_azimuth = self._process_mc_run_header(
+                        eventio_object, data_lists
+                    )
                 elif isinstance(eventio_object, MCShower):
                     self._process_mc_shower(
                         eventio_object, data_lists, array_altitude, array_azimuth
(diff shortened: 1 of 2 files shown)


orelgueta:
Not sure if this is related, but the title made me think of the option for divergent/convergent pointing. Not something you should worry about now, but if you are building machinery to keep track of the telescope/array pointings, see if you can already take into account the option of individual telescope pointings (i.e., cases in which they do not all point in the same direction).

GernotMaier (approved):
All good, thanks!

src/simtools/production_configuration/extract_mc_event_data.py
```diff
@@ -192,10 +189,58 @@ def _process_mc_event(self, eventio_object, data_lists):
...
+        Parameters
+        ----------
+        tracking_positions : list of dict
+            List of tracking positions with "altitude" and "azimuth" keys.
+        data_lists : dict
+            Data lists to update.
+        """
+        altitudes = [pos["altitude"] for pos in tracking_positions]
+        azimuths = [pos["azimuth"] for pos in tracking_positions]
+        if isinstance(altitudes[0], list):
```
GernotMaier, line 230: Can you confirm that the length of the lists are always >0?

GernotMaier (approved):
one minor question - otherwise all good!

src/simtools/production_configuration/extract_mc_event_data.py
tobiaskleiner: Thanks! Yes this is the case always.

PR #1442, opened 2025-03-19
Robust getting of username in metadata collector

GernotMaier:
Metadata collector gets the user name from the system in case it is not defined as `SIMTOOLS_USER_NAME` in the environment (or passed through the command line).

This failed in case no user is defined (as on the gitlab AIV test environment).

Add an exception and issue a warning in these cases.

### DIFF
--- a/tests/unit_tests/data_model/test_metadata_collector.py
+++ b/tests/unit_tests/data_model/test_metadata_collector.py
@@ -490,7 +490,7 @@ def test_fill_contact_meta_failed_system(args_dict_site, caplog, monkeypatch):
     """Test filling contact metadata when system username lookup fails"""
 
     def mock_getuser():
-        raise Exception("Failed to get username")
+        raise KeyError("Failed to get username")
 
     monkeypatch.setattr(getpass, "getuser", mock_getuser)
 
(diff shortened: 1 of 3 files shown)


tests/unit_tests/data_model/test_metadata_collector.py
```diff
@@ -472,3 +472,33 @@ def test_clean_meta_data():
...
+def test_fill_contact_meta_from_system(args_dict_site, caplog):
+    """Test filling contact metadata from system when no name is provided"""
+    contact_dict = {}
+    collector = metadata_collector.MetadataCollector(args_dict=args_dict_site)
+
+    with caplog.at_level(logging.WARNING):
+        collector._fill_contact_meta(contact_dict)
+
+    assert "No user name provided, take user info from system level." in caplog.text
+    assert contact_dict["name"] == getpass.getuser()
```
orelgueta, line 486: Will you ever run the tests on the AIV system (or something similar)? This line would fail in that case, so perhaps you want to add a try..except here as well.

orelgueta (approved):
Small comment/question.

tests/unit_tests/data_model/test_metadata_collector.py
GernotMaier: At this point, it is not planned to run unit tests.

Anything, this is a good remark and I protected the statement with a try .. except: pass statement.

PR #1446, opened 2025-03-20
Add fitting of afterpulse spectrum to derive single pe application.

GernotMaier:
Add option to fit afterpulse spectrum to the derive single pe application. Fitted is a exponential function with linear decay term. Decay factor (`k`) can be fixed; this might be necessary if measurements of afterpulses do not reach large single pe amplitudes,

New configuration parameters for `derive_photon_electron_spectrum.py`:

- `fit_afterpulse` - turn on afterpulse fitting
- `fit_afterpulse_decay_factor` - this will fix the decay factor `k` to the given value (for LSTN, 15 is a good value)
- `afterpulse_amplitude_range` - fit function will be used in this range to predict afterpulse amplitudes forwarded to `norm_spe`

The integration test in `tests/integration_tests/config/derive_photon_electron_spectrum_lst_ecsv.yml` has been adopted and applies a fit now.

There is a bit of fine tuning for LSTN right now; expect that some smaller changes will be necessary when using other data.

Added also a fix in run_applications handling Paths/str for the output path naming.


src/simtools/applications/derive_photon_electron_spectrum.py
```diff
@@ -69,7 +69,7 @@ def _parse(label):
     )
     config.parser.add_argument(
         "--max_amplitude",
-        help="Maximum amplitude in amplitude spectrum",
+        help="Maximum amplitude in pe for amplitude spectrum",
```
orelgueta, line 72: probably better "for single p.e. spectrum"?

src/simtools/applications/derive_photon_electron_spectrum.py
```diff
@@ -82,10 +82,24 @@ def _parse(label):
         required=False,
     )
     config.parser.add_argument(
-        "--threshold_afterpulse_spectrum",
-        help="Threshold for afterpulse ratio calculation",
+        "--afterpulse_amplitude_range",
+        help="Amplitude range in pe for afterpulse calculation",
```
orelgueta, line 86: Is this range for calculating the afterpulse? (we don't actually calculate it.) 
Perhaps it is the range for fitting? In the other PR the limit from below was defaulted to 4.0 which makes more sense to me.

src/simtools/applications/derive_photon_electron_spectrum.py
```diff
@@ -82,10 +82,24 @@ def _parse(label):
...
+        required=False,
+    )
+    config.parser.add_argument(
+        "--fit_afterpulse",
+        help="Fit afterpulse spectrum with an exponential decay function.",
+        action="store_true",
+        required=False,
+    )
+    config.parser.add_argument(
+        "--fit_afterpulse_decay_factor",
```
orelgueta, line 99: Should this be `--fix_afterpulse_decay_factor`? If so, perhaps a better name would be `afterpulse_decay_factor_fixed_value`? Though I am really not sure so please don't accept it just cause I suggested.

src/simtools/applications/run_application.py
```diff
@@ -137,9 +137,7 @@ def read_application_configuration(configuration_file, steps, logger):
     application_config = gen.collect_data_from_file(configuration_file).get("CTA_SIMPIPE")
     place_holder = "__SETTING_WORKFLOW__"
     workflow_dir, setting_workflow = get_subdirectory_name(configuration_file)
-    output_path = (
-        workflow_dir.with_name(workflow_dir.name.replace("input", "output")) / setting_workflow
-    )
+    output_path = Path(str(workflow_dir).replace("input", "output") + "/" + setting_workflow)
```
orelgueta, line 140: We went back to string manipulations? =)

src/simtools/camera/single_photon_electron_spectrum.py
```diff
@@ -92,10 +102,21 @@ def write_single_pe_spectrum(self):
...
+    def _derive_spectrum_norm_spe(
+        self, input_spectrum, afterpulse_spectrum, afterpulse_fitted_spectrum
+    ):
         """
         Derive single photon electron spectrum using sim_telarray tool 'norm_spe'.
 
+        Parameters
+        ----------
+        input_spectrum : str
+            Input file with amplitude spectrum.
```
orelgueta, line 114: Is this the prompt spectrum? Perhaps better to call it that if so.

src/simtools/camera/single_photon_electron_spectrum.py
```diff
@@ -123,7 +146,7 @@ def _derive_spectrum_norm_spe(self):
         if tmp_ap_file:
             command.extend(["-a", f"{tmp_ap_file.name}"])
             command.extend(["-s", f"{self.args_dict.get('scale_afterpulse_spectrum', 1.0)}"])
```
orelgueta, line 148: Same comment as in the other PR about the default value.

src/simtools/camera/single_photon_electron_spectrum.py
```diff
@@ -180,3 +207,113 @@ def _get_input_data(self, input_file, frequency_column):
...
+        """
+        ap_min = self.args_dict["afterpulse_amplitude_range"][0]
+        fix_k = self.args_dict.get("fit_afterpulse_decay_factor")
+
+        x, y, y_err = self._read_afterpulse_spectrum_for_fit(
+            self.args_dict.get("afterpulse_spectrum"), ap_min
+        )
+        fit_func, p0, bounds = self.afterpulse_fit_function(fix_k=fix_k)
+
+        result = curve_fit(fit_func, x, y, sigma=y_err, p0=p0, bounds=bounds, absolute_sigma=True)
```
orelgueta, line 231: There is an assumption here that we always get uncertainties (and that they are in absolute values). Not sure we can always trust to have them, so perhaps a check would be useful.

tests/integration_tests/config/derive_photon_electron_spectrum_lst_ecsv.yml
```diff
@@ -9,14 +9,17 @@ CTA_SIMPIPE:
       AFTERPULSE_SPECTRUM: ./tests/resources/LST_afterpulses.ecsv
       STEP_SIZE: 0.02
       USE_NORM_SPE: true
+      FIT_AFTERPULSE: true
+      FIT_AFTERPULSE_DECAY_FACTOR: 15
+      AFTERPULSE_AMPLITUDE_RANGE: [1, 42]
```
orelgueta, line 14: Range is different here than the default and the test below.

tests/unit_tests/camera/test_single_photon_electron_spectrum.py
```diff
@@ -21,6 +22,7 @@ def spe_spectrum():
         "max_amplitude": 1.0,
         "afterpulse_spectrum": None,
         "input_spectrum": "input_spectrum",
+        "afterpulse_amplitude_range": [4.0, 42.0],
```
orelgueta, line 25: Range is different (but consistent with previous PR)

orelgueta (commented):
Minor comments, not approving yet though

src/simtools/applications/derive_photon_electron_spectrum.py
GernotMaier: Fixed.

src/simtools/applications/derive_photon_electron_spectrum.py
GernotMaier: - fit is done in the range of available data points (so no user input on this)
- `afterpulse_amplitude_range` is the range over which the fitted function is used to calculate afterpulses and forward it to norm_spe

Any suggestion on how to name it better? I don't have a good idea...

src/simtools/applications/derive_photon_electron_spectrum.py
GernotMaier: It should be 'fix'... I've taken your suggestion.

src/simtools/camera/single_photon_electron_spectrum.py
GernotMaier: It doesn't have to. It can be prompt + afterpulse, in case no afterpulse spectrum is given. I've added 

`            Input file with amplitude spectrum (prompt spectrum only if afterpulse spectrum is given).`

src/simtools/camera/single_photon_electron_spectrum.py
GernotMaier: Fixed there.

src/simtools/camera/single_photon_electron_spectrum.py
GernotMaier: Is this unreasonable to require?

`_read_afterpulse_spectrum_for_fit` will fail with column "frequency stdev (afterpulsing)" not found. My assumption was that this is enough information for the user. Should I add more?

tests/integration_tests/config/derive_photon_electron_spectrum_lst_ecsv.yml
GernotMaier: Yes - an integration test can/should(?) use values different from default.

src/simtools/applications/derive_photon_electron_spectrum.py
orelgueta: Name is fine (or it could be `afterpulse_spectrum_range` if you want). Perhaps the point is about the word "calculation", because I don't expect we calculate the afterpulsing spectrum. Either way, it's not a big deal, you can leave it as is as well.

src/simtools/applications/run_application.py
orelgueta: The previous one should also work, you just need to cast to path as in

`Path(workflow_dir).with_name(workflow_dir.name.replace("input", "output")) / setting_workflow`

One shouldn't do string manipulations for paths...

src/simtools/camera/single_photon_electron_spectrum.py
orelgueta: OK!

src/simtools/camera/single_photon_electron_spectrum.py
orelgueta: No, that's fine. Whether it's unreasonable to require I don't know, we don't have any procedure set up for these measurements.

tests/integration_tests/config/derive_photon_electron_spectrum_lst_ecsv.yml
orelgueta: OK!

tests/unit_tests/camera/test_single_photon_electron_spectrum.py
orelgueta: I wasn't worried about using different values, just making sure it is not a mistake (and connected to the point about what this range actually is).

src/simtools/applications/run_application.py
GernotMaier: Fixed it - I think our issue was that we did not expect that `Path.parts` returns strings (obviously...these are parts of a Path).

PR #1450, opened 2025-03-20
Refactor light emission application

tobiaskleiner:
PR to refactor the light emission application.

Moves most of the functionality into helper functions and reduces some of the duplications.


GernotMaier (changes requested):
Could some code from the simulate_light_emission application be moved to modules? Application code is not unit tested, which makes me a bit nervous.

Otherwise all good.

tobiaskleiner:
@GernotMaier thanks for the review. Moved now most of the functionality into modules and added few more tests.

GernotMaier (approved):
All good now, thanks a lot!

PR #1451, opened 2025-03-20
Use consistently numpy dtypes for model parameters

### LINKED ISSUE #1216: Sync type definition in schemas and model parameter files
Model parameter file use numpy-style type definitions (e.g., int64, float64), while schema files require python types. This should by synchronized.

My suggestion is to consistently use numpy types, to ensure the required precision in some cases.

GernotMaier:
Model parameter data are currently defined sometimes as python and sometimes as numpy types. No obvious reason why one or the other is used.

This merge request changes the schemas to use consistently numpy types. There is a corresponding merge request for the model parameter database here: https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/merge_requests/22

To list the define files types:

```
cd src/simtools/schemas/model_parameters
grep -h "\- type" * | sort -u
```

Closes #1216

### DIFF
--- a/src/simtools/schemas/model_parameter_and_data_schema.metaschema.yml
+++ b/src/simtools/schemas/model_parameter_and_data_schema.metaschema.yml
@@ -362,8 +362,10 @@ definitions:
       - double
       - float64
       - int
+      - int64
       - string
       - uint
+      - uint64
       - file
     title: TableColumnType
   Unit:
--- a/docs/changes/1451.maintenance.md
+++ b/docs/changes/1451.maintenance.md
@@ -1,0 +1,1 @@
+Use consistently numpy dtypes for model parameters.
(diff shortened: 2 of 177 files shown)


src/simtools/schemas/model_parameter_and_data_schema.metaschema.yml
```diff
@@ -362,8 +362,10 @@ definitions:
       - double
```
tobiaskleiner, line 362: Should we then remove the non numpy dtypes here?

GernotMaier: I kept this for now, as our current database is using doubles. We don't have a rule, but a little bit of backwards compatibility is probably required.

tobiaskleiner (approved):
Thanks @GernotMaier, all good.

PR #1452, opened 2025-03-20
Catch and raise exception when reading files

GernotMaier:
`simtools.utils.general.collect_data_from_file()` is used to read dictionary from files for various data types. When failing, file names are not printed, which makes debugging difficult.

As an example - trying to validate a large number of schema files can be done with

```
simtools-validate-file-using-schema --file_directory "../simulation-models/simulation-models/model_parameters/" --data_type model_parameter
```

But when failing, no file name is printed in the error message.

This PR changes this - catching broad exceptions (note that we are reading different file type; so this can be a file not found or a json encoder error) and printing the correct statements.


src/simtools/utils/general.py
```diff
@@ -180,13 +180,18 @@ def collect_data_from_file(file_name, yaml_document=None):
...
-        if suffix in [".yml", ".yaml"]:
-            return _collect_data_from_yaml_file(file, file_name, yaml_document)
+    try:
+        with open(file_name, encoding="utf-8") as file:
+            if suffix == ".json":
+                return json.load(file)
+            if suffix == ".list":
+                return [line.strip() for line in file.readlines()]
+            if suffix in [".yml", ".yaml"]:
+                return _collect_data_from_yaml_file(file, file_name, yaml_document)
```
orelgueta, line 190: I find it a bit strange that first we open the file and then test the suffix and decide on the reading method. Would the following code be better?

```
def collect_data_from_file(file_name, yaml_document=None):
    """
    Collect data from file based on its extension.
    Parameters
    ----------
    file_name: str
        Name of the yaml/json/ascii file.
    yaml_document: None, int
        Return list of yaml documents or a single document (for yaml files with several documents).
    Returns
    -------
    data: dict or list
        Data as dict or list.
    """
    if is_url(file_name):
        return collect_data_from_http(file_name)

    file_handlers = {
        '.json': lambda f: json.load(f),
        '.list': lambda f: [line.strip() for line in f.readlines()],
        '.yml': lambda f: _collect_data_from_yaml_file(f, file_name, yaml_document),
        '.yaml': lambda f: _collect_data_from_yaml_file(f, file_name, yaml_document)
    }

    suffix = Path(file_name).suffix.lower()
    handler = file_handlers.get(suffix)

    if not handler:
        return None

    try:
        with open(file_name, encoding="utf-8") as file:
            return handler(file)
    except Exception as exc:  # pylint: disable=broad-except
        error_msg = f"Failed to read file {file_name}: {exc}"
        _logger.error(error_msg)
        raise type(exc)(error_msg) from exc
```

orelgueta (approved):
Tiny suggestion for modified code, but not a blocking one (same functionality), so I approve already.

src/simtools/utils/general.py
GernotMaier: Agree that this is slightly more readable, but 7 lines longer...

Would also had to add at least the check for file_name is None. I've tried the code above and it requires to change 4 unit tests with secondary relation to collect_data_from_file. I would prefer therefore to leave it as it is - hope this is fine.

PR #1454, opened 2025-03-20
Run scheduled nightly tests on production DB

### LINKED ISSUE #1434: Add a test to check that the GitLab model repo and the DB are identical
We should have a (periodic?) test which checks that the main branch in the [GitLab simulation model repo](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models) is identical/equivalent to the entries in the DB.

GernotMaier:
This has been tested by excluding not `schedule` but `workflow_dispatch` and run the unit in the usually PR CI plus triggered by hand:

- usually CI uploads to local DB https://github.com/gammasim/simtools/actions/runs/13987110547/job/39162743143
- by hand: no upload to local DB https://github.com/gammasim/simtools/actions/runs/13987145563/job/39162849793

Note that the mongoDB service is running in both cases - although not used for when using the production DB. 

Closes #1434

### DIFF
--- a/.github/workflows/CI-integrationtests.yml
+++ b/.github/workflows/CI-integrationtests.yml
@@ -68,8 +68,8 @@ jobs:
         run: |
           echo "PATH=\$PATH:/usr/bin:/usr/local/bin:$SIMTOOLS_SIMTEL_PATH" >> "$GITHUB_ENV"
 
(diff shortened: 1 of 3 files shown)


orelgueta (approved):
All good!

PR #1455, opened 2025-03-20
Avoid output.err output.out files written during unit tests

GernotMaier:
The unit test `tests/unit_tests/job_execution/test_job_manager.py::test_submit_local_real_failure` writes two files into the local directory when running: output.err and output.out.

This PR avoids that by mocking builtins_open.


EshitaJoshi (approved):
Thanks for the PR!

PR #1457, opened 2025-03-24
Fix failing test of production images - set consistent date

GernotMaier:
Fix of failing tests of production images, see e.g., https://github.com/gammasim/simtools/actions/runs/14013865711/job/39252381358 

Underlying issue that we define the build matrix three times in the workflow file (didn't find a more elegant solution) - and all three have to be identical. We have commented out different versions at different steps; this is fixed now.

(they are commented out as right now there is no need to build all possible combinations of versions and build args).

Fixed pipeline: https://github.com/gammasim/simtools/actions/runs/14030126370/job/39276114170

### DIFF
--- a/.github/workflows/build-simtools-production-image.yml
+++ b/.github/workflows/build-simtools-production-image.yml
@@ -136,8 +136,8 @@ jobs:
     strategy:
       matrix:
         version:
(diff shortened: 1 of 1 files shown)


EshitaJoshi (approved):

GernotMaier:
Thanks @EshitaJoshi

PR #1461, opened 2025-03-25
Extract the CORSIKA version and add an application that prints the various versions

orelgueta:
The CORSIKA version is extracted by calling it and searching for it in the stdout. In order to test this for real, without mocking, an application is added to print all of the versions (simtools/sim_telarray/CORSIKA/DB) and also save them to a json file.

The validation of the output from the application is missing. At the moment I don't see a simple way to add a comprehensive check without reusing the same functionality we are testing, which defeats the purpose. If the reviewer thinks it's critical to have, we can try to come up with a way to implement the validation (or a reduced version of it). Otherwise, I suggest we live without this validation.


GernotMaier (approved):
Nice, thanks!

Fixed one typo and removed on obsolete line of comments.

Would be nice if `simtools-print-version` without any options would work. Added a comment to #1055 that this should be fixed.

orelgueta:
> Would be nice if `simtools-print-version` without any options would work. Added a comment to #1055 that this should be fixed.

It actually did originally, but one of the problems is that the schema validation of the configuration file requires a "CONFIGURATION" field. Since I anyway added the file name, I decided not to follow this issue further, but I agree it would be nice.

Thanks for the review!

PR #1464, opened 2025-03-26
Change to using files for array layout and event data filepaths in corsika limits derivation

tobiaskleiner:


GernotMaier:
@tobiaskleiner  - can you upload an example of the ecsv output file?

tobiaskleiner:
> @tobiaskleiner - can you upload an example of the ecsv output file?
Changed fileending to txt for upload.
[corsika_simulation_limits_lookup.txt](https://github.com/user-attachments/files/19487555/corsika_simulation_limits_lookup.txt)

GernotMaier:
Thanks for this - two comments for the future:

- not convinced that we need the file name in the table. The metadata should be responsible for the traceability
- in future we probably will replace the telescope list by a (sub)array name. Lists are hard to browse.

GernotMaier (approved):
All good, thanks for this!

PR #1465, opened 2025-03-27
Retrieve CTAO layout definitions. Merge with model parameter value and write to disk.

### LINKED ISSUE #1224: Derive array layout model parameter from common CTAO definitions.
CTAO defines common sub arrays [here](https://gitlab.cta-observatory.org/cta-computing/common/identifiers/-/blob/main/subarray-ids.json?ref_type=heads).

Add an application to use these definitions to write the model parameter `array_layouts.json`.

At a later stage, this information will be retrieved from a SOSS database.

### LINKED ISSUE #1317: Store arrays in model parameters as arrays and not as strings to simplify code.

GernotMaier:
Add small simtool which syncs the array layout definition used in CTAO computing ([CTAO gitlab](https://gitlab.cta-observatory.org/cta-computing/common/identifiers/-/blob/main/subarray-ids.json?ref_type=heads)) with the `array_layouts` parameters. Compares both names and list of telescopes. Updates the model parameter entries and add layouts from CTAO computing. Writes out a model parameter-style json file with updated array layouts.

Note that future versions of this tool might access a dedicated CTAO database for this information.

Closes #1224


src/simtools/layout/ctao_array_layouts.py
```diff
@@ -0,0 +1,166 @@
...
+    layouts_per_site = []
+
+    for array in sub_arrays.get("subarrays", []):
+        elements = []
+        for ids in array.get("array_element_ids", []):
+            element_name = _get_array_element_name(ids, array_element_ids)
+            if names.get_site_from_array_element_name(element_name) != site:
+                break
+            elements.append(element_name)
+        if len(elements) > 0:
```
tobiaskleiner, line 74: Can we have subarrays with zero elements?

src/simtools/layout/ctao_array_layouts.py
```diff
@@ -0,0 +1,166 @@
...
+            }
+            layouts_per_site.append(array_layout)
+
+    _logger.info(f"CTAO array layout definition: {layouts_per_site}")
+    return layouts_per_site
+
+
+def _get_array_element_name(ids, array_element_ids):
+    """Return array element name for common identifier."""
+    for element in array_element_ids.get("array_elements", []):
```
tobiaskleiner, line 87: When is the default empty list used here?

tests/unit_tests/layout/test_ctao_array_layouts.py
```diff
@@ -0,0 +1,214 @@
...
+    site = "south"
+    layouts = cta_array_layouts._get_layouts_per_site(
+        site,
+        {"subarrays": [{"name": "array1", "array_element_ids": ["T1", "T2"]}]},
+        array_element_ids,
+    )
+    assert len(layouts) == 0
+
+
+def test_retrieve_array_layouts_from_url():
```
tobiaskleiner, line 189: Would it make sense to actually test the retrival from the common identifiers repository and not only mock it?

tobiaskleiner (changes requested):
Thanks @GernotMaier! Few minor comments, but all good otherwise.

src/simtools/layout/ctao_array_layouts.py
GernotMaier: The issue is that we get here data from outside of simtools and we don't have control. I therefore added this check to be safe.

src/simtools/layout/ctao_array_layouts.py
GernotMaier: When the key "array_elements" is not defined. Again - just to be safe as we are reading information from outside of simtools.

src/simtools/layout/ctao_array_layouts.py
```diff
@@ -0,0 +1,166 @@
...
+    return None
+
+
+def merge_array_layouts(layouts_1, layouts_2):
+    """
+    Compare array layout dictionaries and merge them.
+
+    Parameters
+    ----------
+    layout_1 : dict
```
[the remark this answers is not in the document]
GernotMaier: Don't understand your comment.

src/simtools/layout/ctao_array_layouts.py
```diff
@@ -0,0 +1,166 @@
...
+
+def write_array_layouts(array_layouts, args_dict, db_config):
+    """
+    Write array layouts as model parameter.
+
+    Parameters
+    ----------
+    args_dict : dict
+        Command line arguments.
+    array_layouts : dict
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

tests/unit_tests/layout/test_ctao_array_layouts.py
GernotMaier: Unit tests should whenever possible run without external dependencies like database connections or file retrieval. This is why we want to mock here. The integration test is testing that we can retrieve the file from the gitlab.

GernotMaier:
@tobiaskleiner - thanks for the review. There is one comment I didn't understand, can you help me with that?

src/simtools/layout/ctao_array_layouts.py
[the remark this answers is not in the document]
tobiaskleiner: The naming is different here in the comment than in the definition above (two times missing s)

src/simtools/layout/ctao_array_layouts.py
tobiaskleiner: ok

src/simtools/layout/ctao_array_layouts.py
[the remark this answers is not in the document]
GernotMaier: OK - fixed!

tobiaskleiner (approved):

GernotMaier:
Thanks for the review @tobiaskleiner

PR #1466, opened 2025-03-27
Store arrays in model parameters as arrays and not as strings. 

### LINKED ISSUE #1317: Store arrays in model parameters as arrays and not as strings to simplify code.

GernotMaier:
Consistently write lists/arrays as such and not as simtel-style strings.

- remove all conversions to strings with the exception simtel_config_writer (there it is needed, as sim_telarray requires strings)
- kept the string-to-list conversion on the "reading side" - this ensures backwards compatibility
- updated model parameter schema to allow list of strings (or list of Nones) for the unit field

Related to [MR 23 in SimulationModels](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/merge_requests/23).

Closes #1317.

### DIFF
--- a/database_scripts/upload_from_model_repository_to_db.sh
+++ b/database_scripts/upload_from_model_repository_to_db.sh
@@ -7,7 +7,7 @@
 set -e
 
 DB_SIMULATION_MODEL_URL="https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models.git"
-DB_SIMULATION_MODEL_BRANCH="strings-to-arrays"
+DB_SIMULATION_MODEL_BRANCH="main"
 
 # Check that this script is not sourced but executed
 if [[ "${BASH_SOURCE[0]}" != "${0}" ]]; then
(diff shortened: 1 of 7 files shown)


EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thanks for the quick review - @EshitaJoshi !

PR #1459, opened 2025-03-25
Add site parameter reporting

EshitaJoshi:
This PR adds reporting functionality for OBS-North and OBS-South parameters.

It can be run using the --all flags like this:

`python src/simtools/applications/docs_produce_array_element_report.py --observatory --all_sites --all_model_versions`

or individually like this:

`python src/simtools/applications/docs_produce_array_element_report.py --observatory --site North --model_version 5.0.0`

PR #1467, opened 2025-03-27
Fix building of prod image from branch.

GernotMaier:
Testing related to PR #1459 showed that the production images are not build correctly from a branch - simtools is always installed from main (despite having set the BUILD_BRANCH build argument correctly). This PR fixes this issue.


EshitaJoshi (approved):
Thanks for the PR and the quick fix!

PR #1468, opened 2025-03-27
Remove error message duplication in general.collect_data_from_file

GernotMaier:
I had been very confused by the large number of error messages when running schema validation in the simulation models repository (e.g., see [this job](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/jobs/307059)) - error messages without job failure was unexpected.

Reason was that with both used `logger.error` and `raise FileNotFoundError(...)` to issue the same error message. The catching of the error in another class did not prevent logger from issuing the error. This has been fixed by removing the logger.error statement.


EshitaJoshi (approved):

GernotMaier:
Thanks @EshitaJoshi

PR #1473, opened 2025-03-31
Ensure files upload to DB follow unicode

### LINKED ISSUE #1463: Simulation model table files should follow unicode
e.g. currently `Benn_LaPalma_sky_converted.lis` is not following unicode.

This means we have to:

- add a check of the files before uploading to the simulation models repository
- remove any code in simtools which tries to handle this situation with try...except statements.

GernotMaier:
Closes #1463. Related to [gitlab merge request 32](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/merge_requests/32).

Tests will fail with the current simulation model version 0.4.0 - this is correct, as not all files follow unicode.


tests/unit_tests/utils/test_general.py
```diff
@@ -912,3 +912,24 @@ def test_is_valid_boolean_type():
     assert not gen._is_valid_boolean_type(np.int32, None)
     assert not gen._is_valid_boolean_type(np.float32, None)
     assert not gen._is_valid_boolean_type(np.str_, None)
+
+
+def test_is_utf8_file(tmp_test_directory):
```
orelgueta, line 917: Note that it might be worth merging this test with the `test_read_file_encoded_in_utf_or_latin` test cause you can use the same files created on the fly to test both.

GernotMaier: Very good point. I've removed the separated test_is_utf8 function and merged the tests.

EshitaJoshi (approved):
Thanks for the PR!

PR #1475, opened 2025-04-01
Primary degraded map is a parameter applied for all telescope types.

GernotMaier:
This was on oversight before - primary degraded map can also be used for single-mirror telescopes, see the screenshot below from the sim_telarray manual:

Screenshot 2025-04-01 at 10 14 11

### DIFF
--- a/docs/changes/1475.bugfix.md
+++ b/docs/changes/1475.bugfix.md
@@ -1,0 +1,1 @@
+Primary degraded map is a parameter applied for all telescope types.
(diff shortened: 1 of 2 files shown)


orelgueta (commented):
You asked for my review but the PR is still a draft (and has integration tests which fail).

Anyway, it's all fine, except one question. Should we change the schema version to 0.2.0?

GernotMaier:
I didn't expect the integration tests to fail..this is why I didn't move it to 'ready for review'. Sorry, shouldn't have asked for a review. 

The schema version should stay 0.1.0 (this is the only one defined for [model_parameter_and_data_schema.metaschema.yml](https://github.com/gammasim/simtools/blob/main/src/simtools/schemas/model_parameter_and_data_schema.metaschema.yml)

orelgueta:
OK, please let me know when the tests pass and you want me to review then (though fixing those is not related to this PR, so I assume there will be no further changes).

PR #1481, opened 2025-04-01
Adding links to array layout plots

EshitaJoshi:
Minor PR to fix broken links and add references to array layout plots.


src/simtools/reporting/docs_read_parameters.py
```diff
@@ -375,11 +375,13 @@ def produce_model_parameter_reports(self):
 
                 file.write("\n")
                 if comparison_data.get(parameter)[0]["file_flag"]:
-                    file.write(f"![Parameter plot.](_images/{self.array_element}_{parameter}.png)")
+                    file.write(
+                        f"![Parameter plot.](../../_images/{self.array_element}_{parameter}.png)"
```
GernotMaier, line 379: Those relative paths are not nice - especially as they appear several times (which will make any updates / changes) cumbersome. Could you think of a better way on how to organize this It seems now it depends on from where we execute the simtool.

GernotMaier (changes requested):
see comment below - everything else is nice!

src/simtools/reporting/docs_read_parameters.py
EshitaJoshi: Added IMAGE_PATH to avoid writing the path multiple times. Also added a reference to the plots for the parameters in the parameter files for a given model version.

GernotMaier (approved):
All good, thanks!

PR #1488, opened 2025-04-04
Skip testing of DB upload for simpipe integration tests.

GernotMaier:
Simpipe doesn't allow to open a sandbox database (or any new database). We therefore skip the integration tests uploading to the DB (these are anyway applications for developers, and not for users).


orelgueta:
> (these are anyway applications for developers, and not for users)

That is true for this DB and for now. If and when we have a simulation production DB, we will probably need to also test uploading to it in all frameworks.

orelgueta (approved):

PR #1489, opened 2025-04-04
Separate TelescopeModel and SiteModel for sim_telarray configuration writer

GernotMaier:
TelescopeModel and SiteModel have not been separated in the ModelParameter and SimtelConfigWriter routines - site parameters have been added simply to the telescope model. This had some side effects, like that checks for site parameters where necessary when writing telescope parameters. 

This PR refactors the code and introduces a clear separation of TelescopeModel and SiteModel. Unfortunately, this required quite some changes, but I think the code is now easier to read and definitely more consistent.

Notes:

- `ModelParameter.get_config_file` has been removed and replaced by `get_config_file` (a property) and `write_sim_telarray_config_file` (this hopefully clarifies from where we call the config file writer).
- there are a couple of checks e.g. for `array_triggers` parameter as part of TelescopeModel in the code. These will be fixed in a follow up PR.


src/simtools/model/model_parameter.py
```diff
@@ -446,40 +438,67 @@ def export_parameter_file(self, par_name, file_path):
...
+            parameters=pars_from_db,
+            dest=destination_path or self.config_file_directory,
+        )
         self._is_exported_model_files_up_to_date = True
 
-    def export_config_file(self):
-        """Export the config file used by sim_telarray."""
-        if not self._is_exported_model_files_up_to_date:
-            self.export_model_files()
+    def write_sim_telarray_config_file(self, add_model=None):
```
orelgueta, line 474: I think it would be better to call it `additional_model`, just because I would have expected `add_model` to be a boolean. Clearly not important though.

src/simtools/model/model_parameter.py
```diff
@@ -446,40 +438,67 @@ def export_parameter_file(self, par_name, file_path):
...
+        Write the sim_telarray configuration file.
+
+        Parameters
+        ----------
+        add_model: TelescopeModel or SiteModel
+            Model object for additional parameter to be written to the config file.
+        """
+        self.export_model_files(update_if_necessary=True)
+
+        self.parameters.update(self._simulation_config_parameters.get("simtel", {}))
```
orelgueta, line 485: Should these two lines be reversed? Or perhaps with this order you avoid exporting simulation config  parameters?

src/simtools/model/model_parameter.py
```diff
@@ -446,40 +438,67 @@ def export_parameter_file(self, par_name, file_path):
...
-            config_parameters=self._simulation_config_parameters["sim_telarray"],
         )
 
     @property
     def config_file_directory(self):
         """Directory for configure files. Configure, if necessary."""
         if self._config_file_directory is None:
             self._set_config_file_directory_and_name()
-        return self._config_file_directory
+        return self._config_file_directory or self._set_config_file_directory_and_name()
```
orelgueta, line 501: Wouldn't this make the code above it checking if `None` redundant? I suggest applying the same lazy init to the config_file_path while we are at it.

src/simtools/ray_tracing/mirror_panel_psf.py
```diff
@@ -63,37 +63,32 @@ def _define_telescope_model(self, label, db_config):
...
-            Dictionary with database configuration.
-
         Returns
         -------
-        tel TelescopeModel
-            telescope model
-
+        tel : TelescopeModel
+            The telescope model.
+        site_model : any
```
orelgueta, line 70: any?

src/simtools/runners/corsika_simtel_runner.py
```diff
@@ -157,8 +157,10 @@ def _make_run_command(self, run_number=None, input_file=None):
         except TypeError:  # allow for self.label to be None
             weak_pointing = False
 
+        self.corsika_config.array_model.export_all_simtel_config_files()
```
orelgueta, line 160: I am worried that this called from `corsika_cofnig` rather than from the `simulator_array`, but it might be OK. We will see once I merge this to the multipipe branch.

src/simtools/simtel/simulator_ray_tracing.py
```diff
@@ -137,7 +141,13 @@ def _load_required_files(self, force_simulate):
             self._write_out_single_pixel_camera_file()
 
     def _make_run_command(self, run_number=None, input_file=None):  # pylint: disable=unused-argument
-        """Generate sim_telarray run command."""
+        """
+        Generate simtel_array run command. Export sim_telarray configuration file(s).
+
+        Parameters are not relevant for the ray tracing simulation.
```
orelgueta, line 147: I suggest to be a little more specific. That is "The `run_number` and `input_file` parameters are not relevant for the ray tracing simulations and are given as arguments only to comply with the structure of the parent class function" (if I even understood it correctly...)

orelgueta (commented):
Very nice. Simplifies the structure, some of the code and makes things clearer. Minor comments.

src/simtools/model/model_parameter.py
GernotMaier: Agree. Fixed.

src/simtools/model/model_parameter.py
GernotMaier: It doesn't matter at this point, as `configuration_corsika` and `configuration_sim_telarray` does not have any files to export. But you are right - fixed it.

src/simtools/model/model_parameter.py
GernotMaier: Yes. Fixed it.

src/simtools/ray_tracing/mirror_panel_psf.py
GernotMaier: Glitch in Copilot...fixed it.

src/simtools/runners/corsika_simtel_runner.py
GernotMaier: Ok, we wait and see if it is an issue?

orelgueta: Yes, we'll see later.

src/simtools/model/model_parameter.py
```diff
@@ -495,10 +497,10 @@ def write_sim_telarray_config_file(self, add_model=None):
 
     @property
     def config_file_directory(self):
-        """Directory for configure files. Configure, if necessary."""
+        """Directory for configuration files. Configure if not yet set."""
         if self._config_file_directory is None:
             self._set_config_file_directory_and_name()
-        return self._config_file_directory or self._set_config_file_directory_and_name()
+        return self._config_file_directory
```
orelgueta, line 503: I thought the opposite, i.e., that 

```
return self._config_file_directory or self._set_config_file_directory_and_name()
```

would cover both cases and you won't need 

```
if self._config_file_directory is None:
            self._set_config_file_directory_and_name()
```

But it doesn't matter, we can keep it like this.

orelgueta (approved):

PR #1492, opened 2025-04-07
Fix unit tests for skipping tests when using production DB

GernotMaier:
Possibly fixes unit tests in testing routines which allows to skip integration tests depending on certain DB configurations (see #1488).

Nightly tests fail: https://github.com/gammasim/simtools/actions/runs/14287388882/job/40044146465#step:13:2013

I think this fixes the issues, although I did not manage to get a failing test with my local setup (tried different DB configurations; random execution of tests).


EshitaJoshi (approved):

GernotMaier:
Thanks @EshitaJoshi

PR #1495, opened 2025-04-07
Remove condition in simulator light emission and patch open

tobiaskleiner:
Removes the test condition by patching path open for reading the configuration file.


GernotMaier (approved):
All good, thanks!

PR #1496, opened 2025-04-07
First PR for support multiple model versions

orelgueta:
This PR adds the possibility to write out the configuration of multiple models for sim_telarray to run in multipipe mode. That technically already works, but I am pretty sure further optimization of file names, output redirection, etc. are needed. Nevertheless, since this first step already works and the tests pass (locally at least), I think it's best to split the work to multiple PRs and start with this one.

### DIFF
--- a/src/simtools/corsika/corsika_config.py
+++ b/src/simtools/corsika/corsika_config.py
@@ -124,9 +124,12 @@ class CorsikaConfig:
         if db_config is None:  # all following parameter require DB
             return config
 
-        db_model_parameters = ModelParameter(
-            mongo_db_config=db_config, model_version=args_dict["model_version"]
-        )
+        model_version = args_dict.get("model_version", None)
+        # If the user provided multiple model versions, we take the first one
+        # because for CORSIKA config we need only one and it doesn't matter which
+        if isinstance(model_version, list):
+            model_version = model_version[0]
+        db_model_parameters = ModelParameter(mongo_db_config=db_config, model_version=model_version)
         parameters_from_db = db_model_parameters.get_simulation_software_parameters("corsika")
 
         config["INTERACTION_FLAGS"] = self._corsika_configuration_interaction_flags(
--- a/src/simtools/runners/corsika_simtel_runner.py
+++ b/src/simtools/runners/corsika_simtel_runner.py
@@ -20,7 +20,7 @@ class CorsikaSimtelRunner:
 
     Parameters
     ----------
-    corsika_config : list
+    corsika_config : CorsikaConfig or list of CorsikaConfig
         A list of "CorsikaConfig" instances which
         contain the CORSIKA configuration parameters.
     simtel_path : str or Path
@@ -43,8 +43,10 @@ class CorsikaSimtelRunner:
         sim_telarray_seeds=None,
     ):
         self._logger = logging.getLogger(__name__)
+        if not isinstance(corsika_config, list):
+            corsika_config = [corsika_config]
         self.corsika_config = corsika_config
-        # the main corsika config is the one used to define the CORSIKA specific configuration.
+        # the main corsika config is the one used to define the CORSIKA specific parameters.
         # The others are used for the array configurations. The name "main" is a bit misleading.
         self.main_corsika_config = corsika_config[0]
         self._simtel_path = simtel_path
@@ -162,9 +164,9 @@ class CorsikaSimtelRunner:
         input_file: str
             Full path of the input CORSIKA file.
             Use '-' to tell sim_telarray to read from standard output
-        corsika_config: CorsikaConfig
+        corsika_config: CorsikaConfig, optional
             CORSIKA configuration.
-        simulator_array: SimulatorArray
+        simulator_array: SimulatorArray, optional
             SimulatorArray instance.
 
         Returns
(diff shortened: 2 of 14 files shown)


orelgueta:
@GernotMaier, before reviewing this, we should check if there are conflicts with your branch. I don't think it will be too bad, but it's worth checking before reviewing.

src/simtools/corsika/corsika_config.py
```diff
@@ -124,9 +124,12 @@ def fill_corsika_configuration(self, args_dict, db_config=None):
         if db_config is None:  # all following parameter require DB
             return config
 
-        db_model_parameters = ModelParameter(
-            mongo_db_config=db_config, model_version=args_dict["model_version"]
-        )
+        model_version = args_dict.get("model_version", None)
+        # If the user provided multiple model versions, we take the first one
```
GernotMaier, line 128: Do we need a test that the corsika configs are actually the same for all model_versions? It is likely to be the same, but we can't be sure, or?

src/simtools/simulator.py
```diff
@@ -114,16 +114,27 @@ def _initialize_array_model(self, mongo_db_config):
...
-            ArrayModel object.
-        """
-        return ArrayModel(
-            label=self.label,
-            site=self.args_dict.get("site"),
-            layout_name=self.args_dict.get("array_layout_name"),
-            mongo_db_config=mongo_db_config,
-            model_version=self.args_dict.get("model_version", None),
-        )
+        list
```
GernotMaier, line 117: To be consistent in naming, change this to `_initialize_array_models`

src/simtools/simulator.py
```diff
@@ -114,16 +114,27 @@ def _initialize_array_model(self, mongo_db_config):
...
+            List of ArrayModel objects.
+        """
+        model_versions = self.args_dict.get("model_version", None)
+        array_models = []
```
GernotMaier, line 121: Replace by 

```
    versions = versions if isinstance(versions, list) else [versions]

    return [
        ArrayModel(
            label=self.label,
            site=self.args_dict.get("site"),
            layout_name=self.args_dict.get("array_layout_name"),
            mongo_db_config=mongo_db_config,
            model_version=v,
        )
        for v in versions
    ]
```

src/simtools/runners/corsika_simtel_runner.py
```diff
@@ -42,26 +43,37 @@ def __init__(
         sim_telarray_seeds=None,
     ):
         self._logger = logging.getLogger(__name__)
+        if not isinstance(corsika_config, list):
+            corsika_config = [corsika_config]
         self.corsika_config = corsika_config
+        # the main corsika config is the one used to define the CORSIKA specific parameters.
+        # The others are used for the array configurations. The name "main" is a bit misleading.
+        self.main_corsika_config = corsika_config[0]
```
GernotMaier, line 51: I suggest to call it `base_corsika_config`. This indicates that it might be overwritten.

GernotMaier (changes requested):
Thanks for this - as you mentioned this is a start.

I don't like the many tests if model_version is a list or not a list (and especially that it affects unrelated code like the `docs_produce_array_element_report`.

Wouldn't it be better to try and do the check for lists in the command line parser? Or is this only relevant for `simulate_prod`, maybe use another command line argument like `model_versions_lists`? 

A few other minor comments below.

src/simtools/simulator.py
orelgueta: Done.

src/simtools/simulator.py
orelgueta: Done

src/simtools/runners/corsika_simtel_runner.py
orelgueta: Done, though not because I think it will actually be overwritten, but rather because "base" is a better name.

orelgueta:
> > I don't like the many tests if model_version is a list or not a list (and especially that it affects unrelated code like the `docs_produce_array_element_report`.
> 
> Because we use `nargs="+"` in the argparser, it is technically always a list. 

Unfortunately that's not true in the case of the tests. When testing, we provide simple strings to the various functions (i.e., we don't go through the argparser). I don't know if it makes sense to change all of the tests cause of it or keep the checks if a list or not. What do you think?

src/simtools/corsika/corsika_config.py
orelgueta: Apparently we do need it, yes. It helped me find an error in the CORSIKA config of Prod5. I therefore added a function that asserts they are the same for parameters which are relevant. I doubt we will ever see a mismatch in the future, but it doesn't seem to hurt to check it (thanks to the cache).

orelgueta:
I managed to reduce the various checks for lists and lengths of lists by adding that check to the configurator. I still kept the protection for model_version not being a list in a few places because they are harmless. 

The issue I have now is that I cannot find a good way to test the added code in the configurator. I might ask for your help with it since you are more familiar with that piece of code.

src/simtools/configuration/configurator.py
```diff
@@ -163,6 +163,7 @@ def initialize(
             self.config["activity_id"] = str(uuid.uuid4())
         if self.config["label"] is None:
             self.config["label"] = self.label
+        self._initialize_model_versions()
```
orelgueta, line 166: This complaint on code coverage we are going to ignore I assume.

GernotMaier (approved):
All good now, thanks!

PR #1500, opened 2025-04-09
Add design parameters to reports

EshitaJoshi:
This PR adds design parameters to all array element reports and highlights which parameters are overwritten and which parameters are inherited from the design model.

Also implemented text wrapping to ensure text stays within the defined column widths and the column doesn't become wider than the page.

Changes to the reports can be viewed [here](http://cta-science.gitlab-pages.cta-observatory.org/-/simulations/simulation-model/model-parameters-documentation/-/jobs/320439/artifacts/public/productions/6.0.0/LSTN-01.html).


GernotMaier (approved):
Thank you @EshitaJoshi - the bold / italics is clear enough and this looks good.

EshitaJoshi:
Thanks for the review @GernotMaier !

PR #1503, opened 2025-04-09
Clipping long data files

EshitaJoshi:
Only include the first 30 lines of a .dat file and add a link to the Simulation Models repository.


GernotMaier (changes requested):
No tests? Even for those small changes, tests would be nice.

EshitaJoshi:
> No tests? Even for those small changes, tests would be nice.

Added a test to check that only the first 30 lines are being written to the md file. Checked line count and content.

GernotMaier (approved):
Thanks - perfect!

PR #1505, opened 2025-04-10
Compare metadata written by sim_telarray with array model parameters.

### LINKED ISSUE #456: Add integration test that input is same as metaparameter output
We should have an integration test for the application that produces a simulation configuration that then runs a small simulation and checks that all of the metaparameters are the same as the input parameters (where available, not all parameters are metaparameters).

GernotMaier:
Add functionality to compare the metadata written by sim_telarray into its data file with values used to write the sim_telarray configuration files. This tests that the configuration has been correctly propagated to sim_telarray and used in the correct way.

In detail, this PR adds:

- a new module called `simtools.simtel.simtel_io_metadata` with two main functions:
  - read metadata from sim_telarray file
  - allow to get the sim_telarray telescope ID for a given CTA telescope naming (e.g., LSTN-02 to '2')
- a new module called `simtools.testing.sim_telarray_metadata` which compares for site and telescope parameter the values of a given model with those read from the metadata file

Closes #456


src/simtools/testing/sim_telarray_metadata.py
```diff
@@ -0,0 +1,164 @@
...
+            f"number of telescopes in array model ({len(array_model.telescope_model)})"
+        )
+
+    for telescope_name in array_model.telescope_model.keys():
+        if not get_sim_telarray_telescope_id(telescope_name, file):
+            raise ValueError(f"Telescope {telescope_name} not found in sim_telarray file metadata")
+
+    telescope_parameter_mismatch = [
+        _assert_model_parameters(telescope_meta[i], model)
+        for i, (_, model) in enumerate(array_model.telescope_model.items(), start=1)
```
orelgueta, line 43: Probably worth replacing with `for i, model in enumerate(array_model.telescope_model.values(), start=1)`

src/simtools/testing/sim_telarray_metadata.py
```diff
@@ -0,0 +1,164 @@
...
+    ]
+
+    # ensure printout of all mismatches, not only those found first
```
orelgueta, line 46: This comment is true for the second `if` statement, but if there's a mismatch in the site parameters, we won't reach the telescope parameters. Either we collect all errors and raise the ValueError afterwards, or perhaps join the mismatch lists so the `any` applies to the site one as well? Otherwise I don't mind if you simply want to move this comment to above the telescope mismatch check, at least there it will be correct.

src/simtools/testing/sim_telarray_metadata.py
```diff
@@ -0,0 +1,164 @@
...
+        sim_telarray_name = _sim_telarray_name_from_parameter_name(param)
+        if sim_telarray_name in metadata.keys():
+            parameter_type = model.parameters[param]["type"]
+            if parameter_type not in ("string", "dict", "boolean"):
+                value, _ = config_reader.extract_value_from_sim_telarray_column(
+                    [metadata[sim_telarray_name]], parameter_type
+                )
+            else:
+                value = metadata[sim_telarray_name]
+                value = (int)(value) if value.isnumeric() else value
```
orelgueta, line 91: Why is it always cast to `int` here? Don't we have some floats in the configuration?

src/simtools/testing/sim_telarray_metadata.py
```diff
@@ -0,0 +1,164 @@
...
+    """Check if floats and ints are equal."""
+    if isinstance(value1, np.ndarray | list) and isinstance(value2, np.ndarray | list):
+        return bool(np.allclose(np.array(value1), np.array(value2), rtol=1e-10))
+    if isinstance(value1, list) and isinstance(value2, float | int | np.integer | np.floating):
+        if all(x == value1[0] for x in value1):
+            return bool(np.isclose(float(value1[0]), float(value2), rtol=1e-10))
+    if isinstance(value2, list) and isinstance(value1, float | int | np.integer | np.floating):
+        if all(x == value2[0] for x in value2):
+            return bool(np.isclose(float(value1), float(value2[0]), rtol=1e-10))
+    return bool(np.isclose(float(value1), float(value2), rtol=1e-10))
```
orelgueta, line 164: We could consider moving these two functions to general utilities (in particular if the rtol is made an argument).

tests/unit_tests/simtel/test_simtel_io_metdata.py
```diff
@@ -0,0 +1,37 @@
...
+@pytest.fixture
+def test_sim_telarray_file():
+    return (
+        "tests/resources/"
+        "run000010_gamma_za20deg_azm000deg_North_test_layout_6.0.0"
+        "_test-production-North.simtel.zst"
+    )
+
+
+def test_read_sim_telarray_metadata(test_sim_telarray_file):
```
orelgueta, line 20: Is it worth also testing that some random values in the file are as expected?

tests/unit_tests/testing/test_sim_telarray_metadata.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+
+@pytest.fixture
+def test_metadata(test_metadata_file):
+    return read_sim_telarray_metadata(test_metadata_file)
+
+
+def test_assert_sim_telarray_metadata_success(test_metadata_file):
+    """Test assert_sim_telarray_metadata with successful comparison."""
```
orelgueta, line 33: This test is not for a success case, right?

orelgueta (commented):
This is very nice and simple. A few minor comments.

src/simtools/simtel/simtel_io_metadata.py
```diff
@@ -0,0 +1,88 @@
...
+        meta = {k.lower().lstrip("*"): v for k, v in meta.items()}
+        return {k: v.strip() if isinstance(v, str) else v for k, v in meta.items()}
+
+    # keys to lower case and strip leading '*', trailing spaces
+    return clean_meta(global_meta), {
+        tel_id: clean_meta(meta) for tel_id, meta in telescope_meta.items()
+    }
+
+
+def get_sim_telarray_telescope_id(telescope_name, file):
```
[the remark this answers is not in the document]
GernotMaier: I have asked around (AI) and the answer is consistent with what we discussed in person.

```
I would recommend removing the @cache decorator from this function since the expensive file I/O operation is already cached by the parent function.
````

On the bidirectional map: I would rather wait until we need it. Don't think we should implement code because it might be useful at some point. Will be quick to update it when needed.

File: src/simtools/testing/sim_telarray_metadata.py (hunk 1 above)
Comment (GernotMaier) [reply src/simtools/testing/sim_telarray_metadata.py:43]: OK, done!

File: src/simtools/testing/sim_telarray_metadata.py (hunk 2 above)
Comment (GernotMaier) [reply src/simtools/testing/sim_telarray_metadata.py:46]: I am aware of that. Your suggestion is good to merge the site / telescope mismatches and then check them. This allows to print all errors.

File: src/simtools/testing/sim_telarray_metadata.py (hunk 4 above)
Comment (GernotMaier) [reply src/simtools/testing/sim_telarray_metadata.py:164]: Yes and no - this is a bit fine tuned to the model parameters (e.g., the conversion from json types to python types), the "none", "None" conversion, the possibility for tuble input) and the style we store the data (numpy arrays). 

Do you have another application in mind where this might be useful?

File: tests/unit_tests/testing/test_sim_telarray_metadata.py (hunk 6 above)
Comment (GernotMaier) [reply tests/unit_tests/testing/test_sim_telarray_metadata.py:33]: Correct. Fixed this.

**Comment by GernotMaier** [comment]
Thanks for the review @orelgueta . I've addressed most of your suggestions - there are two items which might need a bit more discussions. Let me know if this should be part of this pull request.

File: src/simtools/simtel/simtel_io_metadata.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (orelgueta) [reply src/simtools/simtel/simtel_io_metadata.py:65]: Yes, as we discussed in person, I agree with both points.

File: src/simtools/testing/sim_telarray_metadata.py (hunk 3 above)
Comment (orelgueta) [reply src/simtools/testing/sim_telarray_metadata.py:86]: OK, that is indeed a bit confusing (and made me think of negative integers as well), but this code works, so no complaints.

File: src/simtools/testing/sim_telarray_metadata.py (hunk 4 above)
Comment (orelgueta) [reply src/simtools/testing/sim_telarray_metadata.py:159]: No, no application, it just looked like a fairly general function (in particular the second one). I really don't mind it staying here though, so I am resolving this comment.

File: src/simtools/testing/sim_telarray_metadata.py [hunk 8]
```diff
@@ -40,20 +40,15 @@ def assert_sim_telarray_metadata(file, array_model):
...
     # ensure printout of all mismatches, not only those found first
-    if len(site_parameter_mismatch) > 0:
-        raise ValueError(
-            f"Site model parameters do not match sim_telarray metadata: "
-            f"{site_parameter_mismatch} "
-            f"telescope_parameter_mismatch: {telescope_parameter_mismatch}"
-        )
+    telescope_parameter_mismatch.append(site_parameter_mismatch)
     if any(len(m) > 0 for m in telescope_parameter_mismatch):
+        mismatches = [m for m in telescope_parameter_mismatch if len(m) > 0]
```
orelgueta, line 49: Is this needed now? It wasn't needed before as far as I understood (i.e., it was printing the errors for all of them one by one before as well).

orelgueta (approved):
All good now, thanks! I added one question but it's not blocking of course, so I am approving.

Not sure which two points require discussion, it seems we agree on everything already.

src/simtools/testing/sim_telarray_metadata.py
GernotMaier: This line cleans up the output. Otherwise empty list brackets `[]` are printed for each telescope. Simply to make the error message easier to read.

PR #1494, opened 2025-04-07
Spelling of `sim_telarray`

GernotMaier:
Addresses partly the inconsistency in naming of `sim_telarray` in simtools (see issue #1343):

- fixes errors in spelling of `sim_telarray` - we used 67 times `simtel_array`.
- add a function to `simtools.utils.names` to get list of allowed simulation software: `names.simulation_software()`.
  - use this list in `model_parameters` to get parameters relevant for a certain simulation software.

PR #1506, opened 2025-04-14
Change warnings on missing parameters to debug statements

GernotMaier:
Pull #1494 introduced reading the list of simulation softwares from the schema files (to avoid hardwired lists). A now noticed side effect is when reading model parameters are the following warning messages:

```
WARNING::model_parameter(l355)::_load_simulation_software_parameter::No testeff parameters found for North, LSTN-03 (model version 6.0.0).  (Query Unknown simulation software: testeff)
WARNING::model_parameter(l355)::_load_simulation_software_parameter::No simtools parameters found for North, LSTN-04 (model version 6.0.0).  (Query Unknown simulation software: simtools)
```

The software packages `testeff` and `simtools` do not have configuration entries in the model data base (as e.g., `configuration_sim_telarray` and `configuration_corsika`).

Removed the log.WARNING statement, as a ValueError is ok in this case (thought adding a log.DEBUG statement - but there would be a lot printout not helping debugging).

### DIFF
--- a/src/simtools/model/model_parameter.py
+++ b/src/simtools/model/model_parameter.py
@@ -351,12 +351,8 @@ class ModelParameter:
                         simulation_software=simulation_software,
                     )
                 )
-            except ValueError as exc:
-                self._logger.warning(
-                    f"No {simulation_software} parameters found for "
-                    f"{self.site}, {self.name} (model version {self.model_version}). "
-                    f" (Query {exc})"
-                )
+            except ValueError:
+                pass
 
     def _load_parameters_from_db(self):
         """Read parameters from DB and store them in _parameters."""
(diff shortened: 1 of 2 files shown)


src/simtools/model/model_parameter.py
```diff
@@ -351,12 +351,8 @@ def _load_simulation_software_parameter(self):
                         simulation_software=simulation_software,
                     )
                 )
-            except ValueError as exc:
```
orelgueta, line 354: Should we consider printing out the warning just once, when the first failure occurs, instead of completely ignoring it?

It could be done with a class variable which will be used as a flag to print or not the warning.

orelgueta (approved):
Approving, but with a suggestion for an improvement in case you agree it's a good idea.

src/simtools/model/model_parameter.py
GernotMaier: I am against it. It happens each time when this function is called, so this is expected behavior.

orelgueta: OK!

PR #1511, opened 2025-04-16
Add random seeds from file for instrument configuration.

GernotMaier:
Add option to write a list of random seeds and pass them on to `sim_telarray` to be used for generating the same instrument settings (optics, camera, ...) for multiple runs.

Changes:

- change command line option `SIM_TELARRAY_SEEDS` to `SIM_TELARRAY_INSTRUMENT_SEEDS` (for consistent naming)
- add command line option `SIM_TELARRAY_RANDOM_INSTRUMENT_INSTANCES` to set the number of instrument instances 
   - if this parameter is not set; nothing changes compared to previous versions of simtools
   - if this is set: write for each model version into the sim_telarray configuration file a file called `random_seeds.txt` with seeds generated in `simtel_config_writer`

### DIFF
--- a/src/simtools/applications/simulate_prod.py
+++ b/src/simtools/applications/simulate_prod.py
@@ -129,7 +129,8 @@ def _parse(description=None):
         "--sim_telarray_instrument_seeds",
         help=(
             "Random seed used for sim_telarray instrument setup. "
-            "If '--sim_telarray_random_instrument_instances is not set: use as sim_telarray seed. "
+            "If '--sim_telarray_random_instrument_instances is not set: use as sim_telarray seed "
+            " ('random_seed' parameter). "
             "Otherwise: use as base seed for the generation of random instrument instance seeds."
         ),
         type=str,
--- a/src/simtools/applications/simulate_prod.py
+++ b/src/simtools/applications/simulate_prod.py
@@ -122,13 +122,23 @@ def _parse(description=None):
         required=False,
         default=False,
     )
-    config.parser.add_argument(
-        "--sim_telarray_seeds",
+    sim_telarray_seed_group = config.parser.add_argument_group(
+        title="Random seeds for sim_telarray instrument setup",
+    )
+    sim_telarray_seed_group.add_argument(
+        "--sim_telarray_instrument_seeds",
         help=(
-            "Use the provided random seed for sim_telarray instrument setup."
-            "Used only for testing purposes for now."
+            "Random seed used for sim_telarray instrument setup. "
+            "If '--sim_telarray_random_instrument_instances is not set: use as sim_telarray seed. "
+            "Otherwise: use as base seed for the generation of random instrument instance seeds."
         ),
         type=str,
+        required=False,
+    )
+    sim_telarray_seed_group.add_argument(
+        "--sim_telarray_random_instrument_instances",
+        help="Number of random instrument instances initialized in sim_telarray.",
+        type=int,
         required=False,
     )
     return config.initialize(
--- a/src/simtools/runners/corsika_simtel_runner.py
+++ b/src/simtools/runners/corsika_simtel_runner.py
@@ -31,6 +31,8 @@ class CorsikaSimtelRunner:
         Use seeds based on run number and primary particle. If False, use sim_telarray seeds.
     use_multipipe : bool
         Use multipipe to run CORSIKA and sim_telarray.
+    sim_telarray_seeds : dict
+        Dictionary with configuration for sim_telarray random instrument setup.
     """
 
     def __init__(
@@ -201,8 +203,17 @@ class CorsikaSimtelRunner:
             ),
         )
         command += simulator_array.get_config_option("random_state", "none")
-        if self.sim_telarray_seeds:
-            command += simulator_array.get_config_option("random_seed", self.sim_telarray_seeds)
+
+        if self.sim_telarray_seeds.get("random_instances"):
+            config_dir = Path(corsika_config.array_model.config_file_path).parent
+            command += simulator_array.get_config_option(
+                "random_seed",
+                f"file-by-run:{config_dir}/{self.sim_telarray_seeds['seed_file_name']},auto",
+            )
+        elif self.sim_telarray_seeds.get("seed"):
+            command += simulator_array.get_config_option(
+                "random_seed", self.sim_telarray_seeds["seed"]
+            )
         command += simulator_array.get_config_option("show", "all")
         command += simulator_array.get_config_option(
             "output_file",
--- a/src/simtools/runners/corsika_simtel_runner.py
+++ b/src/simtools/runners/corsika_simtel_runner.py
@@ -204,13 +204,13 @@ class CorsikaSimtelRunner:
         )
         command += simulator_array.get_config_option("random_state", "none")
 
-        if self.sim_telarray_seeds.get("random_instances"):
+        if self.sim_telarray_seeds and self.sim_telarray_seeds.get("random_instances"):
             config_dir = Path(corsika_config.array_model.config_file_path).parent
             command += simulator_array.get_config_option(
                 "random_seed",
                 f"file-by-run:{config_dir}/{self.sim_telarray_seeds['seed_file_name']},auto",
             )
-        elif self.sim_telarray_seeds.get("seed"):
+        elif self.sim_telarray_seeds and self.sim_telarray_seeds.get("seed"):
             command += simulator_array.get_config_option(
                 "random_seed", self.sim_telarray_seeds["seed"]
             )
--- a/src/simtools/runners/corsika_simtel_runner.py
+++ b/src/simtools/runners/corsika_simtel_runner.py
@@ -174,6 +174,7 @@ class CorsikaSimtelRunner:
         str:
             Command to run sim_telarray.
         """
+        config_dir = corsika_config.array_model.get_config_directory()
         try:
             weak_pointing = any(pointing in self.label for pointing in ["divergent", "convergent"])
         except TypeError:  # allow for self.label to be None
@@ -183,7 +184,7 @@ class CorsikaSimtelRunner:
 
         command = str(self._simtel_path.joinpath("sim_telarray/bin/sim_telarray"))
         command += f" -c {corsika_config.array_model.config_file_path}"
-        command += f" -I{corsika_config.array_model.get_config_directory()}"
+        command += f" -I{config_dir}"
         command += simulator_array.get_config_option(
             "telescope_theta", corsika_config.zenith_angle, weak_option=weak_pointing
         )
@@ -205,7 +206,6 @@ class CorsikaSimtelRunner:
         command += simulator_array.get_config_option("random_state", "none")
 
         if self.sim_telarray_seeds and self.sim_telarray_seeds.get("random_instances"):
-            config_dir = Path(corsika_config.array_model.config_file_path).parent
             command += simulator_array.get_config_option(
                 "random_seed",
                 f"file-by-run:{config_dir}/{self.sim_telarray_seeds['seed_file_name']},auto",
--- a/src/simtools/simtel/simtel_config_writer.py
+++ b/src/simtools/simtel/simtel_config_writer.py
@@ -221,7 +221,7 @@ class SimtelConfigWriter:
                 file.write(f"# include <{tel_config_file}>\n\n")
             file.write("#endif \n\n")  # configuration files need to end with \n\n
 
-        if sim_telarray_seeds:
+        if sim_telarray_seeds and sim_telarray_seeds.get("random_instances"):
             self._write_random_seeds_file(sim_telarray_seeds, Path(config_file_path).parent)
 
     def _write_random_seeds_file(self, sim_telarray_seeds, config_file_directory):
--- a/src/simtools/simtel/simtel_config_writer.py
+++ b/src/simtools/simtel/simtel_config_writer.py
@@ -2,6 +2,7 @@
 """Configuration file writer for sim_telarray."""
 
 import logging
+import random
 from pathlib import Path
 
 import astropy.units as u
@@ -170,7 +171,9 @@ class SimtelConfigWriter:
 
         return meta_parameters
 
-    def write_array_config_file(self, config_file_path, telescope_model, site_model):
+    def write_array_config_file(
+        self, config_file_path, telescope_model, site_model, sim_telarray_seeds=None
+    ):
         """
         Write the sim_telarray config file for an array of telescopes.
 
@@ -182,6 +185,8 @@ class SimtelConfigWriter:
             Dictionary of TelescopeModel's instances as used by the ArrayModel instance.
         site_model: Site model
             Site model.
+        sim_telarray_seeds: dict
+            Dictionary with configuration for sim_telarray random instrument setup.
         """
         with open(config_file_path, "w", encoding="utf-8") as file:
             self._write_header(file, "ARRAY CONFIGURATION FILE")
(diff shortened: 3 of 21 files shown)


orelgueta:
>   * if this is set: write for each model version into the sim_telarray configuration file a file called `random_seeds.txt` with seeds generated in `simtel_config_writer`

Please name the file with a more descriptive name. That is, use the name of the parameter as a prefix and perhaps the seed used to produce it (but that will include a bit of complexity in the code).

orelgueta:
> > > * if this is set: write for each model version into the sim_telarray configuration file a file called `random_seeds.txt` with seeds generated in `simtel_config_writer`
> > 
> > 
> > Please name the file with a more descriptive name. That is, use the name of the parameter as a prefix and perhaps the seed used to produce it (but that will include a bit of complexity in the code).
> 
> What do you mean by 'parameter'? The parameter is random_seed, or?

I meant the argument name, not "parameter", my bad. So I suggested to call the file something like `sim_telarray_instrument_seeds_12344.txt`. I will see how it is implemented now while reviewing anyway.

src/simtools/applications/simulate_prod.py
```diff
@@ -122,15 +122,25 @@ def _parse(description=None):
...
+    sim_telarray_seed_group = config.parser.add_argument_group(
+        title="Random seeds for sim_telarray instrument setup",
+    )
+    sim_telarray_seed_group.add_argument(
+        "--sim_telarray_instrument_seeds",
         help=(
-            "Use the provided random seed for sim_telarray instrument setup."
-            "Used only for testing purposes for now."
+            "Random seed used for sim_telarray instrument setup. "
+            "If '--sim_telarray_random_instrument_instances is not set: use as sim_telarray seed. "
```
orelgueta, line 132: Since the original sim_telarray seed option was changed, this description is not clear enough. I suggest to at least mention it is provided to as the `random_seed` sim_telarray parameter so one can search the sim_telarray documentation.  Maybe even copying a sentence or two from there would be helpful.

src/simtools/simtel/simtel_config_writer.py
```diff
@@ -216,6 +221,38 @@ def write_array_config_file(self, config_file_path, telescope_model, site_model)
                 file.write(f"# include <{tel_config_file}>\n\n")
             file.write("#endif \n\n")  # configuration files need to end with \n\n
 
+        if sim_telarray_seeds and sim_telarray_seeds.get("random_instances"):
+            self._write_random_seeds_file(sim_telarray_seeds, Path(config_file_path).parent)
```
orelgueta, line 225: `Path(config_file_path).parent` should be defined as a variable since this command repeats twice.

src/simtools/simtel/simtel_config_writer.py
```diff
@@ -216,6 +221,38 @@ def write_array_config_file(self, config_file_path, telescope_model, site_model)
...
+        random_integers = [
+            random.randint(0, 2**32 - 1) for _ in range(sim_telarray_seeds["random_instances"])
+        ]
+        with open(
+            config_file_directory / sim_telarray_seeds["seed_file_name"], "w", encoding="utf-8"
+        ) as file:
+            file.write(
+                "# Random seeds for instrument configuration generated with seed "
+                f"{sim_telarray_seeds['seed']}"
+                f" (model version {self._model_version}, site {self._site})\n"
```
orelgueta, line 251: Should probably add the version of simtools here as well since this code might change in the future. Also, you mentioned the model version and the site, but the pointing is also important for this, is it simply not easily available at this stage?

src/simtools/simulator.py
```diff
@@ -75,6 +75,11 @@ def __init__(
         self._submit_options = self.args_dict.get("submit_options", None)
         self._extra_commands = extra_commands
 
+        self.sim_telarray_seeds = {
+            "seed": self.args_dict.get("sim_telarray_instrument_seeds"),
+            "random_instances": self.args_dict.get("sim_telarray_random_instrument_instances"),
+            "seed_file_name": "random_seeds.txt",  # name only; will be place in config dir
```
orelgueta, line 81: This is the name I suggested to change `"random_seeds.txt"`. In production it won't make a difference, but in local running/testing, it could be useful to have different descriptive names.

tests/integration_tests/config/simulate_prod_gamma_20_deg_South_check_output.yml
```diff
@@ -19,7 +19,7 @@ CTA_SIMPIPE:
       CORE_SCATTER: 10 700 m
       DATA_DIRECTORY: simtools-data
       OUTPUT_PATH: simtools-output
-      SIM_TELARRAY_SEEDS: 1745,290
+      SIM_TELARRAY_INSTRUMENT_SEEDS: 1745,290
```
orelgueta, line 22: The only thing which I am not entirely happy about is that the second seed here is not only for the instrument (unlike the first one), so the name is a bit misleading. However, I think it's not a big deal and with a more detailed description of the argument it should be fine.

tests/unit_tests/simtel/test_simtel_config_writer.py
```diff
@@ -171,3 +187,26 @@ def test_get_sim_telarray_metadata_without_model_parameters(simtel_config_writer
 
     with pytest.raises(ValueError, match=r"^Unknown metadata type"):
         simtel_config_writer._get_sim_telarray_metadata("unknown", None)
+
+
+def test_write_random_seeds_file(simtel_config_writer, tmp_test_directory):
+    config_file_directory = Path(tmp_test_directory) / "config"
```
orelgueta, line 193: This would be a non-standard structure of the testing directory, no? I would expect this file to be written to the model directory and that should already exist.

orelgueta (commented):
Minor comments, one probable mistake.

What's missing is an integration test which actually runs and checks that the random seeds are used. It would probably require two tests, one for each run number and adding the `random_seed` to the metaparameters, but I think it's a necessary test. It can be done in a separate PR though.

GernotMaier:
> > > > * if this is set: write for each model version into the sim_telarray configuration file a file called `random_seeds.txt` with seeds generated in `simtel_config_writer`
> > > 
> > > 
> > > Please name the file with a more descriptive name. That is, use the name of the parameter as a prefix and perhaps the seed used to produce it (but that will include a bit of complexity in the code).
> > 
> > 
> > What do you mean by 'parameter'? The parameter is random_seed, or?
> 
> I meant the argument name, not "parameter", my bad. So I suggested to call the file something like `sim_telarray_instrument_seeds_12344.txt`. I will see how it is implemented now while reviewing anyway.

What does the 12344 mean?

The seed is now in the comment given as first line.

GernotMaier:
> Very nice implementation, thanks!
> 
> Minor comments, one probable mistake.
> 
> What's missing is an integration test which actually runs and checks that the random seeds are used. It would probably require two tests, one for each run number and adding the `random_seed` to the metaparameters, but I think it's a necessary test. It can be done in a separate PR though.

Agree, see #1513

src/simtools/simtel/simulator_array.py
```diff
@@ -82,8 +85,14 @@ def _make_run_command(self, run_number=None, input_file=None):
         command += super().get_config_option("histogram_file", histogram_file)
         command += super().get_config_option("output_file", output_file)
         command += super().get_config_option("random_state", "none")
-        if self.sim_telarray_seeds:
-            command += super().get_config_option("random_seed", self.sim_telarray_seeds)
+        if self.sim_telarray_seeds and self.sim_telarray_seeds.get("random_instrument_instances"):
+            config_dir = Path(self.corsika_config.array_model.config_file_path).parent
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/runners/corsika_simtel_runner.py
```diff
@@ -201,8 +203,17 @@ def _make_run_command(self, run_number, input_file, corsika_config, simulator_ar
             ),
         )
         command += simulator_array.get_config_option("random_state", "none")
-        if self.sim_telarray_seeds:
-            command += simulator_array.get_config_option("random_seed", self.sim_telarray_seeds)
+
+        if self.sim_telarray_seeds and self.sim_telarray_seeds.get("random_instances"):
+            config_dir = Path(corsika_config.array_model.config_file_path).parent
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/applications/simulate_prod.py
GernotMaier: Improved the help message (already very long...)

src/simtools/simtel/simtel_config_writer.py
GernotMaier: Yes, it is not available. And why do we need the pointing? Isn't it enough to keep the seed available to repeat the random number generation if needed?

(and yes, most of the configuration parameters like zenith or sim_telarray version are not available at this point - one of the difficulties of this PR)

src/simtools/simulator.py
GernotMaier: Changed.

tests/integration_tests/config/simulate_prod_gamma_20_deg_South_check_output.yml
GernotMaier: OK. Should I keep it or change it?

tests/unit_tests/simtel/test_simtel_config_writer.py
GernotMaier: Changed it to `model`. You are correct that the directory exists, but in a unit tests it might be better to not depend on directories made in code outside of the tested one. Anyway - for the test not so important.

orelgueta:
> > > > > * if this is set: write for each model version into the sim_telarray configuration file a file called `random_seeds.txt` with seeds generated in `simtel_config_writer`
> > > > 
> > > > 
> > > > Please name the file with a more descriptive name. That is, use the name of the parameter as a prefix and perhaps the seed used to produce it (but that will include a bit of complexity in the code).
> > > 
> > > 
> > > What do you mean by 'parameter'? The parameter is random_seed, or?
> > 
> > 
> > I meant the argument name, not "parameter", my bad. So I suggested to call the file something like `sim_telarray_instrument_seeds_12344.txt`. I will see how it is implemented now while reviewing anyway.
> 
> What does the 12344 mean?
> 
> The seed is now in the comment given as first line.

12344 would have been the seed, but as I said I wrote it before the review and now I see it is not practical since it's too long and is included within the file. The file name I think is still worth it though.

src/simtools/simtel/simtel_config_writer.py
GernotMaier: I've changed (although I don't think it improves the code...think complexity stays the same).

tests/integration_tests/config/simulate_prod_gamma_20_deg_North_check_output.yml
```diff
@@ -19,7 +19,7 @@ CTA_SIMPIPE:
       CORE_SCATTER: 10 500 m
       DATA_DIRECTORY: simtools-data
       OUTPUT_PATH: simtools-output
-      SIM_TELARRAY_SEEDS: 1745,290
+      SIM_TELARRAY_RANDOM_INSTRUMENT_INSTANCES: 100
```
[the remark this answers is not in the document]
orelgueta: In these tests yes because that allows us to check that the output makes sense (trigger_time, pe_sum, photons). With different seeds and just five showers you can easily get a case in which by chance those tests won't pass.

GernotMaier:
@orelgueta - suggest to open an issue for the point where we want feedback from Konrad. If everything else is fine, we should merge. It will be a single line to change in case this is not needed.

src/simtools/simtel/simtel_config_writer.py
orelgueta: Because you added the model version and the site, I assumed you wanted to make it easier to understand the seed. I would easily understand not including them at all (or just the model version).

However, the more important comment here is the version of simtools.

src/simtools/simtel/simulator_array.py
```diff
@@ -67,12 +67,13 @@ def _make_run_command(self, run_number=None, input_file=None):
         str
             Command to run sim_telarray.
         """
+        config_file_path = self.corsika_config.array_model.config_file_path
```
orelgueta, line 70: Why not `self.corsika_config.array_model.get_config_directory()`? Isn't the file exported to the model directory? That is what I would expect (and suggest to do).

src/simtools/simtel/simulator_array.py
```diff
@@ -67,12 +67,13 @@ def _make_run_command(self, run_number=None, input_file=None):
...
         self._log_file = self.get_file_name(file_type="log", run_number=run_number)
         histogram_file = self.get_file_name(file_type="histogram", run_number=run_number)
         output_file = self.get_file_name(file_type="output", run_number=run_number)
 
         command = str(self._simtel_path.joinpath("sim_telarray/bin/sim_telarray"))
-        command += f" -c {self.corsika_config.array_model.config_file_path}"
+        command += f" -c {config_file_path}"
```
orelgueta, line 76: See comment above.

src/simtools/simtel/simulator_array.py
```diff
@@ -86,7 +87,7 @@ def _make_run_command(self, run_number=None, input_file=None):
         command += super().get_config_option("output_file", output_file)
         command += super().get_config_option("random_state", "none")
         if self.sim_telarray_seeds and self.sim_telarray_seeds.get("random_instrument_instances"):
-            config_dir = Path(self.corsika_config.array_model.config_file_path).parent
+            config_dir = Path(config_file_path).parent
```
orelgueta, line 90: See comment above.

orelgueta (commented):
Intermediate 2nd review while I am here.

tests/unit_tests/simtel/test_simtel_config_writer.py
orelgueta: It's created as part of the `tmp_test_directory` fixture, so it's probably safe (or changes would be required in many places anyway). Either way, thanks!

tests/integration_tests/config/simulate_prod_gamma_20_deg_South_check_output.yml
orelgueta: We can keep it

src/simtools/simtel/simtel_config_writer.py
GernotMaier: OK. Do you want to change something? If yes what, not clear to me.

src/simtools/simtel/simulator_array.py
GernotMaier: Fixed.

src/simtools/simtel/simulator_array.py
GernotMaier: This is the directory - so I've changed it.

src/simtools/simtel/simtel_config_writer.py
EshitaJoshi: I also think writing the version of simtools to the file along with the seed and model version/site is a good idea

GernotMaier: But what do we learn from it? Even model and site is not needed and I am inclined to remove it. This is a file with a list of integers generated randomly using the seed indicated in the comment. Maybe a use case would be nice to know where additional information would be relevant?

orelgueta: If you just give me the seed without the version of simtools, how do I know how the list of seeds were produced in the case the code which produces them changes? For example, what if tomorrow we decide to reduce the range to 16 bit because it makes sim_telarray run faster? Adding the version of simtools to the header is more important than all other information.

src/simtools/simtel/simulator_array.py
GernotMaier: Ah - correct and fixed (was used before; but now introduced config_file_dir.

GernotMaier:
@orelgueta - thanks for the discussions...please check the integration test changes to the `simulate_prod_gamma_20_deg_North.yml simulate_prod_gamma_20_deg_North_check_output.yml` files.

src/simtools/simtel/simtel_config_writer.py
GernotMaier: Discussed this in person with @orelgueta - will add the seed used to generate the file to the sim_telarray metaparameters to ensure reproducibility.

orelgueta (approved):
Yes, all good now, thanks!

GernotMaier:
Thanks @orelgueta , this was more work then expected.

@EshitaJoshi , is there anything left from your review?

EshitaJoshi (approved):

PR #1520, opened 2025-04-22
Multiple versions integration test

orelgueta:
This adds an integration test which runs using two model versions. It required also adding that option in the command line, which resulted in a few changes which do not seem directly related. Also, the packing of the output files is taken care of in this PR, together with copying the CORSIKA log file so that it appears in all tarballs with log files for this particular production.

### DIFF
--- a/src/simtools/simulator.py
+++ b/src/simtools/simulator.py
@@ -741,7 +741,10 @@ class Simulator:
         original_version = next(
             model.model_version
             for model in self.array_models
-            if model.model_version in original_log.name
+            if re.search(
+                rf"(?<![0-9A-Za-z]){re.escape(model.model_version)}(?![0-9A-Za-z])",
+                original_log.name,
+            )
         )
 
         for model in self.array_models:
--- a/src/simtools/simulator.py
+++ b/src/simtools/simulator.py
@@ -1,5 +1,6 @@
 """Simulator class for managing simulations of showers and array of telescopes."""
 
+import gzip
 import logging
 import re
 import shutil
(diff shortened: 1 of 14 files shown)


src/simtools/simulator.py
```diff
@@ -683,3 +723,44 @@ def validate_metadata(self):
...
+        corsika_log_files: list
+            List containing the original CORSIKA log file path.
+
+        """
+        original_log = Path(corsika_log_files[0])
+        # Find which model version the original log belongs to
+        original_version = next(
+            model.model_version
+            for model in self.array_models
+            if model.model_version in original_log.name
```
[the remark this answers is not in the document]
orelgueta: Slightly modified version of this was implemented since this does not work for the file names we have.

src/simtools/testing/validate_output.py
```diff
@@ -64,9 +64,20 @@ def _validate_output_files(config, integration_test):
...
+            if version in test_simtel_cfg_files:
+                test_simtel_cfg_file = test_simtel_cfg_files[version]
+                break
+    elif isinstance(from_config_file, list):
+        for version in from_config_file:
+            if version in test_simtel_cfg_files:
+                test_simtel_cfg_file = test_simtel_cfg_files[version]
+                break
+    else:
+        test_simtel_cfg_file = test_simtel_cfg_files.get(from_command_line or from_config_file)
```
orelgueta, line 80: This is really ugly code, but it's more or less the only way to protect against the case when one of the model versions is a list. A simpler option is to simply check first if `test_simtel_cfg_files` is not empty and if not, continue with simpler old code, otherwise return. That is,

```
test_simtel_cfg_files = integration_test.get("TEST_SIMTEL_CFG_FILES", {})
if not test_simtel_cfg_files:
    return

test_simtel_cfg_file = integration_test.get("TEST_SIMTEL_CFG_FILES", {}).get(
        from_command_line or from_config_file
    )
...
```

Currently we do not have a case of multiple versions and `TEST_SIMTEL_CFG_FILES`, so nothing will break, but it will if we ever add a test like that.

GernotMaier:
@orelgueta - do we want to ignore the code coverage or do you want to look again.

orelgueta:
Oh this is quite annoying, code coverage doesn't run when the PR is still in draft. I was looking for those before setting it as ready for review and when I didn't see any, I thought it meant coverage is not an issue. I will fix it and let you know when it's ready for review.

tests/unit_tests/testing/test_helpers.py
```diff
@@ -67,3 +67,73 @@ def test_new_testeff_version_file_not_found(builtins_open):
                 FileNotFoundError, match="The testeff executable could not be found."
             ):
                 helpers._new_testeff_version()
+
+
+def test_skip_multiple_version_test_single_version():
+    config = {"CONFIGURATION": {"MODEL_VERSION": "v1"}}
```
GernotMaier, line 73: I know it doesn't matter...just note that we are using semantic versioning everywhere...

src/simtools/testing/validate_output.py
GernotMaier: Would this be better? Less repetitive.

```
def _test_simtel_cfg_files(config, integration_test, from_command_line, from_config_file):
    """Test simtel cfg files."""
    cfg_files = integration_test.get("TEST_SIMTEL_CFG_FILES", {})
    sources = from_command_line if isinstance(from_command_line, list) else (
              from_config_file if isinstance(from_config_file, list) else [from_command_line or from_config_file])
    
    for version in sources:
        cfg = cfg_files.get(version)
        if cfg:
            _validate_simtel_cfg_files(config, cfg)
            break
```

src/simtools/simulator.py
```diff
@@ -644,26 +674,36 @@ def pack_for_register(self, directory_for_grid_upload=None):
         )
         directory_for_grid_upload.mkdir(parents=True, exist_ok=True)
 
-        tar_file_name = directory_for_grid_upload.joinpath(tar_file_name)
+        # If there are more than one model version,
+        # duplicate the corsika log file to have one for each model version with the "right name".
```
GernotMaier, line 678: Not sure about the log file duplication. This is the best solution? Or can we have the corsika log file separate to all sim_telarray logfiles?

src/simtools/simulator.py
```diff
@@ -644,26 +674,36 @@ def pack_for_register(self, directory_for_grid_upload=None):
...
+        if len(self.array_models) > 1 and corsika_log_files:
```
GernotMaier, line 679: My feeling is that sonarq will definitely complain about the level of complexity of the `pack_for_register` function. Any change to simplify? Making it more modular?

GernotMaier (changes requested):
Thanks for this! I have a couple of suggestions, but in principle approved...

src/simtools/simulator.py
orelgueta: Whether it is the best solution, I cannot really tell you. However, I can definitely say it is the most convenient option. The log files are fairly small and if we can always provide the full set of log files with each data file, I think it's a good idea. Otherwise the user will have to search for the corresponding CORSIKA log file separately, which is much less fun with grid tools.

That being said, we never really saved the CORSIKA log files thus far. So my argument above isn't the strongest.

src/simtools/simulator.py
orelgueta: I actually separated the code in `_copy_corsika_log_file_for_all_versions` in particular because of that. The rest of the changes do not add much complexity, we just loop over the models. I could separate that to a different function as well, but I would prefer to wait for SQ to complain before I do it...

src/simtools/testing/validate_output.py
orelgueta: Yes, this is shorter and less repetitive, so I will change to it. However, the original question was whether the concept of protecting against the lists even warrants the extra complication at this point. 
I take it from your suggestion that you think so and we should keep it.

tests/unit_tests/testing/test_helpers.py
orelgueta: You are right, this is something I have complained about myself as well. I changed it.

src/simtools/testing/validate_output.py
GernotMaier: Yes, keep it.

GernotMaier (approved):
All good, thanks!

PR #1522, opened 2025-04-23
Remove duplication of reading yaml/json files in simtools.utils.general

GernotMaier:
Exactly the same logic was used to read files from disk or url and implemented twice. Merged this by using a function.

Also some minor doc string and error message improvements to general.py

### DIFF
--- a/docs/changes/1522.maintenance.md
+++ b/docs/changes/1522.maintenance.md
@@ -1,0 +1,1 @@
+Remove duplication of reading yaml/json files in simtools.utils.general.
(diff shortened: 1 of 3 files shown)


EshitaJoshi:
No test for the new function?

GernotMaier:
No - we have so many tests for `collect_data_from_file` and `collect_data_from_url` for all different file types, that this is completely covered (but in principle you are correct - unit should be tested individually; but this is a minor merge of existing code).

EshitaJoshi (approved):
Looks good, no new code is added and it's covered indirectly by tests so I'm ok with not adding additional tests.

PR #1524, opened 2025-04-23
improve model parameter description (schemas, versions) in documentation.

GernotMaier:


GernotMaier:
@orelgueta - hope this helps a bit with the schemas....

docs/source/user-guide/model_parameters.md
```diff
@@ -2,49 +2,71 @@
...
-The schema files and especially the model parameter descriptions are derived from (and planned to be synchronized with) the [sim_telarray manual](https://www.mpi-hd.mpg.de/hfm/~bernlohr/sim_telarray/).
+Two types of schema definitions need to be distinguished:
 
-The schema files are in human readable YAML format and follow a fixed meta metaschema (see [model_parameter_and_data_schema.metaschema.yml](https://github.com/gammasim/simtools/blob/main/simtools/schemas/model_parameter_and_data_schema.metaschema.yml); found in simtools schema directory).
+1. **Meta schema:** The meta schema defines the structure of the model parameter file. It is a multi-document YAML file including several schema versions. The meta schema is used to validate the model parameter files. The `schema_version` field is used to identify the correct meta schema.
+2. **Model parameter schema:** The model parameter schema defines the properties of the model parameter. It is a human-readable YAML file that describes the model parameter in detail. The `model_parameter_schema_version` field is used to identify the correct model parameter schema (optional field; default value is `0.1.0`).
```
orelgueta, line 63: Why is it optional? Why did we start from 0.1.0?

orelgueta (approved):
Reading it again definitely helps, but I still don't understand why the model parameter schema is option or why it wasn't specified until now. Either way, this PR is fine of course.

docs/source/user-guide/model_parameters.md
GernotMaier: `model_parameter_schema_version` has been forgotten....nothing really to understand. Until now, all model parameter schema files had a single version only, so not surprising we didn't notice.

All model parameter schema files have a version of `0.1.0`. This is how we usually start. Question is now:

- do we want to change all of them to `1.0.0` (or `2.0.0` for` corsika_starting_grammage`), as we are quite sure about the approach
- do we want to add  `model_parameter_schema_version` to all json dicts uploaded to the model parameter repository / database?

I do think we have other schema files with version numbers of `0.1.0` or similar. Do we want to change them all to `1.0.0`?

orelgueta: Let's keep the `0.1.0` for now, not a big deal, especially since simtools itself is not yet at `1.0.0` (but we are close). I would add `model_parameter_schema_version` to all json dicts, it seems natural to have it there.

PR #1537, opened 2025-04-29
Fix short description of laser events

orelgueta:
Trivial change to fix the short description.

For the record, this parameter does not belong to the ILLN/ILLS (see [issue !18 in the simulation models repo](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/issues/18)), but I think it can wait to be fixed once we implement real calibration devices.


EshitaJoshi (approved):
Thanks for the PR!

PR #1539, opened 2025-04-30
Introduce handling of telescope-dependent CORSIKA starting grammage

### LINKED ISSUE #1328: CORSIKA parameters depending on the primary
The current code and DB structure do not support CORSIKA parameters which depend on the primary, e.g., muons.

For example, the CORSIKA parameter `corsika_starting_grammage` (`FIXCHI`) is different for muons (and the site/telescope in the case of muons). How should we deal with it in the DB/code?

GernotMaier:
Fixing `FIXCHI` (CORSIKA starting grammage) is mostly important for muon simulations. The value is telescope-type dependent.

Related to [gitlab merge request 45](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/merge_requests/45) - please look at the data structure there to understand the change in functionality.

Which telescope to choose?

- in most cases, muon simulations will be single-telescope simulations (meaning: the telescope type used is obviously the one simulated)
- for array-style muon simulations with different telescope types, we choose the starting value highest up in the atmosphere (min value)

Closes #1328

### DIFF
--- a/docs/changes/1539.feature.md
+++ b/docs/changes/1539.feature.md
@@ -1,0 +1,1 @@
+Allow `corsika_starting_grammage` to be telescope-type dependent.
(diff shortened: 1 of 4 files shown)


EshitaJoshi (approved):
Thanks for the PR!

PR #1544, opened 2025-05-02
Docker improvements: linter application; bug fix in extra flags

### LINKED ISSUE #1538: Dockerfile linting not working?
We are using the 'hadolint' tool for the linting of the Dockerfile (see CI-linter.yml).

My feeling is that this is not running, at least I couldn't find it in any of the log files of the Github action.

Running it locally, it gives a couple of warnings:

```
hadolint Dockerfile-prod-opt
Dockerfile-prod-opt:14 DL3033 warning: Specify version with `yum install -y <package>-<version>`.
Dockerfile-prod-opt:28 DL3003 warning: Use WORKDIR to switch to a directory
Dockerfile-prod-opt:28 SC3010 warning: In POSIX sh, [[ ]] is undefined.
Dockerfile-prod-opt:41 SC2154 warning: avx_512_FLAG is referenced but not assigned.
Dockerfile-prod-opt:64 SC2086 info: Double quote to prevent globbing and word splitting.
Dockerfile-prod-opt:79 SC3010 warning: In POSIX sh, [[ ]] is undefined.
Dockerfile-prod-opt:79 DL3003 warning: Use WORKDIR to switch to a directory
Dockerfile-prod-opt:108 DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
Dockerfile-prod-opt:121 DL3013 warning: Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>` or `pip install --requirement <requirements file>`
Dockerfile-prod-opt:121 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
Dockerfile-prod-opt:121 SC1091 info: Not following: File not included in mock.
Dockerfile-prod-opt:121 DL3042 warning: Avoid use of cache directory with pip. Use `pip install --no-cache-dir <package>`
```

(nothing serious - but as always code is easier to maintain without warnings)

GernotMaier:
This maintenance PR:

1. fixes a bug in setting the avx 512 instruction (used an undefined variable `avx_512_FLAG`)
2. apply hadolint linter comments
3. synchronized 'Dockerfile-prod-opt` as much as possible with the docker file in SimPipe (see [here](https://gitlab.cta-observatory.org/cta-computing/dpps/simpipe/simpipe/-/blob/main/Dockerfile?ref_type=heads) and [merge request](https://gitlab.cta-observatory.org/cta-computing/dpps/simpipe/simpipe/-/merge_requests/24))
4. add a docstring on top of the docker files
5. fix CI-linter workflow and run hadolint in a separate step; tested that this actually works (in contrast to superlinter; which seems not to pick up the errors / issues in the docker file) with this [action workflow](https://github.com/gammasim/simtools/actions/runs/14797476811/job/41548021222)

On 3.: need to discuss this how to avoid having almost identical dockerfiles here in simtools and in SimPipe.

Closes #1538

### DIFF
--- a/docker/Dockerfile-dev
+++ b/docker/Dockerfile-dev
@@ -1,9 +1,19 @@
+# Dockerfile for simtools development.
+#
+# Based on the CORSIKA and sim_telarray image.
+# All dependencies for simtools are installed,
+# but not simtools itself. It is expected that
+# simtools is installed in an external directory
+# mounted in the container.
+#
+# hadolint global ignore=DL3007,DL3013,DL3041
+# (ignore warning about using latest - this is a dev image)
 ARG BUILD_OPT="prod6-baseline"
 ARG HADRONIC_MODEL="qgs2"
 ARG SIMTEL_VERSION="240927"
 ARG CORSIKA_VERSION="77550"
 ARG BERNLOHR_VERSION="1.68"
-FROM ghcr.io/gammasim/simtools-corsika-sim-telarray-${SIMTEL_VERSION}-corsika-${CORSIKA_VERSION}-bernlohr-${BERNLOHR_VERSION}-${BUILD_OPT}-${HADRONIC_MODEL}
+FROM ghcr.io/gammasim/simtools-corsika-sim-telarray-${SIMTEL_VERSION}-corsika-${CORSIKA_VERSION}-bernlohr-${BERNLOHR_VERSION}-${BUILD_OPT}-${HADRONIC_MODEL}:latest
 WORKDIR /workdir
 ARG PYTHON_VERSION="3.12"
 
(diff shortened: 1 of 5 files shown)


docker/Dockerfile-dev
```diff
@@ -1,9 +1,19 @@
+# Dockerfile for simtools development.
+#
+# Based on the CORSIKA and sim_telarray image.
+# All dependencies for simtools are installed,
+# but not simtools itself. It is expected that
+# simtools is installed in an external directory
+# mounted in the container.
+#
+# hadolint global ignore=DL3007,DL3013,DL3041
+# (ignore warning about using latest - this is a dev image)
```
orelgueta, line 10: Suggest to expand a bit while we are at it:
```
Ignoring the following since this is a dev container.
# - DL3007, DL3041: ignore warnings about using latest
# - DL3013: ignore warnings about installing non-specific versions with microndf)
```

docker/Dockerfile-dev
```diff
@@ -12,11 +22,12 @@ RUN microdnf update -y && microdnf install -y \
     python${PYTHON_VERSION}-pip python${PYTHON_VERSION}-devel wget && \
     microdnf clean all
 
+# hadolint ignore=SC1091
```
orelgueta, line 25: If you explained the previous ones, explaining this might also make sense, but it's far from critical (because even the warning itself is a bit weird to me).

GernotMaier: see previous comments. Added.

GernotMaier:
@orelgueta - any other comments?

orelgueta (approved):
I didn't really review it before, I just looked at it while it was still in draft. Now I did, all good.

GernotMaier:
You need to explain to my the difference between "review" and "looking at it" at some point. But thanks!

PR #1549, opened 2025-05-05
Add comment on grayskull to documentation

GernotMaier:
Triggered by today's manual changes on the conda gammasimtools-feedstock.

Please let me know if we should add anything else to this section.

### DIFF
--- a/docs/source/developer-guide/maintainer_documentation.md
+++ b/docs/source/developer-guide/maintainer_documentation.md
@@ -33,3 +33,5 @@ The conda feedstock for gammasimtools is [this repository](https://github.com/co
 A new merge request is automatically created for new releases on pypi.
 
 New, updated, or removed command-line tools require manual modifications in the `recipe/meta.yaml` file.
+Changed dependencies require manual modifications in a similar way.
+A template for the required changes can be obtained using the [grayskull](https://pypi.org/project/grayskull/) tool: `grayskull pypi gammasimtools` generates a `meta.yaml` file with the required changes.


orelgueta (approved):
We could also try to enforce any PR that touches pyproject.toml to update also meta.yaml. I got the following suggestion from copilot:
```
name: Check Conda-Forge Feedstock Updates

on:
  pull_request:
    paths:
      - "pyproject.toml"
      - "setup.py"

jobs:
  check-feedstock:
    runs-on: ubuntu-latest
    steps:
      - name: Check Feedstock Update
        run: |
          # Define the feedstock repo (replace with the correct repo name)
          FEEDSTOCK_REPO="conda-forge/<feedstock-repo-name>"

          # Check if the meta.yaml file was updated in the latest PRs
          UPDATED=$(curl -s "https://api.github.com/repos/$FEEDSTOCK_REPO/pulls?state=open" | grep -i "meta.yaml")

          if [ -z "$UPDATED" ]; then
            echo "meta.yaml was not updated. Please ensure it is in sync with the upstream changes."
            exit 1
          else
            echo "meta.yaml was updated in the feedstock. All good!"
          fi
```

However, I am not sure if this works well because the way the conda-forge bot works is a bit confusing. Does it always use the last main to create the fork? Would periodically updating the main in the feedstock be enough?

None of it is important for this PR, so I approve it already.

GernotMaier:
Thanks for this. I am not confident enough with conda-forge to judge if this works.

orelgueta:
OK, so let's just try to remember to propagate changes to pyproject.toml in meta.yaml. If that already works, we can go back and implement this check. Either way I don't think we need to go beyond that to automise the process. I will merge this now anyway.

PR #1557, opened 2025-05-07
Correct data type for 'dsum_threshold' model parameter.

GernotMaier:
dsum_threshold should be an integer.


orelgueta:
Not marked as ready (perhaps cause the GitLab MR should happen first?), but the changes look good, so it can be approved.

orelgueta (approved):
OK, so I approve this in the meantime.

PR #1559, opened 2025-05-09
Reduce number of log-messages (INFO) for uploaded model parameters to DB

GernotMaier:
Uploading the model parameters to the DB results in a huge number of log messages (several messages per parameter and per telescope), which makes debugging hard (especially on the github/gitlab CI, where it means endless scrolling). Additional, some messages are almost identical (e.g., both the `db_model_upload` and the `db_handler` module is reporting `Adding production table for..`, while one is calling the other).

This PR changes the log-level for some of the messages from INFO to DEBUG to reduce the number of messages.

### DIFF
--- a/docs/changes/1559.maintenance.md
+++ b/docs/changes/1559.maintenance.md
@@ -1,0 +1,1 @@
+Reduce number of log-messages (INFO) for uploaded model parameters to DB.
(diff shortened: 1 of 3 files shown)


EshitaJoshi (approved):

orelgueta:
Agree that this is definitely needed, thanks!

PR #1561, opened 2025-05-13
Add the sequential command line argument 

### LINKED ISSUE #1552: Add a command line option for running sequentially
multipipe_corsika has a `--sequential` argument which tells CORSIKA/sim_telarray to run using only one core (to some extent, gzip is still running in parallel, but it's not a problem). That is the way we run on the grid at the moment, so we should have the same option in simtools. Need to implement this command line argument and propagate it to `multipipe_corsika`.

orelgueta:
Add the sequential option for simulations using multipipe. With this option the CORSIKA and sim_telarray instances run in sequential order as far as possible. This mode is useful particularly on the grid, where typically we request a single core per job.

Closes #1552.

### DIFF
--- a/src/simtools/applications/simulate_prod.py
+++ b/src/simtools/applications/simulate_prod.py
@@ -142,6 +142,17 @@ def _parse(description=None):
         type=int,
         required=False,
     )
+    sim_telarray_seed_group.add_argument(
+        "--sequential",
+        help=(
+            "If set to true, run CORSIKA and sim_telarray instances in sequential order "
+            "as far as possible. This mode is useful particularly on the grid, "
+            "where typically we request a single core per job. "
+            "If not set, the CORSIKA and sim_telarray instances are run in parallel."
+        ),
+        type=bool,
+        default=False,
+    )
     return config.initialize(
         db_config=True,
         job_submission=True,
--- a/src/simtools/applications/simulate_prod.py
+++ b/src/simtools/applications/simulate_prod.py
@@ -122,6 +122,17 @@ def _parse(description=None):
         required=False,
         default=False,
     )
+    config.parser.add_argument(
+        "--sequential",
+        help=(
+            "If set to true, run CORSIKA and sim_telarray instances in sequential order "
+            "as far as possible. This mode is useful particularly on the grid, "
+            "where typically we request a single core per job. "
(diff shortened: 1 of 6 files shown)


src/simtools/applications/simulate_prod.py
```diff
@@ -122,6 +122,17 @@ def _parse(description=None):
         required=False,
         default=False,
     )
+    config.parser.add_argument(
```
GernotMaier, line 125: Wondering if the default shouldn't be 'true'? We will run all major productions on the grid with the single core requirement.

src/simtools/applications/simulate_prod.py
```diff
@@ -122,6 +122,17 @@ def _parse(description=None):
...
+        "--sequential",
+        help=(
+            "If set to true, run CORSIKA and sim_telarray instances in sequential order "
```
GernotMaier, line 128: Is this actually correct? One cannot set this parameter to 'set to true', but only set the flag 'sequential'. Suggestion simply "Run CORSIKA and sim_telarray instances in ...".

Further down, I suggest to remove the sentence "This mode is useful particularly on the grid, "
            "where typically we request a single core per job.".  or replace it by a more technical description (without `we` - who is that?). e.g. `Enables single-core mode; otherwise, CORSIKA and sim_telarray run in parallel. `

GernotMaier (changes requested):
Thanks! Minor comment on the help messages, otherwise good.

src/simtools/applications/simulate_prod.py
orelgueta: You are right, I wrote this when the argument was different and also not in a technical enough way. I took your suggested sentence almost as is.

GernotMaier (approved):

src/simtools/applications/simulate_prod.py
orelgueta: I don't know. On the grid we will always run with a configuration file which can always have this kind of setting explicit (which is a small bonus as well). However, when running locally or during testing, it's easier to keep the configuration short and use as many cores as possible to speed things up.

It's not a strong opinion, but I think it makes more sense this way.

src/simtools/applications/simulate_prod.py
GernotMaier: All good - fine with me.

PR #1562, opened 2025-05-14
Improve the run number arguments

### LINKED ISSUE #1323: run_number option missing?
I think we need to add the option for the user to set the `run_number` on top of the `start_run_number` in `simulate_prod.py` or in the command line argument parser. 

When running on the grid, run numbers are created from the number of jobs already run (simple counter). That needs to be added to the `start_run_number` within simtools, where `start_run_number` can be defined in the configuration file. This is similarly done in the `htcondor_script_generator.py`, where in the grid case it is all done in one step, so it should be within e.g., `simulate_prod.py`.

Thoughts? Objections?

orelgueta:
Replace the "run_number_start" argument with "run_number_offset" (clearer) and re-introduce the "run_number" argument to allow specifying a run number (to which the offset is added).

Closes #1323.

### DIFF
--- a/src/simtools/job_execution/htcondor_script_generator.py
+++ b/src/simtools/job_execution/htcondor_script_generator.py
@@ -123,7 +123,8 @@ simtools-simulate-prod \\
     --nshow {args_dict["nshow"]} \\
     --energy_range {energy_range_string} \\
     --core_scatter {core_scatter_string} \\
-    --run_number_start $((process_id + {args_dict["run_number_start"]})) \\
+    --run_number_offset $((process_id)) \\
+    --run_number {args_dict["run_number"]} \\
     --number_of_runs 1 \\
     --submit_engine \"local\" \\
     --data_directory /tmp/simtools-data \\
--- a/src/simtools/job_execution/htcondor_script_generator.py
+++ b/src/simtools/job_execution/htcondor_script_generator.py
@@ -123,8 +123,8 @@ simtools-simulate-prod \\
     --nshow {args_dict["nshow"]} \\
     --energy_range {energy_range_string} \\
     --core_scatter {core_scatter_string} \\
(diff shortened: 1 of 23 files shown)


src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -123,7 +123,8 @@ def _get_submit_script(args_dict):
     --nshow {args_dict["nshow"]} \\
     --energy_range {energy_range_string} \\
     --core_scatter {core_scatter_string} \\
-    --run_number_start $((process_id + {args_dict["run_number_start"]})) \\
+    --run_number_offset $((process_id)) \\
```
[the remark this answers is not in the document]
orelgueta: This was indeed a mistake, but the solution is not correct. The commit [83ca7f2](https://github.com/gammasim/simtools/pull/1562/commits/83ca7f2414d129391825ec43996555b6ac9e9b94) fixes this.

tests/integration_tests/config/simulate_prod_htcondor_generator_gamma_20_deg_North.yml
GernotMaier, line 1: For HT Condor, we agreed to keep the `NUMBER_OF_RUNS` and there is no need for the offset argument.

orelgueta: But `number_of_runs` is still here. The offset is not needed in this case, but I specified it in this file just as a way to test that functionality as well. I specified it in a few other cases as well, but more for fun rather than necessity.

The only thing which I agree is not ideal is that `run_number` is the first run number in HTCondor case and the other runs will be `run_number + i` up to `run_number + number_of_runs`. The name `run_number` here is not entirely clear, but perhaps we can clarify it in the help message of that application?

PR #1568, opened 2025-05-15
Improve file names

### LINKED ISSUE #1133: sim_telarray/CORSIKA output file names are different to current productions
After the refactoring the filenames have a different structure than in the current official productions. Check if going back to the old style does not introduce complications in the code.

orelgueta:
Can't give a better description than what copilot came up with (see below).

Closes #1133.

## Pull Request Overview

This PR standardizes file naming by switching hyphens to underscores, reordering prefixes (e.g., `gamma_run...`), and updating all related generation calls and test configs.

- Refactor `generate_file_name` and its invocations to use underscores only.
- Change subdirectory names (`ray-tracing` → `ray_tracing`, `camera-efficiency` → `camera_efficiency`) in I/O handlers.
- Update 13 integration test YAML configs to match the new naming pattern.


src/simtools/ray_tracing/ray_tracing.py
```diff
@@ -636,8 +636,9 @@ def _generate_file_name(
         self, file_type, suffix, off_axis_angle=None, mirror_number=None, extra_label=None
     ):
         """Generate file name for output files."""
+        file_type_prefix = file_type if file_type == "ray_tracing" else f"ray_tracing_{file_type}"
```
GernotMaier, line 639: Can you help me with this line? I don't understand it.

GernotMaier (approved):
All good, thanks!

src/simtools/ray_tracing/ray_tracing.py
orelgueta: Sometimes we get the file type as `ray_tracing` (and that would be the prefix of the file) and sometimes we just get `photons`. In order to make sure the file prefix always starts with `ray_tracing`, avoiding redundancy when `file_type` is already `ray_tracing`, I added this line. It's not pretty, but it's the shortest way I found how to do it.

If you think it's clearer, I can replace it with 
```
file_type_prefix = f"ray_tracing_{file_type}" if file_type != "ray_tracing" else file_type
```

GernotMaier: OK, thanks - this is clear now.

PR #1570, opened 2025-05-17
Add reading of primary_id from sim_telarray file; add conversion from eventio primary_id in PrimaryParticle

GernotMaier:
Two additions:

- eventio has a slightly different particle ID than CORSIKA (see manual). Add the conversion from event_io to the other usual particle IDs (corsika, PDG) in PrimaryParticle
- add reading of primary_id from sim_telarray file and add it to the run header information returned. Note that to get this value, the first MC shower needs to be read (assuming all particle types in a simulation run are the same).

### DIFF
--- a/docs/changes/1566.feature.md
+++ b/docs/changes/1566.feature.md
@@ -1,0 +1,1 @@
+Add reading and conversion of primary_id from sim_telarray output.
--- a/src/simtools/corsika/primary_particle.py
+++ b/src/simtools/corsika/primary_particle.py
@@ -20,9 +20,15 @@ class PrimaryParticle:
         The actual ID of the primary particle.
     """
 
+    _valid_types = {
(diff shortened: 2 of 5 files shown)


GernotMaier:
@tobiaskleiner - could you check this one please? I am building on this PR and a cascade of open PRs are difficult to handle.

orelgueta (approved):
Since I already looked at it the other day, I can approve it. I will just comment (for a future PR), that we will need to add also iron and oxygen (at least). Far from urgent though.

GernotMaier:
> Since I already looked at it the other day, I can approve it. I will just comment (for a future PR), that we will need to add also iron and oxygen (at least). Far from urgent though.

I suggest to write issues on missing stuff.

GernotMaier:
Thanks for the review @orelgueta

Issue #1577, opened 2025-05-21
Missing features for `production_derive_corsika_limits.py` (partly after update to astropy tables)


GernotMaier:
PR #1573 should be followed by pull requests implementing:

- [x] (PR #1580) reading of several event data files for a single point in the parameter space (e.g., each grid job is writing in the cleanup stage the event data files and we want to read many of them at once) **or** a small application to merge many of those files for the same parameter space values (to get e.g. a single file for gamma, 20 deg, 180 deg az, full moon).
- [x] (PR #1580) plotting of histograms is overwriting existing ones when processing several files
- [x] (PR #1580) writing of corsika limits should include some reasonable rounding (e.g., the core scatter radius to units of 25 or 50 m) 
- [x] (PR #1616 and PR #1619) use of array layouts defined for each production model instead of a list of telescopes

GernotMaier:
PR #1580 adds an application to merge files. However, `production-derive-corsika-limits` requires some optimization to read large table.

PR #1580, opened 2025-05-22
Generalize table reading and writing. Add application to merge tables.

GernotMaier:
Main objective is to allow to use the reduced event data in a 'production scenario' with sufficient statistics and to improve the results of `simtools-production-derive-corsika-limits` values and plotting).

## Generating reduced event data

The production scenario is:

- each job during a simulation production is generating a reduced event data file with three tables ("SHOWERS", "TRIGGERS", "FILE_INFO") in HDF5 format using `simtools-generate-simtel-event-data`.
- these files can be merged into a single file using the new `simtools-merge-tables` application.

 `simtools-merge-tables` is a small application to merge tables from several files into one file - either FITS or HDF5 fromat.

```
# using a file list
simtools-merge-tables  --input_list tt.txt --table_names SHOWERS TRIGGERS FILE_INFO --output_file mmer.hdf5

# listing files on the command line
simtools-merge-tables  --input simtools-output/hdf5/gamma_20deg_0deg_run000001___cta-prod6-2156m-LaPalma-moon+magic_cone10.simtel.zst.hdf5 simtools-output/hdf5/gamma_20deg_0deg_run000009___cta-prod6-2156m-LaPalma-moon+magic_cone10.simtel.zst.hdf5 --table_names SHOWERS TRIGGERS FILE_INFO --output_file mmer.hdf5
```

Note that `simtools-merge-tables` technically does not merge the tables, but copies them into the file with a simple numbering scheme (e.g., 'SHOWERS_1', 'SHOWER_2', ...). This allows later to read the tables easier in chunks, without having to load the full (in merge files potentially very large) table into the memory (astropy.table.Table.read loads always the full table).

Added a new module called `simtools.io_operations.io_table_handler`, which provides some simple functions to read, write, and merge astropy tables.

## Derive CORSIKA limits

Adjusted the modules `simtools.io_operations.io_table_handler`, `simtools.simtel.simtel_io_event_reader`, `simtools.production.derive_corsika_limits_grid` and `simtools.production.derive_corsika_limits` to the new table reading schema.

- full propagation of units from the reduced event data file
- efficient reading with reduced memory footprint (see above): this required to have a single function to fill the histograms (even those plotted).
- round energy, core, and viewcone limits to reasonable values (1 GeV, 25 m, 0.25 deg)
- improved plotting and avoid overwriting of plots for different array layouts.

## Tests

Tested this with 514 diffuse gamma runs. The merging into a compressed 15 GB HD5 file with all the tables is comparable to the time it takes to copy 15 GB.

Reading the file using `simtools-production-derive-corsika-limits` works without loading all values into the memory (the compressed 15 GB would not fit into the memory) and took about 90 min on the working group server. Note that the main inefficiency at this point is that the tables are read for each list of telescopes given. This could be further optimized by filling the histograms for all array layouts at the same time.

Here on of the results for the full North array and a loss fraction of 1.e-3 (which is far too big, we probably will use 1e-7 or even lower).

core_vs_energy_distribution_full

 

Addresses most issue listed in #1577


src/simtools/io_operations/io_table_handler.py
```diff
@@ -0,0 +1,243 @@
...
+    file_type = file_type or read_table_file_type([output_file])
+    if output_file.exists():
+        output_file.unlink()
+    hdus = [fits.PrimaryHDU()]
+    if isinstance(tables, dict):
+        tables = list(tables.values())
+    for table in tables:
+        _table_name = table.meta.get("EXTNAME")
+        _logger.info(f"Writing table {_table_name} of length {len(table)} to {output_file}.")
+        table.info()
```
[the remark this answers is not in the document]
GernotMaier: Removed.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
   "ctao-dpps-cosmic-ray-spectra",
   "ctapipe",
   "eventio",
+  "h5py",
```
orelgueta, line 36: Reminder, add this also in conda feedstock

src/simtools/production_configuration/derive_corsika_limits_grid.py
```diff
@@ -154,3 +146,44 @@ def _create_results_table(results, loss_fraction):
...
+
+def _round_value(key, val):
+    """Round value based on key type."""
+    if key == "lower_energy_limit":
+        return np.floor(val * 1e3) / 1e3
+    if key == "upper_radius_limit":
+        return np.ceil(val / 25) * 25
+    if key == "viewcone_radius":
+        return np.ceil(val / 0.25) * 0.25
+    return val
```
orelgueta, line 176: Don't these depend on the units? I know we control the input, but perhaps a quick check before rounding would be good?

orelgueta (approved):
One comment/question that I think is not blocking. Otherwise all good, thanks!

One small general comment, there's a chance this will have to be expanded a bit when we start using importance sampling.

PR #1582, opened 2025-05-22
Copy CORSIKA log in binary mode

orelgueta:
For some reason sometimes on the grid the copying of the CORSIKA log fails. Copying it directly in binary mode should avoid this error, regardless of the underlying reason. It is very hard to debug the underlying reason as well because I cannot reproduce it locally. Either way, this change should not hurt.

### DIFF
--- a/docs/changes/1582.maintenance.md
+++ b/docs/changes/1582.maintenance.md
@@ -1,0 +1,1 @@
+Copy the CORSIKA log in binary mode instead of text mode.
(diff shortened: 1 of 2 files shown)


EshitaJoshi (approved):
Binary mode is a good idea for faithful copies i think. Though, it's weird that the error cant be reproduced locally. I wonder if the underlying reason is due to any platform differences.

orelgueta:
That's what I am trying to find out. It's even weirder because it happens on some runs and not others. Copying the log file will the job to finish and save the log file so I can look at it.
Thanks for the review!

PR #1586, opened 2025-05-28
Remove obsolete code in JobManager.

GernotMaier:
Remove obsolete code regarding different submit engines and handling of HT Condor / gridengine  submission. This has changed over time:

- all simulations are run using the 'local' submit engine (local, grid, HT Condor)
- gridengine is not supported anymore (as we don't have a test system)
- HT Condor submission is done using the `simtools-simulate-prod-htcondor-generator` script and submission file generator

Removed code related to setting of submit engine and submission.


EshitaJoshi (approved):
Thanks for the PR!

PR #1588, opened 2025-05-28
Write reduced event data file as part of the cleanup stage of the simtools-simulate-prod application.

GernotMaier:
**To be merged after #1580**

Add functionality to write reduced event data tables after a simulation job. Loops over all output files and writes the tables.

Pack files when using the `pack_for_grid_register` option.


src/simtools/simulator.py
```diff
@@ -460,62 +461,53 @@ def _fill_results(self, file, run_number):
             run number
 
         """
-        keys = ["output", "sub_out", "log", "input", "hist", "corsika_log"]
+        keys = ["output", "sub_out", "log", "input", "hist", "corsika_log", "event_data"]
         results = {key: [] for key in keys}
 
+        def get(name, **kwargs):
```
orelgueta, line 467: `get` is too general a name for this function I think. Maybe `get_file_name_str`?

orelgueta (commented):
I am surprised to learn with this PR that we don't actually use the `.simtel.zst` suffix for sim_telarray output files. Even though it is referred to in various places in simtools, it looks like it is not actually used, and instead we use the generic `output` option which just has the `.zst` suffix. This issue is not directly related to this PR, but it is connected because I expected the suffix of the `event_data` files to be `*.event_data.hdf5`. I think we should add that to make it clearer already from the name of the file (and we can discuss the name `event_data` as well). 

If you agree with that, we can implement both changes as part of this PR or we do it in two steps, first merging this and then removing the generic `output` type.

GernotMaier:
@orelgueta - I wondered for a while that the simtel output file has no suffix. 

I will try to find out how it is set, but honestly the file naming setting is unnecessary complex and requires improvements. It just about setting a file naming and it requires to follow through several function calls over different modules.

docs/changes/1588.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix suffix for sim_telarray output files: no correctly `simtel.zstd` (was `.zstd` only)
```
orelgueta, line 1: Should be `.zst`, not `.zstd`

src/simtools/runners/runner_services.py
```diff
@@ -191,7 +191,7 @@ def _get_data_file_path(self, file_type, file_name, run_number):
             "output": ".zst",
             "corsika_output": ".corsika.zst",
             "simtel_output": ".simtel.zst",
-            "event_data": ".hdf5",
+            "event_data": ".event_data.hdf5",
```
orelgueta, line 194: Should this name be discussed? Are we happy with it? If it only contains information about triggered telescopes, why not include trigger in the name? Something like `trigger_data`, `trigger_events`, `trig_evt`, `triggered_telescopes`, `trig_tel`, etc.?

GernotMaier: And I am not happy about it, but since weeks we cannot come up with something better.

GernotMaier: I think we are using somewhere 'reduced' in the code already. Would `reduced_event_data` be too long?

orelgueta: No, let's go for `reduced_event_data`. There is a limit on the grid for the length of file names, but let's hope we are not at that limit that six more characters get us there.

GernotMaier: Changed it (and changing file names is really a bit of a nightmare, as one has to find it in several places in the code...)

orelgueta (approved):
Thanks for the patience =)

PR #1593, opened 2025-05-30
Fix duplication of sim_telarray test files in './tests/resources' and corresponding test fixtures.

### LINKED ISSUE #1567: Duplicated fixtures or files
We have two almost identically named test files, [test_simtel_io_histograms.py](https://github.com/gammasim/simtools/blob/a0c884e9ab0c1a3c8217e14f223de97bf71c641c/tests/unit_tests/simtel/test_simtel_io_histograms.py) and [test_simtel_io_histogram.py](https://github.com/gammasim/simtools/blob/a0c884e9ab0c1a3c8217e14f223de97bf71c641c/tests/unit_tests/simtel/test_simtel_io_histogram.py). 

The content of the two files is quite different, so perhaps both are necessary, I didn't dive too deeply. If they are both necessary, then at least we should remove the duplication of the following fixtures that exist in both by merging the files or moving the fixtures to conftest.

https://github.com/gammasim/simtools/blob/a0c884e9ab0c1a3c8217e14f223de97bf71c641c/tests/unit_tests/simtel/test_simtel_io_histogram.py#L17-L30

GernotMaier:
Maintenance pull request to avoid duplication.

To issues we might consider to solve:

- [ ]  could also get rid of `tests/resources/run000010_gamma_za20deg_azm000deg_North_test_layout_6.0.0_test-production-North.simtel.zst`, if we update the proton sim_telarray files used to test the trigger rates and add metadata fields to them.
- [ ] files names are not consistent with the file names generated and expected by grid production (not important for these tests, but would simply be consistent)

Closes #1567


docs/changes/1592.maintenance.md
```diff
@@ -0,0 +1 @@
+Fix duplication of sim_telarray test files in './tests/resources' and corresponding test fixtures.
```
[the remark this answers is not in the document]
GernotMaier: Brief is good.

GernotMaier:
@orelgueta  - would need your feedback on the two points mentioned above. I assume that the proton simtel files are 'normal' proton simulations?

orelgueta:
Agree that both points are good to address and yes the proton simulations should be normal simulations. Replacing it should definitely be possible and beneficial.

GernotMaier:
I've opened issue #1600 - suggest to review this PR and to the rest in a separate step.

docs/changes/1592.maintenance.md
orelgueta, line 1: I assume this file should be deleted.

tests/unit_tests/testing/test_sim_telarray_metadata.py
```diff
@@ -17,21 +17,11 @@
...
-        "tests/resources/"
-        "run000010_gamma_za20deg_azm000deg_North_test_layout_6.0.0"
-        "_test-production-North.simtel.zst"
-    )
-
-
-@pytest.fixture
-def test_metadata(test_metadata_file):
-    return read_sim_telarray_metadata(test_metadata_file)
+def test_metadata(sim_telarray_file_gamma):
```
orelgueta, line 20: Is this the right file for this test? The rest of the file tests a different module. If anything, it should probably move to the test file above (`test_simtel_io_metdata.py`) and make its name clearer like `test_read_sim_telarray_metadata_type`. Though if you ask me, I would just remove it completely, there's a test already that this function returns two values.

orelgueta (commented):
I don't mind splitting the issue into two PRs, so this can be merged without addressing the file names (once the comments are resolved).

docs/changes/1592.maintenance.md
GernotMaier: Correct. Removed the file.

PR #1594, opened 2025-05-30
Add custom copilot instructions

### LINKED ISSUE #1526: Adding custom instructions for Copilot
All  (?) developers are using Copilot now. Suggest to add a custom instruction files to obtain more consistent and improved responses, see [here](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot) for instructions.

Instructions should be in natural language and should help Copilot to give the correct suggestions. Let's collect in this issue instructions before adding it to a file (`.github/copilot-instructions.md`).

First suggestions:

```
Always use semantic versions for version strings.

Unit test files should be in the ./tests/unit_tests/ directory and the directory structure there should follow that of the python package.
For unit tests, do not use test classes but simple test functions.
```

GernotMaier:
Add a minimal instruction set for copilot.

See [here](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot) and [here](https://code.visualstudio.com/docs/copilot/copilot-customization) for the documentation.

In VS Code, ensure that in the copilot section this is set:

Screenshot 2025-05-30 at 16 09 56

Closes #1526


GernotMaier:
I should say that I have a hard time seeing that this actually works. My impression is that the AI is very moody and one day it works perfectly and follows all instructions, and the next one (like today), it just doesn't care. 

If people thing we don't need it, then I remove it.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,19 @@
...
+- Use f-strings for formatting of log messages and other strings.
+- Avoid hardcoding file paths
+##  Tests
+
+
+Unit test files should be in the ./tests/unit_tests/ directory and the directory structure there should follow that of the python package.
+For unit tests, do not use test classes but simple test functions.
+When writing unit tests, take care of the indentation of the method and the correct formatting.
+
+Check conftest.py for the pytest configuration and common fixtures.
```
orelgueta, line 19: Why not use bullet points like above?

orelgueta (commented):
I think we should live with these instructions for a while and we'll see if they help. If after a few weeks we agree that it didn't make a difference, we can remove it. Though I expect copilot to improve with time and that include taking these instructions into account.

orelgueta (approved):

PR #1598, opened 2025-06-04
Improve printout for assertion error for testing.

GernotMaier:
Improve the error output for integration tests in case the expected output files are not found. Add an explicit printout of the file not found.

(applied additionally a change of 'file' to '_file' in the unit test file, as the linter complaint about mirroring a builtin)


orelgueta (approved):
Changes look fine. If tests pass, can be merged.

PR #1560, opened 2025-05-12
Change documentation theme to pydata and add version switcher. 

GernotMaier:
Switch to pydata theme to be in sync with CTAO sphinx themes (very small change).

Add version switcher, to allow to access the docs from different versions.

This means:

- for a release, the documentation files are moved to e.g. "v0.16.0" in the [gh-pages branch](https://github.com/gammasim/simtools/tree/gh-pages).
- for a merge request (or manual run), the documentation files are moved into the main directory (as we had until now)
- new files in "./docs/_static/switcher.json", which lists the versions to be listed in the menu (see [pydata-thema documentation](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html)). 

No changes to contents of the documentation (future PR).

To see the changes: https://gammasim.github.io/simtools/index.html looks slightly different with a version switcher besides the 'simtools' in the top left. You can change to 0.16 as an example - note that as the old theme does not have this switcher, you have to use 'back' in your browser to come back to the main page.

PR #1606, opened 2025-06-10
Add upcoming release to version list for version switcher. Correct URL to json file.

GernotMaier:
#1560 add the concept of the version switcher to the documentation. 

- forgot to update to to-be-release 0.17.0 (this should be replaced by a CI step)
- fixed url in conf.py to the json file defining the versions
- changed from 'dev' to 'latest' for simplicity.


EshitaJoshi (approved):
Thanks for the PR!

PR #1608, opened 2025-06-10
Removing warning in unit test

EshitaJoshi:
Removing plt.tight_layout() to remove the warning:
```tests/unit_tests/visualization/test_plot_pixels.py::test_plot_pixel_layout_from_file
  /workdir/external/simtools/src/simtools/visualization/plot_pixels.py:249: UserWarning: Tight layout not applied. The left and right margins cannot be made large enough to accommodate all Axes decorations.
    plt.tight_layout()
```
    
The tight layout is not applied anyway so removing it doesn't change how the figures look.


GernotMaier (approved):
I think 1-line reviews are my new favorites.

PR #1617, opened 2025-06-27
Reproducible test data for proton simtel files

GernotMaier:
The two proton simtel files in `./tests/resources` have been generated a very long time ago by Raul and are up-to-date for different reasons (e.g., missing sim_telarray metadata). Very little information was available on how to generate them.

This pull request:

- updates the two proton files with a new CORSIKA/sim_telarray/simtools version
- adds for each file the simtools config file to generate them and the log/histogram tar ball for reproducibility

Needed quite a few changes in the units and integration tests using those file (e.g., obviously the trigger rates changes as we use a different array).


docs/changes/1617.maintence.md
orelgueta, line 1: File with wrong file name

tests/integration_tests/config/generate_simtel_event_data_hdf5.yml
```diff
@@ -4,7 +4,7 @@ CTA_SIMPIPE:
   - APPLICATION: simtools-generate-simtel-event-data
     TEST_NAME: hdf5
     CONFIGURATION:
-      INPUT: 'tests/resources/*proton*.simtel.zst'
+      INPUT: 'tests/resources/*proton*North_alpha_*.simtel.zst'
```
orelgueta, line 7: Is this a good idea to have wildcards here? In the future you might have more files which fit it and run this test all the time on them without meaning to.

tests/resources/proton_run000201_za20deg_azm000deg_North_alpha_6.0.0_test_file/config.yml
```diff
@@ -0,0 +1,26 @@
...
+      MODEL_VERSION: 6.0.0
+      SITE: North
+      ARRAY_LAYOUT_NAME: alpha
+      PRIMARY: proton
+      RUN_NUMBER: 201
+      AZIMUTH_ANGLE: 0
+      ZENITH_ANGLE: 20
+      NSHOW: 50
+      VIEW_CONE: 0 deg 10 deg
+      ENERGY_RANGE: 30 GeV 600000 GeV
```
orelgueta, line 17: How did you choose this energy range? Starting from 30 GeV I could understand to increase the chance for triggering (but it would be a mistake for calculating trigger rate). However, why go up to 600 TeV in such a test file?

tests/resources/proton_run000201_za20deg_azm000deg_North_alpha_6.0.0_test_file/config.yml
```diff
@@ -0,0 +1,26 @@
...
+      CORE_SCATTER: 20 2000 m
```
orelgueta, line 18: 2000 m would be a bit too much, no? We use 1900 m for the hyper array in Prod6.

tests/resources/proton_run000202_za20deg_azm000deg_North_alpha_6.0.0_test_file/config.yml
orelgueta, line 1: Same questions about the values used as above

tests/resources/proton_za20deg_azm000deg_North_alpha_6.0.0_reduced_event_data/config_hdf5.yml
```diff
@@ -0,0 +1,17 @@
+---
+CTA_SIMPIPE:
+  APPLICATIONS:
+  - APPLICATION: simtools-generate-simtel-event-data
+    TEST_NAME: hdf5
+    CONFIGURATION:
+      INPUT: 'tests/resources/*proton*North_alpha_*.simtel.zst'
```
orelgueta, line 7: Same question about the wildcards

tests/resources/simtel_output_files.txt
orelgueta, line 1: This file is a bit weird and from a quick check it looks like it could be easily just replaced with an explicit list in a couple of places. Can be done in a different PR of course.

tests/resources/proton_za20deg_azm000deg_North_alpha_6.0.0_reduced_event_data/config_fits.log
orelgueta, line 1: If you keep this log file, I would save also the sim_telarray and CORSIKA log files. Also, I would gzip all the log files together so they take less space in the repo (assuming no incremental changes are expected for them).

docs/changes/1617.maintence.md
GernotMaier: Fixed.

tests/integration_tests/config/generate_simtel_event_data_hdf5.yml
GernotMaier: Fixed.

tests/resources/proton_run000201_za20deg_azm000deg_North_alpha_6.0.0_test_file/config.yml
GernotMaier: 30 GeV are need to get some triggered events. I didn't change the upper end from the template I've used from the integration tests. Is this an issue?

tests/resources/proton_run000201_za20deg_azm000deg_North_alpha_6.0.0_test_file/config.yml
GernotMaier: Again - if we get triggers, does it matter? Also we want to use for the limit derivation a very wide definition. (but I agree with you that 2000 m are a lot)

GernotMaier:
@orelgueta  - thanks a lot for the review (although I had to remind you!).

Open questions are the ranges and if it matters. Please let me know.

tests/resources/proton_run000201_za20deg_azm000deg_North_alpha_6.0.0_test_file/config.yml
orelgueta: Just to avoid waste and having this example there, I would change it, but won't complain if you decide to keep.

orelgueta (approved):
Approving while still thinking the high energy limit and size of scatter radius should be reduced.

tests/resources/simtel_output_files.txt
GernotMaier: see issue #1621

PR #1623, opened 2025-06-30
Set correct site naming (not `NORTH` but `North`)

GernotMaier:
These files seems to not be used in integration or unit tests, but serve more as examples.


GernotMaier:
Thanks @EshitaJoshi

PR #1625, opened 2025-07-01
Ruff update requires to fix RUF005 issue

GernotMaier:
Ruff update in pre-commit requires to address [RUF005](https://docs.astral.sh/ruff/rules/collection-literal-concatenation/).


GernotMaier:
Thanks @EshitaJoshi

PR #1629, opened 2025-07-03
Replace file lists `tests/resources/simtel_output_files.txt` used for testing by explicitly listing of files.

### LINKED ISSUE #1621: Replace text file with list of files
I think it would be cleaner and more transparent to replace [tests/resources/simtel_output_files.txt](https://github.com/gammasim/simtools/blob/fc5b7d51385b5489317fbd3e740b0edbc288f3a6/tests/resources/simtel_output_files.txt) with an explicit list of files.

GernotMaier:
Closes #1621 — I'm still not convinced, but it's based on a customer complaint.

### DIFF
--- a/docs/changes/1629.maintenance.md
+++ b/docs/changes/1629.maintenance.md
@@ -1,0 +1,1 @@
+Replace file lists `tests/resources/simtel_output_files.txt` used for testing by explicitly listing of files.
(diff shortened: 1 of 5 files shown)


orelgueta:
Very nice, thanks!

Issue #1630, opened 2025-07-03
Standardize key naming conventions in schema files for consistency and reusability


GernotMaier:
The schema files in `src/simtools/schema` use inconsistent key naming conventions. Some keys use capital letters while others use lower case, and there is a mix of snake_case with other styles. This inconsistency makes it difficult to combine schemas and avoid code duplication. It would be beneficial to standardize the key naming conventions across all schema files to improve maintainability and reusability. This is especially important given that jsonschema does not have an option to ignore cases.

This is a bit of work, my suggestion is to move to lower case snake_case everywhere, for the simple reason that everything connected to model parameters is already written this way (although I think capital keys would increase readability)

PR #1635, opened 2025-07-04
Default plotting for tables; lower cases for keys in plotting configuration schema.

GernotMaier:
This is a follow-up PR to #1628:

- change schema for plotting configurations to lower case (as discussed in #1630)
- move examples from tests/resources/plot_table_data to the model parameters schema as examples.

I did not check that all model parameters to be plotted have now a default plotting configuration. Please check and open a separate issue on that (I think we have not covered yet incident-angle reflectivity as it is the case for the SSTs).

For testing of a telescope parameter, do e.g.:

```
python src/simtools/applications/plot_tabular_data_for_model_parameter.py --parameter mirror_reflectivity --parameter_version 1.0.0 --site South --telescope SSTS-design --plot_type all --output_path simtools-output --use_plain_output_path
```

For testing of a site parameter, do e.g.:

```
python src/simtools/applications/plot_tabular_data_for_model_parameter.py --parameter atmospheric_transmission --parameter_version 1.0.0 --site North --plot_type all --output_path simtools-output --use_plain_output_path
```


EshitaJoshi (approved):
Thanks for the PR!

Issue #1634, opened 2025-07-04
Derive CORSIKA limits: file list defines implicit grid.


GernotMaier:
The input file list of reduced event data defines implicit the grid of observational parameters. 

- not sure if we want this behaviour (probably fine when working with merged HDF5 files only)
- not transparent to the user (nowhere documented)

As deriving the CORSIKA limits is quite slow for a reasonable number of events, my suggestion is to change the functionality such that running `simtools-production-derive-corsika-limits` derives the limits for a single grid point only for each run (still for different array layouts). We then can have small follow up script merging the derived limits.

PR #1641, opened 2025-07-08
Derive corsika limit application to be used with single input file

GernotMaier:
`simtools-production-derive-corsika-limits` should be applied to a single input file only, as it calculate the limits for a single grid point, see issue #1634.

Remove her the possibility to give a list as files as input.


GernotMaier:
@tobiaskleiner - this one is now ready for review (took a bit longer, as git was confused by the renaming - seems like a renaming of A -> B and C -> B is confusing for git)

tobiaskleiner (approved):
Thanks @GernotMaier, vimdiff gave almost no difference - all good.

PR #1643, opened 2025-07-08
Add warning in simtools-derive-corsika-limits in case we are at the limit of the simulations

### LINKED ISSUE #1633: Derive corsika limit should issue warning if limits are on the edge of the available parameter space.
Example:

- `simtools-production-derive-corsika-limits` derives a max viewcone radius of 10 deg
- the underlying simulations use a view cone radius of 10 deg.

This should result as a warning, as there is no way the application can reliably derive a good limit in this case (same for core, energy).

GernotMaier:
...just hope that someone is looking at warnings....

Closes #1633


src/simtools/production_configuration/derive_corsika_limits.py
```diff
@@ -217,12 +217,23 @@ def compute_lower_energy_limit(self, loss_fraction):
...
         )
+        return self._is_close(
+            energy_min,
+            self.file_info["energy_min"].to("TeV") if "energy_min" in self.file_info else None,
+            "Lower energy limit is equal to the minimum energy of ",
+        )
+
+    def _is_close(self, value, reference, warning_text):
+        """Check if the value is close to the reference value and log a warning if so."""
+        if reference is not None and np.isclose(value.value, reference.value):
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/production_configuration/derive_corsika_limits.py
```diff
@@ -217,12 +217,23 @@ def compute_lower_energy_limit(self, loss_fraction):
...
         )
+        return self._is_close(
+            energy_min,
+            self.file_info["energy_min"].to("TeV") if "energy_min" in self.file_info else None,
+            "Lower energy limit is equal to the minimum energy of",
+        )
+
+    def _is_close(self, value, reference, warning_text):
+        """Check if the value is close to the reference value and log a warning if so."""
+        if reference is not None and np.isclose(value.value, reference.value):
```
tobiaskleiner, line 234: Do you want to specify a tolerance here or we go with the default of 1e-05?

src/simtools/production_configuration/derive_corsika_limits.py
```diff
@@ -256,6 +267,13 @@ def compute_upper_radius_limit(self, loss_fraction):
             )
             * u.m
         )
+        return self._is_close(
+            radius_limit,
+            self.file_info["core_scatter_max"].to("m")
+            if "core_scatter_max" in self.file_info
+            else None,
+            "Upper radius limit is equal to the maximum core scatter distance of",
```
tobiaskleiner, line 275: Equal or close? We would log this also if it close but not yet equal (depending on the binning).

tobiaskleiner (commented):
Thanks for the PR @GernotMaier. I think this is good to have, have added some comments.

src/simtools/production_configuration/derive_corsika_limits.py
GernotMaier: I think for all application here (view cone, energy range, scatter area) this value is good - we essentially have to catch only the case when we run against the limit.

src/simtools/production_configuration/derive_corsika_limits.py
GernotMaier: Technically you are right. But practically, I would consider a core scatter radius of e.g. 10000 m and 10001 m as equal. To you want to me to change it or can we tolerate this slight inconsistency?

PR #1646, opened 2025-07-09
Write incrementally to disk in simtel io event writer

tobiaskleiner:
The current implementation is extremely memory inefficient: A short test with 100 production files resulted already in a memory usage of >100 GB, due to all tables being kept in memory. Especially if we have order of 10e5 grid points this will be a problem. This PR adds a sequential write option to clear memory after each file is read and processed.


GernotMaier:
@tobiaskleiner - Is this for the merge or the writing stage? The use case is to write a single file always, but merge them later using `simtools-merge-tables`

PR #1650, opened 2025-07-13
Fix schema version string

GernotMaier:
(not sure why this was not picked in the CI of #1645)

### DIFF
--- a/src/simtools/schemas/model_parameters/adjust_gain.schema.yml
+++ b/src/simtools/schemas/model_parameters/adjust_gain.schema.yml
@@ -1,7 +1,7 @@
 %YAML 1.2
 ---
 title: Schema for adjust_gain model parameter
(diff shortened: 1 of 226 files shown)


tests/unit_tests/utils/test_general.py
```diff
@@ -413,7 +415,7 @@ def test_collect_data_dict_from_json():
 
 
 def test_collect_data_from_http():
-    _file = "tests/resources/num_gains.schema.yml"
+    _file = "src/simtools/schemas/model_parameters/num_gains.schema.yml"
```
EshitaJoshi, line 418: This should also be ```MODEL_PARAMETER_SCHEMA_PATH``` for consistency, right?

EshitaJoshi (approved):
Thanks for the PR!

PR #1686, opened 2025-08-18
Ensure that run numbers start at 1 for cases no offset if given (HT Condor job submission).

GernotMaier:
Three bug fixes related to the HT Condor submission scripts: 

- ensure that run numbers start at 1 when no `run_number_offset` is given (as then the run number is the condor process_id
- fix setting or array layout name to non-list. 
- added missing view_cone parameter


src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -101,9 +100,20 @@ def _get_submit_script(args_dict):
     )
     core_scatter = args_dict["core_scatter"]
     core_scatter_string = f'"{core_scatter[0]} {core_scatter[1].to(u.m).value} m"'
+    view_cone = args_dict["view_cone"]
+    view_cone_string = f'"{view_cone[0].to(u.deg).value} deg {view_cone[1].to(u.deg).value} deg"'
```
tobiaskleiner, line 104: suggest to use builtin formatting 
view_cone_string = f'"{view_cone[0].to(u.deg):.2f} {view_cone[1].to(u.deg):.2f}"'

tobiaskleiner (commented):
@GernotMaier thanks for the PR. One minor comment.

src/simtools/job_execution/htcondor_script_generator.py
GernotMaier: Good point - implemented.

GernotMaier:
@tobiaskleiner - thanks for the review. I've applied the request changes. Let me know if this is ok.

PR #1687, opened 2025-08-19
Add camera incident angle application

tobiaskleiner:
This PR adds an application to run ray tracing simulations to calculate the incidence angle of photons arriving at the camera plane and at the primary and secondary (if available) mirrors.

Outputs the photon files, incident angle files and histograms showing the distribution of incident angles.

Requires to compile with -DDEBUG_TRACE_99

Coverage reported is at ~87% due to sonar including also application code currently.

### DIFF
--- a/src/simtools/applications/calculate_incident_angles.py
+++ b/src/simtools/applications/calculate_incident_angles.py
@@ -169,10 +169,13 @@ def main():
         debug_plots=args_dict.get("debug_plots", False),
     )
     total = sum(len(t) for t in results_by_offset.values())
-    logger.info(
+    summary_msg = (
         f"Calculated incident angles for {len(results_by_offset)} offsets,\n"
         f"total photon statistics {total}"
     )
+    if total < 1_000_000:
+        summary_msg += " (below 1e6; results may be statistically unstable)"
+    logger.info(summary_msg)
 
 
 if __name__ == "__main__":


tobiaskleiner:
Looks like it could be between primary and secondary mirror. But there is also a small drop at ~8.5 deg for 0 offset visible in the primiary mirror distribution.

Also do you know why there is such a large spike at about 11.5 deg?

incident_angles_primary_multi_calculate_incident_angles


incident_angles_secondary_multi_calculate_incident_angles

GernotMaier:
The tool should calculate incident angles on all surfaces (mirror planes / focal plane). 

I think you need to dig a bit to understand the 11.5 deg - not sure what it is.

tobiaskleiner:
And this plot here which indicates that the gaps mirrored onto m2 are indeed responsible for the drop in the incident angle plot at an angle of about 27 deg. They are linked to the gaps that we see in the xy plot at a radial distance of about 0.38m:

incident_radius_secondary_multi_calculate_incident_angles_SSTS-04

tobiaskleiner:
And for completeness a 5 deg off axis 2d plot of photon statistics at the two mirrors:

image

image

GernotMaier:
Sphinx gives me the following warnings:

```
/workdir/external/simtools/src/simtools/applications/calculate_incident_angles.py:docstring of simtools.applications.calculate_incident_angles:70: WARNING: Duplicate explicit target name: "plot_calculate_incident_angles_plot". [docutils]
/workdir/external/simtools/src/simtools/applications/calculate_incident_angles.py:docstring of simtools.applications.calculate_incident_angles:77: WARNING: Duplicate explicit target name: "plot_calculate_incident_angles_plot". [docutils]
/workdir/external/simtools/src/simtools/applications/calculate_incident_angles.py:docstring of simtools.applications.calculate_incident_angles:83: WARNING: Duplicate explicit target name: "plot_calculate_incident_angles_plot". [docutils]```

File: src/simtools/ray_tracing/incident_angles.py [hunk 1]
```diff
@@ -0,0 +1,592 @@
...
+        simtel_path,
+        db_config,
+        config_data,
+        output_dir,
+        label=None,
+        perfect_mirror=False,
+        overwrite_rdna=False,
+        mirror_reflection_random_angle=None,
+        algn=None,
+        test=False,
```
GernotMaier, line 61: Suggest to remove test, as it is confusing.

e.g. we define an integration test with 1000 photons. But setting test true, means that a different number of photons are generated. As we have nowadays configuration files for tests, there is no need for this parameter.

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,592 @@
...
+            self.label,
+            db_config,
+            config_data["site"],
+            config_data["telescope"],
+            config_data["model_version"],
+        )
+
+        self.rt_params = self._setup_rt_params(config_data)
+
+    def __repr__(self):
```
GernotMaier, line 101: Nice. But do actually use it?

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,592 @@
...
+        self.logger.info("Running sim_telarray PSF-style simulation for incident angles")
+
+        self.telescope_model.write_sim_telarray_config_file(additional_model=self.site_model)
+
+        photons_file, stars_file, log_file = self._prepare_psf_io_files()
+        run_script = self._write_run_script(photons_file, stars_file, log_file)
+        self._run_script(run_script, log_file)
+
+        data = self._compute_incidence_angles_from_imaging_list(photons_file)
+        self.results = QTable()
```
GernotMaier, line 134: Simplify with e.g.

```
self.results = QTable()
self.results["angle_incidence_focal"] = data["angle_incidence_focal_deg"] * u.deg

if self.calculate_primary_secondary_angles:
    field_map = {
        "angle_incidence_primary_deg": ("angle_incidence_primary", u.deg),
        "angle_incidence_secondary_deg": ("angle_incidence_secondary", u.deg),
        "primary_hit_radius_m": ("primary_hit_radius", u.m),
        "secondary_hit_radius_m": ("secondary_hit_radius", u.m),
        "primary_hit_x_m": ("primary_hit_x", u.m),
        "primary_hit_y_m": ("primary_hit_y", u.m),
        "secondary_hit_x_m": ("secondary_hit_x", u.m),
        "secondary_hit_y_m": ("secondary_hit_y", u.m),
    }
    for key, (name, unit) in field_map.items():
        if key in data:
            self.results[name] = data[key] * unit
```

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,592 @@
...
+        suffix = self._label_suffix()
+        photons_file = self.photons_dir / f"incident_angles_photons_{suffix}.lis"
+        stars_file = self.photons_dir / f"incident_angles_stars_{suffix}.lis"
+        log_file = self.logs_dir / f"incident_angles_{suffix}.log"
+
+        if photons_file.exists():
+            try:
+                photons_file.unlink()
+            except OSError as err:
+                self.logger.warning(f"Failed to remove existing photons file {photons_file}: {err}")
```
GernotMaier, line 188: Why is this a warning only? Wouldn't that be an indication of going something more wrong when we can't delete the file? My suspicion is that this should be an error.

src/simtools/ray_tracing/incident_angles.py
GernotMaier, line 1: Suggest to look at the log message, there seems to be three times the same information (running, running, executing):

```
INFO::incident_angles(l170)::run_for_offsets::Running for off-axis angle 2 deg with label calculate_incident_angles
INFO::incident_angles(l125)::run::Running sim_telarray PSF-style simulation for incident angles
INFO::incident_angles(l275)::_run_script::Executing simtools-output/calculate_incident_angles/scripts/run_incident_angles_calculate_incident_angles_MSTN-04_off2.sh (logging to simtools-output/calculate_incident_angles/logs/incident_angles_calculate_incident_angles_MSTN-04_off2.log)
```

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,592 @@
...
+            result["angle_incidence_secondary_deg"] = secondary
+            result["primary_hit_radius_m"] = radius_m
+            result["secondary_hit_radius_m"] = secondary_radius_m
+            result["primary_hit_x_m"] = primary_hit_x_m
+            result["primary_hit_y_m"] = primary_hit_y_m
+            result["secondary_hit_x_m"] = secondary_hit_x_m
+            result["secondary_hit_y_m"] = secondary_hit_y_m
+        return result
+
+    def _discover_column_indices(self, photons_file):
```
GernotMaier, line 350: This is a bit hard to understand. What does 'discovered' mean? Found?

Ideal would be if there would be unit test called `test_discover_column_indices....` with some clear examples.

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,592 @@
...
+            indices.get("primary"),
+            indices.get("secondary"),
+            indices.get("prim_x"),
+            indices.get("prim_y"),
+            indices.get("sec_x"),
+            indices.get("sec_y"),
+        )
+
+    def _default_column_indices(self):
+        """Return default 0-based indices matching legacy positions.
```
GernotMaier, line 385: What does 'legacy' mean in this context?

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,592 @@
...
+    def _compute_incidence_angles_from_imaging_list(self, photons_file):
+        """
+        Compute incidence angles from imaging list.
+
+        Column positions may differ between telescope types and sim_telarray builds.
+        Parse header lines ("#   Column N: ...") to discover indices; otherwise
+        fall back to legacy positions (1-based): focal=26, primary=32, secondary=36,
+        primary X/Y = 29/30, secondary X/Y = 33/34.
+        """
+        (
```
GernotMaier, line 290: Simplify the code and let col_idx be returned by _discover_column_indices

src/simtools/applications/calculate_incident_angles.py
```diff
@@ -0,0 +1,274 @@
...
+- off_axis_angles (float, optional)
+    One or more off-axis angles in degrees (space-separated). Default: [0.0].
+- source_distance (float, optional)
+    Source distance in kilometers. Default: 10.0.
+- number_of_rays (int, optional)
+    Number of star photons to trace per run. Default: 10000.
+- camera_shift (float, optional)
+    Camera shift along optical axis. Reserved for future use. Default: 0.0.
+- overwrite_rdna (flag, optional)
+    Overwrite mirror_reflection_random_angle with 0 deg when not provided.
```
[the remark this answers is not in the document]
tobiaskleiner: Ok good suggestion. Flag has been removed.

src/simtools/applications/calculate_incident_angles.py
```diff
@@ -0,0 +1,274 @@
...
+- mirror_reflection_random_angle (float, optional)
+    Explicit mirror_reflection_random_angle in degrees; overrides overwrite_rdna.
+- algn (float, optional)
```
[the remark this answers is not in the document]
tobiaskleiner: fixed

src/simtools/applications/calculate_incident_angles.py
```diff
@@ -0,0 +1,274 @@
...
+    Accuracy parameter for mirror alignment distributions. Default: 0.005.
+- use_prod4 (flag, optional)
+    Use prod-4 (SST: prod-5) config files instead of prod-6.
+- flip_mirror_layout (flag, optional)
+    Flip mirror layout x/y (if supported by the configuration).
+- no_optimisation (flag, optional)
```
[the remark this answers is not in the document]
tobiaskleiner: removed

src/simtools/applications/calculate_incident_angles.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+def _parse(label):
+    """Parse command line configuration."""
+    config = configurator.Configurator(
+        label=label,
+        description="Calculate incident angles using sim_telarray ray tracing.",
+    )
+
+    config.parser.add_argument(
```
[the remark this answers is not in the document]
tobiaskleiner: So frist we need to give a zenith angle for the location of the light source. 

Only thing that is (currently) different would be atmospheric transmission (and this should have no difference on the incident angles, except in terms of total photon statistics). I would let it here since we might be interested to compare statistics and we could have some other zenith dependent parameters in the future that could lead to differences in the distribution.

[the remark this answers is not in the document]
tobiaskleiner: Was thinking in terms of zenith angle dependent model parameters, that could affect the distribution. But ok can be removed for the moment.

src/simtools/ray_tracing/incident_angles.py
tobiaskleiner: Renamed and renamed the unit test that tests this method.

src/simtools/ray_tracing/incident_angles.py
tobiaskleiner: These photon files are a bit messy as the column indices are changing between telescopes. This function here is to declare the initial values as is the case for the SST. Then we assign them new later if the column headers found are at a different index.

src/simtools/applications/calculate_incident_angles.py
```diff
@@ -14,39 +13,26 @@
...
-    Camera shift along optical axis. Reserved for future use. Default: 0.0.
-- overwrite_rdna (flag, optional)
-    Overwrite mirror_reflection_random_angle with 0 deg when not provided.
+- number_of_photons (int, optional)
+    Number of photons of the light source to trace per run. Default: 10000.
 - mirror_reflection_random_angle (float, optional)
-    Explicit mirror_reflection_random_angle in degrees; overrides overwrite_rdna.
-- algn (float, optional)
+    Explicit mirror_reflection_random_angle in degrees.
+- mirror_alignment_random (float, optional)
```
GernotMaier, line 34: Not sure we discussed this (Github doesn't show my old comments). 

Is the application using by default the mirror random refleciton angle and alignment parameters from the database?

src/simtools/applications/calculate_incident_angles.py
```diff
@@ -0,0 +1,194 @@
...
+    results_by_offset = calculator.run_for_offsets(offsets)
+    plot_incident_angles(
+        results_by_offset,
+        output_dir,
+        label_with_telescope,
+        debug_plots=args_dict.get("debug_plots", False),
+    )
+    total = sum(len(t) for t in results_by_offset.values())
+    logger.info(
+        f"Calculated incident angles for {len(results_by_offset)} offsets,\n"
```
GernotMaier, line 188: As discussed: photon statistics is not the important quantity. If you want to give a quick feedback, then better to print the mean of the distribution - the average incidence angle is important feedback.

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
+"""Calculate photon incident angles on focal plane and primary/secondary mirrors.
+
+Parses the imaging list (``.lis``) produced by sim_telarray_debug_trace and uses
+Angle of incidence at focal surface, with respect to the optical axis [deg],
+Angle of incidence on(to) primary mirror [deg], and
```
GernotMaier, line 5: Why `on(to)` and not `on to`?

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+    -----
+    Additional options are read from ``config_data`` when present:
+    - ``perfect_mirror`` (bool, default False)
+    - ``mirror_reflection_random_angle`` (float, optional)
+    - ``mirror_alignment_random`` (float, optional)
+    - ``calculate_primary_secondary_angles`` (bool, default True)
+    """
+
+    # Use fixed zenith angle (degrees) for incident-angle simulations.
+    ZENITH_ANGLE_DEG = 45.0
```
GernotMaier, line 47: Why did you choose 45 deg?

It is neither the simplest case (0 deg) or the average case (60-70 deg).

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        cfg = config_data
+        self.perfect_mirror = cfg.get("perfect_mirror", False)
+        self.mirror_reflection_random_angle = cfg.get("mirror_reflection_random_angle")
+        self.mirror_alignment_random = cfg.get("mirror_alignment_random")
+        self.calculate_primary_secondary_angles = cfg.get(
+            "calculate_primary_secondary_angles", True
+        )
+        self.results = None
+
+        self.output_dir.mkdir(parents=True, exist_ok=True)
```
GernotMaier, line 72: Just for info: this will all go after implement #1745

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+            )
+            opts.append(
+                cfg("mirror_align_random_vertical", f"{self.mirror_alignment_random},28.,0.0,0.0")
+            )
+
+        opts += [
+            cfg("IMAGING_LIST", str(photons_file)),
+            cfg("stars", str(stars_file)),
+            cfg("altitude", self.site_model.get_parameter_value("corsika_observation_level")),
+            cfg("telescope_theta", theta + off),
```
GernotMaier, line 259: Do I understand it correctly is e.g. zenith is 45 deg, we simulate [45 deg + offset]? Is this is an issue for telescopes with non-assymmetric response due to shadowing?

Add this point we don't have this in the raytracine, but as Konrad discussed recently we will need that for the SSTs. I want to make sure that we don't simulate here exactly following one of the support beams.

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+            opts += [
+                "-DPERFECT_DISH=1",
+                "-C random_focal_length=0",
+                "-C mirror_reflection_random_angle=0",
+                "-C mirror_align_random_distance=0",
+                "-C mirror_align_random_horizontal=0,28,0,0",
+                "-C mirror_align_random_vertical=0,28,0,0",
+            ]
+
+        if self.mirror_reflection_random_angle is not None:
```
GernotMaier, line 245: Similar to early - does this overwrite parameters from the database when given from the command line?

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        Returns
+        -------
+        bool
+            True when the token is present as a separate word; False otherwise.
+        """
+        desc_l = desc.lower()
+        axis_l = axis.lower()
+        return bool(re.search(r"(^|\s)" + re.escape(axis_l) + r"(\s|$)", desc_l))
+
+    def _maybe_set_reflection_index(self, desc, num, indices):
```
GernotMaier, line 465: Rename to `_set_reflection_index_if_match` (maybe does not sound deterministic)

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        idx : int | None
+            Index into ``parts``.
+
+        Returns
+        -------
+        tuple[bool, float]
+            Tuple of ``(ok, value)`` where ``ok`` is False when parsing fails
+            or the index is out of range; in that case ``value`` is 0.0.
+        """
+        if idx is None or idx < 0 or idx >= len(parts):
```
GernotMaier, line 507: Would this be simpler (similar next function):

```
    try:
        return True, float(parts[idx])
    except (IndexError, ValueError, TypeError):
        return False, 0.0
```

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        Parameters
+        ----------
+        parts : list[str]
+            Tokenized input row.
+        col_idx : dict[str, int]
+            Indices for hit coordinate columns.
+        radius_m, primary_hit_x_m, primary_hit_y_m : list | None
+            Output arrays to append into.
+        """
+        x_ok, x_cm = self._parse_float(parts, col_idx.get("prim_x"))
```
GernotMaier, line 632: I think the following would be easier to read:

```
    x_ok, x_cm = self._parse_float(parts, col_idx.get("prim_x"))
    y_ok, y_cm = self._parse_float(parts, col_idx.get("prim_y"))

    if x_ok and y_ok:
        x_m, y_m = x_cm / 100.0, y_cm / 100.0
        r_m = (x_cm**2 + y_cm**2) ** 0.5 / 100.0
    else:
        x_m = y_m = r_m = float("nan")

    if radius_m is not None:
        radius_m.append(r_m)
    if primary_hit_x_m is not None:
        primary_hit_x_m.append(x_m)
    if primary_hit_y_m is not None:
        primary_hit_y_m.append(y_m)
```

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        Parameters
+        ----------
+        parts : list[str]
+            Tokenized input row.
+        col_idx : dict[str, int]
+            Indices for hit coordinate columns.
+        secondary_radius_m, secondary_hit_x_m, secondary_hit_y_m : list | None
+            Output arrays to append into.
+        """
+        sx_ok, sx_cm = self._parse_float(parts, col_idx.get("sec_x"))
```
GernotMaier, line 664: see previous comment. Simplify for readability.

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        self.results.write(output_file, format="ascii.ecsv", overwrite=True)
+
+    def export_results(self):
+        """Export the results ECSV and a short text summary file."""
+        if self.results is None or len(self.results) == 0:
+            self.logger.error("Cannot export results because they do not exist")
+            return
+        table_file = self.results_dir / f"incident_angles_{self._label_suffix()}.ecsv"
+        self.results.write(table_file, format="ascii.ecsv", overwrite=True)
+        summary_file = self.results_dir / f"incident_angles_summary_{self._label_suffix()}.txt"
```
GernotMaier, line 728: Trying to understand better the summary file. Is this metadata? Those values are from the tool configuration? Then they should be saved as metadata. 

Otherwise it would be better to save this not as txt, but as yaml or json to guarantee machine readability.

src/simtools/visualization/plot_incident_angles.py
GernotMaier, line 1: Could you look into simtools.visualize.plot_tables and use this for plotting? I do think it should serve your purpose, otherwise adapt. As some of the distribution of incident angle is a model parameter, the corresponding model parameter schema files for the plotting instructions (see field `plot_configuration`). 

Using plot_tables should replace parts of plot_incident_angles.

src/simtools/visualization/plot_incident_angles.py
```diff
@@ -0,0 +1,419 @@
+#!/usr/bin/python3
+"""Plot incident angle histograms for focal, primary, and secondary mirrors.
+
+Also plots the primary-mirror hit radius if available.
```
GernotMaier, line 4: Remove 'Also'

src/simtools/visualization/plot_incident_angles.py
```diff
@@ -0,0 +1,419 @@
...
+import logging
+from pathlib import Path
+
+import astropy.units as u
+import matplotlib.pyplot as plt
+import numpy as np
+
+__all__ = ["plot_incident_angles"]
+
+Y_AXIS_BIN_COUNT_LABEL = "Count / Bin"
```
GernotMaier, line 16: As mentioned earlier, distributions should be normalized to integral one.

GernotMaier (changes requested):
Thanks a lot for the changes - this improved significantly the readability and it is easier now to understand what is going on.

Below some more comments (I am adding also an issue with next steps).

src/simtools/ray_tracing/incident_angles.py
tobiaskleiner: It is simply not important as you said, but I can change it to any angle you wish.

src/simtools/ray_tracing/incident_angles.py
tobiaskleiner: Yes, and I could remove this if not needed at all.

src/simtools/visualization/plot_incident_angles.py
tobiaskleiner: I am not sure what part of the incident angles plotting code you mean exactly, but plot_tables it is very different functionality from looking briefly at it.

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -0,0 +1,735 @@
...
+        self.logger.info("Executing %s (logging to %s)", script_path, log_file)
+        try:
+            subprocess.check_call([str(script_path)])
+        except subprocess.CalledProcessError as exc:
+            raise RuntimeError(f"Incident angles run failed, see log: {log_file}") from exc
+
+    def _compute_incidence_angles_from_imaging_list(self, photons_file):
+        """Compute incidence angles from an imaging list file.
+
+        Column positions may differ between telescope types and sim_telarray builds.
```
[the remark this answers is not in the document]
tobiaskleiner: No the second case, different versions of it.

src/simtools/applications/calculate_incident_angles.py
tobiaskleiner: Should we always use the one from DB or should we allow to overwrite this here?

src/simtools/visualization/plot_incident_angles.py
tobiaskleiner: fixed

src/simtools/visualization/plot_incident_angles.py
tobiaskleiner: fixed

src/simtools/ray_tracing/incident_angles.py
tobiaskleiner: fixed

src/simtools/ray_tracing/incident_angles.py
tobiaskleiner: fixed

src/simtools/ray_tracing/incident_angles.py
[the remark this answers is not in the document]
GernotMaier: Can you please provide more details: so the version of sim-telarray we are using (240205, 240927, 250304) are giving different output? 

- if it is version, please call it versions and not builds

src/simtools/visualization/plot_incident_angles.py
GernotMaier: I'll do it, it is actually straightforward. We cannot duplicate.

src/simtools/applications/calculate_incident_angles.py
GernotMaier: For now, we always use the DB. We discussed this item in last week's meeting: there is a need for a common functionality to overwrite parameters read from the DB. But we definitely don't want individual tools to do this in their own style.

src/simtools/visualization/plot_incident_angles.py
GernotMaier: Do you normalize in plot or calculate incident angle? It should be done in calculate incidient angle, as we want to write out there as well the results to the simulation model.

GernotMaier:
@tobiaskleiner - looked through your comments - most of it resolved, but could you check the remaining once?

PR #1690, opened 2025-08-20
Single telescope array trigger

GernotMaier:
The array trigger file for sim_telarray is written 'on the fly' by `SimtelConfigWriter:_write_array_triggers_file`. The issue of the code was that always the stereo trigger requirement was written out, even for single telescope configurations.

Changed this to read the correct configuration from the `array_triggers` model parameter (e.g. ``MSTN_single_telescope`).

Check https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/blob/main/simulation-models/model_parameters/OBS-South/array_triggers/array_triggers-1.0.0.json?ref_type=heads to see the array trigger simulation model parameter.

### DIFF
--- a/docs/changes/1690.bugfix.md
+++ b/docs/changes/1690.bugfix.md
@@ -1,0 +1,1 @@
+Fix array trigger writing for single telescope simulations.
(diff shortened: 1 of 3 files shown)


EshitaJoshi (approved):
Thanks for the PR!

PR #1700, opened 2025-08-25
Fix sonar reported security hotspots

GernotMaier:
Fix sonar reported security hotspots, see [here](https://sonar-ctao.zeuthen.desy.de/security_hotspots?id=gammasim_simtools_0d23837b-8b2d-4e54-9a98-2f1bde681f14).

This triggered a couple of good issues, especially to simplify the directory creation. We should generally avoid building directory names several times at different places in the code. Unified here the sim_telarray configuration file directory.

@EshitaJoshi  - please check the regex in docs_read_parameters.py. I think I got it right, but always hard to see what is happening (this is why putting regex into a small service function with a doctoring and corresponding unit test makes the code more maintainable).


EshitaJoshi (approved):
Thanks for the PR! The regex works well, I generated the reports again and checked that the markdown links were still working as expected.

PR #1710, opened 2025-08-27
Fix release link template in towncrier.

GernotMaier:
A surprising number of links in the towncrier-generate CHANGELOG did not work (see #1707) and it turns out that the link template in the towncrier configuration is wrong. This PR fixes this.


EshitaJoshi (approved):

GernotMaier:
Thanks for this @EshitaJoshi

Issue #972, opened 2024-06-11
Unit tests in applications


GernotMaier:
Some applications actually have a couple of functions, but we currently do not run any testing there.

Suggestion is:

- review the application code and see if some of the functionality can be moved into modules (and tested)
- understand if we want to add unit tests to those functions remaining in the application code

Note that code in applications without tests lead to lower coverage: we should exclude those from the coverage calculation if we are the opinion that it is not needed.

GernotMaier:
Below the list of all simtools applications. Unchecked are those which need a review of the code in the applications and needs review.


- [ ] calculate_trigger_rate.py
- [x] convert_all_model_parameters_from_simtel.py
- [ ] convert_geo_coordinates_of_array_elements.py
- [x] convert_model_parameter_from_simtel.py
- [x] db_add_file_to_db.py
- [x] db_add_simulation_model_from_repository_to_db.py
- [x] db_add_value_from_json_to_db.py
- [x] db_get_array_layouts_from_db.py
- [x] db_get_file_from_db.py
- [x] db_get_parameter_from_db.py
- [x] db_inspect_databases.py
- [x] derive_ctao_array_layouts.py
- [x] derive_mirror_rnda.py
- [x] derive_photon_electron_spectrum.py
- [ ] derive_psf_parameters.py
- [x] docs_produce_array_element_report.py
- [x] docs_produce_calibration_reports.py
- [x] docs_produce_model_parameter_reports.py
- [x] docs_produce_simulation_configuration_report.py
- [x] generate_array_config.py
- [ ] generate_corsika_histograms.py
- [x] generate_default_metadata.py
- [ ] generate_regular_arrays.py
- [ ] generate_sim_telarray_histograms.py
- [x] generate_simtel_event_data.py
- [x] merge_tables.py
- [ ] plot_array_layout.py
- [x] plot_tabular_data_for_model_parameter.py
- [x] plot_tabular_data.py
- [x] print_version.py
- [x] production_derive_corsika_limits.py
- [x] production_derive_statistics.py
- [ ] production_generate_grid.py
- [ ] production_merge_corsika_limits.py
- [x] run_application.py
- [ ] simulate_light_emission.py
- [x] simulate_prod_htcondor_generator.py
- [x] simulate_prod.py
- [x] submit_array_layouts.py
- [x] submit_data_from_external.py
- [x] submit_model_parameter_from_external.py
- [x] validate_camera_efficiency.py
- [ ] validate_camera_fov.py
[...]

PR #1736, opened 2025-09-03
Move application code to modules and add testing - plot simtel events

tobiaskleiner:
This is the first part of a series of PRs to adress issue #972.

Moved all application code to modules and added unit tests for the moved code.

Unit tests and integration tests pass locally. So everything should be fine to merge. Tests fail here due to known issue with:
test_create_d80_vs_offaxis_plot

But maybe better to cross check.


tobiaskleiner:
Thanks for the review @GernotMaier, all points adressed and merging now.

PR #1738, opened 2025-09-03
 Move application code to modules and add testing - corsika histograms

tobiaskleiner:


tobiaskleiner:
@GernotMaier thanks for the review. Moved and renamed to `plot_corsika_histograms`.

PR #1741, opened 2025-09-05
Complete the list of generated vector-optimzed images

GernotMaier:
Add avx512f and sse4 to the most recent v77 versions.


.github/workflows/build-simtools-production-image.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       - name: Download corsika-simtel-package
         run: |
           wget --no-verbose ${{ env.CLOUD_URL }}/${{ secrets[format('CLOUD_SIMTEL_{0}', matrix.simtel_version)] }}/download -O corsika_simtelarray_${{ matrix.simtel_version }}.tar.gz
+          wget --no-verbose ${{ env.CLOUD_URL }}/${{ secrets.CLOUD_AUTOCONF }}/download -O autoconf.tar.gz
```
orelgueta, line 36: I wanted to make this comment on #1722, but I ask here, should the tarball we download be for a specific version? Later in the compilation step we anyway call `RUN cd autoconf-2.71`, so we might as well specify the version also in the tarball.

It is far from critical anyway, but I thought I should mention.

orelgueta (approved):
All good, thanks!

I did make a comment that might require a change, but it's definitely a non-blocking one.

.github/workflows/build-simtools-production-image.yml
orelgueta: I don't mind either of the solutions, but I will offer another option as well, which is to use the `ARG` option like we did with all other versions.

PR #1772, opened 2025-09-19
Unit tests should not write into the source tree

GernotMaier:
Some unit tests write into the source tree. This should not be done, we always use `tmp_test_directory`.

Doc-CI fails, fixed in #1773


GernotMaier:
Thanks @tobiaskleiner

PR #1777, opened 2025-09-22
Fix model versions interpreted as floats

### LINKED ISSUE #1767: Integration tests not always running file tests
Integration tests can include tests for correct output files. There seems to be an issue with at least simulate_prod_gamma_20_deg_South_multiple_model_versions.yml, where the tests on `test_output_files` are not performed.

The issue seems to be [this line](https://github.com/gammasim/simtools/blob/bcbb675659eeefb1438740c8e5948f279e253ce9/src/simtools/testing/validate_output.py#L43), where we compare strings with floats: `['6.0', '6.1'] [6.0, 6.1]`. Need to ensure somewhere that model versions are always interpreted as strings. 

This might be a new issue, as we did not allow before to have `major.minor` only in the versioning.

GernotMaier:
PR #1768 introduces major.minor model versions, with simtools choosing the latest patch version automatically. 

This PR addresses issue #1767, which points out that some file comparison as part of the integration tests are not running anymore due to the interpretation of major.minor (meaning `6.0` is a float, but `6.0.0` is a string). Required the adaption of several integration tests.

Import to note is that this PR requires to update the sim_telarray configuration files in `tests/resources/sim_telarray_configurations`.

This will be needed to be done in future for each patch update of the simulation models.

Closes #1767


EshitaJoshi (approved):
Looks good, thanks!

GernotMaier:
Thanks @EshitaJoshi

PR #1803, opened 2025-09-29
Remove 'instrument:type' and 'instrument:site' in model parameter schema files.

GernotMaier:
Remove these fields as in the data model used now we:

- define the relation instrument - model parameter and site - model parameter in the production tables in the simulation models.

simtools does not use this information. Additionally, the fields are not updated, are incomplete or outdated.

Note! The unit test "tests/unit_tests/io/test_ascii_handler.py::test_collect_dict_from_url" must fail (as it compares the meta_schema_url result with the schema. They will pass as soon as this PR is merged to main.

### DIFF
--- a/docs/changes/1803.maintenance.md
+++ b/docs/changes/1803.maintenance.md
@@ -1,0 +1,1 @@
+Remove 'instrument:type' and 'instrument:site' in model parameter schema files.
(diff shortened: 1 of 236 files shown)


GernotMaier:
@orelgueta  - please take note and protest / comment if necessary (not on the details, on the general points)

PR #1809, opened 2025-09-30
Introduction application control module

GernotMaier:
Application startup (and in future shutdown) requires always the same 1-5 steps (e.g., setting up logger, io_handler; parsing configuration parameters). This PR introduces an application control functionality in `simtools.application_control` to have fixed startup (and in future shutdown) functionality.

Motivation for this was that I wanted to add that applications prints at startup the simtools and DB version used - but that would have required changed in every single application! In future, this can simply added to `simtools.application_control.startup_application`. A similar functionality can be implemented for shutdown.

In this initial step, this required changes in every single application. Quite repetitive, but once and never again. I've reformatted and synchronized also the docstrings, etc, in this go (e.g., no need to repeat in every single application the very long docstring for `_parse`.

For the review:

- review `src/simtools/application_control.py`
- parse through the changes in the applications (no need to do a line-by-line review)

### DIFF
--- a/src/simtools/application_control.py
+++ b/src/simtools/application_control.py
@@ -1,10 +1,21 @@
 """Application control utilities for startup and shutdown simtools applications."""
 
 import logging
+from dataclasses import dataclass
 from pathlib import Path
 
 import simtools.utils.general as gen
 from simtools.io import io_handler
+
+
+@dataclass
+class ApplicationContext:
+    """Container for common application context elements."""
+
+    args: dict
+    db_config: dict
+    logger: logging.Logger
+    io_handler: io_handler.IOHandler | None
 
 
 def startup_application(parse_function, setup_io_handler=True, logger_name=None):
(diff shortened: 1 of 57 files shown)


src/simtools/application_control.py
```diff
@@ -0,0 +1,102 @@
...
+
+    if logger_name:
+        logger = logging.getLogger(logger_name)
+    else:
+        logger = logging.getLogger()
+    logger.setLevel(gen.get_log_level_from_user(args_dict["log_level"]))
+
+    io_handler_instance = io_handler.IOHandler() if setup_io_handler else None
+
+    return args_dict, db_config, logger, io_handler_instance
```
tobiaskleiner, line 75: Suggest to use a dataclass for the return instead of a 4tuple:


```
from dataclasses import dataclass
from typing import Optional
import logging
from simtools.io.io_handler import IOHandler

@dataclass
class ApplicationContext:
    args: dict
    db_config: dict
    logger: logging.Logger
    io_handler: Optional[IOHandler]


```

and to return this in startup_application:
```
return ApplicationContext(
        args=args_dict,
        db_config=db_config,
        logger=logger,
        io_handler=io_handler_instance,
    )
```


and in the applications something like:
```

app_context = startup_application(_parse, setup_io_handler=False)

app_context.logger.info("Starting production simulations")
run_simulations(app_context.args, app_context.db_config)


```

This would make it easier maintainable and avoids ordering issues.

GernotMaier: Very good point, this make it much easier. Added!

PR #1817, opened 2025-10-01
Add integration tests for different telescope flashers

tobiaskleiner:
-description in title -

`--branch implement-realistic-flasher-model-parameters` will be removed after review


GernotMaier:
Please check if we need to chosen statistics - many of the added integration tests appear in the least of slowest integration tests. Maybe less events / photon statistics are ok for these tests?

Otherwise fine to be merged.

PR #1851, opened 2025-10-21
Add flasher pulse shape writing

tobiaskleiner:


GernotMaier (changes requested):
Thanks for this. 

Some quick comments - we also look if using the following for pulse shape definitions is simpler:

https://en.wikipedia.org/wiki/Exponentially_modified_Gaussian_distribution

src/simtools/simtel/simtel_config_writer.py
```diff
@@ -11,9 +11,62 @@
...
+    logger.debug(
+        f"Generating lightpulse table with rise10-90={width_ns}"
+        f" ns, fall90-10={exp_decay_ns} ns, dt={dt_ns} ns"
+    )
+
+    t, y = generate_pulse_from_risefall(
+        width_ns, exp_decay_ns, dt_ns=dt_ns, duration_sigma=duration_sigma
+    )
+
+    file_path = Path(file_path)
```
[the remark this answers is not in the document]
tobiaskleiner: Ok i take now the path from iohandler, and write the pulse shapes in a subdirectory "pulse_shapes". Or do you want to write it in the same folder as the model parameters?

tobiaskleiner:
@GernotMaier please have a look at the comments and let me know if this is fine.

src/simtools/simtel/simtel_config_writer.py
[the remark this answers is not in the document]
GernotMaier: This is ok.

src/simtools/simtel/pulse_shapes.py
```diff
@@ -0,0 +1,154 @@
+"""Pulse shape computations for light emission simulations for flasher."""
+
+from __future__ import annotations
```
GernotMaier: Is this one needed? I don't see annotations (apart from the docstring lines 68/70)

src/simtools/simtel/pulse_shapes.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        tau = float(res_tau.x)
+    else:
+        tau = brentq(fall_residual, a, b, maxiter=200)
+
+    _logger.info(f"Solved pulse parameters: sigma={sigma:.6g}, tau={tau:.6g}")
+    return sigma, float(tau)
+
+
+def generate_pulse_from_risefall(
```
GernotMaier: Docstring incomplete (parameters).

Also think this is not the correct PEP naming, should be `generate_pulse_from_rise_fall`. Would `generate_pulse_from_rise_fall_times` too long? It would be a little bit clearer.

src/simtools/simtel/pulse_shapes.py
```diff
@@ -0,0 +1,154 @@
...
+    rise_width_ns,
+    fall_width_ns,
+    dt_ns=0.1,
+    duration_sigma=8.0,
+    rise_range=(0.1, 0.9),
+    fall_range=(0.9, 0.1),
+):
+    """Get (t, y) by solving parameters from generic rise/fall specs and convolving.
+
+    Defaults correspond to 10-90 rise and 90-10 fall widths.
```
GernotMaier: `to 10-90% rise and 90-10% fall times`

src/simtools/simtel/pulse_shapes.py
```diff
@@ -0,0 +1,154 @@
...
+    t_hi = np.interp(y_high, y_rev, t_rev)
+    t_lo = np.interp(y_low, y_rev, t_rev)
+    return t_lo - t_hi
+
+
+def _gaussian(t, sigma):
+    return np.exp(-0.5 * (t / max(sigma, 1e-9)) ** 2)
+
+
+def _exp_decay(t, tau):
```
GernotMaier: I think this can be simplified to:

```
tau = max(tau, 1e-9)
 return np.where(t >= 0, np.exp(-t / tau), 0.0)
```

src/simtools/simtel/pulse_shapes.py
```diff
@@ -0,0 +1,154 @@
...
+import logging
+
+import numpy as np
+from scipy.optimize import brentq, minimize_scalar
+from scipy.signal import fftconvolve
+
+_logger = logging.getLogger(__name__)
+
+
+def _rise_width(t, y, y_low=0.1, y_high=0.9):
```
GernotMaier: Small note - no direct need to do something. But _rise_width and _fall_width looks almost identical.

GernotMaier:
Coverage on new code is a bit low - can we add 1-2 tests?

https://sonar-ctao.zeuthen.desy.de/component_measures?id=gammasim_simtools_0d23837b-8b2d-4e54-9a98-2f1bde681f14&metric=new_coverage&pullRequest=1851&view=list

PR #1856, opened 2025-10-22
Bugfix in validation of generated sim_telarray configuration

### LINKED ISSUE #1854: Bug in integration tests: comparison of sim_telarray configuration file not done properly
Integration tests generating sim_telarray configuration files compare the generated integration tests with references files found in `tests/resources/sim_telarray_configurations/`.

If I read the code correctly, then `_compare_simtel_cfg_files` in `simtools.testing.validate_output` is missing an `assert` statement, and no failure is reported. This would be a notable bug in our testing procedure.

Noticed while running pytest with debug statements printed to the screen. Was wondering why e.g.

```
ERROR    simtools.testing.validate_output:validate_output.py:374 Line counts differ: tests/resources/sim_telarray_configurations/CTA-beta-South-6.0_test.cfg (205 lines), /tmp/pytest-of-root/pytest-13/test-data0/simtools-simulate-prod-gamma_40_deg_south_check_output/simtools-output/model/6.0/CTA-beta-South-6.0_check_output.cfg (232 lines).
```

do not lead to a test failure.

Test case:

```
pytest -s --log-cli-level=DEBUG --no-cov --model_version=6.0 tests/integration_tests/test_applications_from_config.py::test_applications_from_config[simtools-simulate-prod_gamma_40_deg_south_check_output]
```

GernotMaier:
Missing assert statement allows these tests to always pass - even if there are differences between generated and reference sim_telarray configuration files. This is obviously a bug.

There are significant differences in the configuration files and this PR includes the updates to the most recent files.

Closes #1854

### DIFF
--- a/docs/changes/1855.maintenance.md
+++ b/docs/changes/1855.maintenance.md
@@ -1,0 +1,1 @@
+Review and update all XPASS unit tests.
--- a/src/simtools/data_model/schema.py
+++ b/src/simtools/data_model/schema.py
@@ -95,7 +95,7 @@ def get_model_parameter_schema_version(schema_version=None):
 
 
 def validate_dict_using_schema(
-    data, schema_file=None, json_schema=None, ignore_software_version=False
+    data, schema_file=None, json_schema=None, ignore_software_version=False, offline=False
 ):
     """
     Validate a data dictionary against a schema.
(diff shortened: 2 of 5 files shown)


GernotMaier:
@orelgueta - I assigned it to you to make sure that all the updates to the reference configuration files are ok.

orelgueta (approved):
Are we worried about the fact that it is all done with `5.0` and `6.0`, which means that you might get different values in the future when we publish a bug fix version for either of them? That is, maybe the value of one of the parameters will be different because it changes from 6.0.2 to 6.0.3?

Either way, I am approving already because if the issue above is not a problem, my one comment isn't blocking.

PR #1862, opened 2025-10-27
Fix plotting in UTM coordinates / added calibration devices.

GernotMaier:
A couple of updates required for plotting list of array elements as provided through the [CTAO Array Elements](https://gitlab.cta-observatory.org/cta-science/array-element-positions) website. Note that this plotting stuff is a bit fine tuned and it continues to be ridiculous that we need a module called ‘legend_handler’. 

- added functionality to plot calibration devices
- fixed functionality to plot in UTM coordinates. Requires to use `--bounds exact` to use the exact bounds (the default is to use symmetric bounds with +=max in x and y)
- allows to set plotting limits on the command line for x and y (e.g. `--x_lim 366700 366900 --y_lim 7268700 7268900`)

### DIFF
--- a/src/simtools/visualization/plot_array_layout.py
+++ b/src/simtools/visualization/plot_array_layout.py
@@ -40,10 +40,10 @@ def plot_array_layout(
     grayed_out_elements=None,
     highlighted_elements=None,
     legend_location="best",
-    bounds_mode: str = "exact",
-    padding: float = 0.1,
-    x_lim: tuple[float, float] | None = None,
-    y_lim: tuple[float, float] | None = None,
+    bounds_mode="exact",
+    padding=0.1,
+    x_lim=None,
+    y_lim=None,
 ):
     """
     Plot telescope array layout.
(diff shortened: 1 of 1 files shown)


src/simtools/visualization/plot_array_layout.py
```diff
@@ -23,6 +40,10 @@ def plot_array_layout(
     grayed_out_elements=None,
     highlighted_elements=None,
     legend_location="best",
+    bounds_mode: str = "exact",
+    padding: float = 0.1,
+    x_lim: tuple[float, float] | None = None,
```
tobiaskleiner: Didn't you say you don't like type hinting?

tobiaskleiner (approved):
Minor comment about type hints, otherwise all good and thanks @GernotMaier.

src/simtools/visualization/plot_array_layout.py
GernotMaier: Fixed (this was the AI...)

GernotMaier:
@tobiaskleiner - thanks for the review!

Issue #1771, opened 2025-09-19
Remove repetition in psf optimisation code


EshitaJoshi:
Almost every function call includes repetitively the same parameters (site_model, telescope_model). This is not nice and calls for a class.

Also increase speed of the unit test:
1.44s call tests/unit_tests/ray_tracing/test_psf_parameter_optimisation.py::test__handle_monte_carlo_analysis[True-True]

PR #1864, opened 2025-10-27
Introduce optimizer class

### LINKED ISSUE #1770: Improve exception handling in gradient descent algorithm
Current implementation sets gradient to 0 when ValueError or RuntimeError get raised, this is bad because it hides the error and implies that we are at a turning point when we are not.

EshitaJoshi:
Moving and refactoring code into an optimizer class.
Improved exception handling - now return None instead of setting gradients to 0 if the simulation fails to run.
Cache ray tracing simulation results in case the gradient descent algorithm steps to a set of parameter values which we already ran a simulation for (expected to happen more as we get closer to the minima).

Speed has improved a lot since the refactoring.
Unit tests are also faster now, all passing in 1.96s (with the slowest test taking 0.60s).

Closes #1770, #1771 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ This changelog is generated using [Towncrier](https://towncrier.readthedocs.io/)
 
 <!-- towncrier release notes start -->
 
+## [v0.24.0](https://github.com/gammasim/simtools/releases/tag/v0.24.0) - 2025-10-28
+
+### Bugfixes
+
+- Bugfix in validation of generated sim_telarray configuration: missing assert on the outcome of the tests (tests always passed, even with differences). ([#1856](https://github.com/gammasim/simtools/pull/1856))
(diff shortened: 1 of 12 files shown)


src/simtools/ray_tracing/psf_parameter_optimisation.py
```diff
@@ -32,6 +33,739 @@
...
+    ----------
+    tel_model : TelescopeModel
+        Telescope model object containing parameter configurations.
+    site_model : SiteModel
+        Site model object with environmental conditions.
+    args_dict : dict
+        Dictionary containing simulation configuration arguments.
+    data_to_plot : dict
+        Dictionary containing measured PSF data under "measured" key.
+    radius : array-like
```
GernotMaier: Is 'array-like' a data type?

src/simtools/ray_tracing/psf_parameter_optimisation.py
```diff
@@ -32,6 +33,739 @@
...
+            Number of Monte Carlo simulations to run.
+
+        Returns
+        -------
+        tuple
+            Monte Carlo analysis results.
+        """
+        if self.data_to_plot is None or self.radius is None:
+            logger.error("No PSF measurement data provided. Cannot analyze Monte Carlo error.")
+            return None, None, [], None, None, [], None, None, []
```
GernotMaier: Same here with this tuple. Very hard to read. Use dict or data class.

EshitaJoshi:
@GernotMaier thank you for the review! I have removed redundant comments/docstrings and will now merge the branch. The cleanup option and changing returns from tuples to dict or data class will be implemented in the very next PR.

PR #1883, opened 2025-11-04
Verify that the request and actually simulated number of events is identical for CORSIKA and sim_telarray simulations.

### LINKED ISSUE #1874: Validate the number of produced showers/events
A weird issue at CSCS raised the need to verify the number of events in the output file agrees with the expected number of events. What we observed is that sometimes we can get a kill signal that allows sim_telarray to exit gracefully with the following error:

```
Premature or unexpected end of run(s) after 387440 events (19373 showers) while expecting 2000000 events (100000 / 100000 showers).
```

The output file then looks normal, but contains significantly less events than expected. 

I suggest we add a validation test at the end of a production run like we used to have. In the legacy production system it was based on sim_telarray tools and took the number of events from the header of the file (see attached file with the code). Now we could either use that or, in the case `save_reduced_event_lists` is set to `true`, we could actually count the number of events in the file.

[dirac_simtel_check.txt](https://github.com/user-attachments/files/23211776/dirac_simtel_check.txt)

GernotMaier:
The main purpose of this pull request is to address #1874 and test at the end of a simulation run that the number of requested and simulated event is identical. Otherwise raise a ValueError.

Two tests are implemented:

- read the number of headers for `MCShower` and `MCEvent` from the sim_telarray files
- read the length of the `SHOWERS` table from the reduced event list file

Both results are compared with the number of expected events.

Additional minor changes:

- cleanup of the final printout of the simulation is part of this PR with a bit more consistent printout (now called `simulator.report()`.
- 'hist' key in list of output files in simtel.simulator is changed to 'histogram' (to be consistent with the getter in the runner utilities.

Closes #1874

### DIFF
--- a/docs/changes/1851.feature.md
+++ b/docs/changes/1851.feature.md
@@ -1,0 +1,1 @@
+Add pulse shape table writer for light emission flasher simulations.
--- a/docs/source/api-reference/sim_telarray.md
+++ b/docs/source/api-reference/sim_telarray.md
@@ -109,3 +109,12 @@ Support modules for running sim_telarray.
 .. automodule:: simtel.simulator_ray_tracing
    :members:
 ```
+
+## pulse_shapes
+
+(pulse_shapes-1)=
+
+```{eval-rst}
+.. automodule:: simtel.pulse_shapes
+   :members:
+```
--- a/src/simtools/applications/simulate_flasher.py
+++ b/src/simtools/applications/simulate_flasher.py
@@ -29,8 +29,9 @@ Example Usage
     .. code-block:: console
 
         simtools-simulate-flasher --run_mode full_simulation \
-        --light_source MSFx-FlashCam --model_version 6.0.0 \
-        --telescope MSTS-04 --site South
+        --light_source MSFx-NectarCam --model_version 6.0 \
+        --telescope MSTS-04 --site South --run_number 1 \
+        --array_layout_name 1mst
 
 Command Line Arguments
 ----------------------
--- a/src/simtools/simtel/pulse_shapes.py
+++ b/src/simtools/simtel/pulse_shapes.py
@@ -1,0 +1,273 @@
+"""Pulse shape computations for light emission simulations for flasher."""
+
+import logging
+
+import numpy as np
+from scipy.optimize import least_squares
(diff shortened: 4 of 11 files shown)


EshitaJoshi (approved):
Looks good to me, thanks

src/simtools/simulator.py
```diff
@@ -786,3 +761,92 @@ def _get_calibration_device_types(run_mode):
...
+        ------
+        ValueError
+            If the number of simulated events does not match the expected number.
+        """
+        for file in self.get_file_list(file_type="simtel_output"):
+            shower_events, mc_events = get_simulated_events(file)
+            if shower_events != expected_shower_events:
+                raise ValueError(
+                    f"Number of simulated shower events ({shower_events}) does not match "
+                    f"the expected number ({expected_shower_events}) in file {file}."
```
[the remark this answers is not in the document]
GernotMaier: Very good point. I've changed the logic and collect all errors and print them after the loop over all files. @EshitaJoshi  - let me know if the changes are ok.

src/simtools/simulator.py
```diff
@@ -786,3 +761,111 @@ def _get_calibration_device_types(run_mode):
...
+            f"Expected number of shower events: {expected_shower_events}, "
+            f"expected number of MC events: {expected_mc_events}"
+        )
+
+        if self.simulation_software in ["corsika_sim_telarray", "sim_telarray"]:
+            self._verify_simulated_events_in_sim_telarray(
+                expected_shower_events, expected_mc_events
+            )
+        if self.args_dict.get("save_reduced_event_lists"):
+            self._verify_simulated_events_in_reduced_event_lists(expected_mc_events)
```
orelgueta: Do I understand correctly that if the reduced event lists is saved, we count the number of events in the file twice? It might not cost that much time (to be seen), but is probably redundant, no? I think that if we save the reduced event lists, it's enough check only with `_verify_simulated_events_in_reduced_event_lists`

GernotMaier: Yes - do we really couple writing of event lists and the testing of sim_telarray files? I think they should be logical decoupled and the overhead of counting objects in a file should not be large.

PR #1885, opened 2025-11-04
Add tests that certain strings do not appear in the log files

GernotMaier:
PR #1882 added the functionality to test certain patterns in the CORSIKA log file.

Here, we original wanted to only update the integration tests to check the CORSIKA log files for the 'CURVED' option. Noticed while doing is that we need also the opposite test: that 'CURVED' is not set. Unfortunately CORSIKA is not printing out something like 'flat atmosphere'. This is the reason to add a small update to allow to list a list of `forbidden_pattern` - strings which should not appear in the log file.

SonarQ is complaining for the unit tests in general when writing or opening a tar file (flagged as unsafe). This is not an issue, as we are writing the tar files ourselves. Added therefore the corresponding exceptions to the tests (in the past: we flagged each tests as 'ok' on the sonarpage). Sorry about this change here, somewhat unrelated.


GernotMaier:
Thanks @EshitaJoshi

PR #1886, opened 2025-11-04
Adding mirror plots

### LINKED ISSUE #1638: Add missing default plotting configs
Some parameters are missing default plotting configurations that need to be added.
They are:

mirror_list
optics_properties
lightguide_efficiency_vs_wavelength

For SCTs:
fake_mirror_list
primary_mirror_segmentation
secondary_mirror_reflectivity

EshitaJoshi:
Adding the ```plot_mirrors``` module to the visualization library and close #1638 .
Adding the following plots to the auto report generation:

LSTN-01
mirror_CTA-N-LST1_v2019-03-31_rotated

MST
mirror_CTA-100_1-20-86-0-04

SST
astri_M1_segments_pre-production

SCT primary and secondary segments
sct_primary_segments
sct_secondary_segments

### DIFF
--- a/docs/changes/1894.feature.md
+++ b/docs/changes/1894.feature.md
@@ -1,0 +1,1 @@
+Add application for pulse shape parameter derivation and move existing functionality there.
--- a/docs/source/user-guide/applications.md
+++ b/docs/source/user-guide/applications.md
@@ -65,6 +65,7 @@ simtools-db-upload-model-repository <applications/simtools-db-upload-model-repos
 simtools-derive-ctao-array-layouts <applications/simtools-derive-ctao-array-layouts>
 simtools-derive-mirror-rnda <applications/simtools-derive-mirror-rnda>
 simtools-derive-photon-electron-spectrum <applications/simtools-derive-photon-electron-spectrum>
+simtools-derive-pulse-shape-parameters <applications/simtools-derive-pulse-shape-parameters>
 simtools-derive-psf-parameters <applications/simtools-derive-psf-parameters>
 simtools-derive-trigger-rates <applications/simtools-derive-trigger-rates>
 simtools-docs-produce-array-element-report <applications/simtools-docs-produce-array-element-report>
--- a/docs/source/user-guide/applications/simtools-derive-pulse-shape-parameters.rst
+++ b/docs/source/user-guide/applications/simtools-derive-pulse-shape-parameters.rst
@@ -1,0 +1,6 @@
+
+simtools-derive-pulse-shape-parameters
+======================================
+
+.. automodule:: derive_pulse_shape_parameters
+   :members:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,6 +89,7 @@ scripts.simtools-derive-ctao-array-layouts = "simtools.applications.derive_ctao_
 scripts.simtools-derive-mirror-rnda = "simtools.applications.derive_mirror_rnda:main"
 scripts.simtools-derive-photon-electron-spectrum = "simtools.applications.derive_photon_electron_spectrum:main"
 scripts.simtools-derive-psf-parameters = "simtools.applications.derive_psf_parameters:main"
+scripts.simtools-derive-pulse-shape-parameters = "simtools.applications.derive_pulse_shape_parameters:main"
 scripts.simtools-derive-trigger-rates = "simtools.applications.derive_trigger_rates:main"
 scripts.simtools-docs-produce-array-element-report = "simtools.applications.docs_produce_array_element_report:main"
 scripts.simtools-docs-produce-calibration-reports = "simtools.applications.docs_produce_calibration_reports:main"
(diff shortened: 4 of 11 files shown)


src/simtools/reporting/docs_read_parameters.py
```diff
@@ -175,7 +214,7 @@ def _convert_to_md(self, parameter, parameter_version, input_file):
         markdown_output_file = output_data_path / output_file_name
 
         if not markdown_output_file.exists():
-            outpath = Path(io_handler.IOHandler().get_output_directory().parent / "_images")
+            outpath = Path(io_handler.IOHandler().get_output_directory() / "_images")
```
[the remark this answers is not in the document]
EshitaJoshi: Fixed

EshitaJoshi:
> What does 'West' mean? Does it depend on the hemisphere? I just want to know...

This is all it says about it in ```mirror_CTA-100_1.20-86-0.04.dat```, I don't have any more information than that..
Should I remove the coordinate axes from the plots?

EshitaJoshi:
Thanks so much @orelgueta!
 
> Do I see correctly that you already rotated the LST and MST plots?
>
> For the LST indeed we found this error back in 2023 so we modified the file we received (coordinate systems are complicated). Once you use the new file, does the coordinate system make sense? Is there a reason to indicate the camera coordinate system on the mirror plot?

Yeah I already rotated them. I guess there's no reason to show the coordinate system now that we have added the rotation.

> The SST mirror is still 90 degrees rotated (see picture below).

Thanks, rotated that too:
astri_M1_segments_pre-production

 
> Unless we get individual measurements of the SCT mirror panels, any rotation error is not important there. The size of the mirrors in your plots is different to what I have in mine, but in this case I cannot claim mine are correct, I didn't thoroughly check the SCT mirror plots (at least not in the last 6 years).

These are the values updated in 2022 which I got from ```sct_primary_segments.dat``` and ```sct_secondary_segments.dat```. These files also have older values commented out in them, so I assume your plots are using those values instead. These files have the following note:

```
# Updated to use mean radii of inner and outer edge of segments
# as calculated from spreadsheet 'Segments.xlsx' (2022-05-30).
```

> Now, answering the original question, I would suggest to remove the axes mention on the plot and just add a text that says "for an observer facing the mirrors". What is important about the coordinate systems though, is to make sure that any transformation that's applied to one mirror list to get it to look right is applied in the same way to all (primary) mirror lists. Otherwise it means one of them will be wrong when running sim_telarray simulations.

Ok, changed. I have now rotated every mirror plot by 90 deg clockwise.

orelgueta:
Hi @EshitaJoshi, thanks for all the confirmations!
It could very well be that I plotted old SCT files, I don't trust my plots for that.

The new 90 degree rotated plots look good. We should note that this must be a part of the validation of new mirror lists that we receive. They must be rotated just like you did and compared to a picture of the telescope or the previous plot if the shape is not expected to change.

src/simtools/visualization/plot_mirrors.py
```diff
@@ -0,0 +1,845 @@
...
+        Path where to save the plot (without extension)
+    db_config : dict, optional
+        Database configuration dictionary
+
+    Returns
+    -------
+    None
+        The function saves the plot to the specified output file.
+    """
+    db = db_handler.DatabaseHandler(db_config=db_config)
```
GernotMaier: One item I've overlooked (thanks Orel for pointing out). We try to "hide" the implementation of db_handler from all code but the model_parameter module. It is more consistent with the rest of the code to define a TelescopeModel instance here - in a very similar way, roughly like:

```
    tel_model = TelescopeModel(
        site=app_context.args["site"],
        telescope_name=app_context.args["telescope"],
        db_config=app_context.db_config,
        model_version=app_context.args["model_version"],
        label=label,
    )
    tel_model.export_model_files()
```

This will ensure that if we have to change something fundamental in the db_handler, there is no need for many changes throughout the code.

EshitaJoshi: Thanks, added the TelescopeModel

src/simtools/visualization/plot_mirrors.py
```diff
@@ -153,6 +150,7 @@ def plot_mirror_layout(mirrors, telescope_model_name, title=None):
 
     x_pos = mirrors.mirror_table["mirror_x"].to("cm").value
     y_pos = mirrors.mirror_table["mirror_y"].to("cm").value
+    x_pos, y_pos = y_pos, -x_pos
```
GernotMaier: Suggest to add a comment here to motivate the rotation. We otherwise won't remember.

src/simtools/visualization/plot_mirrors.py
```diff
@@ -163,6 +161,11 @@ def plot_mirror_layout(mirrors, telescope_model_name, title=None):
         else list(range(len(x_pos)))
     )
 
+    # MST mirrors are numbered from 1 at the bottom to N at the top
```
GernotMaier: I don't understand this one - why do we need to re-assign mirror ids?

PR #1890, opened 2025-11-05
Check that simulated events is identical to request for CORSIKA simulations

GernotMaier:
Similar to PR #1883 : check CORSIKA output (not log file - note that in #1883 we don't write a CORSIKA file but pipe directly to sim_telarray) for the number of event headers and cross check with number of requested events. Allow for a small mismatch (as experience on the grid shows) - default tolerance is 0.1%.

Add a short integration test to run CORSIKA simulations only.

One small annoying change: file names are defined in several places (need to change!) and not exactly the same. CORSIKA files have a three digit printout for zenith (`040deg`), while all other files are happy with two (`40deg`). Change the CORSIKA ones to two.

### DIFF
--- a/docs/changes/1889.feature.md
+++ b/docs/changes/1889.feature.md
@@ -1,1 +1,0 @@
-Add test to verify that number of CORSIKA events is consistent with requested number of events.


EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thanks for the quick review @EshitaJoshi

PR #1900, opened 2025-11-12
Integrate flasher pulse shape model parameters

tobiaskleiner:
Updates the light emission simulator to use the new format of the pulse shape model parameters.

### DIFF
--- a/tests/unit_tests/simtel/test_simulator_light_emission.py
+++ b/tests/unit_tests/simtel/test_simulator_light_emission.py
@@ -251,6 +251,23 @@ def test__get_pulse_shape_string_for_sim_telarray(simulator_instance):
         "flasher_pulse_shape"
     )
 
+    # Reset mocks for third test (exponential branch)
+    simulator_instance.calibration_model.reset_mock()
+
+    # Test exponential pulse shape with decay only
+    simulator_instance.calibration_model.get_parameter_value.return_value = [
+        "Exponential",
+        0.0,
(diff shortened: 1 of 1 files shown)


docs/changes/1900.maintenance.md
GernotMaier: Tiny comment: hard to know that the 'new parameters' are - maybe write explicitly.

src/simtools/simtel/simulator_light_emission.py
```diff
@@ -349,16 +349,14 @@ def _add_flasher_command_options(self):
...
-        pulse_shape = self.calibration_model.get_parameter_value("flasher_pulse_shape")
-        width_q = self.calibration_model.get_parameter_value_with_unit("flasher_pulse_width")
-        exp_q = self.calibration_model.get_parameter_value_with_unit("flasher_pulse_exp_decay")
-        if (
-            isinstance(exp_q, u.Quantity)
-            and isinstance(width_q, u.Quantity)
-            and pulse_shape == "Gauss-Exponential"
-        ):
+
+        if shape_name == "Gauss-Exponential" and width_ns > 0 and exp_ns > 0:
```
GernotMaier: I am wondering if we should raise an error here if shape_name is Gauss-Exponential, but width or length is not >0.

tobiaskleiner:
Thanks for the review @GernotMaier. I have adressed both of them - mergin now.

PR #1923, opened 2025-11-28
Corsika naming interaction models

GernotMaier:
PR #1921 introduces a flexible definition of the sim_telarray and corsika paths. It also introduced `CORSIKA_EXECUTABLE` to define the executable name.

As @orelgueta pointed out, this requires detailed knowledge by the user and this PR introduces an improved naming with e.g.

- `SIMTOOLS_CORSIKA_HE_INTERACTION=qgs3`
- `SIMTOOLS_CORSIKA_LE_INTERACTION=urqmd`

to define the low and high energy interaction models. `simtools.setting` is building the correct path.

Added also a check that this corsika path exists.

This is an 'add-on' to PR #1921 - therefore changed directly the changelog from 1921 and not introduced a new one.

### DIFF
--- a/src/simtools/settings.py
+++ b/src/simtools/settings.py
@@ -33,18 +33,24 @@ class _Config:
         """
         self._args = MappingProxyType(args) if args is not None else {}
         self._db_config = MappingProxyType(db_config) if db_config is not None else {}
-        self._sim_telarray_path = os.getenv(
-            "SIMTOOLS_SIMTEL_PATH",
-            args.get("simtel_path", None) if args is not None else None,
+        self._sim_telarray_path = (
+            args.get("simtel_path")
+            if args is not None and "simtel_path" in args
+            else os.getenv("SIMTOOLS_SIMTEL_PATH")
         )
-        self._sim_telarray_exe = os.getenv(
-            "SIMTOOLS_SIMTEL_EXECUTABLE",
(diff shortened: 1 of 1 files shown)


src/simtools/settings.py
```diff
@@ -45,10 +45,31 @@ def load(self, args=None, db_config=None):
...
+        self._corsika_exe = self._get_corsika_exec() if self._corsika_path is not None else None
+
+    def _get_corsika_exec(self):
+        """
+        Get the CORSIKA executable from environment variable or command line argument.
+
+        Build the executable name based on configured interaction models. Fall back to
+        legacy naming (simply "corsika") if models are not specified.
+        """
+        he_model = os.getenv(
```
orelgueta: Isn't this reading of the env here contradict the usage of `_fill_from_environmental_variables` in `configurator.py`?

GernotMaier: We need the os.getenv to allow to have applications with now command line parameters (e.g. `simtools-print-version`).

However, you are right that the call is inverted and I've fixed this. It should be:

```
        he_model = (
            self._args.get("corsika_he_interaction")
            if self._args is not None and "corsika_he_interaction" in self._args
            else os.getenv("SIMTOOLS_CORSIKA_HE_INTERACTION")
        )
```

This allows to configure the HE model through the command line and overwrite the ENV variables. Thanks for the hint!

PR #1940, opened 2025-12-09
Improve illuminator integration tests

tobiaskleiner:
Add pattern search in logfiles during integration tests.
Fix bug when isotropic is used for angular distribution and width 0 is falsely added.


tobiaskleiner:
Thanks for the review @GernotMaier!

PR #1956, opened 2025-12-18
Remove duplication and add an action for the DB upload

GernotMaier:
This will never work.


GernotMaier:
I am closing this - it actually is not working.

PR #1999, opened 2026-01-28
Adding a label to ray tracing

EshitaJoshi:
Previously RayTracing and SimulatorRayTracing were getting their labels from the telescope model, which would be cached. This is usually ok, but for processes being run in parallel the cached label was being used and files were being over-written by all the parallel workers.

Adding here the ability to simply pass a label to RayTracing which gets used for all the output file naming, instead of reading the cached label from the telescope model.


GernotMaier (approved):
No comments, thanks!

PR #2034, opened 2026-02-09
Remove mrra optimization here

EshitaJoshi:
The optimization has been moved to ```derive_mirror_rnda```, and so ```mirror_reflection_random_angle``` is being removed as a parameter in this optimization.


src/simtools/applications/derive_psf_parameters.py
```diff
@@ -169,7 +161,7 @@ def _parse():
             "(not used with --monte_carlo_analysis)."
         ),
         type=float,
-        default=0.01,
+        default=0.0001,
```
GernotMaier: This is an unusual small learning rate, or? Just a comment, no need to change.

EshitaJoshi: It is small, but after removing rnda from this function the larger learning rate was too large. The learning rate is adaptive anyway so it's not going to stay the same, this is just the starting rate for the first step.

PR #2113, opened 2026-04-13
Improve activity and `run_application` metadata

GernotMaier:
Several important improvements regarding metadata / reproducibility.

Best to look at the simulation model parameter setting changes related to this PR:

- [input example](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-model-parameter-setting/-/tree/remove-dates/input/LSTN-design/pm_photoelectron_spectrum/019d776b-e24c-741d-bc05-e3f6f7ec77c7?ref_type=heads)
- [output model meta parameter example](https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-model-parameter-setting/-/blob/remove-dates/output/LSTN-design/pm_photoelectron_spectrum/019d776b-e24c-741d-bc05-e3f6f7ec77c7/pm_photoelectron_spectrum/pm_photoelectron_spectrum-2.0.1.meta.yml?ref_type=heads)

Every process of running a simtools application generates now a `activity_id` (UUID7). This is also written into the `activity:id` field in the metadata:

```
  activity:
    name: plot_tabular_data
    type: software
    id: 019d82c8-a267-7353-bdf1-8c8fb71ebe29
    software:
      name: simtools
      version: 0.29.1.dev142+g3150e8719
    start: '2026-04-12T17:41:21+00:00'
    end: '2026-04-12T17:41:21+00:00'
```
 
For the simtools-run-application we execute several simtools after each other in the process of the setting workflows.

- change directory structure in the input directories of the setting workflows and replace dates by UUID7s.
- these IDs are workflow IDs and are propagated to the metadata
- add the list of activities to the model parameter metadata file as `context:associated_activities`, e.g.\

```
    associated_activities:
    - name: simtools-derive-photon-electron-spectrum
      activity_id: 019d82c8-a266-7ba9-b064-c637329871a6
    - name: simtools-plot-tabular-data
      activity_id: 019d82c8-a267-7353-bdf1-8c8fb71ebe29
    - name: simtools-submit-model-parameter-from-external
      activity_id: 019d82c8-a268-78a1-aead-59360a46c1cd
```

The activity ID added to the model parameter metadata is above workflow ID, e.g.:

```
activity:
    name: setting_workflow
    type: software
    id: 019d776b-e24c-741d-bc05-e3f6f7ec77c7
    software:
      name: simtools
      version: 0.29.1.dev142+g3150e8719
    start: '2026-04-12T17:41:20+00:00'
    end: '2026-04-12T17:41:23+00:00'
    runtime_environment: null
```

Add `runtime_environment` field to copy this entry from the input configuration to the model parameter metadata file.

Additional added some simplification in the log file generation.

Technical notes:

- use UUID7 as their are sortable. From python 3.14 on, the `uuid` package can generate UUIDs version 7. As we allow python >=3.12, we need to add a dependency (yes, we use `uuid6` to generate UUID7)
- generalized the uuid generation in general.py to allow easy change in case we decide for a different ID in future.


src/simtools/application_control.py
```diff
@@ -59,7 +58,9 @@ def setup_logging(logger_name=None, log_level="INFO", log_file=None):
         log_file_path = Path(log_file)
         if log_file_path.parent:
             log_file_path.parent.mkdir(parents=True, exist_ok=True)
-        file_format = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+        file_format = logging.Formatter(
+            f"{config.activity_id} - %(name)s - %(levelname)s - %(message)s"
```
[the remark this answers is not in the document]
GernotMaier: No - adding more to the prefix will make it ever harder to read.

src/simtools/applications/run_application.py
```diff
@@ -94,14 +95,14 @@ def main():
...
-        initialization_kwargs={"db_config": True},
+        initialization_kwargs={"db_config": True, "paths": False},
         startup_kwargs={
             "setup_io_handler": False,
             "resolve_sim_software_executables": False,
         },
     )
 
-    simtools_runner.run_applications(app_context.args, app_context.logger)
+    simtools_runner.run_applications(app_context.args)
```
tobiaskleiner, line 105: Is this intentional to remove the logger injection here?

src/simtools/schemas/metadata.metaschema.yml
```diff
@@ -571,6 +586,33 @@ definitions:
...
+                    - type: string
+                    - type: "null"
+                  default: null
+                activity_id:
+                  description: |-
+                    Identifier of the associated activity.
+                  anyOf:
+                    - type: string
+                    - type: "null"
+                    - type: number
```
tobiaskleiner: why number? UUID should not be numeric

src/simtools/schemas/metadata.metaschema.yml
```diff
@@ -1145,3 +1202,30 @@ definitions:
...
+                    - type: string
+                    - type: "null"
+                  default: null
+                ACTIVITY_ID:
+                  description: |-
+                    Identifier of the associated activity.
+                  anyOf:
+                    - type: string
+                    - type: "null"
+                    - type: number
```
tobiaskleiner, line 1226: also here

src/simtools/data_model/workflow_metadata.py
```diff
@@ -0,0 +1,110 @@
...
+    metadata_args["label"] = "setting_workflow"
+    metadata_args["activity_id"] = workflow_activity_id
+    metadata_args["activity_start"] = workflow_start.isoformat(timespec="seconds")
+    metadata_args["activity_end"] = workflow_end.isoformat(timespec="seconds")
+    metadata_args["runtime_environment"] = deepcopy(runtime_environment)
+    metadata_args["site"] = workflow_context.get("site")
+    metadata_args["instrument"] = workflow_context.get("instrument")
+
+    collector = MetadataCollector(metadata_args, clean_meta=False)
+    return collector.get_top_level_metadata().get("cta", {}).get("activity", {})
```
tobiaskleiner, line 54: Can we assume cta is there?

src/simtools/runners/simtools_runner.py
```diff
@@ -32,25 +49,64 @@ def run_applications(args_dict, logger):
...
-            file.write("=" * 80 + "\n")
-            file.write(f"Application: {app}\nSTDOUT:\n{result.stdout}\nSTDERR:\n{result.stderr}\n")
-
-
-def _read_application_configuration(configuration_file, steps, logger):
+        try:
+            for config in configurations:
+                app = config.get("application")
+                if not config.get("run_application"):
+                    logger.info(f"Skipping application: {app}")
```
tobiaskleiner, line 56: Would it make sense to skip the application, but still continue with the workflow?

src/simtools/data_model/workflow_metadata.py
```diff
@@ -0,0 +1,110 @@
...
+    None
+        Function updates file in place when it exists.
+    """
+    metadata_path = Path(metadata_file)
+    if not metadata_path.exists():
+        logger.debug(f"Model-parameter metadata file does not exist: {metadata_path}")
+        return
+
+    metadata = ascii_handler.collect_data_from_file(metadata_path)
+    metadata = gen.change_dict_keys_case(metadata, True)
```
tobiaskleiner, line 84: Why do we need this here actually? Or in other words why do we allow ambigouity with cases?

src/simtools/data_model/workflow_metadata.py
```diff
@@ -0,0 +1,110 @@
...
+    dict
+        Activity block to be injected into model-parameter metadata files.
+    """
+    metadata_args = dict(args_dict)
+    metadata_args["label"] = "setting_workflow"
+    metadata_args["activity_id"] = workflow_activity_id
+    metadata_args["activity_start"] = workflow_start.isoformat(timespec="seconds")
+    metadata_args["activity_end"] = workflow_end.isoformat(timespec="seconds")
+    metadata_args["runtime_environment"] = deepcopy(runtime_environment)
+    metadata_args["site"] = workflow_context.get("site")
```
tobiaskleiner, line 50: all of these could return None. Should we be more careful here?

src/simtools/runners/simtools_runner.py
```diff
@@ -92,15 +153,71 @@ def _read_application_configuration(configuration_file, steps, logger):
...
+        return app_configuration["log_file"]
+    output_path = app_configuration.get("output_path")
+    if output_path is None:
+        return None
+    return Path(output_path) / f"{application}-{counter:02d}.log"
+
+
+def _get_model_parameter_metadata_file(application, app_configuration):
+    """Return expected metadata file for model-parameter submission applications."""
+    if application != "simtools-submit-model-parameter-from-external":
```
tobiaskleiner: Hardcoded but probably fine

src/simtools/runners/simtools_runner.py
```diff
@@ -92,15 +153,71 @@ def _read_application_configuration(configuration_file, steps, logger):
...
+        List of application configurations.
+
+    Returns
+    -------
+    dict
+        Context dict with 'site' and 'instrument' keys.
+    """
+    return {
+        "site": _get_workflow_configuration_value(configurations, "site"),
+        "instrument": _get_workflow_configuration_value(configurations, "instrument")
```
tobiaskleiner, line 213: The fallback here to telescope should be mentioned in the docstring. And is this actually intended?

src/simtools/schemas/metadata.metaschema.yml
```diff
@@ -571,6 +586,33 @@ definitions:
...
+            items:
+              type: object
+              additionalProperties: false
+              properties:
+                activity_name:
+                  description: |-
+                    Name of the associated activity.
+                  anyOf:
+                    - type: string
+                    - type: "null"
```
tobiaskleiner, line 603: activity_name is required, but here you allow "null" as value which makes this a bit pointless in my opinion. Should these fields be strictly required as non-null strings instead?

src/simtools/schemas/metadata.metaschema.yml
```diff
@@ -1145,3 +1202,30 @@ definitions:
...
+                  default: null
+                ACTIVITY_ID:
+                  description: |-
+                    Identifier of the associated activity.
+                  anyOf:
+                    - type: string
+                    - type: "null"
```
tobiaskleiner, line 1225: also here

src/simtools/utils/general.py
```diff
@@ -969,3 +970,99 @@ def load_environment_variables(env_file=".env", env_list=None):
         cleaned_value = env_value.split("#")[0].strip().replace('"', "").replace("'", "")
         env_values[key.removeprefix("SIMTOOLS_").lower()] = cleaned_value
     return env_values
+
+
+def uuid():
```
tobiaskleiner: So this replaces the builtin uuid in python? Maybe better to name it differently..

src/simtools/utils/general.py
```diff
@@ -969,3 +970,99 @@ def load_environment_variables(env_file=".env", env_list=None):
...
+    try:
+        anchor_index = path.parts.index(anchor)
+        subdirs = path.parts[anchor_index + 1 : -1]
+    except (ValueError, IndexError) as exc:
+        raise ValueError(f"Could not find subdirectory under '{anchor}': {exc}") from exc
+
+    if len(subdirs) == 0:
+        raise ValueError(f"Could not find subdirectory under '{anchor}'")
+
+    return "/".join(subdirs)
```
tobiaskleiner: Path(*subdirs) is better

src/simtools/utils/general.py
```diff
@@ -969,3 +970,99 @@ def load_environment_variables(env_file=".env", env_list=None):
...
+    -------
+    str
+        Subdirectory path between anchor and file name, joined with '/'.
+
+    Raises
+    ------
+    ValueError
+        If anchor is not present or no subdirectories are found after the anchor.
+    """
+    path = Path(path).resolve()
```
tobiaskleiner: just careful here that this also resolves symlinks

tobiaskleiner (changes requested):
Thanks @GernotMaier, this is a good refactor for uuids and metadata. I have added a few comments about consistency and some potential checks.

src/simtools/applications/run_application.py
GernotMaier: yes - to have the usage of logger in simtools_runner consistent with other modules.

src/simtools/schemas/metadata.metaschema.yml
GernotMaier: Fixed.

src/simtools/schemas/metadata.metaschema.yml
GernotMaier: I am not changing this - as this is an older version of the metaschema and used before we consistently applied UUIDs. I don't want to break something (although unlikely).

src/simtools/data_model/workflow_metadata.py
GernotMaier: You are right that in general we do not hardwire the observatory name. This case is a bit different, as we only one metadata schema (SimtoolsOutputMetadata), which is for CTA (and has this key included). Using a different metadata scheme would require a bit of work - and we can do this when this is becoming relevant.

src/simtools/runners/simtools_runner.py
GernotMaier: But this is exactly what happens - each config is a step in the workflow. What we do here is we keep single steps (there is functionality to give e.g., on the command line to say: run step 1, 3, 5 of the workflow)

src/simtools/data_model/workflow_metadata.py
GernotMaier: I have seen metadata definition from CTA with upper, lower, mixed case. simtools used in the past upper case, but moved to lower. I think this is fine to be robust here.

src/simtools/data_model/workflow_metadata.py
GernotMaier: Yes, and this is fine. At least for now workflows shouldn't fail due to missing metadata entries on site or telescope name.

src/simtools/runners/simtools_runner.py
GernotMaier: Yes, the naming is not consistent within simtools and the ctao metadata definition. 'instrument' is foreseen by CTAO, this is why it is given preference.

src/simtools/runners/simtools_runner.py
GernotMaier: Actually not needed. Removed it.

src/simtools/schemas/metadata.metaschema.yml
GernotMaier: No, there is a difference - the field is required here, meaning it is a reminder that either there was no activity (e.g., something is set by hand) or we don't know the activity id.

src/simtools/schemas/metadata.metaschema.yml
GernotMaier: see above

src/simtools/utils/general.py
GernotMaier: Good point - renamed it to get_uuid()

src/simtools/utils/general.py
GernotMaier: Fixed.

src/simtools/utils/general.py
GernotMaier: Good point - I actually want to keep the original path and not resolve it. Remove the `.resolve()`.

GernotMaier:
@tobiaskleiner - thanks for the careful review! I've implemented / commented on everything, let me know if this is fine.

tobiaskleiner:
Thanks @GernotMaier for the answers and the fixes. Can be merged now.

tobiaskleiner (approved):

PR #2119, opened 2026-04-15
improve legend font size in plots

GernotMaier:
Matplotlib-generated plots seem to be by design ugly. 

This is a simple fix to avoid legends which cover completely the actual plot - we lower the font size to get:

fadc_pulse_shape_comparison

In comparison, before we had:

fadc_pulse_shape_comparison

Also changed a warning about units in axis title to a debug statement (not something important for the user)

### DIFF
--- a/docs/changes/2119.maintenance.md
+++ b/docs/changes/2119.maintenance.md
@@ -1,0 +1,1 @@
+Change legend font size in tabular plotting.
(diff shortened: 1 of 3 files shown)


tobiaskleiner (approved):
Thanks @GernotMaier!

PR #2128, opened 2026-04-21
Add production limits for list of telescopes

tobiaskleiner:
This PR adds the option to pass a list of telescopes to compute the production limits. 

Fixes reading of NSB values from the merged tables.

### DIFF
--- a/docs/changes/2128.feature.md
+++ b/docs/changes/2128.feature.md
@@ -1,0 +1,1 @@
+Allow custom list of telescopes for corsika limits derivation.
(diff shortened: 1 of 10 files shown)


GernotMaier (changes requested):
Thanks for this.

Please check if commandline_parser:_add_model_option_layout() does not give you already the required command line option.

tobiaskleiner:
> Thanks for this.
> 
> Please check if commandline_parser:_add_model_option_layout() does not give you already the required command line option.

Right and this is exactly what is being used here, but you are right that it should not appear as arguement in the dosctring, as it is implicit. Let me know if this is fine now.

GernotMaier (approved):
All good now, thanks!

tobiaskleiner:
Thanks @GernotMaier!

Issue #2112, opened 2026-04-13
Integration test failures - IERS download


GernotMaier:
https://github.com/gammasim/simtools/actions/runs/24271486556/job/70877234663

I think there is a mechanism to disable the downloading of the IERS files - maybe we should do this for the tests?

GernotMaier:
Closed by #2129

PR #2129, opened 2026-04-23
Fix IERS download issue in integration tests CI

tobiaskleiner:
Disable Astropy IERS auto-download in offline/test mode to prevent network timeouts and spurious warnings causing test failures.

### DIFF
--- a/docs/changes/2129.bugfix.md
+++ b/docs/changes/2129.bugfix.md
@@ -1,0 +1,1 @@
+Disable Astropy IERS auto-download in offline/test mode to prevent network timeouts and spurious warnings causing test failures.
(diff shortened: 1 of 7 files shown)


GernotMaier (changes requested):
Thanks!

Would it make sense to move this into application_control (maybe called from startup_application) and do once at startup, independent of which application is called? Any application is calling this function.

tobiaskleiner:
Thanks @GernotMaier  for the review. I have moved the part to applications control and removed the settings in each application. Please let me know if this is fine now. Closes #2112.

GernotMaier (approved):
Thanks!

GernotMaier:
@tobiaskleiner - I am merging. Enjoy your vaccation!

PR #2146, opened 2026-04-30
Allow to specifiy model version dependent configuration parameters in configuration files

GernotMaier:
Add model dependent configuration parameters in the configuration files (used e.g., by integration tests):

```
  array_layout_name:
      by_version:
        "<7.0.0":
        - 1lst
        - alpha
        ">=7.0.0":
        - LSTN-01
        - CTAO-North-Alpha
```

This allows to run integration tests for different model versions, even if parameter versions change. Fixes e.g. [these runs](https://github.com/gammasim/simtools/actions/runs/25160419424/job/73753141659?pr=2132).

### DIFF
--- a/src/simtools/version.py
+++ b/src/simtools/version.py
@@ -315,26 +315,27 @@ def resolve_by_version(config, model_version):
     versions = model_version if isinstance(model_version, list) else [model_version]
     parsed_versions = [Version(str(version_item)) for version_item in versions]
 
+    def resolve_by_version_field(by_version_dict, parsed_versions, key):
+        """Resolve a single by_version field for all versions, enforcing consistency."""
+        matched_values = [_resolve_single_version(by_version_dict, pv) for pv in parsed_versions]
+        first_match = matched_values[0]
+        if not all(match == first_match for match in matched_values):
+            raise ValueError(
+                f"Inconsistent by_version resolution for key '{key}' and model versions "
+                f"{versions}: {matched_values}"
+            )
+        return first_match
+
+    def _resolve_single_version(by_version_dict, parsed_version):
+        for constraint, result in by_version_dict.items():
+            if matches_version_constraint(str(constraint), parsed_version):
+                return result
+        return None
+
     resolved = {}
     for key, value in config.items():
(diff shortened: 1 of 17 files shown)


src/simtools/version.py
```diff
@@ -267,3 +267,75 @@ def check_version_constraint(version_string, constraint):
...
+        Constraint string such as "<7.0.0", ">=6.0.0", "==5.2.1".
+    parsed_version : packaging.version.Version
+        Parsed version to test.
+
+    Returns
+    -------
+    bool
+        True if the version satisfies the constraint.
+    """
+    return check_version_constraint(str(parsed_version), str(constraint).strip())
```
EshitaJoshi: This function just calls the other function, wouldn't it be better to just directly call ```check_version_constraint``` with the appropriate string inputs? We don't need an extra 15 lines of docstring for this

src/simtools/version.py
```diff
@@ -267,3 +267,75 @@ def check_version_constraint(version_string, constraint):
...
+        matched_values = [_resolve_single_version(by_version_dict, pv) for pv in parsed_versions]
+        first_match = matched_values[0]
+        if not all(match == first_match for match in matched_values):
+            raise ValueError(
+                f"Inconsistent by_version resolution for key '{key}' and model versions "
+                f"{versions}: {matched_values}"
+            )
+        return first_match
+
+    def _resolve_single_version(by_version_dict, parsed_version):
```
EshitaJoshi: This doesnt need to be hidden as it's inside another function already

EshitaJoshi (approved):
Thanks for the PR!

src/simtools/version.py
GernotMaier: Fixed. This is by-the-way an issue introduced due to sonar and the complexity complains: I do think we loose readability by splitting a simple flow with several if statements into multiple functions. Anyway, changed.

src/simtools/version.py
GernotMaier: Good point - removed.

GernotMaier:
Thanks @EshitaJoshi  - simplified as suggested. Will wait for the tests and merge then.

PR #2166, opened 2026-05-05
Reintroducing Site reports

EshitaJoshi:
Associated MR on gitlab: https://gitlab.cta-observatory.org/cta-science/simulations/simulation-model/simulation-models/-/merge_requests/126 .

The reports [here](http://cta-science.gitlab-pages.cta-observatory.org/-/simulations/simulation-model/simulation-models/-/jobs/701075/artifacts/docs/build/html/productions/7.0.0/OBS-North.html) were generated using this branch.


src/simtools/applications/plot_array_layout.py
GernotMaier, line 1: We do not want a lot of functional code in the application scripts. Can you move those new functions into a module?

It is probably either src/simtools/visualization/plot_array_layout.py or src/simtools/layout/array_layout_utils.py.

src/simtools/reporting/docs_read_parameters.py
```diff
@@ -1098,8 +1098,8 @@ def _write_array_layouts_section(self, file, layouts):
             for element in sorted(elements):
                 file.write(f"| [{element}]({element}.md) |\n")
             file.write("\n")
-            version = self.model_version.replace(".", "-")
-            filename = f"OBS-{self.site}_{layout_name}_{version}.png"
+            version = self.model_version[:3]
```
GernotMaier: Remind me: did we decide to plot only major.minor and not patch versions? This is probably the only place where this is the case.

src/simtools/simtel/simtel_table_reader.py
```diff
@@ -248,6 +248,11 @@ def _data_columns_nsb_spectrum():
     )
 
 
+def _data_columns_nsb_reference_spectrum():
```
GernotMaier, line 251: Very confusing but ok. I expect that at some time we will give up simtel_table_reader.

GernotMaier (changes requested):
Thanks for this - two minor requests.

src/simtools/applications/plot_array_layout.py
EshitaJoshi: Done, moved to src/simtools/visualization/plot_array_layout.py

EshitaJoshi:
@GernotMaier Can you have another look at this please and let me know if it's ok to merge?

GernotMaier (approved):
Thanks - looks good.

Someone else might not be able to sleep because of the mix of underscore and hyphens (e.g. in ` array_layout_subsystem_msts_North_ground_6-0-2.png`). I don't mind.

EshitaJoshi:
> Someone else might not be able to sleep because of the mix of underscore and hyphens (e.g. in ` array_layout_subsystem_msts_North_ground_6-0-2.png`). I don't mind.

The issue is that when we have a filename ending with ```major.minor.patch``` version, the code seems to interpret the ```.patch``` as the existing file suffix and replaces it with ```.png``` so using dashes felt simpler, but we can change it at any point if someone complains

GernotMaier:
> > Someone else might not be able to sleep because of the mix of underscore and hyphens (e.g. in ` array_layout_subsystem_msts_North_ground_6-0-2.png`). I don't mind.
> 
> The issue is that when we have a filename ending with `major.minor.patch` version, the code seems to interpret the `.patch` as the existing file suffix and replaces it with `.png` so using dashes felt simpler, but we can change it at any point if someone complains

Yes - but you could use consistently underscores `rray_layout_subsystem_msts_North_ground_6_0_2.png`

PR #2201, opened 2026-05-26
Set the number of showers and number of runs in the generated production grid

GernotMaier:
Add functionality to set the number of showers and number of runs in the generated production grid.

Allows to specify either:
- `showers_per_run` and `number_of_runs`
or
- `total_showers` and `showers_per_run`

Add support for zenith-angle scaling of total showers, as provided by the LST team. The scaling factor can be configured by a new CL option `--zenith_angle_scaling_factor`.

Updated the test file for corsika limits with a more recent versions. As this does not include a NSB axes, add functionality to support both 2D (energy, zenith) and 3D (energy, zenith, NSB) limits files.

### Prompt

The following prompt has been used to generate the changes.

Implement support for flexible shower statistics in:

`src/simtools/applications/production_generate_grid.py`

**Requirements**

0. Rename CLI parameter:
- `nshow` -> `showers_per_run`

Apply this change consistently to:
- CLI handling
- written ECSV tables
- ECSV reading/parsing
- internal variable names where appropriate

No need to keep backward compatibility.

---

1. Existing behavior

If:
- `showers_per_run`
- `number_of_runs`

are provided, keep current behavior:

Each grid point gets:
- `number_of_runs` runs
- each with `showers_per_run` showers

---

2. Support total_showers

Add support for:
- `total_showers`
- `showers_per_run`

Behavior:
- compute:
  `number_of_runs = ceil(total_showers / showers_per_run)`
- if there is a remainder:
  - create a final run with fewer showers

Example:
- `total_showers = 2500`
- `showers_per_run = 1000`

Produces:
- run1 = 1000
- run2 = 1000
- run3 = 500

Store the actual number of showers per run in the generated grid.

---

3. Zenith-angle scaling

Add new CLI option:

```text
--total_showers_scaling [fixed|zenith_scaled]
```

Default:

- fixed

Behavior:

- fixed
  use the provided total_showers
- zenith_scaled
  scale showers per grid point using:

```
scaled_total_showers =
total_showers * exp(3.9781 * (cos(ZD) - 1))
```

where:

ZD = zenith distance in radians for the grid point

Apply scaling before calculating number of runs.


src/simtools/production_configuration/simulation_jobs.py
```diff
@@ -453,25 +456,61 @@ def _clip_max_quantity(configured_max, lookup_max):
...
+    total_showers_scaling,
+    cos_scaling_factor=DEFAULT_ZENITH_ANGLE_SCALING_FACTOR,
+):
+    """
+    Return total showers adjusted for the selected scaling mode.
+
+    Scaling modes:
+
+    - "fixed": total showers is unchanged.
+    - "zenith_scaled": total showers is scaled by the cosine of the zenith angle
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/production_configuration/simulation_jobs.py
```diff
@@ -391,32 +392,34 @@ def calculate_log_energy_midpoint(energy_range_pair):
...
+    showers_per_run_reference_energy,
+    showers_per_run_power_index=None,
     reference_energy=None,
 ):
-    """Return an energy-dependent nshow value."""
-    if baseline_nshow < 1:
-        raise ValueError("baseline_nshow must be a positive integer.")
+    """Return an energy-dependent showers per run value."""
+    if showers_per_run_reference_energy < 1:
+        raise ValueError("showers_per_run_reference_energy must be a positive integer.")
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

src/simtools/applications/simulate_prod.py
```diff
@@ -29,7 +29,7 @@
         so East is 90 degrees.
     zenith_angle (float, required)
         Zenith angle in degrees.
-    nshow (int, optional)
+    showers_per_run(int, optional)
         Number of showers to simulate.
         The Number of simulated events depends on the number of times a shower is reused in the
         telescope simulation. The number provided here is before any reuse factors.
```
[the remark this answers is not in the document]
GernotMaier: Fixed.

orelgueta:
Comment based on the PR description first:

> Support total_showers
> Add support for:
> 
> total_showers
> showers_per_run
> Behavior:
> 
> compute:
> number_of_runs = ceil(total_showers / showers_per_run)
> if there is a remainder:
> create a final run with fewer showers
> Example:
> 
> total_showers = 2500
> showers_per_run = 1000
> Produces:
> 
> run1 = 1000
> run2 = 1000
> run3 = 500
> Store the actual number of showers per run in the generated grid.

I don't think this should be the behaviour. The difference between the remainder and just rounding up is small enough not to worry about the extra resources. I think those extra resources are worth it in order to always be able to assume how many events are present in each file.

src/simtools/configuration/defaults.py
```diff
@@ -13,3 +13,6 @@
 # Valid simulation software identifiers and the default choice.
 SIMULATION_SOFTWARE_CHOICES = ("corsika", "sim_telarray", "corsika_sim_telarray")
 SIMULATION_SOFTWARE_DEFAULT = "corsika_sim_telarray"
+
+# Default exponential factor for zenith-dependent total shower scaling (derived by the LST team).
```
orelgueta: It would probably be different for the full array. Maybe it should just say that in the comment here for now, but later this might actually be a production configuration parameter.

src/simtools/applications/production_generate_grid.py
```diff
@@ -102,25 +107,43 @@ def _add_arguments(parser):
...
-            "of each energy_range entry."
+            "Scaling factor for zenith-dependent total_showers scaling. "
+            "Used only when --total_showers_scaling is 'zenith_scaled'."
         ),
         type=float,
         required=False,
-        default=None,
+        default=defaults.ZENITH_ANGLE_SCALING_FACTOR_DEFAULT,
     )
     parser.add_argument(
```
orelgueta, line 145: This is a great feature which I was about to suggest before opening the review and am happy to see it is included. The only comment I have here is that perhaps the description can be a bit more verbose to explain what it does and why (it took me a while to understand from the code). Also, it could be useful to mention that when using this feature, the baseline number of shower per run is taken from the run with the lowest energy threshold (which generally translates to the smallest zenith angle).

Then regarding the exact implementation (apologies for writing it here instead of where the code is), at the moment it is done based on the geometric mean using the full energy range, assuming the maximum energy does not change. However, with the change of the maximum energy as we increase the zenith, this scaling might become too aggressive. If I understand correctly, taking the min/max values for the energies of 20 deg and 60 deg, the reduction in the number of showers per run is roughly a factor 25, which is way too much. In reality we only reduce by a factor 2-3. We can achieve that in this specific case by setting the power law to -0.5, but I don't know if that works for all zenith angles. Perhaps modifying the code to take into account only the energy threshold would be easier, not sure.

orelgueta (commented):
Very nice!
One very minor comment and one which is a bit more complex, but technically not necessarily blocking if we choose a different parameter value.

GernotMaier:
> Comment based on the PR description first:
> 
> > Support total_showers
> > Add support for:
> > total_showers
> > showers_per_run
> > Behavior:
> > compute:
> > number_of_runs = ceil(total_showers / showers_per_run)
> > if there is a remainder:
> > create a final run with fewer showers
> > Example:
> > total_showers = 2500
> > showers_per_run = 1000
> > Produces:
> > run1 = 1000
> > run2 = 1000
> > run3 = 500
> > Store the actual number of showers per run in the generated grid.
> 
> I don't think this should be the behaviour. The difference between the remainder and just rounding up is small enough not to worry about the extra resources. I think those extra resources are worth it in order to always be able to assume how many events are present in each file.

Can you clarify what the issue is? That run3 has less events than the other runs? Just want to understand. What would you prefer? Simply adding as the last run a full run - better to have a bit more statistics then less?

src/simtools/configuration/defaults.py
GernotMaier: Improved the comment to say this.

src/simtools/applications/production_generate_grid.py
GernotMaier: Does it depend on the energy?

`total_showers * np.exp(cos_scaling_factor * (cos_zenith - 1))`

GernotMaier:
@orel - thanks for the review. I did not understand your comments, can you have another look?

orelgueta:
> Can you clarify what the issue is? That run3 has less events than the other runs? Just want to understand. What would you prefer? Simply adding as the last run a full run - better to have a bit more statistics then less?

Yes, run3 should have the same statistics as the rest. Better to have more than strictly following the request from the user.

src/simtools/applications/production_generate_grid.py
orelgueta: Sorry, I think the issue is that I marked the line where you add the argument for energy scaling, but github decided to show that and the lines above for the zenith scaling. This entire comment is about the energy scaling of the number of showers per run. Does reading the comment knowing that help?

GernotMaier:
@orelgueta  - thanks for all the feedback.

Changed:

- updated docstring and argpars help with improved explanation of the power-law scaling of the number of events (note that until now I used this mostly for fixed energy simulations)
- changed the calculation of the number of events such that the remainder is a full set of sims:

total_showers = 2500
showers_per_run = 1000
Produces:
run1 = 1000
run2 = 1000
run3 = 1000

I print out a warning now for the case that this changes the total_showers numbers.

GernotMaier:
@orelgueta - I've added now to the docstring of production_generate_grid.py that the power law scaling should be used for fixed energies only. We will extend this functionality in future.

GernotMaier:
Let me know if anything is missing then or if we can go ahead and merge.

orelgueta (approved):

PR #2210, opened 2026-05-28
Fix TypeError for object-dtype columns in HDF5 writer

### LINKED ISSUE #2208: Error when writing reduced error list files
In the middle of writing many reduced file lists from the half-moon files of Prod6 I noticed the following error printed to screen:

INFO: Writing table FILE_INFO of length 1 to /lustre/fs23/group/cta/users/ogueta/cta/reduced_event_list_prod6/gamma_z52_south_moon10p/gamma_52deg_180deg_run001208___cta-prod6-2156m-LaPalma-moon+magic.reduced_event_data.hdf5
  File "/workdir/env/bin/simtools-write-reduced-event-lists", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/workdir/env/lib64/python3.12/site-packages/simtools/applications/write_reduced_event_lists.py", line 51, in main
    Simulator.write_reduced_event_lists(
  File "/workdir/env/lib64/python3.12/site-packages/simtools/simulator.py", line 408, in write_reduced_event_lists
    table_handler.write_tables(
  File "/workdir/env/lib64/python3.12/site-packages/simtools/io/table_handler.py", line 270, in write_tables
    write_table_in_hdf5(table, output_file, _table_name)
  File "/workdir/env/lib64/python3.12/site-packages/simtools/io/table_handler.py", line 305, in write_table_in_hdf5
    dset = f.create_dataset(
           ^^^^^^^^^^^^^^^^^
  File "/workdir/env/lib64/python3.12/site-packages/h5py/_hl/group.py", line 240, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workdir/env/lib64/python3.12/site-packages/h5py/_hl/dataset.py", line 102, in make_new_dset
    tid = h5t.py_create(dtype, logical=1)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h5py/h5t.pyx", line 1748, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1772, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1799, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1702, in h5py.h5t._c_compound
  File "h5py/h5t.pyx", line 1772, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1839, in h5py.h5t.py_create
TypeError: Object dtype dtype('O') has no native HDF5 equivalent

GernotMaier:
## Summary

- Extend dtype conversion in `write_table_in_hdf5` to handle `dtype('O')` (Python object/str) columns, in addition to the existing `dtype('U')` (unicode) handling
- Adds a unit test reproducing the failure with a `file_name` column of object dtype

## Root Cause

The `FILE_INFO` table contains a `file_name` column defined as `str`, which numpy represents as `dtype('O')`. HDF5 has no native equivalent for this dtype. The existing fix only converted `dtype.kind == 'U'` columns; `dtype.kind == 'O'` was missed.

Fixes #2208

### DIFF
--- a/docs/changes/2210.bugfix.md
+++ b/docs/changes/2210.bugfix.md
@@ -1,1 +1,1 @@
-Fix TypeError when writing HDF5 tables with object-dtype string columns (e.g. `file_name` in `FILE_INFO`).
+Add 'moon' NSB level to legacy simtel writer. Fix TypeError when writing HDF5 tables with object-dtype string columns (e.g. `file_name` in `FILE_INFO`).
--- a/docs/changes/2210.bugfix.md
+++ b/docs/changes/2210.bugfix.md
@@ -1,0 +1,1 @@
+Fix TypeError when writing HDF5 tables with object-dtype string columns (e.g. `file_name` in `FILE_INFO`).
--- a/src/simtools/io/table_handler.py
+++ b/src/simtools/io/table_handler.py
@@ -346,7 +346,7 @@ def write_table_in_hdf5(table, output_file, table_name):
     None
     """
     for col in table.colnames:
-        if table[col].dtype.kind == "U":  # hdf5 does not support unicode
+        if table[col].dtype.kind in ("U", "O"):  # hdf5 does not support unicode or object dtypes
             table[col] = table[col].astype("S")
 
     with h5py.File(output_file, "a") as f:
--- a/src/simtools/sim_events/writer.py
+++ b/src/simtools/sim_events/writer.py
@@ -412,7 +412,8 @@ class EventDataWriter:
         ValueError
             If no NSB level keyword is found in the file name.
         """
-        nsb_levels = {"dark": 0.24, "half": 0.835, "full": 1.2, "moon": 0.835}
+        nsb_levels = {"dark": 0.24, "half": 0.835, "full": 1.2}
+        nsb_levels["moon"] = nsb_levels["half"]  # moon uses same level as half
 
         for key, value in nsb_levels.items():
             try:
--- a/src/simtools/sim_events/writer.py
+++ b/src/simtools/sim_events/writer.py
@@ -392,7 +392,8 @@ class EventDataWriter:
         """
         Return NSB level from file name.
 
-        Hardwired values are used for "dark", "half", and "full" NSB levels.
+        Hardwired values are used for "dark", "half", "full", and "moon" NSB levels.
+        "moon" is treated as equivalent to "half" (0.835).
         Allows to read legacy sim_telarray files without 'nsb_integrated_flux'
         metadata field.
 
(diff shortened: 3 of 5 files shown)


GernotMaier:
@orelgueta - can you check this? I don't have your full config, so I can't test it with the file you have used.

orelgueta:
Yes, this fixes the issue. The question is whether we know why this happens for the moon files and whether it's OK to just skip those objects.

GernotMaier:
@orelgueta 

It doesn't fix the issue - the actually problem is in sim_events.writer:

```
    def _get_nsb_level_from_file_name(self, file):
        """
        Return NSB level from file name.

        Hardwired values are used for "dark", "half", and "full" NSB levels.
        Allows to read legacy sim_telarray files without 'nsb_integrated_flux'
        metadata field.

        Parameters
        ----------
        file : str
            File name to extract NSB level from.

        Returns
        -------
        float
            NSB level extracted from file name.
        """
        nsb_levels = {"dark": 0.24, "half": 0.835, "full": 1.2}

        for key, value in nsb_levels.items():
            try:
                if key in file.lower():
                    self._logger.warning(f"NSB level set to hardwired value of {value} for {file}")
                    return value
            except AttributeError as exc:
                raise AttributeError("Invalid file name.") from exc

        self._logger.warning(f"No NSB level found in {file}, defaulting to None")
        return None
```

Did you change this now again? 

```gamma_52deg_180deg_run001208___cta-prod6-2156m-LaPalma-moon+magic.simtel.zst``` - what should I do with 'moon'? 

We really should move as soon as possible to use simtools for production. All this legacy stuff is error prone - so I am glad this broke!

**Comment by orelgueta** [comment]
In the legacy simulations "moon" always stood for half-moon conditions. Not ideal, but it's the last time we use the legacy software, so the pain is temporary.

**Comment by GernotMaier** [comment]
The real culprit is nsb_level, not file_name:

   - _get_nsb_level_from_file_name only recognises "dark", "half", or "full" in the filename → returns a float
   - The filename moon+magic.simtel.zst matches none of those → returns None
   - When Astropy builds the table with nsb_level=None, numpy stores that column as dtype=object
   - h5py cannot write dtype=object, causing the TypeError

  Any other filename that also lacks "dark"/"half"/"full" would trigger the same bug. The fix (catching dtype.kind == "O" for all columns) correctly handles it — None gets cast to a byte-string b"None", which at least doesn't crash.

I am adding now that moon and half is equivalent.

**Comment by orelgueta** [comment]
OK! So that would solve the underlying issue, thanks!

**Comment by GernotMaier** [comment]
@orelgueta - I've added now a 'moon' level to the legacy file treatment. And I added a error in case this is not successful, I don't think we should be silent here.

File: docs/changes/2210.bugfix.md [hunk 1]
```diff
@@ -0,0 +1 @@
+Fix TypeError when writing HDF5 tables with object-dtype string columns (e.g. `file_name` in `FILE_INFO`).
```
orelgueta: This should probably be expanded after the second change, but it's not critical.

orelgueta (approved):

docs/changes/2210.bugfix.md
GernotMaier: Fixed and improved.

PR #2232, opened 2026-06-02
Fix `instrument` field in model parameter schema and restrict it to allowed names defined in `common_definitions.schema.yml`.

GernotMaier:
(that was clearly missed before...)


orelgueta (approved):
Thanks!

PR #2245, opened 2026-06-08
Fix binning effect at lower corsika limit

GernotMaier:
The lower energy limit for CORSIKA is derived from the distribution of triggered events.

In rare cases, we obtained a value which was lower then lower energy limit of the broad-range simulations due to a binning effect (if the broad-range e_min is in the same bin as the lowest energy triggered event, the algorithm picked the lower energy bin edge).

This PR changes the algorithm such that in these cases we select the broad-range e_min.

### DIFF
--- a/src/simtools/production_configuration/derive_corsika_limits.py
+++ b/src/simtools/production_configuration/derive_corsika_limits.py
@@ -645,11 +645,11 @@ def _enforce_minimum_value(candidate, minimum):
     if minimum is None:
         return candidate
 
-    if hasattr(candidate, "unit") and hasattr(minimum, "to"):
+    if isinstance(candidate, u.Quantity) and isinstance(minimum, u.Quantity):
         minimum = minimum.to(candidate.unit)
-    elif hasattr(candidate, "unit") and not hasattr(minimum, "to"):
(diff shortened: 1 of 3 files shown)


src/simtools/production_configuration/derive_corsika_limits.py
```diff
@@ -640,10 +640,28 @@ def _process_result_row(res, cols, columns, units):
                 units[k] = None
 
 
-def _round_value(key, val):
+def _enforce_minimum_value(candidate, minimum):
+    """Ensure candidate is not below minimum while preserving units when present."""
+    if minimum is None:
+        return candidate
```
EshitaJoshi, line 646: Maybe log a warning that we are returning candidate because minimum was not set/none?

EshitaJoshi (approved):
Thanks for the PR!

src/simtools/production_configuration/derive_corsika_limits.py
GernotMaier: Good point - but the limits are tested later (after line 800) before returning them and we get a log message from there. I don't think we need to have it in both cases

GernotMaier:
Thanks @EshitaJoshi !

Issue #1233, opened 2024-11-12
Production configuration: Add documentation


tobiaskleiner:
Add the following to the production configuration tool:
- description of the assumptions, default values, statistical methods.
- metric documentation
- science cases documentation

PR #2246, opened 2026-06-08
Production grid limit interpolation

### LINKED ISSUE #2204: Production generation: max energy of simulations
The maximum energy in ERANG is currently understood as being a fixed values without any scaling (e.g. as function of zenith angle).

Discuss / implement a functional scaling.

GernotMaier:
Improved support for CORSIKA production grid limits, including new max energy scaling options (Closes #2204), and general code robustness. 

General logic now:

- set limits through `energy_range`, `core_scatter`, `view_cone`
- if `corsika_limits` is set (pointing to a table with limits derived): use those limits (interpolate), but keep those from the previous line as absolute limits (e.g., `core_scatter` gives the absolute max value for the scatter range)
- for `energy_max_scaling: -2.5 200 TeV`, scale the max energy by 200 TeV x cos(Zd)**-2.5 (but use the value in `energy_range` as absolute max - in the plot below it is 500 TeV)

Improved also the plotting, adding beside the alt/az also 1D projections.

Add also the plotting of the CORSIKA limits, e.g., 

production_grid_zenith_profile_core_scatter_max

production_grid_zenith_profile_energy_max

production_grid_zenith_profile_energy_min


src/simtools/applications/production_generate_grid.py
```diff
@@ -234,6 +242,27 @@ def _add_arguments(parser):
...
+            "energy_max_scaling_reference * cos(zenith_angle) ** power_index. "
+            "Provide: <power_index> <reference_energy_value> <reference_energy_unit> "
+            "(for example: --energy_max_scaling -2.5 300 TeV)."
+        ),
+        nargs=3,
+        type=str,
+        metavar=("POWER_INDEX", "REFERENCE_ENERGY_VALUE", "REFERENCE_ENERGY_UNIT"),
+        required=False,
+        default=None,
+    )
```
orelgueta, line 259: This isn't a very intuitive scaling. Since in most cases the scaling will not change, shouldn't we have a reasonable default that will be used most of the time?

Also, should the point about the maximum which is explain in the PR description be mentioned here as well?

src/simtools/production_configuration/corsika_limits_lookup.py
```diff
@@ -64,6 +78,11 @@ def load_matching_lookup_arrays(self):
             Lookup arrays for interpolation.
         """
         lookup_table = Table.read(self.lookup_table, format="ascii.ecsv")
+        logger.info(
+            "Loaded lookup table with %d rows and columns: %s",
+            len(lookup_table),
+            lookup_table.colnames,
+        )
```
orelgueta, line 83: Doesn't follow f-string convention (even listed in copliot instructions).

src/simtools/production_configuration/corsika_limits_lookup.py
```diff
@@ -186,19 +223,28 @@ def interpolate_grid_limits(self, target_values):
...
-            ).reshape(
+            )
+            if np.isnan(interpolated).any():
+                nearest_values = griddata(
+                    interpolation_points,
+                    self._repeat_wrapped_lookup_values(values),
+                    interpolation_target,
+                    method="nearest",
+                )
+                interpolated = np.where(np.isnan(interpolated), nearest_values, interpolated)
```
orelgueta: Quite interesting that this two step process is necessary to avoid `NaN` output, but I assume you checked that this is the simplest way.

orelgueta: Looking into it further (I thought perhaps the `fill_value` argument could be useful), there isn't a simpler option. However, I did get the following recommendation from copilot that will simplify this two-step process somewhat:

```
def create_interpolator_with_nearest_fallback(points, values):
    """Create a linear interpolator with nearest-neighbor fallback."""
    linear_interp = LinearNDInterpolator(points, values, fill_value=np.nan)
    nearest_interp = NearestNDInterpolator(points, values)
    
    def interpolate(x):
        result = linear_interp(x)
        nan_mask = np.isnan(result)
        if np.any(nan_mask):
            result[nan_mask] = nearest_interp(x)[nan_mask]
        return result
    
    return interpolate
```

This is the comment that came with that code snippet:
> This would avoid storing two separate interpolator dictionaries and the explicit nan-checking logic scattered throughout the code. The current implementation is functionally correct but could indeed be more compact.

src/simtools/production_configuration/simulation_jobs.py
```diff
@@ -587,6 +593,47 @@ def calculate_zenith_scaled_showers_per_run(
     raise ValueError(f"Unknown showers_per_run_scaling mode: {showers_per_run_scaling}")
 
 
+def scale_energy_max_for_zenith_angle(
+    zenith_angle,
+    energy_range_pair,
+    energy_max_scaling=None,
+):
+    """Scale energy_max with zenith angle and return an updated energy range pair."""
```
orelgueta: Docstring misses parameters and such for a public method

src/simtools/production_configuration/simulation_jobs.py
```diff
@@ -587,6 +593,47 @@ def calculate_zenith_scaled_showers_per_run(
...
+    scaled_energy_max = scaling_reference * (cos_zenith**energy_max_scaling_index)
+    if scaled_energy_max <= energy_min.to(scaled_energy_max.unit):
+        return None
+
+    return energy_min, scaled_energy_max.to(energy_range_max.unit)
+
+
+def _as_quantity(value, unit):
+    """Return value as Quantity, using unit for plain numeric values."""
+    return value if isinstance(value, u.Quantity) else value * unit
```
orelgueta: Kind of surprised that you need this function instead of using Astropy's functionality. Asking about it, I got the following clarification (rephrased):

> If the intent is to keep the original unit of a variable which is already a Quantity (just ensure it has a unit), the current implementation is fine. But if unit conversion is acceptable, you could simplify to just u.Quantity(value, unit)

Based on the usage above, I think using just `u.Quantity(value, unit)` would be fine, but I am not 100% sure.

Either way, if you keep this function, it should probably be in general utilities, no?

src/simtools/production_configuration/simulation_jobs.py
```diff
@@ -621,6 +687,22 @@ def _clip_max_quantity(configured_max, lookup_max):
...
+    elif len(tokens) == 1 and isinstance(tokens[0], str):
+        tokens = shlex.split(tokens[0])
+
+    if len(tokens) != 3:
+        raise ValueError(
+            f"{parameter_name} must be provided as "
+            "<power_index> <reference_energy_value> <reference_energy_unit>."
+        )
+
+    return float(tokens[0]), u.Quantity(f"{tokens[1]} {tokens[2]}")
```
orelgueta, line 720: Should this function be in command line parser?

orelgueta (commented):
Quite a lot of improvements. Just a few minor comments.

src/simtools/applications/production_generate_grid.py
GernotMaier: The example is from Abelardo's slides. Should we do something else?

The default is 'None' - meaning for all zenith angles the same scaling. 

And there will be a follow up PR with a lot of detailed documentation (for the complete process; addresses also #1233)

src/simtools/production_configuration/corsika_limits_lookup.py
GernotMaier: Fixed.

src/simtools/production_configuration/corsika_limits_lookup.py
GernotMaier: Very good point I've removed the Nan structure and the duplication.

src/simtools/production_configuration/simulation_jobs.py
GernotMaier: Fixed.

src/simtools/production_configuration/simulation_jobs.py
GernotMaier: Actually redundant: `value_conversion.get_value_as_quantity` does the same. Fixed.

src/simtools/applications/production_generate_grid.py
orelgueta: > The example is from Abelardo's slides. Should we do something else?

No, the concept is fine of course.

> The default is 'None' - meaning for all zenith angles the same scaling.

Meaning that all zenith angles use the same energy range, right?

If you think the values you have now from Abelardo or from your studies can fit most cases, I would set those as a default value.

Either way, the the point about the maximum energy cut off which is explained in the PR description should be mentioned in the help message as well.

src/simtools/production_configuration/simulation_jobs.py
GernotMaier: This is a very specialised parser used only in this context, while in command line parser we have those which are used by several application. This is why I put it here, ok?

orelgueta: OK!

orelgueta (approved):
The comment on the input argument is still pending, but it's a fairly small point, so I approve already.

src/simtools/applications/production_generate_grid.py
GernotMaier: For 'none', we use the value given in energy_range.

I've updated the help messages (and I agree with the feeling that we have a large number of parameters in this application with many options)

GernotMaier:
Thanks @orelgueta - I've updated the help message.

src/simtools/applications/production_generate_grid.py
orelgueta: OK, if a plot like this is included in the Production Configuration Report the user looks at while setting up a production, I assume that user will remember to set this parameter to something sensible. All good.

image

PR #2248, opened 2026-06-08
Generate htcondor submit files to make bias curves

EshitaJoshi:
Give it the apptainer image path and run with:

```
simtools-generate-bias-curve-submissions  --site North     --model_version 7.0.0     --array_layout_name LSTN-01     --azimuth_angle 0.0     --zenith_angle 20.0     --showers_per_run 1000     --core_scatter "20 1900 m"     --view_cone "0 deg 5 deg"     --number_of_runs 10     --apptainer_image path-to-image.sif  --label test    --output_path simtools-output/bias_curves
```

It will then generate ```bias_curves/nsb/htcondor_submit``` and ```bias_curves/proton/htcondor_submit``` with the .condor and .sh files respectively that can be submitted. Remember to copy the ```env.txt``` file into each htcondor_submit folder!


GernotMaier:
Thanks @EshitaJoshi  for the draft on the HTCondor / overwrite model parameter scanning. I think there is too much duplication - parameter_scan_generator.py is a variation of htcondor_script_generator.py and we would have to maintain in several places the same functionality. My suggestion is to combine this and get a job list file (as discussed yesterday), but with the overwrite model parameter as parameter:

```
run_number  zenith_angle  ...  overwrite_model_parameters                 scan_label
1           20 deg        ...  overwrite_threshold_220.yaml                threshold_220
2           20 deg        ...  overwrite_threshold_220.yaml                threshold_220
3           20 deg        ...  overwrite_threshold_230.yaml
```

To use as much of the existing code (and also what the user is familiar with), the workflow would be:

```
simtools-production-generate-grid --output_file base_grid.ecsv ...
simtools-generate-parameter-scan-grid \
  --job_grid_file base_grid.ecsv \
  --scan_config scan.yaml \
  --output_file scan_grid.ecsv
simtools-simulate-prod-htcondor-generator \
  --job_grid_file scan_grid.ecsv \
  --output_path htcondor_submit \
  --apptainer_image ...
  ```
  
  This way you don't need to duplicate the basic job preparation, and the job scripts only needs to be extended to use (if available) the `overwrite_model_parameters` columns.
  
 I paste below the diagnostic of the AI:
 
Yes. [parameter_scan_generator.py](/Users/maierg/DESYCloud/CTA/Pipelines/MC-Pipeline/gammasim-tools/simtools/src/simtools/job_execution/parameter_scan_generator.py:156) is largely a second HTCondor generator beside [htcondor_script_generator.py](/Users/maierg/DESYCloud/CTA/Pipelines/MC-Pipeline/gammasim-tools/simtools/src/simtools/job_execution/htcondor_script_generator.py:196). The unique part is only the scan logic: cartesian parameter combinations and overwrite YAML generation. The submit file, params file, executable script, image resolution, log directories, and queue-from handling should stay in the existing HTCondor path.

Main concerns with the new diff:
- It bypasses the production grid flow from [production_generate_grid.py](/Users/maierg/DESYCloud/CTA/Pipelines/MC-Pipeline/gammasim-tools/simtools/src/simtools/applications/production_generate_grid.py:246), so it loses existing quantity parsing, run expansion, CORSIKA-limit clipping, shower scaling, metadata, and grid validation.
- It duplicates submit rendering from [htcondor_script_generator.py](/Users/maierg/DESYCloud/CTA/Pipelines/MC-Pipeline/gammasim-tools/simtools/src/simtools/job_execution/htcondor_script_generator.py:255).
- It uses `--nshow` in the generated script, while the established production path uses `--showers_per_run`.
- It only uses the first Apptainer image from a mapping, so it drops the existing multi-image behavior.
- It hardcodes HTCondor log paths and has its own params format, so future fixes would need to be made twice.

My suggested implementation plan:

1. Keep `simtools-production-generate-grid` as the source of simulation job rows.

2. Reframe parameter scan generation as “expand an existing job grid”, not “generate HTCondor submission files”. Add a small scan-grid step that:
   - reads a base `job_grid.ecsv` with `read_job_grid`;
   - reads the scan YAML;
   - generates one overwrite YAML per scan combination;
   - duplicates each base grid row for each scan combination;
   - adds optional row fields like `overwrite_model_parameters` and `label_suffix`;
   - writes a new `scan_job_grid.ecsv`.

3. Extend [job_grid_io.py](/Users/maierg/DESYCloud/CTA/Pipelines/MC-Pipeline/gammasim-tools/simtools/src/simtools/production_configuration/job_grid_io.py:12) to preserve optional string fields, at least:
   - `overwrite_model_parameters`
   - `label_suffix` or `scan_label`

4. Extend [htcondor_script_generator.py](/Users/maierg/DESYCloud/CTA/Pipelines/MC-Pipeline/gammasim-tools/simtools/src/simtools/job_execution/htcondor_script_generator.py:21) to include optional queued fields when present:
   - pass `--overwrite_model_parameters "$overwrite_model_parameters"` only if set;
   - append the scan label suffix to `job_label`;
   - keep all existing output, image-label grouping, `htcondor_log_path`, `simulation_output`, and params-file behavior.

And for full clarity, this is the prompt I've used:

```
Carefully analyse the diff to main and compare that to the case where we:

-  generate a grid of simulation parameters with src/simtools/applications/production_generate_grid.py
- prepare a submission file, script and condor file for HT Condor with src/simtools/applications/simulate_prod_htcondor_generator.py

Is src/simtools/job_execution/htcondor_script_generator.py and src/simtools/job_execution/parameter_scan_generator.py quite similar to each other?

Plan and make a suggestion on how to implement the parameter scan using as much of the existing functionality.
```

File: src/simtools/applications/generate_parameter_scan_grid.py [hunk 1]
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/python3
+
+r"""
+Generate a parameter scan job grid by expanding a base production grid.
+
+Reads a base job grid ECSV from ``simtools-production-generate-grid``, generates
+one overwrite YAML file per scan parameter combination, and writes an expanded
+``scan_grid.ecsv``. The resulting file is passed directly to
```
GernotMaier, line 8: Why do need to pass it directly to the HT Condor generator? This mixes a bit configuration steps and the actual backend.

src/simtools/job_execution/parameter_scan_generator.py
```diff
@@ -0,0 +1,216 @@
...
+    return overwrite_data
+
+
+def _generate_overwrite_file(overwrite_base, param_combo, combo_name, work_dir, label):
+    """Generate overwrite YAML file for one parameter combination."""
+    overwrite_data = _build_overwrite_data(overwrite_base, param_combo)
+
+    safe_label = _format_value_for_name(label)
+    overwrite_file = work_dir / f"overwrite_{safe_label}_{combo_name}.yaml"
+    with open(overwrite_file, "w", encoding="utf-8") as file_handle:
```
GernotMaier: Do not use yaml dump directly (we don't do that anywhere). Always use ascii_handler.write_data_to_file (and collect_data_from_file() for reading), as this allows to have in one place any necessary extensions.

src/simtools/job_execution/parameter_scan_generator.py
```diff
@@ -0,0 +1,216 @@
...
+    duplicated for every combination with ``overwrite_model_parameters`` and
+    ``scan_label`` columns added.
+    """
+    scan_config_path = Path(scan_config_path)
+    output_file = Path(output_file)
+    output_dir = output_file.parent
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    with open(scan_config_path, encoding="utf-8") as file_handle:
+        scan_config = yaml.safe_load(file_handle)
```
GernotMaier: use ascii_handler

src/simtools/job_execution/parameter_scan_generator.py
```diff
@@ -0,0 +1,216 @@
...
+            {
+                "combo": combo,
+                "name": "_".join(combo_name_parts),
+            }
+        )
+
+    return combinations
+
+
+def expand_job_grid_with_scan(base_grid_file, scan_config_path, output_file):
```
GernotMaier, line 178: Public functions should have numpy doc style description of parameters.

src/simtools/job_execution/parameter_scan_generator.py
```diff
@@ -0,0 +1,216 @@
...
+def _combo_name_part(param_spec, value):
+    """Return the label component for a single scan parameter value."""
+    scan_label = _format_value_for_name(param_spec.get("label", param_spec["name"]))
+    scan_value = _format_value_for_name(value)
+    return f"{scan_label}{param_spec.get('label_separator', '_')}{scan_value}"
+
+
+def _generate_parameter_combinations(param_specs):
+    """Generate all cartesian combinations of parameter values."""
+    value_lists = [p["values"] for p in param_specs]
```
GernotMaier, line 153: Use

```
values = general.ensure_list(param_spec["values"])
if not values:
    raise ValueError("'values' must contain at least one scan value.")
```

Otherwise a "220" is expanded to [2,2,0]

src/simtools/job_execution/parameter_scan_generator.py
GernotMaier, line 1: How do we ensure that scan_config_path has the correct structure and keys? Is there a new schema needed? 

I think the simulation_models_info.schema.yml covers the "overwrite" part, but maybe add

parameter_scan_config.schema.yml, validating:
top-level label
parameter_scan.overwrite
parameter_scan.parameters
parameter_scan.job_grid_updates

src/simtools/applications/generate_bias_curve_submissions.py
GernotMaier, line 1: We discussed in Slack that we want to separate the configuration step (simtools-production-generate-grid, simtools-generate-parameter-scan-grid) and the submission step.

src/simtools/job_execution/bias_curve_submissions.py
GernotMaier, line 1: Most of the logic here is not related at all to job submission, but to bias curves. I think job_execution is the wrong place for it, do you have a suggestion for a better place? Where do you put the bias curve evaluation code? Maybe it should go there.

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+
+from simtools.model.site_model import SiteModel
+from simtools.runners import simtools_runner
+from simtools.utils import names
+
+_logger = logging.getLogger(__name__)
+
+_PARAMETER_VERSION = "2.0.0"
+
+_NSB_ENERGY_RANGE = "20 MeV 25 MeV"
```
GernotMaier, line 31: Are these hardwired values actual used? I don't think we want that - both for versions, ranges, and thresholds. Isn't this all in the scan_config.yaml file?

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+
+    return [telescope]
+
+
+def _threshold_param_name(telescope):
+    """Return the trigger-threshold parameter name for a telescope."""
+    if not telescope:
+        raise ValueError("Cannot determine threshold parameter for empty telescope name.")
+
+    telescope_type = telescope[0].upper()
```
GernotMaier: Not ideal to identify a telescope with the first character only! This is not future proofed. Also would a "S" messing (for SST or SCT??)

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+    for key in required_args:
+        if args.get(key) in (None, ""):
+            raise ValueError(f"Missing required argument: --{key}")
+
+
+def generate_bias_curve_submissions(args):
+    """Generate NSB and proton bias-curve HTCondor submissions from CLI args."""
+    _validate_required_args(args)
+
+    args["telescopes"] = _resolve_telescopes_from_layout(args)
```
GernotMaier, line 347: Minor: generally I wouldn't expect that the command line args is modified / overwritten / extended at a later stage in the code. This is a bit of unexpected behavour.

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+    if len(args["telescopes"]) == 1:
+        args["telescope"] = str(args["telescopes"][0])
```
GernotMaier, line 349: Isn't the length of the array in all cases 1, given _resolve_telescopes_from_layout?

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+
+    output_root = Path(args.get("output_path") or ".").expanduser().resolve()
+    output_root.mkdir(parents=True, exist_ok=True)
```
GernotMaier: use io_handler.get_output_path()

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+        }
+    )
+
+
+def _generate_curve_submissions(curve_name, curve_definition, args, output_root):
+    """Generate grid, scan config, scan grid, and HTCondor scripts for one curve."""
+    curve_directory = output_root / curve_name
+    curve_directory.mkdir(parents=True, exist_ok=True)
+
+    telescope = args["telescopes"][0]
```
GernotMaier, line 303: We defined args["telescope"]

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+    if curve_name == "proton":
+        return _base_proton_overwrite(
+            telescope=telescope,
+            model_version=args["model_version"],
+        )
+
+    raise ValueError(f"Unsupported curve name '{curve_name}'.")
+
+
+def _write_yaml(file_path, content):
```
GernotMaier, line 209: use ascii_handler to write files.

src/simtools/job_execution/bias_curve_submissions.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def _base_proton_overwrite(telescope, model_version):
+    """Build the proton base overwrite block before threshold insertion."""
+    return {
+        "model_version": model_version,
+        "model_update": "patch_update",
+        "model_version_history": [model_version],
+        "description": "Tune for proton telescope trigger scan",
+        "changes": {telescope: {}},
```
GernotMaier, line 159: I think we need also here the NSB overwrite. NSB is always simulated and affects the trigger.

tests/unit_tests/job_execution/test_job_manager.py
```diff
@@ -333,3 +333,60 @@ def test_raise_job_execution_error_no_logs(mocker):
         jm._raise_job_execution_error(exc, None, None, None)
 
     assert "See excerpt from log file above" in str(exc_info.value)
+
+
+# Coverage edge cases for job_manager.py
```
GernotMaier, line 338: Is this related to changes in this PR? I don't see any changes in job_manager.

(and if they are good tests, remove this comment.

tests/unit_tests/job_execution/test_bias_curve_submissions.py
```diff
@@ -0,0 +1,364 @@
...
+            bias_curve_submissions._resolve_telescopes_from_layout(args)
+
+
+@pytest.mark.parametrize(
+    ("telescope", "expected_param", "expected_values"),
+    [
+        (
+            "LSTN-01",
+            "asum_threshold",
+            [220, 230, 240, 250, 260, 270, 280, 290, 300, 320, 340, 360],
```
GernotMaier, line 108: See comment in the code also on those hardwired values. If it is needed, please don't duplicated the values both in the production and in the test code.

tests/unit_tests/job_execution/test_bias_curve_submissions.py
```diff
@@ -0,0 +1,364 @@
...
+    assert telescope_changes["min_photons"]["value"] == 0
+    assert telescope_changes["min_photoelectrons"]["value"] == 0
+    assert "asum_threshold" not in telescope_changes
+
+    assert overwrite["changes"]["OBS-North"]["nsb_scaling_factor"]["value"] == 2
+    assert scan_config["parameter_scan"]["parameters"] == [
+        bias_curve_submissions._parameter_scan_entry(args["telescope"])
+    ]
+    assert scan_config["parameter_scan"]["job_grid_updates"] == {"telescope": "LSTN-01"}
+    assert scan_config["parameter_scan"]["job_grid_updates"] == {"telescope": "LSTN-01"}
```
GernotMaier, line 188: Duplicated line - either remove or correct the test.

tests/unit_tests/job_execution/test_bias_curve_submissions.py
```diff
@@ -0,0 +1,364 @@
...
+    assert bias_curve_submissions._threshold_values_for_telescope(telescope) == expected_values
+
+
+def test_threshold_param_name_rejects_empty_telescope_name():
+    with pytest.raises(ValueError, match="empty telescope name"):
+        bias_curve_submissions._threshold_param_name("")
+
+
+def test_threshold_param_name_rejects_unknown_telescope_type():
+    with pytest.raises(ValueError, match="Expected telescope name to start with L, M, or S"):
```
GernotMaier, line 128: I think 'S' is not implemented (and we should work with single characters)

src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -350,16 +414,16 @@ def _get_submit_script(args_dict):
 corsika_he_interaction="{bash_indices["corsika_he_interaction"]}"
 run_number="{bash_indices["run_number"]}"
 pack_for_grid_register="{bash_indices["pack_for_grid_register"]}"
-energy_range_tag="{energy_range_tag}"
-job_label="{label}_${{corsika_he_interaction}}-${{corsika_le_interaction}}_${{energy_range_tag}}"
+job_label="{label}"
+{scan_label_block}{overwrite_parameters_block}{telescope_block}
```
GernotMaier, line 416: Does this break the original functionality? e.g. preparing a grid with different interaction models or a list of fixed energies?

src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -350,16 +414,16 @@ def _get_submit_script(args_dict):
...
 
 simtools-simulate-prod \\
     --simulation_software {args_dict["simulation_software"]} \\
     --label "$job_label" \\
     --model_version "$model_version" \\
     --site {args_dict["site"]} \\
     --array_layout_name "$array_layout_name" \\
-    --primary "$primary" \\
+{telescope_argument}    --primary "$primary" \\
```
GernotMaier, line 424: Missing spaces?

src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -371,22 +435,20 @@ def _get_submit_script(args_dict):
     --run_number "$run_number" \\
     --run_number_offset {run_number_offset} \\
     --save_reduced_event_lists \\
-    --output_path /tmp/simtools-output \\
+{overwrite_parameters_argument}    --output_path /tmp/simtools-output \\
```
GernotMaier, line 436: Missing spaces?

src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -188,6 +199,18 @@ def _write_params_file(params_file_path, label_job_specs):
                 _format_param_value(job_spec["run_number"], "run_number"),
                 _format_param_value(job_spec["pack_for_grid_register"], "pack_for_grid_register"),
             ]
+
+            if "overwrite_model_parameters" in params_fields:
+                row.append(
+                    _format_param_value(
```
GernotMaier, line 202: job_spec.get("overwrite_model_parameters") is a path? Would _format_param_value change the path? e.g. if there it is something like "my path"?

src/simtools/job_execution/htcondor_script_generator.py
```diff
@@ -308,10 +341,12 @@ def _get_submit_script(args_dict):
     str
         HTCondor submit script content.
     """
-    # Map _PARAMS_FIELDS to bash positional indices ($3, $4, etc.)
+    params_fields = params_fields or _PARAMS_FIELDS
```
GernotMaier: Is this needed? It seems to me that params_field is always set when _get_submit_file is called.

GernotMaier (changes requested):
Very nice!! I have minor comments below plus the separation of configuration and submission. But in general, this is a nice implementation - especially as simtools-production-generate-grid is a generic base-grid generator.

I didn't test the submission step yet (as I had to do this first review offline without HT Condor). Will do this after the updates.

src/simtools/applications/generate_parameter_scan_grid.py
EshitaJoshi: Will stop the application after the scan_grid.ecsv is generated. The user can then pass this on themselves, making this backend agnostic.

src/simtools/job_execution/parameter_scan_generator.py
GernotMaier: Fixed.

src/simtools/job_execution/parameter_scan_generator.py
GernotMaier: Fixed.

PR #2250, opened 2026-06-08
Tag collision

GernotMaier:
PR #2175 introduced the generation of apptainer images additional to the docker-type containers. Unexpected to me, there is a tag collision and the SIF overwrites the docker style image.

I expected that these are really different paths: "docker://${image_tag}" und "oras://${image_tag}". 

To fix this, add a `apptainer` string to the tag for the apptainer SIF images.

See [here](https://github.com/gammasim/simtools/pkgs/container/simtools-prod/versions) for the different container versions generated.


EshitaJoshi (approved):
Thanks for the PR!

GernotMaier:
Thanks @EshitaJoshi

PR #2264, opened 2026-06-12
Document production minor changes

orelgueta:
The scientific notation I think you saw why it is useful.
The other one changes from this plot 

image

to this one

image

I could not see any difference in any other plot as a result of this change.


GernotMaier (approved):

GernotMaier:
Thanks for this @orelgueta

PR #2286, opened 2026-06-24
Try to improve documentation of offset run number

GernotMaier:
I tried to improve the docstrings / help messages.

Note!!

The current `run_number_offset` semantics are not consistent across the production workflow: `Simulator` computes the effective run as `run_number + run_number_offset`, while `production_generate_grid` uses the offset to write absolute grid run numbers starting at `run_number_offset + 1`; if the resulting grid is then executed through the HTCondor generator with the same `run_number_offset`, the offset can be applied a second time. For example, generating a grid with `run_number_offset=100` produces `run_number=101`, but executing that row with `run_number_offset=100` would simulate run 201. We should decide whether offsets belong only at grid-generation time or only at execution time, then document/enforce that contract to prevent double application.

### DIFF
--- a/src/simtools/applications/production_generate_grid.py
+++ b/src/simtools/applications/production_generate_grid.py
@@ -172,7 +172,12 @@ showers_per_run_scaling (str, optional)
     ``showers_per_run * cos(zenith_angle)`` and rounds up to at least one
     shower. Default is ``fixed``.
 run_number_offset (int, optional)
-    Offset added to generated run numbers. Default is ``0``.
+    Number of already assigned run numbers before this generated grid. The
+    first generated run is ``run_number_offset + 1``. For example, use
+    ``run_number_offset=100`` to continue a production after run 100 and start
+    this grid at run 101. The generated grid stores absolute ``run_number``
+    values; do not apply the same offset again when executing the grid.
+    Default is ``0``.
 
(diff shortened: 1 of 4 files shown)


orelgueta:
Thanks for these improvements, it helps make it much clearer.


> The current `run_number_offset` semantics are not consistent across the production workflow: `Simulator` computes the effective run as `run_number + run_number_offset`, while `production_generate_grid` uses the offset to write absolute grid run numbers starting at `run_number_offset + 1`; if the resulting grid is then executed through the HTCondor generator with the same `run_number_offset`, the offset can be applied a second time. For example, generating a grid with `run_number_offset=100` produces `run_number=101`, but executing that row with `run_number_offset=100` would simulate run 201. We should decide whether offsets belong only at grid-generation time or only at execution time, then document/enforce that contract to prevent double application.

I agree there is some inconsistency there, but I think we should leave it like that. Generally when a user uses the applications that create the production configurations, it is not expected to then manually add any arguments to e.g., `simulate_prod`. So the case of applying the offset twice would be considered user error and an unlikely one.
I don't mind further documentation to help with that, but I wouldn't implement code to prevent double application, I think it's overkill.

orelgueta (approved):

GernotMaier:
OK!

PR #2287, opened 2026-06-24
Enforce file name limit for sim_telarray configuration files

### LINKED ISSUE #2278: Catch php limit on maximum file name length
php is reading e.g. include statements in the cfg files using

```
(void) getword(line,&i,mword,80,' ','\0');
```

This limits the length of the file name for the model configuration file. 

We should:

1. catch these limits in simtools before running sim_telarray
2. consider much shorter cfg file names (do we ever 'mix' different configurations in the same directory - if not, let's keep them short)

e.g. affects for PR #2248  - I could successfully run these sims with shorter configuration file names.

GernotMaier:
The idea is that if the generated file name (using site, telescope, labels, extra_labels) is too long, the labels + extralabels are replaced by a random string. 

Closes #2278 .

### DIFF
--- a/docs/changes/2287.feature.md
+++ b/docs/changes/2287.feature.md
@@ -1,1 +1,1 @@
-Enfore file length limits for sim_telarray configuration files.
+Enforce file length limits for sim_telarray configuration files.
--- a/docs/changes/2287.feature.md
+++ b/docs/changes/2287.feature.md
@@ -1,0 +1,1 @@
+Enfore file length limits for sim_telarray configuration files.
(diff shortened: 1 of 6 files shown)


orelgueta (approved):
The fix is good. Can we increase the allowed size in sim_telarray as well so that future versions of both software won't have this limitation?

GernotMaier:
I really don't think it is a limitation. I don't see a single application where we write several configurations into the same model directory. So there is no need to have these file names super long.

PR #2304, opened 2026-06-30
Float point comparison in ruff

GernotMaier:
Add a check for floating point comparison into ruff.

**Note: this is ruff preview, meaning marked as not stable** (but it works very well and I think we should use it)

Note also that it seems to be much better than SonarQ - e.g. SonarQ fails on:

- `assert a == 5/3` 
- `assert a == n.pi`

This PR fixes also a couple of this type of comparison.

For discussion of ruff preview, see [here](https://docs.astral.sh/ruff/preview/#using-rules-that-are-in-preview)

### DIFF
--- a/docs/changes/2304.maintenance.md
+++ b/docs/changes/2304.maintenance.md
@@ -1,0 +1,1 @@
+Add ruff rule on floating point comparison (currently in ruff preview).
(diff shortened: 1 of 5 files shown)


orelgueta (approved):
Is it intentional that this PR is to a different branch rather than main?

Also, in 

> Note also that it seems to be much better than SonarQ - e.g. SonarQ fails on:
> 
> assert a == 5/3
> assert a == n.pi

I assume you mean to say that SQ doesn't catch issues like that, correct?

GernotMaier:
Yes - because it continues on ruff changes and is based therefore on #2302

PR #2368, opened 2026-07-22
Simplified markdown table writer

GernotMaier:
Not a dramatic simplification, but a little bit nicer. I couldn't find a better markdown functionality in astropy.

Closes #2079

### DIFF
--- a/tests/unit_tests/reporting/test_docs_read_parameters.py
+++ b/tests/unit_tests/reporting/test_docs_read_parameters.py
@@ -15,6 +15,15 @@ DESCRIPTION = "Test parameter"
 SHORT_DESC = "Short"
 
 
+def _markdown_table_rows(content):
+    """Return Markdown table rows with only cell-padding whitespace removed."""
+    return [
+        [cell.strip() for cell in line.strip("|").split("|")]
+        for line in content.splitlines()
+        if line.startswith("|")
+    ]
+
+
 def test_get_all_parameter_descriptions(telescope_model_lst, tmp_path):
     args = {
         "telescope": telescope_model_lst.name,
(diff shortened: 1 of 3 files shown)


GernotMaier:
@EshitaJoshi - added you as reviewer, as this was 'your' issue...it feels a bit like cosmetics (but it is a simple change).

tests/unit_tests/reporting/test_docs_read_parameters.py
```diff
@@ -626,9 +635,10 @@ def test__write_parameters_table(tmp_path):
     # Verify list-of-dicts parameter is linked and table is written below
     assert "| dict_param | [View Dict Param](#dict-param) | 4.0.0 |" in output
     assert "## Dict Param" in output
-    assert "| a | b | value |" in output
-    assert "| A | B | 1 m |" in output
-    assert "| X | Y | 2 m |" in output
+    table_rows = _markdown_table_rows(output)
+    assert ["a", "b", "value"] in table_rows
+    assert ["A", "B", "1 m"] in table_rows
+    assert ["X", "Y", "2 m"] in table_rows
```
EshitaJoshi, line 641: I dont understand why the motivation behind the helper function in the tests and changing the testing to this from

```
assert "| A | B | 1 m |" in output
assert "| X | Y | 2 m |" in output
```


The intended output is supposed to be markdown, so we should verify the expected markdown, right?

GernotMaier:
Looking at it again - and we actually don't generate a valid markdown table. Thanks for kind of spotting it @EshitaJoshi 

Review the changes and the original code and the issue I suggest we close this PR without merging and close also the issue. The original point was to use Astropy table, but Astropy has no ascii.markdown functionality. 

Ok with you @EshitaJoshi  ? Sorry for the noise.

EshitaJoshi:
> Looking at it again - and we actually don't generate a valid markdown table. Thanks for kind of spotting it @EshitaJoshi
> 
> Review the changes and the original code and the issue I suggest we close this PR without merging and close also the issue. The original point was to use Astropy table, but Astropy has no ascii.markdown functionality.
> 
> Ok with you @EshitaJoshi ? Sorry for the noise.

Yes, that's good with me, thanks!

GernotMaier:
Thanks - closing it then.

PR #2393, opened 2026-08-02
Trigger rate integration tests

EshitaJoshi:
Adding integration tests for `simtools-generate-bias-curve-submissions` and  `simtools-simulate-prod-htcondor-generator`.


GernotMaier (approved):
Thanks - please change the directory structure and decide on 1 or two threshold required - see the PR in simtools-tests. If this is fixed, we are fine and can merge.

EshitaJoshi:
> Thanks - please change the directory structure and decide on 1 or two threshold required - see the PR in simtools-tests. If this is fixed, we are fine and can merge.

How do you want me to change the directory structure? The application generates `nsb/` and `proton/` folders to keep the relevant files separated and stop them from being overwritten. Right now I'm just putting both folders in `generated/` but i can move them to `generated/bias_curves` instead - is that what you wanted?

GernotMaier:
Yes, I think `generated/bias_curves` would be more consistent. How would one otherwise know that these files are related to bias curves and not some other nsb-related analysis?

EshitaJoshi:
> Yes, I think `generated/bias_curves` would be more consistent. How would one otherwise know that these files are related to bias curves and not some other nsb-related analysis?

Good point, fixed. Thanks for the review, I've made the changes in both PRs. Fixed the directory structure and only checking for one threshold value to keep the generated resource files to a minimum. I will merge this when the tests pass.

PR #2427, opened 2026-08-07
Round pointing angles to two digits in metadata file

orelgueta:
No need to keep too many digits in the DIRAC  metadata dictionary.


EshitaJoshi (approved):
Thanks for the PR!

PR #2486, opened 2026-08-31
Separate the view cone metadata entry

orelgueta:
Separate the view cone metadata entry written for DIRAC to two separate min/max entries.

### DIFF
--- a/src/simtools/production_configuration/job_metadata.py
+++ b/src/simtools/production_configuration/job_metadata.py
@@ -31,7 +31,8 @@ def build_simulation_job_metadata(args_dict, simulator):
         "phiP": round((azimuth_angle + 180.0) % 360.0, 2),
         "thetaP": round(float(args_dict["zenith_angle"].to_value(u.deg)), 2),
         "sct": str(_has_sct(simulator.array_models)),
-        "view_cone": _format_view_cone(view_cone_min, view_cone_max),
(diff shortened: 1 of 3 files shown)


src/simtools/production_configuration/job_metadata.py
```diff
@@ -31,7 +31,8 @@ def build_simulation_job_metadata(args_dict, simulator):
         "phiP": round((azimuth_angle + 180.0) % 360.0, 2),
         "thetaP": round(float(args_dict["zenith_angle"].to_value(u.deg)), 2),
         "sct": str(_has_sct(simulator.array_models)),
-        "view_cone": _format_view_cone(view_cone_min, view_cone_max),
+        "view_cone_min": round(float(view_cone_min.to_value(u.deg)), 2),
```
GernotMaier, line 34: Did the old entry had a `deg` for the units in the view_cone string? Just to be sure that this is a intended change.

GernotMaier (approved):
Have a look at my comment - otherwise good.

src/simtools/production_configuration/job_metadata.py
orelgueta: Yes, it had a `deg` in the string. Now we won't have an explicit unit anymore, but it's worth it to be able to have a float and this is anyway temporary for the current WMS interface. In the future we will hopefully have units and the metadata will be read directly from the output file.

PR #2494, opened 2026-08-31
Remove overwrite files

EshitaJoshi:
Removing the use of overwrite.yaml files to overwrite model parameters, instead we now have the overwrite dictionary in the metadata for the job grid file as follows:

```
# meta: !!omap
# - model_parameter_sets:
#     asum220:
#       LSTN-02:
#         asum_threshold: {value: 220}
#         min_photoelectrons: {value: 0}
#         min_photons: {value: 0}
#       OBS-North:
#         nsb_scaling_factor: {value: 2.0}
```


src/simtools/job_execution/parameter_scan_generator.py
```diff
@@ -176,13 +202,51 @@ def _generate_parameter_combinations(param_specs):
...
+
+    Returns
+    -------
+    dict
+        Dictionary mapping parameter set names to their changes dictionaries
+    """
+    parameter_sets = {}
+
+    for combo_spec in param_combinations:
+        # Build the parameter set name
```
GernotMaier: Tiny comment: no need to have comment lines which say exactly the same as the well named function calls.

src/simtools/applications/simulate_prod.py
GernotMaier, line 1: Adds a lot of code to the application, which we generally try to avoid (but this one had already more code than it should). Is there a possibility to avoid it? Otherwise we do a follow-up cleanup PR.

src/simtools/model/model_parameter.py
```diff
@@ -394,28 +412,100 @@ def _load_parameters_from_db(self):
...
+
+    def _determine_ignore_collections(self):
+        """Determine if we should ignore any collections when applying overrides."""
+        ignore_collections = ("configuration_sim_telarray", "configuration_corsika")
+        if not self.overwrite_model_parameter_dict:
+            return ignore_collections
+
+        if self._has_overrides_for_collections(ignore_collections):
+            return None
+        if self._has_overrides_for_collections(("configuration_sim_telarray",)):
```
GernotMaier, line 436: A bit hard to follow through, but don't we have to do this also if there is a `configuration_corsika` given?

GernotMaier (approved):
Please have a look at the comments - one might be an issue but it is very very hard to understand (which is in the nature of these things).

GernotMaier (approved):
Looks good now, thanks!

PR #2529, opened 2026-09-11
Add simtools-prod-test image

GernotMaier:
Add a generation pipeline for a testing image to avoid frequently failing integration tests due to failing gitlab connections.

This is the 'normal' simtools-prod image with the following additions:

- added simulation-models
- added interaction tables

This image will be used in the CI-integration test.


GernotMaier:
I am closing this - the idea was good, it just doesn't seem to work with a reasonable amount of time invested.

PR #2543, opened 2026-09-18
Incident angle model parameters

### LINKED ISSUE #1976: Photon incident angles as input to camera efficiency - remove as model parameters.
This is a follow-up issue to #1747 and #1938.

The following four distributions are not used as model parameter, but as input to the validate camera efficiency application.:

```
camera_filter_incidence_angle
lightguide_efficiency_vs_incidence_angle
primary_mirror_incidence_angle
secondary_mirror_incidence_angle
```

- [ ] decide if they should be model parameters (meaning stored in the model parameter database) or if we treat them as intermediate data products

If we remove them from the model parameters, we need to:

- change the validate camera efficiency application to read in these values from the command line
- setup a workflow which derives the angles and calculates the model parameters

EshitaJoshi:
Update and cleanup `simtools-derive-incident-angles`, and add derivation and export of `lightguide_efficiency_vs_incidence_angle`.

Closes #1976


src/simtools/applications/derive_incident_angle.py
```diff
@@ -15,22 +17,25 @@
...
-        action="store_true",
+        action=argparse.BooleanOptionalAction,
+        default=False,
         required=False,
     ),
     cli.ArgumentDefinition(
         "calculate_primary_secondary_angles",
         dest="calculate_primary_secondary_angles",
-        help="Compute angles of incidence on primary and secondary mirrors",
+        help="Compute angles of incidence on primary and secondary mirrors (enabled by default)",
```
GernotMaier, line 35: Is this true by default, even for single-mirror telescopes? I am trying to understand what the impact is. Does it e.g., impact the reading of columns using `_default_column_indices`?

src/simtools/ray_tracing/incident_angles.py
```diff
@@ -718,57 +687,160 @@ def _save_results(self):
...
+
+        bin_widths = np.diff(bin_edges)
+        bin_areas = 2.0 * np.pi * np.sin(np.radians(bin_centers)) * np.radians(bin_widths)
+        efficiency = np.divide(
+            detected_counts,
+            bin_areas,
+            out=np.zeros_like(detected_counts, dtype=float),
+            where=bin_areas > 0,
+        )
+        max_efficiency = np.max(efficiency)
```
GernotMaier, line 809: I do struggle quite a bit with this part. Is this really light guide efficiency? Where does "photons transmitted for the light guide / photons incident on light-guide entrance" is applied? Is it possibly that this is only the angle distribution of photons incident on the light-guide entrance?

GernotMaier (changes requested):
See below, probably me not understanding what is going on.

